PR #388, opened 2014-04-30
Literal.toPython() support for xsd:hexBinary

bcogrel:
Following the discussion #386, here is an implementation of the bytes option, where `toPython()` returns a bytes value.


nicholascar:
This PR has been re-issued in PR #979 to cater for Python 2/3 differences and since the original repo has gone away.

PR #405, opened 2014-06-25
Add percent (%) to _unquotedChars in RDFa 1.1. parser

RinkeHoekstra:
If the percent (%) character appears in a URI, the parser should not quote it, as this will lead to double quoted URIs that will no longer resolve. 

For instance:

The HTML page at :

http://link.springer.com/chapter/10.1007%2F978-3-642-25093-4_9

contains a link to an image:

http://link.springer.com/static-content/lookinside/613/chp%253A10.1007%252F978-3-642-25093-4_9/000.png

which becomes:

http://link.springer.com/static-content/lookinside/613/chp%25253A10.1007%25252F978-3-642-25093-4_9/000.png

in the resulting RDF graph. 

Adding the percent character to the _unquotedChars should fix this.


gromgull:
We've moved RDFa out of core rdflib, so closing this. See #828.

Issue #167, opened 2012-02-20
Clarify confusion around type of context element in ConjunctiveGraphs  and context aware stores


ghost:
gromgull, 2011-08-20T07:58:07.000Z

In the ConjunctiveGraph, the context field is assumed to be a graph with the identifier set to the URI of the context, i.e. this is what happens if you create context like this: 

``` python
g=ConjunctiveGraph()
uri1=URIRef("http://example.org/mygraph1")
uri2=URIRef("http://example.org/mygraph2")

bob = URIRef(u'urn:bob')
likes = URIRef(u'urn:likes')
pizza = URIRef(u'urn:pizza')

g.get_context(uri1).add((bob, likes, pizza))
g.get_context(uri2).add((bob, likes, pizza))
```

Now g.contexts() returns a generator over some graphs. 

Now, for code working on the store level, i.e. serializers and parsers, there should perhaps not be any graph objects? 

I came across this when looking at the nquad parser, here: 

https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/parsers/nquads.py#L106

This adds context as simply an URI ref. 

I added a "work-around" to make the conjunctivegraph.contexts generator work "correctly" here: 

https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1075

Is this ok?

kvjrhall:
Note that this is also observed when using `Datastore` instances. In particular,

```
ds = Datastore()f
quads = ds.quads((None, None, None, None))   # Fourth term is identifier

store = SPARQLUpdateStore()
store.addN(quads)  # Expects four term to be graph
```

This seems to have been intended with #409 with the goal of reducing confusion for end users. My personal experience was a large amount of confusion that required me to review the code for both modules in order to identifying what was going wrong.

As a semantic web developer I've been trained to always interpret `triple` or `quad` as a tuple of graph nodes. Ideally, I'd hope that the interface match the expectations of a RDF Dataset as closely as possible and behave as a "set of quads" much the way that a graph can be treated as simply a "set of triples".

The documentation for the referenced methods, their argument names, and their method names are all ambiguous and add to the confusion. If a developer has started working with one of the APIs first (i.e., they are used to `Datastore`) then the first API has "deceived" them by setting their expectations incorrectly.

[...]

PR #409, opened 2014-07-11
Move Store API to work with identifiers, not graphs

gromgull:
This is a long-time coming fix for #167. 

There are two API layers in RDFLib, the Graph level and the Store level. The issue here is the handling of the fourth element of a quad, i.e. the context. Currently both layers use Graph objects to represent this fourth context element, but this leads to lots of trouble (see examples in #167). 

This is an attempt to sort this out by having Graphs handled at the Graph API level, and have stores deal with the graph identifiers only. Hopefully, we can make this change transparent to end-users, but plugin developers will probably notice. 

A list of changes:
- all store.add/remove methods now raise an Exception if passed a graph.
- the Graph API works on Graph objects and will pack/unpack as required.

This is not yet finished, but a PR helps me get an overview of what is left to fix.


gromgull:
I've rebased this onto the six_2to3 branch, and it passes the tests now.

gromgull:
but the PR is for master :sweat_smile:

joernhees:
one can change the base branch of a PR (edit on top), i'll do this...

nicholascar:
I've assigned this to Milestone 6.0.0 and have marked it *critical* as this and graph IDing issues in general are well known annoyances and need a general solution (i.e. for Graph, ConjunctiveGraph, QuotedGraph??, Dataset and Store). 

This PR needs deduplicating with PR #845.

nicholascar:
This PR is long since obsolete so I'm closing

PR #417, opened 2014-08-12
Use <<= instead of deprecated << in SPARQL parser

bogdan-kulynych:
Allows to use newer pyparsing (>=2.0.1) without getting deprecation warnings:

```
pyparsing.py:3546: DeprecationWarning: Operator '<<' is deprecated, use '<<=' instead
  ret << Group( Suppress(opener) + ZeroOrMore( ignoreExpr | ret | content ) + Suppress(closer) )
```

Addresses inveniosoftware/invenio#2077


gromgull:
we'll upgrade to pyparsing 2 for all version for v5 and merge it then.

PR #431, opened 2014-10-30
Added trig unit tests to highlight some current parsing/serializing issues

davechallis:
This adds some TriG unit tests that currently fail (but should pass when issues are resolved).

`testGraphParsing` currently fails due to the TriG parser treating triples in the default graph with and without enclosing braces as 2 separate contexts, whereas they should both be treated as one.

`testRoundTrips` tests whether repeatedly parsing/serializing the same data results in a graph containing the same items.  Currently this fails due to a combination of the issue mentioned above, and due to an issue with handling QNames in the TriG serializer.

`testDefaultGraphSerializesWithoutName` checks that triples in the default graph are serialized correctly.  Currently this fails, as the default graph is serialized as `<None> { ... }`.

I'm planning to look into fixes for these, but thought adding some failing unit tests first would be more useful, in case anyone else wants to look at or fix them too.

### DIFF
--- a/test/test_trig.py
+++ b/test/test_trig.py
@@ -46,3 +46,72 @@ class TestTrig(unittest.TestCase):
         self.assertEqual(len(re.findall(b("p2"), s)), 1)
 
         self.assert_(b('{}') not in s) # no empty graphs!
+
+    def testGraphParsing(self):
+        # should parse into single default graph context
+        data = """
+<http://example.com/thing#thing_a> <http://example.com/knows> <http://example.com/thing#thing_b> .
+"""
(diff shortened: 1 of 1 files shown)


gromgull:
I've rebased this onto current master, and apart from `testRoundtrip` all tests now pass. 

Test RoundTrip fails due to https://github.com/RDFLib/rdflib/issues/436 - and it can't really be fixed in 4.2.2 without changing the API somewhere. 

I'll push this to 5.0

gromgull:
I've added some more commits in #699

PR #438, opened 2014-11-24
Transition from 2to3 to use of six.py

joernhees:
The old porting approach in #374 to six became quite messy because it started by deactivating 2to3 and then tried to fix all errors. Instead let's try to work through a diff of py2 with py3 after 2to3 and deactivate 2to3 one by one whenever a file is done. In the process the tests are meant to always pass.

Branch is unstable!


joernhees:
just rebased branch on top of master after release of 4.2.1 !

joernhees:
yay, that was quite some code knitting...

joernhees:
just rebased branch on top of master in 4.2.1-dev at 53b56b9b049d0649b6946f7de4642355e5e63949

ashleysommer:
removing milestone 6.0.0 from this, it was closed in favor of #519, which is already merged into 5.0.0 since 2017.

PR #443, opened 2014-12-15
Microdata parser: updated the parser to the latest version of the microdata->rdf note (published in December 2014)

joernhees:
re-based microdata-to-rdf-second-edition from six_2to3 branch as described here: https://github.com/RDFLib/rdflib/commit/b082c4851840f4c2c83910f84331fb07fca16f4c#commitcomment-8975378

pull request to run the tests...

### DIFF
--- a/rdflib/plugins/parsers/pyMicrodata/__init__.py
+++ b/rdflib/plugins/parsers/pyMicrodata/__init__.py
@@ -47,7 +47,7 @@ U{W3C® SOFTWARE NOTICE AND LICENSE<href="http://www.w3.org/Consortium/Legal/200
 $Id: __init__.py,v 1.15 2012/09/05 16:40:43 ivan Exp $ $Date: 2012/09/05 16:40:43 $
 """
 
-__version__ = "1.2"
+__version__ = "2.0"
 __author__  = 'Ivan Herman'
 __contact__ = 'Ivan Herman, ivan@w3.org'
 
@@ -60,7 +60,6 @@ else :
 	from StringIO import StringIO
 
 import datetime
-import os
 
 import rdflib
 from rdflib	import URIRef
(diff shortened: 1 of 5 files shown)


joernhees:
Going through this again, the remaining problem is that rdflib and pymicrodata diverged when some issue was fixed in rdflib but not in pymicrodata. Then you made a large update to pymicrodata and that update now seems to break one of the rdflib tests (which was introduced after the fixes).

I think we should definitely do two things:
 1. backport the fix from rdflib to pymicrodata if necessary, get both versions in sync
 2. inspect what's wrong in the failing test (and maybe change it)

Wrt. to 2 you already pointed out that the test might not be the best, which i agree to... it is just a syntactic test for semantic data. The benefit is that it notified us of a problem. Part of this problem seems to have been that some default prefixes changed, which could be desired or not. If you say it's ok, it's not much work to fix the test.

Another thing (and where we definitely started to talk about too many different things at once) is how to make sure we won't have such a problem again. Let's put that decision into #582.

iherman:
Jörn,

as I said in my original reply: I really have no time to deal with this issue now. Is it something that you can take care of?

Ivan

joernhees:
short on time myself, but i'll look into this

joernhees:
OK, having looked into this for a while yesterday and today, this is where i'm at:

- [pyMicrodata](https://github.com/RDFLib/pymicrodata/tree/master/pyMicrodata) contains @iherman's latest version
- there are the following commits that touch `plugins/parsers/pyMicrodata/*` after commits by @iherman only and i've checked they are in [pyMicrodata](https://github.com/RDFLib/pymicrodata/tree/master/pyMicrodata) already:
   - [x] fade112 - Fix local microdata parsing in Python3 as well (also see #377) @dbs
   - [x] d77e3a4 - add file:// to base if it's a filename (also see #406) @gromgull      
   - [x] 0529f0d - Microdata parser: The handling of the <meter> and <data> elements were missing @iherman
   - [x] 227b41f - cleaned up trailing whitespace @gromgull 
     - this is a whitespace only commit, which i'll do anyhow

As the changes are backported already it seems, i'll next copy all the new version back into rdflib and make a PR to see how that goes with our tests.

Finally, when that's figured out, i'll close this and we'll think about which way to resolve this massive timesink once and for all in #582.

gromgull:
See https://github.com/RDFLib/rdflib/pull/828

Issue #400, opened 2014-06-12
Bad UTF-8 handling for NTriple parser


Arthur-VaisseLesteven:
The current NTriple parser does not handle properly utf-8 string.

the following triple trigger an error : 
&lt;http://linkedgeodata.org/triplify/user22701&gt; &lt;http://www.w3.org/2000/01/rdf-schema#label&gt; "CiaránMooney" .

when executing this code :
<pre>
line = codecs.open("file", "r", "utf-8").readline()
parser = NTriples.NTriplesParser(sink = mySink)
parser.parsestring(line)#where line is the previous triple
</pre>
The parser fail to parse the "á" character.
Modifying the parse function of the NTriplesParser class allow to solve this bug.

Current :
<pre>
  def parse(self, f):
        """Parse f as an N-Triples file."""
        if not hasattr(f, 'read'):
            raise ParseError("Item to parse must be a file-like object.")

        f = ascii(f)
        ...
</pre>
Fix:
<pre>
  def parse(self, f):
        """Parse f as an N-Triples file."""
        if not hasattr(f, 'read'):
            raise ParseError("Item to parse must be a file-like object.")

        f = codecs.getreader("utf-8")(f)
        ...
</pre>
Hope this help.

AVL

PR #447, opened 2015-01-03
added support for nt in utf-8 encoding

Kinmarui:
By default it will work like previously but now you can add encoding parameter when parsing file

### usage

g is standard Graph
`g = rdflib.Graph()`

old syntax
`result = g.parse( 'c:\\sample_ascii_file.nt', format="nt") # still works`

new syntax
`result = g.parse( 'c:\\sample_unicode_file.nt', format="nt", encoding="utf-8")`

### DIFF
--- a/rdflib/plugins/parsers/nt.py
+++ b/rdflib/plugins/parsers/nt.py
@@ -20,8 +20,8 @@ class NTParser(Parser):
     def __init__(self):
         super(NTParser, self).__init__()
 
-    def parse(self, source, sink, baseURI=None):
+    def parse(self, source, sink, baseURI=None, encoding="ascii"):
         f = source.getByteStream()  # TODO getCharacterStream?
         parser = NTriplesParser(NTSink(sink))
-        parser.parse(f)
+        parser.parse(f, encoding)
         f.close()
--- a/rdflib/plugins/parsers/ntriples.py
+++ b/rdflib/plugins/parsers/ntriples.py
@@ -10,7 +10,7 @@ from rdflib.term import URIRef as URI
 from rdflib.term import BNode as bNode
 from rdflib.term import Literal
 
-from rdflib.py3compat import cast_bytes, decodeUnicodeEscape, ascii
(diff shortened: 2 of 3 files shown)


joernhees:
thanks for the attempt, but build fails on [py2.6](https://travis-ci.org/RDFLib/rdflib/jobs/45777388) & [py2.7](https://travis-ci.org/RDFLib/rdflib/jobs/45777389)

aside from this i think an encoding flag like this is a bit confusing: (also see #400)

The NTriples parser should be able to automatically handle old style NTriples 1.0 ASCII "\u..." encoding as well as NTriples 1.1 UTF-8 encoding. As NTriples 1.1 is backwards compatible (#400 as well), the default should be accepting UTF-8 encoding and also resolving the "\u..." escapes, making the encoding flag useless.

You could maybe argue that one might want to check if a file is NTriples 1.0, but that case is the opposite of what you tried to implement here. Also, in that case i'd say it's rather part of the format than specifying an encoding, so the format string in the parse method should be something like `format=nt1.0` instead of specifying an encoding.

As mentioned in #400 the story looks different for serialization: there one should have a possibility to specify that one needs a NTriples 1.0 file, default should be NTriples 1.1.

Kinmarui:
thanks for response,

encoding flag was only option that came to my mind because I didn't find a way to determine file encoding inside parser. 
If you know how to determine encoding when I have file stream(BufferedReader) in parser please share even partial solution. Or can I safely use Arthur-VaisseLesteven suggestion posted in https://github.com/RDFLib/rdflib/issues/400 because ascii is a subset of UTF-8 encoding, and I don't need any flag to handle ascii files in any special way?

I also [fixed issue](https://travis-ci.org/RDFLib/rdflib/builds/45850621) with python 2.x.
Anyway this is duplicate of https://github.com/RDFLib/rdflib/issues/400 so you can close this thread and we can move there if you prefer.

joernhees:
yupp, the suggestion in #400 should work be in master as per #449

PR #451, opened 2015-01-29
Bugfix: term.Literal.__add__

pinbe:
As long as Literal(1) is internaly casted by int, Literal(1) + Literal(1) may return Literal(2) instead of Literal('11').

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -655,7 +655,7 @@ class Literal(Identifier):
             except TypeError:
                 pass  # fall-through
 
-        s = unicode.__add__(self, val)
+        s = py + val.toPython()
         return Literal(s, self.language, self.datatype)
 
     def __nonzero__(self):


gromgull:
That is the intended behaviour. Literal comparisons and operations are done in `value space` to make it easy to implement the SPARQL engine. See https://rdflib.readthedocs.org/en/latest/upgrade3to4.html#datatyped-literals

gromgull:
Wait - I may have been too hasty :)

pinbe:
Le 29 janv. 2015 à 15:08, Gunnar Aastrand Grimnes <notifications@github.com> a écrit :

> Wait - I may have been too hasty :)

Thank you your reply. I was preparing a more detailed explanation, just below :-) just bellow.

Regards.
Benoît Pin.


--------

Thank for your reply: You confirms that comparisons and operations should be done in the *value space*.
Actually, this information also confirm the Literal.__add__ method is buggy.
The bug report was probably too short.
Let me explain longer:

Here's two literals that are integer values

```
>>> one = Literal(1)
>>> two = Literal(2)
>>> print one.datatype, two.datatype
http://www.w3.org/2001/XMLSchema#integer http://www.w3.org/2001/XMLSchema#integer
```

Now, let's do the sum with the current rdflib version

```
>>> print one + two
12
```

This buggy result denotes that the sum has been made in the *lexical space*.

After applying the proposed fix, you will obtain the correct result in the *value space*.

```
>>> print one + two
3
>>> print (one + two).datatype
http://www.w3.org/2001/XMLSchema#integer
```

nicholascar:
Replaced by PR 978 which is updated with commits since this PR and also new tests for move addition value and type combinations

PR #459, opened 2015-02-24
SPARQLStore does not transform Literal('') into Literal('None') anymore, fixes #457

### LINKED ISSUE #457: Bad SPARQLStore handling of empty literals
When an empty literal is returned to a SPARQLStore, it becomes an `rdflib.term.Literal(u'None')` instead of an `rdflib.term.Literal(u'')`:

```python
# ./fuseki-server --update --mem /empty_literal

import rdflib
from rdflib.plugins.stores.sparqlstore import SPARQLUpdateStore

store = SPARQLUpdateStore('http://fuseki.til.io:3030/empty_literal/query',
                          'http://fuseki.til.io:3030/empty_literal/update')

graph = rdflib.Graph(store, 'http://example.com/')

graph.add((rdflib.URIRef('http://example.com/s'), 
           rdflib.URIRef('http://example.com/p'), 
           rdflib.Literal('')))

o = tuple(graph)[0][2]
assert o == rdflib.Literal(''), repr(o)
```
raises an `AssertionError: rdflib.term.Literal(u'None')`

The problem seems to be in `rdflib.plugins.stores.sparqlstore:CastToTerm`:

```python
import io
from xml.etree import ElementTree

import rdflib
from rdflib.plugins.stores.sparqlstore import CastToTerm

s = """<ns0:literal xmlns:ns0="http://www.w3.org/2005/sparql-results#" />"""
node = ElementTree.parse(io.BytesIO(s)).getroot()
term = CastToTerm(node)
assert term  == rdflib.Literal(''), repr(term)
```
raises an `AssertionError: rdflib.term.Literal(u'None')`

joernhees:
fixes #457

### DIFF
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -132,10 +132,7 @@ def CastToTerm(node):
         value = node.text if node.text is not None else ''
         if 'datatype' in node.attrib:
             dT = URIRef(node.attrib['datatype'])
-            if False:  # not node.xpath('*'):
-                return Literal('', datatype=dT)
-            else:
-                return Literal(value, datatype=dT)
+            return Literal(value, datatype=dT)
         elif '{http://www.w3.org/XML/1998/namespace}lang' in node.attrib:
             return Literal(value, lang=node.attrib[
                 "{http://www.w3.org/XML/1998/namespace}lang"])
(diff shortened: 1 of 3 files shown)


joernhees:
hrmm, seems as if the introduced test fails on py3

gjhiggins:
Can't recall the origin of the code offhand but that False was in the original commit.

The fragment is obsolete and can be deleted to leave just:

```python
if 'datatype' in node.attrib:
    dT = URIRef(node.attrib['datatype'])
    return Literal(value, datatype=dT)
```

A fix for the test:

```python
from rdflib.py3compat import b
s = b("""<ns0:literal xmlns:ns0="http://www.w3.org/2005/sparql-results#" />""")
```

I'll leave it up to you to decide whether to update the PR or make a separate commit. 

Cheers

Graham

joernhees:
thanks for the feedback. had to leave yesterday, i'll just update the PR.

PR #463, opened 2015-02-25
Replace "A and B or C" by "B if A else C"

bcroq:
With "A and B or C", when B is a context (ie a graph), Python implicitly calls `Graph:__len__`. This call can be quite long, we don't need it.

More informations here: https://docs.python.org/2/reference/datamodel.html#object.__nonzero__


joernhees:
thanks, seems quite legit and makes the code way more readable and faster

Issue #461, opened 2015-02-25
decrease number of skipped tests


joernhees:
currently a lot of tests are skipped (98 as in https://travis-ci.org/RDFLib/rdflib/jobs/52130178) .
- [ ] find out why so many are skipped
- [x] see if travis environments can be configured to run more (e.g., local sparql endpoints)
- [ ] re-enable what's possible

gjhiggins:
Yup.
commit https://github.com/gjhiggins/rdflib/commit/7d3c58ede43125fcceff03f91727c3301e75ab5f
and travis job 
https://travis-ci.org/gjhiggins/rdflib/jobs/52323760
confirming your observation of a test failure for sparqlupdatestore.

Worth a PR or would prefer to integrate it by hand?

joernhees:
wow, cool, thanks for figuring that out.

sure, always worth a pull request... just refer this issue in it

gjhiggins:
Expired.

PR #466, opened 2015-02-28
Running a local fuseki server on travis

joernhees:
see #461


joernhees:
hmm, actually after the tests ran through it seems that now tests for Dataset fail... ConjunctiveGraph and Dataset seem to expect different behaviors from the endpoint, so maybe running both test suites on one invocation of fuseki is not possible?

PR #469, opened 2015-03-05
Don't check for ops in reverseOps before removing it, just try it.

bcroq:


rdflib/plugins/stores/auditable.py
```diff
@@ -56,9 +56,10 @@ def add(self, triple, context, quoted=False):
             context = context.__class__(self.store, context.identifier) if context is not None else None
             ctxId = context.identifier if context is not None else None
             self.reverseOps.append((s, p, o, ctxId, 'remove'))
-            if (s, p, o, ctxId, 'add') in self.reverseOps:
-                self.reverseOps.remove(
-                    (s, p, o, context, 'add'))
+            try:
+                self.reverseOps.remove((s, p, o, ctxId, 'add'))
+            except ValueError:
+                pass
```
joernhees: not so much a comment for this change, but for my understanding: is this to "compress" the reverseOps only?

rdflib/plugins/stores/auditable.py
```diff
@@ -72,26 +73,21 @@ def remove(self, (subject, predicate, object_), context=None):
             if None in [subject, predicate, object_, context]:
                 if ctxId:
                     for s, p, o in context.triples((subject, predicate, object_)):
-                        if (s, p, o, ctxId, 'remove') in self.reverseOps:
+                        try:
                             self.reverseOps.remove((s, p, o, ctxId, 'remove'))
-                        else:
+                        except ValueError:
                             self.reverseOps.append((s, p, o, ctxId, 'add'))
```
joernhees: this somehow seems "asymmetric" to the above add, please re-check

rdflib/plugins/stores/auditable.py
bcroq: In add() there was a bug: it checks if (s, p, o, **ctxId**, 'remove') is in self.reverseOps and if it is found it removes (s, p, o, **context**, 'add')

Then, it is generally a bad idea to check if an element is in a list before removing it. Just try to remove and deal with the ValueError exception.

```python
if x in l:
    l.remove(x)
else:
    l.add(y)
```

becomes:

```python
try:
    l.remove(x)
except ValueError:
    l.add(y)
```
It saves an uneeded call to `__contains__`

rdflib/plugins/stores/auditable.py
bcroq: This is typically the 

```python
if x in l:
    l.remove(x)
else:
    l.add(y)
```
becomes:

```python
try:
    l.remove(x)
except ValueError:
    l.add(y)
```

transformation.

PR #472, opened 2015-03-09
exclude `def main():` functions from test coverage analysis

joernhees:
main() functions in nearly all cases are there for testing the local files only


joernhees:
no effect, so not worth it...

PR #473, opened 2015-03-15
code to convert an rdflib.Graph into a graph_tool.Graph

joernhees:


joernhees:
currently fails due to an import error, as there is no pypi package for graph_tool it seems...
see https://git.skewed.de/count0/graph-tool/issues/208

joernhees:
:( seem travis is still running on Ubuntu 12.04 LTS for which there aren't any graph_tool packages... i guess this just has to wait or be committed without running its tests for now :(
any feedback?

PR #476, opened 2015-03-15
make travis fuseki install more failsafe, fixes #475

### LINKED ISSUE #475: travis fuseki install blocks all tests if download fails
also see #474 and https://travis-ci.org/RDFLib/rdflib/jobs/54469463 for example...

we should try to make this more fail safe / at least find a more stable URI...

joernhees:


joernhees:
k, tested with a broken URI... the message isn't really standing out, but if fuseki is not running we'll get messages about skipped SPARQL tests, so i guess that's enough

PR #480, opened 2015-03-30
trig output fixes

drewp:


rdflib/plugins/serializers/trig.py
```diff
@@ -40,22 +41,6 @@ def preprocess(self):
...
-            if node in self.keywords:
-                continue
-            # Don't use generated prefixes for subjects and objects
-            self.getQName(node, gen_prefix=(i == VERB))
-            if isinstance(node, Literal) and node.datatype:
-                self.getQName(node.datatype, gen_prefix=_GEN_QNAME_FOR_DT)
-        p = triple[1]
-        if isinstance(p, BNode):
-            self._references[p]+=1
-
```
joernhees: did you remove this on purpose? seems it is called from `preprocess` and the one in `TurtleSerializer` looks quite different to me

drewp: TurtleSerializer's one is a super call followed by a repeat of lines 48-57. The super call would run RecursiveSerializer's method, which is a repeat of lines 45-47. Most importantly, the test suite still passes :)

joernhees: ah sorry, i must have looked at the `RecursiveSerializer` one directly... so essentially this was unnecessary code duplication?

rdflib/plugins/serializers/trig.py
```diff
@@ -83,7 +68,10 @@ def serialize(self, stream, base=None, encoding=None,
             if self.default_context and store.identifier==self.default_context:
                 self.write(self.indent() + '\n{')
             else:
-                self.write(self.indent() + '\n<%s> {' % self.getQName(store.identifier))
+                iri = self.getQName(store.identifier)
+                if iri is None:
+                    iri = '<%s>' % store.identifier
+                self.write(self.indent() + '\n%s {' % iri)
```
joernhees: two questions here:
 - should it be possible that `store.identifier` isn't in `getQName` at all?
 - wouldn't this introduce a bug if `iri is not None` cause it's not wrapped in `<%s>` anymore?
    (so shouldn't line 73 be: `iri = '%s' % store.identifier` and 74: `self.write(self.indent() + '\n<%s> {' % iri)` ?

rdflib/plugins/serializers/trig.py
drewp: Yes I think so

rdflib/plugins/serializers/trig.py
drewp: getQName gets to return None if it has no abbreviation, which I test with the URIRef("http://example.com/foo.") case.

As I understand it, a non-None iri (e.g. a successful abbreviation like "ns1:foo") shouldn't be wrapped in <>. That seemed to be a bug before.  See Example 1 in http://www.w3.org/TR/trig/#sec-triple-statements for a demo of no angle brackets.

I pushed an update to get bnode graph ids right too (like "_:a1b2c3 { ... }").

joernhees:
thanks for the changes and comments, i think this looks good now...

PR #482, opened 2015-04-22
fix broken example

ggstuart:
The example updates age but should be updating FOAF.age in order to produce the required output


joernhees:
thanks ;)

PR #486, opened 2015-05-26
Batch commits for SPARQLUpdateStore

jimmccusker:
SPARQLUpdateStore writes an HTTP transaction for each and every triple addition or deletion. When adding significant amounts of data via this API, this results in a lot of network chatter and significant delays while the transactions are completed. This change batches edits up until the user calls commit() or until a read operation is performed. The result should always be the same behavior from any given client's perspective, since all edits are applied before a read occurs.


joernhees:
seems legit to me, but as this changes the current behavior of auto-committing changes, i think we should either put this in rdflib 5.x.x (as it breaks backwards compatibility according to http://semver.org) or maybe you could add an `autocommit` flag to its init which changes the store's behavior. If that flag defaults to `True` the change would be backwards compatible and could go into 4.2.1

jimmccusker:
I'll add the flag then.

joernhees:
:+1:

PR #497, opened 2015-07-17
no relative import in when run as a script

cliffxuan:
original throws an exception
```
  File "plugins/sparql/algebra.py", line 787, in <module>
    from . import parser
ValueError: Attempted relative import in non-package
```

### DIFF
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -101,8 +101,7 @@ class URLInputSource(InputSource):
         # Fix for issue 130 https://github.com/RDFLib/rdflib/issues/130
         self.url = file.geturl()    # in case redirections took place
         self.setPublicId(self.url)
-        self.content_type = file.info().get('content-type')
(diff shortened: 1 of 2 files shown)


joernhees:
could you provide a minimal example of what you're trying to do?

cliffxuan:
just to run the script does not work currently because of the relative import 

```
➜  rdflib (master) python plugins/sparql/algebra.py "select * where {?s ?p ?o}"
```

joernhees:
well, you can actually run it like this (this is a standard relative imports & path issue):
```
python -m plugins.sparql.algebra "select * where {?s ?p ?o}"
```

I agree that we could improve the usability of that as in 9d1318c .

5a895ff seems to be unrelated to me.

joernhees:
@cliffxuan ok, i pulled 9d1318c into master, i'm closing this in case you want to re-open a pull request for 5a895ff

PR #498, opened 2015-07-20
handling no content-type header returned

joernhees:
@cliffxuan moved here from #497

### DIFF
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -101,8 +101,7 @@ class URLInputSource(InputSource):
         # Fix for issue 130 https://github.com/RDFLib/rdflib/issues/130
         self.url = file.geturl()    # in case redirections took place
         self.setPublicId(self.url)
-        self.content_type = file.info().get('content-type')
-        self.content_type = self.content_type.split(";", 1)[0]
+        self.content_type = file.info().get('content-type', '').split(";", 1)[0]
         self.setByteStream(file)
         # TODO: self.setEncoding(encoding)
         self.response_info = file.info() # a mimetools.Message instance


cliffxuan:
the server doesn't necessarily have to return a `content-type` header, and you probably don't want to make this a show stopper?

The following is the http response for `http://dev.w3.org/2000/10/swap/test/cwm/fam-rules.n3`, and it doesn't have a `content-type` header.

```
HTTP/1.1 200 OK
Date: Mon, 20 Jul 2015 12:48:50 GMT
Server: Apache/2.2.22 (Debian)
Last-Modified: Fri, 30 Jan 2004 16:08:09 GMT
ETag: "122aa2-96-3d22471246c40"
Accept-Ranges: bytes
Content-Length: 150
Cache-Control: max-age=172800
Expires: Wed, 22 Jul 2015 12:48:50 GMT
Keep-Alive: timeout=5, max=100
Connection: Keep-Alive
```

joernhees:
it's correct that content-type isn't mandatory: http://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1

the thing i'm not sure about is whether we should return `None` or `''` or just raise an exception right here... what speaks against `''` is the handling of `content_type` in [Graph.parse()](https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1031). If we return `''` it won't enter the default handling of `if format is None`... but then the default handling would be wrong in this case and the comments there show that we should maybe be explicit and raise an exception.

So if at all, i'd rather return `None`:

```diff
         self.content_type = file.info().get('content-type')
-        self.content_type = self.content_type.split(";", 1)[0]
+        if self.content_type is not None:
+             self.content_type = self.content_type.split(";", 1)[0]
```

cliffxuan:
both empty string and `None` are falsy.

```
        if format is None:
            format = source.content_type
        if format is None:
            # raise Exception("Could not determine format for %r. You can" + \
            # "expicitly specify one with the format argument." % source)
            format = "application/rdf+xml"
```

I would change the above to:

```
if not format:
    format = source.content_type or 'application/rdf+xml'
```

http://stackoverflow.com/questions/9573244/most-elegant-way-to-check-if-the-string-is-empty-in-python

joernhees:
True, but this would change our API semantics. We usually treat `None` as default (as in not specified) and `''` as in argument was explicitly set to empty string. See the signature of [Graph.parse(..., format=None, ...)](https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L960).

If a content-type is specified in HTTP it can't be an empty string, so i think it's worse to suddenly introduce a second undefined `''` rather than just return `None`.

In favor of a smaller change i won't touch `Graph.parse()` for this... see #499

joernhees:
still thanks for reporting this and sorry for opting another way

cliffxuan:
that's ok. i don't mind too much, to be honest.

cliffxuan:
a side question, are you also the maintainer of Fuxi? what is the status of it? i created these fixes because i'm playing with the examples in Fuxi.

joernhees:
I'd say @chimezie and @gjhiggins, but see https://github.com/RDFLib/FuXi/issues/9#issuecomment-39822633

Issue #495, opened 2015-07-09
Dataset class documentation refers to non-existant .graphs() method


ssssam:
In a few places the Dataset class docstring refers to a `Dataset.graphs()` method. But there's no such thing. Should it refer to `Dataset.contexts()` instead? Or should `Dataset.graphs()` be added as an alias of `Dataset.contexts()` ?

joernhees:
i think you're right and it should refer to `Dataset.contexts()` but it's probably safest to ask @iherman and @gjhiggins again who modified those [lines...](https://github.com/RDFLib/rdflib/blame/master/rdflib/graph.py#L1560)

gjhiggins:
The program semantics will be Ivan's call, my edits were in pursuit of better PEP8 conformance and confined to the syntax level.

iherman:
I am sorry guys, you are absolutely right; my mistake. I have added ``context`` to the ``Dataset`` to 'hide' the behaviour of the supertype, but the ``graphs`` method should be the obvious method for the user. (Ie, no need to document ``context``, in fact.)

I have added ``graphs``, see PR #504; I let you pull it.

Note that I have simply copied ``context`` verbatim (and have tested it locally). I am not sure whether there is a more elegant way of referring to ``self.context``, because ``self.context`` is an iterator (using ``yield``). Instead of experimenting, I just made a verbatim copy.

Cc: @gjhiggins, @joernhees

gromgull:
I removed this intentionally in https://github.com/RDFLib/rdflib/commit/1ed4feb63219badbe910b8b402377f011a5853e7

as it does exactly what the `.contexts` method does.

gromgull:
I dont feel very strongly about whether we reintroduce the method or fix the docs, but the method should just call .contexts, not duplicate the code.

iherman:
@gromgull 

see [my comment](https://github.com/RDFLib/rdflib/pull/504#issuecomment-126996238) on whether to use ``graphs()`` or not.

As for duplication or not: as [I said](https://github.com/RDFLib/rdflib/issues/495#issuecomment-126903646), just saying

```
def graphs(self,t):
    return self.contexts(t)
```

does not work because ``contexts`` is an iterator. I was not sure what the best way is to do that, maybe

```
def graphs(self,t):
    for s in self.contexts(s) : yield s
```

would work, but I did not have the time to test that (thinking about it further this should probably work!). There is probably a standard python idiom for this, but I do not know from the top of my head.

[...]

PR #504, opened 2015-08-01
Added a 'graphs' method to the Dataset object 

iherman:
This is to handle issue #495.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1625,16 +1625,12 @@ class Dataset(ConjunctiveGraph):
     def contexts(self, triple=None):
         default = False
         for c in super(Dataset, self).contexts(triple):
-            default|=c.identifier == DATASET_DEFAULT_GRAPH_ID
+            default |= c.identifier == DATASET_DEFAULT_GRAPH_ID
             yield c
-        if not default: yield self.graph(DATASET_DEFAULT_GRAPH_ID)
+        if not default:
+            yield self.graph(DATASET_DEFAULT_GRAPH_ID)
 
-    def graphs(self, triple=None) :
-        default = False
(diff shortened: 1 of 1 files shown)


gromgull:
As far as I can recall it was removed intentionally, as the same functionality is available from the `.contexts()` method - inherited from the super-class. Let me check ...

gromgull:
Yep - removed and mentioned in the commit message here: https://github.com/RDFLib/rdflib/commit/1ed4feb63219badbe910b8b402377f011a5853e7

I would vote we fix the docs instead.

iherman:
Well... I respectfully disagree. The very reason for having the ``Dataset`` object is because, if one uses RDFLib in terms of the RDF 1.1 standard, then the notion of "context" is unknown, whereas the term dataset and graph is defined and used all over the place. In this sense, from a usability point of view, the function name ``graphs()`` is clear, whereas the term ``contexts()`` is not. 

I do not remember the history of this, it has been a very long time ago when this was done. It may well be that, originally, I actually used the method name ``graphs()`` and somebody changed to ``contexts()`` without me paying attention. That was my mistake then (I have to admit that I did not really look at rdflib in the past few years, my work focus at W3C have radically changed lately).

I am not sure what the decision process is in this case, and I do not want to make a big fuss about this; not being very active any more I do not think it is my place to do so. I just want to express my strong disagreement...

gromgull:
`decision process` is a big word for what happens in RDFLib :) 

Like you, I am not really using rdflib much myself anymore, and you make a fair point. I have no problem with reintroducing the method as an alias for `contexts`, it should then probably just do `return self.contexts()` though, and not duplicate the code.

Issue #494, opened 2015-07-08
RDGA1 graph canonicalization sometimes collapses distinct BNodes


joernhees:
There seems to be a bug in RDGA1 wrt. graph canonicalization.
Given an input graph `g`, `cg = rdflib.compare.to_canonical_graph(g)` will in some cases construct a canonical version of the graph that can't be isomorphic to `g` as `cg`'s node degrees are different.

Below is an example graph `g` showing this behavior. I'm a bit sorry for the length, it's basically the RDF representation of a SPARQL BGP (see #483 for more info) by using reification for each triple in the SPARQL pattern:

```python
import rdflib
g = rdflib.Graph()
g += [
 (rdflib.term.BNode('Na1a8fbcf755f41c1b5728f326be50994'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#object'),
  rdflib.term.URIRef(u'source')),
 (rdflib.term.BNode('Na1a8fbcf755f41c1b5728f326be50994'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#predicate'),
  rdflib.term.BNode('vcb3')),
 (rdflib.term.BNode('Na1a8fbcf755f41c1b5728f326be50994'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#subject'),
  rdflib.term.BNode('vcb2')),
 (rdflib.term.BNode('Na1a8fbcf755f41c1b5728f326be50994'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#Statement')),
 (rdflib.term.BNode('Na713b02f320d409c806ff0190db324f4'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#object'),
  rdflib.term.URIRef(u'target')),
 (rdflib.term.BNode('Na713b02f320d409c806ff0190db324f4'),
  rdflib.term.URIRef(u'http://www.w3.org/1999/02/22-rdf-syntax-ns#predicate'),
  rdflib.term.BNode('vcb0')),
 (rdflib.term.BNode('Na713b02f320d409c806ff0190db324f4'),
[...]

PR #506, opened 2015-08-09
Fixed bnode collision bug.

jimmccusker:
Fix for #494.

### DIFF
--- a/rdflib/compare.py
+++ b/rdflib/compare.py
@@ -204,7 +204,9 @@ class Color:
                 return unicode(x)
         if isinstance(color, Node):
             return stringify(color)
-        value = sum(map(self.hashfunc, ' '.join([stringify(x) for x in color])))
+        value = 0
+        for triple in color:
+            value += self.hashfunc(' '.join([stringify(x) for x in triple]))
         val = u"%x" % value
         self._hash_cache[color] = val
         return val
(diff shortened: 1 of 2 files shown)


rdflib/compare.py
```diff
@@ -290,7 +292,7 @@ def _initial_color(self):
 
     def _individuate(self, color, individual):
         new_color = list(color.color)
-        new_color.append((len(color.nodes)))
+        new_color.append(tuple([len(color.nodes)]))
```
joernhees: sorry for not spotting this one before :-/

could also be written as `new_color.append((len(color.nodes),))`, which is a bit faster, but maybe premature optimization:

```python
In [1]: %timeit tuple([1337])
The slowest run took 5.53 times longer than the fastest. This could mean that an intermediate result is being cached
1000000 loops, best of 3: 345 ns per loop

In [2]: %timeit (1337,)
The slowest run took 47.08 times longer than the fastest. This could mean that an intermediate result is being cached
10000000 loops, best of 3: 25.3 ns per loop
```

wwaites: On Mon, 10 Aug 2015 10:33:29 -0700, Jörn Hees <notifications@github.com> said:


    > could also be written as
    > `new_color.append((len(color.nodes),))`, which is a bit faster,
    > but maybe premature optimization:

Creating a value with the right datatype instead of creating one with
the wrong datatype and then converting it isn't optimisation, it's
just doing things correctly.

Or so I think anyways...

-w

jimmccusker: I wouldn't object to the change. I couldn't remember the syntax for a single-element tuple.

joernhees: @wwaites: well, the double bracket removal is one of the tripping stones in python, so i can somehow understand why some people prefer `(tuple([...]))` instead of `((...,))`. Without seeing it as much of a problem i still changed it in b8df01f9bc53ccd01d61953108c30afae0f4b36e ;)

PR #513, opened 2015-08-25
SPARQLStore customizable BNode handling

joernhees:
this implements the discussion in #512

- [x] deprecates `CastToTerm`, `TraverseSPARQLResultDOM` and `localName` methods
- [x] deprecates `SPARQLStore(..., bNodeAsURI)` argument
- [x] adds two `SPARQLStore(..., node_to_sparql, node_from_result)` args to customize node transformations
- [x] added a doc example how to transform `BNodes`
- [x] several code cleanups

please check

### DIFF
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -14,8 +14,8 @@ ORDERBY = 'ORDER BY'
 import re
 import collections
 import urllib2
+import warnings
 
-# import warnings
 try:
     from SPARQLWrapper import SPARQLWrapper, XML, POST, GET, URLENCODED, POSTDIRECTLY
 except ImportError:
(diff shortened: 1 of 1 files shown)


rdflib/plugins/stores/sparqlstore.py
```diff
@@ -145,6 +109,84 @@ def CastToTerm(node):
...
+        "Call to deprecated function CastToTerm, use _node_from_result.",
+        category=DeprecationWarning,
+    )
+    return _node_from_result(node)
+
+
+def _node_to_sparql(node):
+    if isinstance(node, BNode):
+        raise Exception(
+            "SPARQLStore does not support BNodes! "
```
uholzer: Maybe this Exception should now say "SPARQLStore does not support BNodes *by default*! See ..."

uholzer:
This has been quite some work it seems. Except for the nitpick in my previous comment, this looks good to me.

rdflib/plugins/stores/sparqlstore.py
joernhees: i thought so too, but as this currently raises an overly broad exception, the only thing people can do is rely on its message to actually handle it. That's why I decided against it for now and didn't change the message.

maybe we should have an issue to change all `raise Exception` to something more appropriate in 5.0.0

joernhees:
eh, it's mostly moves, deprecations and cleanups...

thanks for reviewing, i'll merge this soon if no one objects

ssssam:
This looks like it'll resolve the issue I had in #511 and #512, thanks !!

joernhees:
@ssssam you can try by installing rdflib like this: `pip install -U git+git://github.com/RDFLib/rdflib`

PR #514, opened 2015-08-25
5.0.0-dev branch (unstable)

joernhees:
meta pull request to track and build 5.0.0-dev branch.

Changes that would break backwards compatibility (http://semver.org) will be merged in here to keep track of them.

this branch is not stable (can and probably will be rebased)


niklasl:
Would it be interesting to reinvestigate using the git flow practices going forward? It's quite nice for keeping a consistent order of pending features, release patches (hotfixes) and maintenance branches for old releases (for 4.x.x fixes once 5.0.0 is out).

joernhees:
hmm, i thought we're not that far from git flow and picked the parts that make sense for us in the combination of http://semver.org, github forks, pull requests and dev / maintenance power...

I have to say that github (in comparison to gitlab) does a pretty bad job at visualizing the repo history, but with `gitk --all` i don't find what we do complicated / far from git flow. Also i see it of the way of least surprise for quick contributors. Currently we use:
- tags as "releases"
- master is our "stable develop" branch (no rebases, tests should pass) (can be seen as a simplification of git flow, where master does nothing but "jump". Also we never really had the need / manpower for hotfixes.)
- features branch from the dev branch (master) (very common in github fork/PRs)
- currently master is also 4.2.2-dev (the python version and semver pre-release of 4.2.2)
- 5.0.0-dev (the one this PR tracks) is probably the most confusing:
  - the name comes from semver pre-release / python version...
  - adding to the confusion is that i made this a PR even though it's not a feature, but just a place to see the state of CI tests
  - as can be seen in `gitk --all` this branch simply merges master + 4.2.2 incompatible things
  - master is meant to merge this branch after the 4.2.2 release (which i agree takes a lot longer than expected) with the benefit of having done some of the work for 5.0.0 already
  - if i had to classify this in git flow, i'd say it's a future develop line (which git flow lacks but http://semver.org somewhat mandates)

That said, if there's any way to simplify this, i'm the last to object.

gromgull:
I agree with @joernhees - we don't have exactly the naming of gitflow, but most of it's features. The one things we do lack is indeed right here. This is 5.0, which is master + some incompatible fixes. 

But it's not master any more, since I've added lots of stuff to master since it was branched. My plan would be to rebase it back onto master once in a while and finally when 4.2.2 is released. But then history is changed and people who made PRs onto 5.0 are doomed. 

The alternative is to apply any changes now both here and to master, then we get two commits with the same content but different SHAs.

gromgull:
I also agree that thinking will do maintenance releases for 4.X once 5.0 is out is probably wishful thinking :)

joernhees:
reverting  7bec219 (meaning actually executing the test for trig round trips) doesn't seem to work...

will outsource into separate PR against this branch and revert the revert by rebasing on the merge...

also weird: travis seems to run tests twice ([PR](https://travis-ci.org/RDFLib/rdflib/builds/195376794) and [PUSH](https://travis-ci.org/RDFLib/rdflib/builds/195376775)), PUSH worked, PR seems to have been confused as it ran a weird commit 8927eb0 (seems to be a double merge that i never pushed...)

anyhow, will hope this magically goes away on rebasing to 4204d6b4

niklasl:
Sounds fair enough. I like git-flow for distinctly documenting each step, but the outline @joernhees wrote looks good. Perhaps adding that to the [developers guide](http://rdflib.readthedocs.io/en/stable/developers.html) would be nice?

As for maintenance releases, @gromgull is probably right. It would be good to be somewhat prepared for it anyway though, since performance or security related bugs might crop up. Given that 5.0.0 might break things for someone depending on intricate parts of the internals, the need for a maintenance release could come up. Not likely, but possible.

joernhees:
hmm, seems travis is still confused by this, above changes work fine.

gromgull:
@joernhees yeah - I don't know why it's confused here. I got an email saying the build was fixed.

joernhees:
i think the job running on the release branch 5.0.0-dev (PUSH) is fine, but the one running on this PR might be running with the travis config from master? I think i fixed travis config in a PR before though... weird.

PR #519, opened 2015-08-27
Transition from 2to3 to use of six.py to be merged in 5.0.0-dev

joernhees:
related to and closes #438 but compares to branch 5.0.0-dev instead of master

The old porting approach in #374 to six became quite messy because it started by deactivating 2to3 and then tried to fix all errors. Instead let's try to work through a diff of py2 with py3 after 2to3 and deactivate 2to3 one by one whenever a file is done. In the process the tests are meant to always pass.

Branch is unstable!


gromgull:
Looking at some of the files you already did Jörn, I guess we are happy leaving in `u"blah"` strings. They are allowed in some later py3, so I would just not support any py3 where they are not ok :)

joernhees:
awesome!

gromgull:
Jörn, this is pretty much done. How to proceed? 

I would maybe squash merge into the 5.0.0 branch, so it's all one commit.

joernhees:
hmm, i'm against squashing, as a single huge commit is always harder to merge than individual ones on upcoming conflicts, etc... except for a lying cleaner and shorter history, what's the benefit?

gromgull:
My plan would have been to rebase this back onto wherever it came from, then you have to deal with the merge for each small commit now. Then once setup to merge cleanly you could squash. Then there is no downside to having a clean commit :) 

BUT RDFLib's commit history is hardly pristine, and I don't feel very strongly either way.

joernhees:
yes, i just rebased this on top of 5.0.0-dev (which i did from time to time)...

Squashing it all into one, however destroys info (such as the commit messages and who did what). I'm not very attached to those, but for example via blame it helps you to understand why a line was changed via its commit message (see 9ab4078). Also having rebased this a couple of times already, i found it incredibly helpful to have rebase stop at a small commit in case of a conflict... just immediately let's you focus to 2-3 files instead of 100...

joernhees:
so, this isn't totally done yet: there are some leftover files that were introduced since starting this effort 661a28938ffac1ed9898fd95dc48df2540e30de3 . Last thing to go is the special treatment of py3 in `setup.py`. Already working on this...

gromgull:
So I've converted the rest and removed all special py3 handling from setup et al. 

However, py3compat has grown ridicolously - I think having this: https://github.com/RDFLib/rdflib/blob/six_2to3/rdflib/py3compat.py#L18-L52

is weird, when we use six anyway. I'll do a global search replace and import these straight from six in every place. 

That way we don't import everything but the kitchen sink for every rdflib module.

setup.py
```diff
@@ -1,324 +1,16 @@
...
-# its been automatically converted in the setup process.
-def find_version(filename):
-    _version_re = re.compile(r'__version__ = "(.*)"')
-    for line in open(filename):
-        version_match = _version_re.match(line)
-        if version_match:
-            return version_match.group(1)
 
-version = find_version('rdflib/__init__.py')
+from rdflib import __version__ as version
```
joernhees: this is bad, if any of the requirements listed above isn't present it will fail... at build time we don't want to `import rdflib`

the reason it doesn't fail in CI is cause we do a `pip install -r requirements.txt` in travis... maybe we shouldn't

gromgull: I thought I remembered it being bad for some reason - but figure I'll let travis decide :)

.travis.yml
```diff
@@ -16,7 +16,6 @@ python:
     - 3.6
 
 before_install:
-    - pip install -U setuptools pip  # seems travis comes with a too old setuptools for html5lib
```
gromgull: bugger, this wasn't mean to go in this commit.

gromgull:
Assuming the tests pass in 3 min, this is all ready to be merge as far as I am concerned!

joernhees:
tests pass, i went over the last couple of commits and quickly skimmed the whole (as good as possible with 156 changed files)... in the interest of finishing this ~ 1.5 years process, i'm just going ahead and merging this... other bugs and cleanups can be fixed later

requirements.txt
```diff
@@ -4,3 +4,4 @@ isodate
 pyparsing
 six
 SPARQLWrapper
+doctest-ignore-unicode
```
gromgull: @joernhees after I went to bed I realise that this should have been in `test_requires` in setup.py and not here :D 

And then we could run the tests with `python setup.py nosetests` on travis. 

But we can fix in master!

Issue #492, opened 2015-07-03
TypeError: unorderable types: SequencePath() < SequencePath()


jpotoniec:
Hi,
I believe there is some error in the SPARQL parser during reordering triples with property paths
Minimal working example leading to the error:

```python
#!/usr/bin/env python3

import rdflib

query = '''
prefix owl: <http://www.w3.org/2002/07/owl#> 
prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
select ?x
where
{   
    ?x rdf:rest/rdf:first _:6.
    ?x rdf:rest/rdf:first _:5.
}
'''
print(rdflib.__version__)
g = rdflib.Graph()
result = g.query(query)
```

I am using rdflib installed with pip as described in README. The output of the above script is:
```
4.2.0
Traceback (most recent call last):
  File "./query.py", line 17, in <module>
    result = g.query(query)
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/graph.py", line 1080, in query
    query_object, initBindings, initNs, **kwargs))
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/plugins/sparql/processor.py", line 73, in query
    query = translateQuery(parsetree, base, initNs)
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/plugins/sparql/algebra.py", line 744, in translateQuery
    P, PV = translate(q[1])
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/plugins/sparql/algebra.py", line 533, in translate
    M = translateGroupGraphPattern(q.where)
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/plugins/sparql/algebra.py", line 278, in translateGroupGraphPattern
    g[-1]["triples"] += triples(p.triples)
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/plugins/sparql/algebra.py", line 131, in triples
    for x in range(0, len(l), 3))
  File "/home/smaug/tmp/venv/lib/python3.4/site-packages/rdflib/plugins/sparql/algebra.py", line 113, in reorderTriples
    1], varsknown, varscount), x[1]) for x in l[i:])
TypeError: unorderable types: SequencePath() < SequencePath()
```

[...]

PR #525, opened 2015-09-23
fix sparql path order on python3

joernhees:
see #492


gromgull:
this would have been an excellent opportunity to implement __eq__ as well, it will work now, but things will be equal on `id(object)`, not on two Paths of the same type with the same arguments :)

joernhees:
hehe, i admit, i didn't think about this for a long time... i considered using `functools.total_orderings` on the `Path` class, but then thought that each of the `repr` actually return the internal's `str` not their `repr`. i decided to pretend as little as possible that we do any meaningful comparison of `Path` objects and just solve one problem at a time... we need `<` for py3 sorting, nothing more...

also otherwise:

```python
In [1]: from rdflib.paths import Path, InvPath

In [2]: repr(Path())
Out[2]: '<rdflib.paths.Path instance at 0x10db2ed88>'

In [3]: repr(InvPath(Path()))
Out[3]: 'Path(~<rdflib.paths.Path instance at 0x10db93440>)'

In [4]: repr(InvPath(InvPath(Path())))
Out[4]: 'Path(~Path(~<rdflib.paths.Path instance at 0x10db935f0>))'
```

joernhees:
```python
from functools import total_ordering
@total_ordering
class Path(object):
    def eval(self, graph, subj=None, obj=None):
        raise NotImplementedError()

    def __eq__(self, other):
        if not isinstance(other, Path):
            raise TypeError('incomparable types: %s() == %s()' % (
                repr(self), repr(other)))
        return repr(self) == repr(other)

    def __lt__(self, other):
        if not isinstance(other, Path):
            raise TypeError('unorderable types: %s() < %s()' % (
                repr(self), repr(other)))
        return repr(self) < repr(other)
```
?

PR #526, opened 2015-09-23
adding python 3.5 support

joernhees:


rdflib/plugins/sparql/operators.py
```diff
@@ -260,7 +261,10 @@ def group(self, n):
         cFlag = reduce(pyop.or_, [flagMap.get(f, 0) for f in flags])
 
         # @@FIXME@@ either datatype OR lang, NOT both
-    return Literal(re.sub(unicode(pattern), _r, text, cFlag),
+
+    compat_r = unicode(replacement) if sys.version_info[:2] >= (3, 5) else _r
+
+    return Literal(re.sub(unicode(pattern), compat_r, text, cFlag),
```
joernhees: @gjhiggins could you review this? i'm not sure this is the best / a good solution

gromgull: it's not pretty - and should maybe somehow be in compat.py - but if it works, SHIPIT!

but maybe add a comment? :)

joernhees: ;) alright

PR #531, opened 2015-10-19
Exposed the timeout of SPARQLWrapper

mwatts15:
I want to set a timeout on the connection in SPARQLStore. I can cause `sendall()` in the underlying socket object to timeout through SPARQLWrapper, but SPARQLUpdateStore resets the timeout to 0. This change keeps the timeout in-between updates.

Background:
I have been doing some experiments with sending SPARQL updates of increasing size and hit a limit where the underlying socket in the SPARQLUpdateStore is blocking in `sendall()`. I check with `netstat` and see the state of the connection is "CLOSE_WAIT", so the remote endpoint closed the connection. It seems like `sendall()` should return here; not sure.

### DIFF
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -319,9 +319,11 @@ class SPARQLStore(NSSPARQLWrapper, Store):
                 % (" ".join("?" + str(x) for x in v),
                    " ".join(initBindings[x].n3() for x in v))
 
+        timeout = self.timeout
         self.resetQuery()
         if self._is_contextual(queryGraph):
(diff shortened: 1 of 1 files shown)


joernhees:
i like the idea of improving the configurability of the SPARQLUpdateStore connection, but probably not restricted to the timeout parameter only.

also as you can see, your change breaks current tests if no timeout is set.

mwatts15:
> i like the idea of improving the configurability of the SPARQLUpdateStore connection, but probably not restricted to the timeout parameter only.

What do you have in mind?

wikier:
:+1:

joernhees:
eh, well can't hurt to have this at least until the full config can be restored for each request...

PR #534, opened 2015-10-29
Avoid verbose build logging

lrowe:
For systems that install many packages, very verbose logging for a particular package is a pain when tracking down build errors from other packages.

This removes almost 5000 lines of logs from a standard python3 build.


joernhees:
:+1:

Issue #535, opened 2015-10-29
nquads parser creates one context per triple in the default graph


pchampin:
In nquads, the 4th component (graph id) of each line can be omitted, meaning that the triple belongs to the default graph.

Currently, the nquads parser puts every such triple in a different context, identified by a fresh BNode, which is clearly wrong.

E.g.:
```python
from rdflib import ConjunctiveGraph, Dataset
from rdflib.parser import StringInputSource

ds = ConjunctiveGraph()

src = StringInputSource("""
<http://example.org/s1> <http://example.org/p1> <http://example.org/o1> .
<http://example.org/s2> <http://example.org/p2> <http://example.org/o2> .
<http://example.org/s3> <http://example.org/p3> <http://example.org/o3> <http://example.org/g3> .
""")

ds.parse(src, format="nquads")

assert len(ds) == 3, len(g)
assert len(list(ds.contexts())) == 2, len(list(ds.contexts()))
```

niklasl:
Closely related to #436. We should coordinate solutions.

pchampin:
You are right, those issues related.
The reason I opened a new one is that, regardless of the chosen solution for handling the default graph,
it is clear that those triples should not end up in *different* contexts.

So those are two different problems, fixing one requires fixing the other. Indeed, we should coordinate.

niklasl:
Yes, I agree, this issue for the nquads parser is distinct.

(I'll also look into coordinating a solution for RDFLib/rdflib-jsonld#34 with this. I wrote down #436 when I originally hit that. But attempting to resolve all parsing and serializing of datasets at once might have been too much.)

joernhees:
hmm, one question wrt. the version in which this will be fixed, e.g. by merging #536 :

as we use http://semver.org, is this a bug fix only (so 4.2.2) or an actual backwards incompatible change (5.x.x)?

pchampin:
The current behaviour of the nquads parser is wrong, to a point that no one
could reasonably rely on it.
So I would go for a bug fix.

joernhees:
fixed in #536

PR #536, opened 2015-10-29
fixed issue #535

pchampin:
I had to make a choice in how this issue was fixed:
I decided that the "default graph" for the parser was the one identified by the publicID passed to the ``parse`` method, which seems to make sense.

### DIFF
--- a/test/test_issue535.py
+++ b/test/test_issue535.py
@@ -4,14 +4,15 @@ from rdflib.parser import StringInputSource
 def test_nquads_default_graph():
     ds = ConjunctiveGraph()
 
-    src = StringInputSource("""
+    data = """
     <http://example.org/s1> <http://example.org/p1> <http://example.org/o1> .
     <http://example.org/s2> <http://example.org/p2> <http://example.org/o2> .
     <http://example.org/s3> <http://example.org/p3> <http://example.org/o3> <http://example.org/g3> .
-    """)
+    """
 
     publicID = URIRef("http://example.org/g0")
-    ds.parse(src, format="nquads", publicID=publicID)
+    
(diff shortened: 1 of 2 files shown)


joernhees:
thanks for the PR, seems the failing py3 build is unrelated to this, but rather related to SPARQLWrapper's failing dependency on urlgrabber in py3...

joernhees:
re-running tests after SPARQLWrapper 1.7.1 release...

test/test_issue535.py
```diff
@@ -0,0 +1,18 @@
+from rdflib import ConjunctiveGraph, URIRef
+from rdflib.parser import StringInputSource
+
+def test_nquads_default_graph():
+    ds = ConjunctiveGraph()
+
+    src = StringInputSource("""
```
joernhees: can you try changing this to `data = """...`

test/test_issue535.py
```diff
@@ -0,0 +1,18 @@
...
+    <http://example.org/s1> <http://example.org/p1> <http://example.org/o1> .
+    <http://example.org/s2> <http://example.org/p2> <http://example.org/o2> .
+    <http://example.org/s3> <http://example.org/p3> <http://example.org/o3> <http://example.org/g3> .
+    """)
+
+    publicID = URIRef("http://example.org/g0")
+    ds.parse(src, format="nquads", publicID=publicID)
```
joernhees: and this to `ds.parse(data=data, format="nquads", publicID=publicID)`

joernhees:
test fails on py3, can you try passing the string via `data` arg instead (see inline comments)?

PR #537, opened 2015-11-02
Test-suite and bug-fixes for AuditableStore

pchampin:
I was interested in using AuditableStore, but encountered some bugs.
Then I realized that there was no unit-test for that store, so I wrote some, and I fixed the class until all tests passed.

### DIFF
--- a/test/test_auditable.py
+++ b/test/test_auditable.py
@@ -1,5 +1,6 @@
 # -*- coding=utf8 -*-
 from __future__ import unicode_literals
+from nose.tools import assert_items_equal
 import unittest
 
 from rdflib import Graph, Namespace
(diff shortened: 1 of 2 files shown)


test/test_auditable.py
```diff
@@ -0,0 +1,296 @@
...
+        self.g = Graph()
+        self.g.add((EX.s0, EX.p0, EX.o0))
+        self.g.add((EX.s0, EX.p0, EX.o0bis))
+
+        self.t = Graph(AuditableStore(self.g.store),
+                       self.g.identifier)
+
+    def test_add_commit(self):
+        self.t.add((EX.s1, EX.p1, EX.o1))
+        self.assertItemsEqual(self.t, [
```
joernhees: py3 tests fail cause it seems `assertItemsEqual` is deprecated since 3.2

maybe just rewrite as nose test?

joernhees:
seems that assert_items_equal is even worse (fails on everything but 2.7)... i suggest writing a short helper method in the test which just bases on assert...

gromgull:
I am not sure how much we care about a pretty commit history, but squashing this one to ONE commit would look nice :)

PR #547, opened 2015-11-20
Small but nice SPARQL Optimisation fix

gromgull:
This reorders the triples in a BGP while evaluating, so that the "most
bound" triples are evaluated first. The algebra transformation also does
this, but "statically", i.e. without looking at the data.

At runtime, we may have bindings from other places
(BIND/VALUES/initBindings) which can massively improve query time.

In a BerkelyDB store with ~250k triples, I had a query with two triples
patterns that retrieved 38 out of 25000 possible matching triples with a
particular property. Adding this fix brought query time from 8s to
800ms.

### DIFF
--- a/rdflib/plugins/sparql/evaluate.py
+++ b/rdflib/plugins/sparql/evaluate.py
@@ -214,7 +214,11 @@ def evalPart(ctx, part):
             pass  # the given custome-function did not handle this part
 
     if part.name == 'BGP':
-        return evalBGP(ctx, part.triples)  # NOTE pass part.triples, not part!
+        # Reorder triples patterns by number of bound nodes in the current ctx
+        # Do patterns with more bound nodes first
+        triples = sorted(part.triples, key=lambda t: len([n for n in t if ctx[n] is None]))
+
+        return evalBGP(ctx, triples)
     elif part.name == 'Filter':
         return evalFilter(ctx, part)
     elif part.name == 'Join':


gromgull:
As far as I can see the single failing test `test_sparqlupdatestore` has nothing to do with what I edited :(

gromgull:
It breaks while trying to parse the reply to the query produced from this: https://github.com/RDFLib/rdflib/blob/master/test/test_sparqlupdatestore.py#L237

I don't know where it goes wrong - on insert or query.

rdflib/plugins/sparql/evaluate.py
```diff
@@ -214,7 +214,10 @@ def evalPart(ctx, part):
             pass  # the given custome-function did not handle this part
 
     if part.name == 'BGP':
-        return evalBGP(ctx, part.triples)  # NOTE pass part.triples, not part!
+        # Reorder triples patterns based on 'bindedness' based on current ctx
+        triples = sorted(part.triples, key=lambda t: len([n for n in t if ctx[n] is None]))
```
joernhees: i have problems understanding this... could you maybe expand the comment a bit more?

my guess is `n` is a node in triple `t`, so if `ctx[n] is None` then `n` is unbound, so you select those triples first which have least unbound nodes?

joernhees:
btw: the failing test seems related to the recent SPARQLWrapper updates, more in #550

rdflib/plugins/sparql/evaluate.py
gromgull: Your guess is spot on. I am not sure when this kicks in, for only "static" triples (with not optional/bindings/values/anything) - the code here: https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/sparql/algebra.py#L84 will already have sorted them correctly. 

This code helps when some other source of bindings has given us more bindings since then - in my base it was using initBindings.

gromgull:
I made the comment a bit clearer - if the test-fail is unrelated we can merge :)

PR #550, opened 2015-11-22
master build fails (guess: dependency updates)

joernhees:
this is an empty commit PR (for now) to trigger an additional travis run as it seems that recent builds fail...

see #547 and #548 for example.

### DIFF
--- a/requirements.py2.txt
+++ b/requirements.py2.txt
@@ -1,4 +1,5 @@
 flake8
 isodate
 pyparsing<=1.5.7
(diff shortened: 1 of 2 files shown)


joernhees:
CURRENT STATE JUST FOR TESTING, DO NOT MERGE!

would effectively restrict SPARQLWrapper to 1.7.1 on all systems that install rdflib!

joernhees:
thanks to @wikier for fixing this in keepalive, closing this...

PR #551, opened 2015-11-22
Made ClosedNamespace (and _RDFNamespace) inherit from Namespace

### LINKED ISSUE #542: Consistency Namespace vs. ClosedNamespace
As I understand rdflib.Namespace and ClosedNamespace, they are both meant as auxiliaries to simplify the creation of URIRefs in Python code, e.g. `RDF.type` or `OWL.Class`. In addition, ClosedNamespace provides an additional vocabulary check, e.g., `RDF.spam` will raise an exception.

However, sometimes other uses of Namespace/ClosedNamespace can be useful, e.g., string comparison. At this stage, it is quite inconvenient that Namespace is derived from unicode, whereas ClosedNamespace is derived from the standard object, and the actual namespace string is an instance attribute (uri). In the interest of consistency, I propose that ClosedNamespace should also be derived from unicode (or even from Namespace).

Regards

gromgull:
This fixes #542

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -95,11 +95,6 @@ class Namespace(unicode):
             rt = unicode.__new__(cls, value, 'utf-8')
         return rt
 
-
-    @property
-    def title(self):
-        return URIRef(self + 'title')
-
     def term(self, name):
         # need to handle slices explicitly because of __getitem__ override
         return URIRef(self + (name if isinstance(name, basestring) else ''))
@@ -148,26 +143,33 @@ class URIPattern(unicode):
 
 
 
-class ClosedNamespace(object):
+class ClosedNamespace(Namespace):
     """
     A namespace with a closed list of members
 
     Trying to create terms not listen is an error
     """
 
-    def __init__(self, uri, terms):
-        self.uri = uri
-        self.__uris = {}
+    def __new__(cls, uri, terms):
+        rt = Namespace.__new__(cls, uri)
+
+        rt.__uris = {}
         for t in terms:
-            self.__uris[t] = URIRef(self.uri + t)
+            rt.__uris[t] = URIRef(rt + t)
+
+        return rt
 
     def term(self, name):
         uri = self.__uris.get(name)
         if uri is None:
             raise Exception(
-                "term '%s' not in namespace '%s'" % (name, self.uri))
+                "term '%s' not in namespace '%s'" % (name, self))
         else:
             return uri
+
+    @property
+    def uri(self): # support legacy code from before ClosedNamespace extended unicode
+        return self
 
     def __getitem__(self, key, default=None):
         return self.term(key)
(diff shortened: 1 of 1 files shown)


joernhees:
ERROR: Failure: RuntimeError (maximum recursion depth exceeded in cmp)

gromgull:
Now only the test broken by SPARQLWrapper unicode thingy breaks :)

rdflib/namespace.py
```diff
@@ -95,11 +95,6 @@ def __new__(cls, value):
             rt = unicode.__new__(cls, value, 'utf-8')
         return rt
 
-
-    @property
-    def title(self):
-        return URIRef(self + 'title')
-
```
joernhees: what was that for anyhow?

gromgull: No idea, nothing broke when I removed it :)

niklasl: I think it was to make this:

    >>> from rdflib.namespace import DCTERMS
    >>> DCTERMS.title

result in:

    rdflib.term.URIRef(u'http://purl.org/dc/terms/title')

instead of:

    u'Http://Purl.Org/Dc/Terms/'

joernhees:
```python
In [1]: from rdflib.namespace import DCTERMS
INFO:rdflib:RDFLib Version: 4.2.2-dev

In [2]: DCTERMS.title
Out[2]: <function title>

In [3]: DCTERMS.title()
Out[3]: u'Http://Purl.Org/Dc/Terms/'

In [4]: DCTERMS.something
Out[4]: rdflib.term.URIRef(u'http://purl.org/dc/terms/something')
```

wtf?

rdflib/namespace.py
niklasl: So it should be added back (along with a test) since existing code will break otherwise.

joernhees: will do... too quick with that one...

gromgull: I agree with should restore it, but this is a problem for ask methods on Unicode objects, lower, islower, index etc

niklasl: Note: the title method won't be called (of course), but the value will be `<built-in method upper of Namespace object at ...>` instead of (e.g.) dcterms:title.

(This is a general gotcha with the Namespace utility colliding with unicode method names of course, and I guess it would be more ideal if it used `__getattribute__` to never return bound methods.)

joernhees: it's to override this:
```python
In [7]: unicode('foo').title()
Out[7]: u'Foo'
```

joernhees: maybe we should just not subclass unicode?

joernhees: for now added it back + test for `DCTERMS.title`... in master already

joernhees:
patched with ae1f6397fe47be7bd6912dae8b08c4007eb9d2ac to allow `DCTERMS.title`

rdflib/namespace.py
niklasl: Great, thanks! I jotted down the other thoughts in #556.

gromgull:
"Fixing" this actually broke a bit of random code for me - I used to rely on `ClosedNamespace.title` not returning a method. We should not ship this in a non-major release :)

joernhees:
@gromgull i don't understand, it doesn't for me:
```python
In [1]: import rdflib
INFO:rdflib:RDFLib Version: 4.2.2-dev

In [2]: rdflib.namespace.DCTERMS
Out[2]: Namespace(u'http://purl.org/dc/terms/')

In [3]: rdflib.namespace.DCTERMS.title
Out[3]: rdflib.term.URIRef(u'http://purl.org/dc/terms/title')
```

gromgull:
The problem comes from ClosedNamespace:

```In [117]: c=ClosedNamespace('http://example.org/','count')

In [118]: c.count
Out[118]: <function count>

In [119]: :(```

previously this worked, since ClosedNamespace did not inherit from unicode

joernhees:
ok, i'll roll this back for the 4.2.2 release and move it to 5.0.0

PR #552, opened 2015-11-24
let paths be comparable against all nodes. Fixes #545

### LINKED ISSUE #545: preparedQuery error (specific in Python3)
When I write this query, I got error in Python3.5
Query program is below.

```python:Query.py
from rdflib import Namespace
from rdflib.plugins import sparql
from rdflib.namespace import RDFS, OWL, DC, SKOS
query = sparql.prepareQuery(
        """
            SELECT DISTINCT ?property ?parent
            WHERE{
                ?property a owl:DeprecatedProperty .
                ?property dc:relation ?relation .
                ?property rdfs:subPropertyOf ?parent .
                ?property rdfs:label | skos:altLabel ?label . 
            }
        """,
        initNs = {"rdfs":RDFS, "owl":OWL, "dc":DC, "skos":SKOS})
```
Error log is below.
```log:Error.log
TypeError                                 Traceback (most recent call last)
<ipython-input-59-83c306286d37> in <module>()
     12             }
     13         """,
---> 14         initNs = {"rdfs":RDFS, "owl":OWL, "dc":DC, "skos":SKOS})

/home/UserName/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/rdflib/plugins/sparql/processor.py in prepareQuery(queryString, initNs, base)
     23     Parse and translate a SPARQL Query
     24     """
---> 25     return translateQuery(parseQuery(queryString), base, initNs)
     26 
     27 

/home/UserName/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py in translateQuery(q, base, initNs)
    742         q[1], visitPost=functools.partial(translatePName, prologue=prologue))
    743 
--> 744     P, PV = translate(q[1])
    745     datasetClause = q[1].datasetClause
    746     if q[1].name == 'ConstructQuery':

/home/UserName/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py in translate(q)
    531 
    532     # all query types have a where part
--> 533     M = translateGroupGraphPattern(q.where)
    534 
    535     aggregate = False

/home/UserName/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py in translateGroupGraphPattern(graphPattern)
    276             if not (g and g[-1].name == 'BGP'):
    277                 g.append(BGP())
--> 278             g[-1]["triples"] += triples(p.triples)
    279         else:
    280             g.append(p)

/home/UserName/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py in triples(l)
    129         raise Exception('these aint triples')
    130     return reorderTriples((l[x], l[x + 1], l[x + 2])
--> 131                           for x in range(0, len(l), 3))
    132 
    133 

/home/UserName/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py in reorderTriples(l)
    111     while i < len(l):
    112         l[i:] = sorted((_knownTerms(x[
--> 113                        1], varsknown, varscount), x[1]) for x in l[i:])
    114         t = l[i][0][0]  # top block has this many terms bound
    115         j = 0

TypeError: unorderable types: AlternativePath() < URIRef()
```

However, this preparedQuery has no problem in Python 2.7.10 , and This error is solved if query is modified by grouping triple pattern.
Modified query is below .
```python
from rdflib import Namespace
from rdflib.plugins import sparql
from rdflib.namespace import RDFS, OWL, DC, SKOS
query = sparql.prepareQuery(
        """
            SELECT DISTINCT ?property ?parent
            WHERE{
                ?property a owl:DeprecatedProperty .
                ?property dc:relation ?relation .
                ?property rdfs:subPropertyOf ?parent .
                {?property rdfs:label | skos:altLabel ?label .}
            }
        """,
        initNs = {"rdfs":RDFS, "owl":OWL, "dc":DC, "skos":SKOS})
```

It looks like a bug.
regard.

gromgull:


gromgull:
Hard to say what happens with master failing ... i'll wait for it to be fixed and rebase :)

joernhees:
:+1: sorry for that... #550

joernhees:
py3.4:
```
TypeError: unorderable types: URIRef() < AlternativePath()
```

i guess this is a symmetry problem... if the first operand is not a `Path` it breaks... should we maybe move the whole comparison code to `Node` or `Identifier` and make `Path` a subclass?

also, shouldn't https://github.com/RDFLib/rdflib/blob/master/rdflib/term.py#L171 actually `raise NotImplemented()` instead of `return`ing it?

gromgull:
Don't ask me why, but returning is correct: https://docs.python.org/2/library/constants.html#NotImplemented

I'll look into py3, but not today :)

joernhees:
ah i didn't know it's actually a keyword and confused it with https://docs.python.org/2/library/exceptions.html#exceptions.NotImplementedError which would need to be raised...

gromgull:
The missing part was adding `@functools.total_ordering` - that fixes py3.4, but breaks 2.6 where it doesn't exist :( 

When can we kill 2.6 support? :)

joernhees:
i guess with rdflib 5.0.0 but for now i just fixed it by also defining the other comps...

one other thing i noticed looking at this again though: if possible `repr` is meant to return something that you can paste into a python interpreter re-creating the object... do you see an easy way to change the `__repr__` methods of the current `Path` objects in that way? (i didn't)

PR #557, opened 2015-11-30
Test-suite and bug-fixes for AuditableStore (cleaned history)

pchampin:
This is a squashed version of #537 .
@gromgull I only squashed those commits fixing the test-suite, which indeed didn't look good. However, I kepts separate the commit creating the tests suite, and the different steps in fixing AuditableStore so as to pass the suite. I think it is better that way, but if you prefer, I can also squash all of them into a single commit.


gromgull:
Looks good! Thanks!

PR #562, opened 2015-12-11
DOC: unamed -> unnamed

yarikoptic:
could well be that I am simply missing some basic terminology in RDF which has `unamed` (uniquely named?) of some kind? ;)


gromgull:
that was not intentional - thanks!

Issue #563, opened 2015-12-23
aggregate operators COUNT and SAMPLE should ignore NULL values


pchampin:
Consider the following query:
```
  SELECT ?x (COUNT(?y) as ?ys) (COUNT(?z) as ?zs) WHERE {
    VALUES (?x ?y ?z) {
      (2 6 UNDEF)
      (2 UNDEF 10)
      (3 UNDEF 15)
      (3 9 UNDEF)
    }
  }
  GROUP BY ?x
```
it should return the following tuples:
```
  2 1 1
  3 1 1
```
as, per [the specification](http://www.w3.org/TR/sparql11-query/#defn_aggCount):

    [COUNT] counts the number of times a given expression has a bound, and non-error value

But instead it returns the following tuples:
```
  2 2 2
  3 2 2
```

There is a similar problem with the SAMPLE operator.
I would expect that:
```
  SELECT ?x (SAMPLE(?y) as ?ys) (SAMPLE(?z) as ?zs) WHERE {
    VALUES (?x ?y ?z) {
      (2 6 UNDEF)
      (2 UNDEF 10)
      (3 UNDEF 15)
      (3 9 UNDEF)
    }
  }
  GROUP BY ?x
```
return the following tuples:
```
  2 6 10
  3 9 15
```
but instead I get
```
  2 6 _
  3 _ 15
```
(where ``_`` means NULL).

Here the specification is not as explicit as how to handle NULL values,
but both Virtuoso and Corese give me the expected result, so there seem to be a consensus on the fact that SAMPLE should not return NULL values.

(in fact, when one sampled column contains only NULL values, both Virtuoso and Corese populate it with an artificial 0 value).

joernhees:
good catch

PR #564, opened 2015-12-23
fixing issue #563

pchampin:


rdflib/plugins/sparql/aggregates.py
```diff
@@ -98,10 +100,14 @@ def agg_Count(a, group, bindings):
...
-        bindings[a.res] = _eval(a.vars, iter(group).next())
-    except StopIteration:
-        pass  # no res
+    for ctx in group:
+        val = _eval(a.vars, ctx)
+        if type(val) is not NotBoundError:
+            bindings[a.res] = val
+            break
+    else:
+        bindings[a.res] = Literal(0)
```
joernhees: Returning a `Literal(0)` where we can't find anything seems wrong to me. Could we return `None` instead?

http://lod.openlinksw.com/sparql?default-graph-uri=&qtxt=SELECT+%3Fx+%28SAMPLE%28%3Fy%29+as+%3Fys%29+%28SAMPLE%28%3Fz%29+as+%3Fzs%29+WHERE+{%0D%0A++++VALUES+%28%3Fx+%3Fy+%3Fz%29+{%0D%0A+++++%282+6+UNDEF%29%0D%0A+++++%282+UNDEF+10%29%0D%0A+++++%283+UNDEF+15%29%0D%0A+++++%283+9+UNDEF%29%0D%0A+++++%285+UNDEF+25%29%0D%0A++++}%0D%0A++}%0D%0A++GROUP+BY+%3Fx&should-sponge=&format=text%2Fhtml&CXML_redir_for_subjs=121&CXML_redir_for_hrefs=&timeout=30000&debug=on

joernhees:
@gromgull except for the `Literal(0)` this seems legit to me, any thoughts?

rdflib/plugins/sparql/aggregates.py
gromgull: Sample of nothing should be an error: 

http://www.w3.org/TR/sparql11-query/#defn_aggSample

> Sample({}) = error

joernhees: yes, the problem is that there are various meanings of "error"... IMO they don't mean to remove the whole result row / cancel the whole query, but just the one binding within the row, as shown here:
http://www.w3.org/TR/sparql11-query/#aggregateExample2

So if we encounter such an error, i think it's correct to somehow return the row with a placeholder indicating an unbound variable (which i thought `None` was), rather than removing the result row or simply cancelling the whole query.

At least that's what Virtuoso does in the sample query above, so it must be true ;)

pchampin:
I'm terribly sorry about this mess with ``Literal(0)``. I totally agree that this seems wrong, I did it because I couldn't find the part of the spec where it was specified (thx @gromgull for pointing it out), and I *thought* that Virtuoso and Corese both did that. Obviously, they do not, they both return NULL instead, I still don't know how I messed up to get this impression :-/

PR #567, opened 2015-12-29
fix sparql count & sample for unbound

### LINKED ISSUE #563: aggregate operators COUNT and SAMPLE should ignore NULL values

joernhees:
fixes #563 and builds on top of #564, changing it to return `None` in case `SAMPLE(?var)` can't find a binding for `?var`.

### DIFF
--- a/rdflib/plugins/sparql/aggregates.py
+++ b/rdflib/plugins/sparql/aggregates.py
@@ -89,7 +89,7 @@ def agg_Count(a, group, bindings):
         try:
             if a.vars != '*':
                 val = _eval(a.vars, x)
-                if type(val) is NotBoundError:
+                if isinstance(val, NotBoundError):
                     continue
             c += 1
         except:
@@ -102,7 +102,7 @@ def agg_Count(a, group, bindings):
 def agg_Sample(a, group, bindings):
     for ctx in group:
         val = _eval(a.vars, ctx)
-        if type(val) is not NotBoundError:
+        if not isinstance(val, NotBoundError):
             bindings[a.res] = val
             break
 
--- a/rdflib/plugins/sparql/aggregates.py
+++ b/rdflib/plugins/sparql/aggregates.py
@@ -105,9 +105,6 @@ def agg_Sample(a, group, bindings):
         if type(val) is not NotBoundError:
             bindings[a.res] = val
             break
-    else:
-        bindings[a.res] = Literal(0)
-
 
 
 def agg_GroupConcat(a, group, bindings):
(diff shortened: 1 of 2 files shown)


joernhees:
@gromgull hehe, well just out of curiosity i asked jena:
```bash
$ touch empty.n3
$ sparql --data empty.n3 --query <(echo 'SELECT ?x (SAMPLE(?y) as ?ys) (SAMPLE(?z) as ?zs) WHERE {
  VALUES (?x ?y ?z) {
    (2 6 UNDEF)
    (2 UNDEF 10)
    (3 UNDEF 15)
    (3 9 UNDEF)
    (5 UNDEF 25)
  }
}
GROUP BY ?x
')
---------------
| x | ys | zs |
===============
| 5 |    | 25 |
| 3 |    |    |
| 2 |    |    |
---------------
```

joernhees:
also see https://lists.w3.org/Archives/Public/public-sparql-dev/2016JanMar/0005.html

joernhees:
sooo... according to @afs in https://lists.w3.org/Archives/Public/public-sparql-dev/2016JanMar/0006.html both our `sample` behaviors are correct, but the new one is a bit more useful. the old behavior wrt. `count` definitely was a bug.

PR #568, opened 2015-12-29
SPARQL NotBoundError should be raised instead of returned

joernhees:
working on #567 i noticed that `NotBoundError` (which is an `Exception`) is returned rather than raised in [evalutils.py#L73](https://github.com/joernhees/rdflib/blob/be16dfea6e7e41a120d353b3f8ff71614103bc9c/rdflib/plugins/sparql/evalutils.py#L73).

I tried to change it, but that seems to cause other trouble (as travis will show), which i don't really seem to understand at the moment...

### DIFF
--- a/rdflib/plugins/sparql/aggregates.py
+++ b/rdflib/plugins/sparql/aggregates.py
@@ -89,7 +89,7 @@ def agg_Count(a, group, bindings):
         try:
             if a.vars != '*':
                 val = _eval(a.vars, x)
-                if type(val) is NotBoundError:
+                if isinstance(val, NotBoundError):
                     continue
             c += 1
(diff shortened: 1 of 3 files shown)


gromgull:
>SPARQL NotBoundError should be raised instead of returned #568
 
Why? :)

joernhees:
hah, cause it's raised in [parserutils.py#L80](https://github.com/joernhees/rdflib/blob/6200934ace658ca05de693956b37623ce3a545eb/rdflib/plugins/sparql/parserutils.py#L80) ;)

joernhees:
well, can still be fixed in a later cleanup... closing this for now

PR #573, opened 2016-01-10
Second proposed fix for the broken top_level.txt

### LINKED ISSUE #571: Broken top_level.txt file in distribution
This is what the [setuptools documentation](https://pythonhosted.org/setuptools/formats.html#top-level-txt-conflict-management-metadata) says about the `top_level.txt`file:

> This file is a list of the top-level module or package names provided by the project, one Python identifier per line.
>
> Subpackages are not included; a project containing both a ``foo.bar`` and a ``foo.baz`` would include only one line, ``foo``, in its ``top_level.txt``.

The file distributed with rdflib can be accessed with:

    >>> import pkg_resources
    >>> provider = pkg_resources.get_distribution('rdflib')
    >>> print(provider.get_metadata('top_level.txt'))
    rdflib
    rdflib/extras
    rdflib/plugins
    rdflib/plugins/parsers
    rdflib/plugins/parsers/pyMicrodata
    rdflib/plugins/parsers/pyRdfa
    rdflib/plugins/parsers/pyRdfa/extras
    rdflib/plugins/parsers/pyRdfa/host
    rdflib/plugins/parsers/pyRdfa/rdfs
    rdflib/plugins/parsers/pyRdfa/transform
    rdflib/plugins/serializers
    rdflib/plugins/sparql
    rdflib/plugins/sparql/results
    rdflib/plugins/stores
    rdflib/tools

and is clearly broken as it should only contain le first line.

I think the problem is that the `packages` variable in `setup.py` is not what should be passed to the `packages=` keyword argument.

alexamici:
Minimal change that should fix #571. This time for real.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,20 +82,20 @@ def find_version(filename):
 version = find_version('rdflib/__init__.py')
 
 packages = ['rdflib',
-            'rdflib/extras',
-            'rdflib/plugins',
-            'rdflib/plugins/parsers',
-            'rdflib/plugins/parsers/pyRdfa',
-            'rdflib/plugins/parsers/pyRdfa/transform',
-            'rdflib/plugins/parsers/pyRdfa/extras',
-            'rdflib/plugins/parsers/pyRdfa/host',
-            'rdflib/plugins/parsers/pyRdfa/rdfs',
-            'rdflib/plugins/parsers/pyMicrodata',
-            'rdflib/plugins/serializers',
-            'rdflib/plugins/sparql',
-            'rdflib/plugins/sparql/results',
-            'rdflib/plugins/stores',
-            'rdflib/tools'
+            'rdflib.extras',
+            'rdflib.plugins',
+            'rdflib.plugins.parsers',
+            'rdflib.plugins.parsers.pyRdfa',
+            'rdflib.plugins.parsers.pyRdfa.transform',
+            'rdflib.plugins.parsers.pyRdfa.extras',
+            'rdflib.plugins.parsers.pyRdfa.host',
+            'rdflib.plugins.parsers.pyRdfa.rdfs',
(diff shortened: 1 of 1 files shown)


joernhees:
sorry for asking, but can you point me to where it explicitly says we shouldn't use `/`? i tried finding something normative but failed...

alexamici:
The relevant section is the one that I quoted in #571, you are supposed to include module names (dotted format) not folder paths and only top level modules. In fact setuptools drop all non top level module names and with my patch the top_level.txt file only include rdflib.

I noticed because for a project of mine I parsed the top_level.txt file of around a hundred projects and only rdflib and another project didn't conform to the documentation.

Finally, most projects just use `setuptools.find_packages` that does the job fine. Unless you have a reason not to, I'd advise you to use it as well. I can update the current PR if you wish.

alexamici:
The [Listing whole packages](https://docs.python.org/2/distutils/setupscript.html#listing-whole-packages) section of *Writing the Setup Script* of the main `distutil` documentation is also relevant.

> The packages option tells the Distutils to process (build, distribute, install, etc.) all pure Python modules found in each package mentioned in the packages list. In order to do this, of course, there has to be a correspondence between package names and directories in the filesystem. The default correspondence is the most obvious one, i.e. package distutils is found in the directory distutils relative to the distribution root."

joernhees:
> Finally, most projects just use setuptools.find_packages that does the job fine. Unless you have a reason not to, I'd advise you to use it as well. I can update the current PR if you wish.

if this works, i would happily accept it... not really keen on manually doing stuff that can be done automatically... the setup process of rdflib grew and was tweaked over many years, so any simplification is welcome.

for now i'll accept this pull request, as what we currently do seems to be weird...

PR #577, opened 2016-01-16
Prevent RDFa parser from failing on time elements with child nodes

### LINKED ISSUE #576: "NameError: name 'self' is not defined" (RDFa parser, with TIME element containing another element)
Given an HTML structure like the following:

    <time><a href="http://example.org">2016-01-01</a></time>

the RDFa parser will abort with an unhandled NameError exception because the `_get_literal()` function body references `self` but we are not in an OO context at the time.

I'm working on a fix, with a unit test.

dbs:
The _get_literal() function in the RDFa parser referred to a self instance that
does not exist. This led to a NameError exception if the HTML to be parsed
included a TIME element with one or more child nodes, such as
'<time><em>2016-01-01</em></time>'. Removing the reference to 'self' fixes
the problem and fixes #576.

### DIFF
--- a/rdflib/plugins/parsers/pyRdfa/host/html5.py
+++ b/rdflib/plugins/parsers/pyRdfa/host/html5.py
@@ -172,7 +172,7 @@ def html5_extra_attributes(node, state) :
 			if node.nodeType == node.TEXT_NODE:
 				rc = rc + node.data
 			elif node.nodeType == node.ELEMENT_NODE :
-				rc = rc + self._get_literal(node)
+				rc = rc + _get_literal(node)
 		if state.options.space_preserve :
 			return rc
 		else :
--- a/test/test_issue576.py
+++ b/test/test_issue576.py
@@ -1,0 +1,26 @@
+import rdflib
+
+html = """<!DOCTYPE html>
+<html>
+    <head>
+        <title>Boom</title>
+    </head>
+<body vocab="http://schema.org/" typeof="Book" resource="http://example.com/">
+    <time property="dateCreated"><em>2016-01-01</em></time>
+</body>
+</html>
+"""
+
+
+def test_time_child_element():
+    """
+    Ensure TIME elements that contain child nodes parse cleanly
+    """
+    g = rdflib.Graph()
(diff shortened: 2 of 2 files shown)


joernhees:
hmm, actually same problem as with pyMicrodata: maybe such things should go to https://github.com/RDFLib/PyRDFa or https://github.com/RDFLib/pyrdfa3 so they could be included in upstream test cases? @iherman ?

iherman:
Thanks @dbs and @joernhees. I have made the change in the https://github.com/RDFLib/pyrdfa3 repo (see https://github.com/RDFLib/pyrdfa3/pull/13). The https://github.com/RDFLib/PyRDFa is explicitly marked as 'outdated', so I did not touch it.

iherman:
I do not do merges of PR myself, I leave that to @joernhees ...

avengerpenguin:
I keep hitting this bug over and over. Is v4.2.2 going to be released any time soon please so I can just depend on that version in all my builds?

joernhees:
@avengerpenguin sorry for that... my plan was to fix a couple of SPARQL bugs ([4.2.2](https://github.com/RDFLib/rdflib/milestones/rdflib%204.2.2)) before the release, but i couldn't find enough time to work through them yet... maybe we should just release 4.2.2 and care about the SPARQL issue in a future release?

avengerpenguin:
That's up to you if you feel 4.2.2 is in a fit state to be released. I would normally advocate a bug fix release for each bug just to get fixes out there for people, just as with my situation where I would benefit from this issue being fixed in a proper package and I haven't encountered any of the SPARQL bugs you're holding off the release for. Nothing's stopping you having a 4.2.3 release with those bugs fixed when you have time :-)

If you think 4.2.2 has is no worse than 4.2.1 and it has this issue fixed in it, then please do release it if you can.

avengerpenguin:
I've got my application building ok using the master branch of this repo, so that takes some pressure off me. I guess it'd still be nice to see a release with this fix and any other fixes if 4.2.2-dev has fewer bugs than 4.2.1.

Issue #574, opened 2016-01-10
ImportError: No module named 'pyMicrodata'


dbs:
Under Python 3, invoking the microdata module can throw an uncaught ImportError exception "No module named 'pyMicrodata'", which is rather unexpected for what should be a common operation.

The following currently reproduces the error from the command line:

    rdfpipe -w -i microdata https://ambrose.edu/resources/ambrose-archives

and returns the following traceback:

    Traceback (most recent call last):
      File "/lib/python3.4/site-packages/rdflib/plugins/parsers/pyMicrodata/utils.py", line 297, in __init__
        self.location = urlparse.urljoin(self.data.geturl(),self.headers[URIOpener.CONTENT_LOCATION])
    AttributeError: 'function' object has no attribute 'urljoin'
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/bin/rdfpipe", line 9, in <module>
        load_entry_point('rdflib==4.2.1', 'console_scripts', 'rdfpipe')()
      File "/lib/python3.4/site-packages/rdflib/tools/rdfpipe.py", line 181, in main
        outfile, opts.output_format, ns_bindings)
      File "/lib/python3.4/site-packages/rdflib/tools/rdfpipe.py", line 49, in parse_and_serialize
        graph.parse(fpath, format=use_format, **kws)
      File "/lib/python3.4/site-packages/rdflib/graph.py", line 1468, in parse
        location=location, file=file, data=data, **args)
      File "/lib/python3.4/site-packages/rdflib/graph.py", line 1041, in parse
        parser.parse(source, self, **args)
      File "/lib/python3.4/site-packages/rdflib/plugins/parsers/structureddata.py", line 306, in parse
        vocab_cache=vocab_cache)
      File "/lib/python3.4/site-packages/rdflib/plugins/parsers/structureddata.py", line 252, in _process
        orig_source, graph=graph, rdfOutput=False)
      File "/lib/python3.4/site-packages/rdflib/plugins/parsers/pyMicrodata/__init__.py", line 290, in graph_from_source
        if not rdfOutput : raise e
      File "/lib/python3.4/site-packages/rdflib/plugins/parsers/pyMicrodata/__init__.py", line 262, in graph_from_source
        if not rdfOutput : raise e
      File "/lib/python3.4/site-packages/rdflib/plugins/parsers/pyMicrodata/__init__.py", line 252, in graph_from_source
        input = self._get_input(name)
[...]

PR #578, opened 2016-01-17
Avoid class reference to imported function

dbs:
We import the urljoin() function directly to support both Py2 and Py3, so the
reference to urlparse.urljoin() raises an exception. Don't do that!

Fixes 574.


joernhees:
thanks, we'll see what remains after the pyMicrodata cleanup though ;)

joernhees:
closed #574

joernhees:
@iherman did you sync this?

iherman:
@joernhees if you mean whether my local copy is in sync with the current master (including your change in adding the csvw to the initial context, thanks for this!) then yes, I am in sync. I plan to update the initial context at the end of February, when some more Recommendations will be added to the mix.

PR #584, opened 2016-01-25
Update reference to "Emulating container types"

dbs:
The docs.python.org site changed their URL structures. Update the
reference and the URL accordingly.

Fixes # 575.


dbs:
Augh, sorry for the duplicate of #583 . Closing.

Issue #493, opened 2015-07-06
SPARQL result in XML may not be strictly valid?


pchampin:
The XML serializer for SPARQL results produces a result that looks valid to me, but is not recognized by at least one tool (namely [Corese](http://wimmics.inria.fr/corese)). The problem comes from the fact that the serializer produces something like

``` xml
  <sparql:sparql xmlns:sparql="http://www.w3.org/2005/sparql-results#">
```

while Corese expects an empty NS prefix, as in

``` xml
  <sparql xmlns="http://www.w3.org/2005/sparql-results#">
```

which is actually [what the recommendation says](http://www.w3.org/TR/rdf-sparql-XMLres/#docElement). Again, the fact that the documentation uses XML namespaces makes me think that what RDFLib produces is valid, but the phrasing ("written as follow") makes it ambiguous. I explicitly [asked the question on the public mailing list](https://lists.w3.org/Archives/Public/public-sparql-dev/2015JulSep/0007.html).

Whatever the answer is, I guess it wouldn't hurt to be a little more conservative in RDFLib. If I'm correct, changing [that line](https://github.com/RDFLib/rdflib/blob/7f6c9e73c2b7b84fd38811b3064e3740f96d7976/rdflib/plugins/sparql/results/xmlresults.py#L148) to use the empty prefix rather than the "sparql" prefix would do the trick (sorry, not really able to test it myself right now).

joernhees:
@gromgull any feedback on this?

joernhees:
https://lists.w3.org/Archives/Public/public-sparql-dev/2015JulSep/0009.html says it's Corese's problem...

gromgull:
fixed by #588

PR #588, opened 2016-01-28
changed prefix for SPARQL namespace in XML serialization

pchampin:
Although #493 is not really a bug (the current serialization _is_ valid after all), several tools are not fully compliant and require the namespace prefix to be empty.
As this change help those "not so compliant" tools without hurting "fully compliant" tools, I think it is worth it.
(plus: it reduces the size of the XML serialization...)


gromgull:
why not.

PR #590, opened 2016-01-28
Use find_packages for package discovery.

alexamici:
This is an update of #573 as discussed with @joernhees 

I checked that the `packages` variable is filled with the module names currently hard-coded in `setup.py`.


joernhees:
thanks

PR #594, opened 2016-02-08
Added the csvw prefix to the RDFa initial context.

iherman:
See 

https://lists.w3.org/Archives/Public/semantic-web/2016Feb/0017.html

for the details.

### DIFF
--- a/rdflib/plugins/parsers/pyRdfa/initialcontext.py
+++ b/rdflib/plugins/parsers/pyRdfa/initialcontext.py
@@ -21,6 +21,7 @@ initial_context = {
 }
 
 initial_context["http://www.w3.org/2011/rdfa-context/rdfa-1.1"].ns = {
+	'csvw'		: 'http://www.w3.org/ns/csvw#',
 	'dcat'		: 'http://www.w3.org/ns/dcat#',
     'qb'		: 'http://purl.org/linked-data/cube#',
     'org'		: 'http://www.w3.org/ns/org#',
@@ -99,4 +100,3 @@ initial_context["http://www.w3.org/2011/rdfa-context/xhtml-rdfa-1.1"].terms = {
 	'first'					: 'http://www.w3.org/1999/xhtml/vocab#first',
 	'top'					: 'http://www.w3.org/1999/xhtml/vocab#top',
 }
-


joernhees:
is this really intended for the current revision? if not it would make more sense to rebase and merge this into the 5.0.0-dev branch...

iherman:
> On 9 Feb 2016, at 16:53, Jörn Hees <notifications@github.com> wrote:
> 
> is this really intended for the current revision? if not it would make more sense to rebase and merge this into the 5.0.0-dev branch...
> 
> 

Well… it depends when the 5.0.0 branch will be released. If it is a matter of a few weeks, then it is of course fine. If it is longer, then probably it is better to get this particular issue out as soon as possible; RDFa processors are supposed to update themselves to the latest version of the initial context as soon as possible.

It all depends on the release plans.

joernhees:
i merged this into 5.0.0-dev and it works, see #514 (d2671334b7adbdd5e9f5aea6c54871f57eb34734) ...

wrt. when 5.0.0 is released: as soon as these issues are solved or moved to other milestones...
- [4.2.2](https://github.com/RDFLib/rdflib/milestones/rdflib%204.2.2)
- [5.0.0](https://github.com/RDFLib/rdflib/milestones/rdflib%205.0.0)

niklasl:
How about adding a method for updating the RDFa processor context manually in between releases? Something like this could be added to the `rdflib.plugins.parsers.pyRdfa.initialcontext` module:

```python
from rdflib import Graph, Namespace, RDF

RDFA = Namespace('http://www.w3.org/ns/rdfa#')

def update_initial_context():
    for context_uri, context in initial_context.items():
        graph = Graph().parse(context_uri, format="rdfa")
        wrapper = initial_context[context_uri]
        for mapping, uri in graph.resource(RDFA.uri).subject_objects():
            uri = unicode(uri)
            prefix = mapping.value(RDFA.prefix)
            if prefix:
                prefix = unicode(prefix)
                assert mapping[RDF.type:RDFA.PrefixMapping]
                assert wrapper.ns.setdefault(prefix, uri) == uri
            else:
                term = unicode(mapping.value(RDFA['term']))
                assert mapping[RDF.type:RDFA.TermMapping]
                assert wrapper.terms.setdefault(term, uri) == uri
```

A user could then do:

```python
from rdflib.plugins.parsers.pyRdfa import initialcontext

initialcontext.update_initial_context()
```

(Something nicer might be to call this automatically at times and cache the results. But that'd be more complex, and failures hard to understand, so I suggest beginning with this.)

iherman:
Yeah… I actually had a much more complicated version at a time when we had the notion of a profile in RDFa (which we removed later) which allowed the user to dynamically such a set of prefixes. That version did a little bit what you suggest in the last line, namely checking dates, loading the prefixes and such only when necessary, otherwise using some local cache. But the code has been removed when the feature was removed. (It may be somewhere deep down in one of the pre-RDFlib commits, but it may even precede my move to github.)

As for this thing: there is an update once, say, every 18 months. There may be an oa prefix coming up at the end of this year, then… well, I do not know. Frankly, for a 5 minute update once every 18 months I did not feel like doing this. The new versions of RDFLib are released more often…

But… if you want to do this, of course!:-)

Ivan


> On 7 Mar 2016, at 15:39, Niklas Lindström <notifications@github.com> wrote:
> 
> How about adding a method for updating the RDFa processor context manually in between releases? Something like this could be added to the rdflib.plugins.parsers.pyRdfa.initialcontext module:
> 
> from rdflib import Graph, Namespace, RDF
> 
> RDFA = Namespace('http://www.w3.org/ns/rdfa#')
> 
> def update_initial_context():
>     for context_uri, context in initial_context.items():
>         graph = Graph().parse(context_uri, format="rdfa")
>         wrapper = initial_context[context_uri]
>         for mapping, uri in graph.resource(RDFA.uri).subject_objects():
>             uri = unicode(uri)
>             prefix = mapping.value(RDFA.prefix)
>             if prefix:
>                 prefix = unicode(prefix)
>                 assert mapping[RDF.type:RDFA.PrefixMapping]
>                 assert wrapper.ns.setdefault(prefix, uri) == uri
>             else:
>                 term = unicode(mapping.value(RDFA['term']))
>                 assert mapping[RDF.type:RDFA.TermMapping]
>                 assert wrapper.terms.setdefault(term, uri) == uri
> A user could then do:
> 
> from rdflib.plugins.parsers.pyRdfa import initialcontext
> 
> initialcontext.update_initial_context()
> (Something nicer might be to call this automatically at times and cache the results. But that'd be more complex, and failures hard to understand, so I suggest beginning with this.)
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/RDFLib/rdflib/pull/594#issuecomment-193275304>.
> 


----
Ivan Herman
Bankrashof 108
1183NW Amstelveen, The Netherlands
http://www.ivan-herman.net

joernhees:
that would be an option, even though i have to say that from a dev perspective i remember that i was actually confused by that local caching before (different systems gave different results for the very same code)... so doing this automagically definitely has some disadvantages.

i don't want to revive the discussion about depending on a standalone package here, but that would solve that without magic in a standard way and without the potential to bash w3.org with requests, potential to fail when there's no connection, ...

iherman:
> On 9 Mar 2016, at 20:41, Jörn Hees <notifications@github.com> wrote:
> 
> that would be an option, even though i have to say that from a dev perspective i remember that i was actually confused by that local caching before (different systems gave different results for the very same code)... so doing this automagically definitely has some disadvantages.
> 
> i don't want to revive the discussion about depending on a standalone package here, but that would solve that without magic in a standard way and without the potential to bash w3.org with requests, potential to fail when there's no connection, …
> 
I do not see why this would be the case. If the RDFa module was a standalone package, somebody still has to update it on github if a new prefix comes in. This means that end users would have to update their local copy to have it deployed. Technology wise a separate module would bring no difference.

Of course, a separate module could be considered to be more agile in this sense, ie, the new releases would come up more easily, but that is a matter of release policy rather than anything else.

joernhees:
yes, the benefit lies solely in decomposition and using the features of automated installers such as pip.

The semver.org release policy in RDFLib would probably dictate that such a prefix update has a new major release, as it could potentially break code in a backwards incompatible way if prefixes disappear for example...

iherman:
> On 10 Mar 2016, at 15:19, Jörn Hees <notifications@github.com> wrote:
> 
> yes, the benefit lies solely in decomposition and using the features of automated installers such as pip.
> 
> The semver.org release policy in RDFLib would probably dictate that such a prefix update has a new major release, as it could potentially break code in a backwards incompatible way if prefixes disappear for example...
> 
> 

I think the danger of breaking application is very low. Indeed, there is *no* way a prefix can disappear. Any change is restricted to the *addition* of prefixes, per RDFa policy.

PR #599, opened 2016-02-16
fix unicode encoding errors in sparql processor #446

### LINKED ISSUE #446: UnicodeEncodeError on SPARQLError
Here is code sample how to reproduce this error:

```python
import sys
import pprint
import rdflib

assert sys.version.startswith('2.7.6')
assert rdflib.__version__ == '4.0.1'

graph = rdflib.Graph()
graph.parse('http://dbpedia.org/resource/Metallica')
rows = graph.query('''
    SELECT DISTINCT ?p ?o
    WHERE {
        ?s ?p ?o .
        FILTER(lang(?o) = "" || lang(?o) = "en")
    }
''')

pprint.pprint(list(rows))
```

Running above, gives this:

```
  File "testme.py", line 18, in <module>
    pprint.pprint(list(rows))
  File "rdflib/query.py", line 238, in __len__
    return len(self.bindings)
  File "rdflib/query.py", line 178, in _get_bindings
    self._bindings += list(self._genbindings)
  File "rdflib/plugins/sparql/evaluate.py", line 353, in evalDistinct
    for x in res:
  File "rdflib/plugins/sparql/evaluate.py", line 362, in <genexpr>
    return (row.project(project.PV) for row in res)
  File "rdflib/plugins/sparql/evaluate.py", line 153, in evalFilter
    if _ebv(part.expr, c.forget(ctx)):
  File "rdflib/plugins/sparql/evalutils.py", line 50, in _ebv
    return EBV(expr.eval(ctx))
  File "rdflib/plugins/sparql/parserutils.py", line 194, in eval
    return self._evalfn(ctx)
  File "rdflib/plugins/sparql/operators.py", line 823, in ConditionalOrExpression
    expr = e.expr
  File "rdflib/plugins/sparql/parserutils.py", line 173, in __getattr__
    return self[a]
  File "rdflib/plugins/sparql/parserutils.py", line 163, in __getitem__
    return self._value(OrderedDict.__getitem__(self, a))
  File "rdflib/plugins/sparql/parserutils.py", line 158, in _value
    return value(self.ctx, val, variables)
  File "rdflib/plugins/sparql/parserutils.py", line 62, in value
    return val.eval(ctx)  # recurse?
  File "rdflib/plugins/sparql/parserutils.py", line 194, in eval
    return self._evalfn(ctx)
  File "rdflib/plugins/sparql/operators.py", line 783, in RelationalExpression
    raise SPARQLError('I cannot compare this non-node: %s' % expr)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 61-69: ordinal not in range(128)
```

I got same error on fresh `master` version.

joernhees:
attempt to fix #446.

The problem consist of two problems: in [`parserutils`](https://github.com/RDFLib/rdflib/blob/6200934ace658ca05de693956b37623ce3a545eb/rdflib/plugins/sparql/parserutils.py#L199-L199) a SPARQLError is returned instead of raised, making it end up in `operators.py`, in which it ends up in another `SPARQLError('... %s' % here)` causing `str()` to be called on it... if the original expression contained a unicode char it is thus causing an encoding error:

```
In [7]: Exception('name was: %s' % u'jörn')
Out[7]: Exception(u'name was: j\xf6rn')

In [8]: Exception('something went wrong: %s' % Exception('name was: %s' % u'jörn'))
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-8-b264e31546f5> in <module>()
----> 1 Exception('something went wrong: %s' % Exception('name was: %s' % u'jörn'))

UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 11: ordinal not in range(128)
```

this fixes the problem by using `%r` instead of `%s`, which might be a good idea in general.

### DIFF
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -488,7 +488,7 @@ def Builtin_TIMEZONE(e, ctx):
     """
     dt = datetime(e.arg)
     if not dt.tzinfo:
-        raise SPARQLError('datatime has no timezone: %s' % dt)
+        raise SPARQLError('datatime has no timezone: %r' % dt)
 
     delta = dt.tzinfo.utcoffset(ctx.now)
 
@@ -547,7 +547,7 @@ def Builtin_LANG(e, ctx):
 def Builtin_DATATYPE(e, ctx):
     l = e.arg
     if not isinstance(l, Literal):
-        raise SPARQLError('Can only get datatype of literal: %s' % l)
+        raise SPARQLError('Can only get datatype of literal: %r' % l)
     if l.language:
         return RDF_langString
     if not l.datatype and not l.language:
@@ -603,7 +603,7 @@ def Function(e, ctx):
                 return Literal(x, datatype=XSD.string)
             else:
                 raise SPARQLError(
-                    "Cannot cast term %s of type %s" % (x, type(x)))
+                    "Cannot cast term %r of type %r" % (x, type(x)))
 
         if not isinstance(x, Literal):
             raise SPARQLError(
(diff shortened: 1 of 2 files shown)


gromgull:
> a SPARQLError is returned instead of raised, 

this is intentional - it's a marker for a certain type of error that occured, that isn't really "exceptional"

gromgull:
we could argue that we should use a different object as error marker though, and not one that is ALSO used as an exception

joernhees:
yupp, it seems weird to me to return exceptions, but i already tried raising it or returning `object()` or `None`... both cause a lot of failed tests, which is why i opted to just make the exception wrapping more robust...

i think we can live with this for now and move the "don't return exceptions" part to a more boring time ;)

PR #600, opened 2016-02-16
Graph.update() uses own namespaces if none given, fixes #579

### LINKED ISSUE #579: Inconsistent behavor of `initN` in `update` and `query`
The [doc](http://rdflib.readthedocs.org/en/latest/namespaces_and_bindings.html) says

>  If you pass no `initNs` argument, the namespaces registered with the graphs `namespace_manager` are used

However, it doesn't seem true for `update`.

```{python}
from rdflib import Graph, URIRef, Literal, Namespace
from rdflib.namespace import FOAF, RDF

g = Graph()
g.bind('foaf', FOAF)
n = Namespace("http://myname/")
g.add((n.bob, FOAF.name, Literal('bb')))
# query is successful.
g.query("select ?n where { ?n foaf:name 'bb' . }")
# update is not.
g.update("""
    delete { ?e ?p ?o }
    where {
        ?e foaf:name 'ss' .
    }""")
Exception: Unknown namespace prefix : foaf
```

joernhees:
see #579

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1084,10 +1084,12 @@ class Graph(Node):
             query_object, initBindings, initNs, **kwargs))
 
     def update(self, update_object, processor='sparql',
-              initNs={}, initBindings={},
+              initNs=None, initBindings=None,
               use_store_provided=True, **kwargs):
-        """
-        """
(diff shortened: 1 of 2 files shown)


gromgull:
👍

PR #601, opened 2016-02-27
SAPRQL: bind variables to other variables in subgroups

### LINKED ISSUE #580: BIND not working as expected in SPARQL
Ok, consider the following data:

``` turtle
@prefix s: <http://schema.org/> .
@prefix x: <http://example.org/> .

x:john a s:Person ; s:name "John" .
x:jane a s:Person ; s:name "Jane" ; s:knows x:john.
x:ecorp a s:Organization ; s:name "Evil Corp" ; s:employee x:jane . 
```

and the following query:

``` sparql
PREFIX s: <http://schema.org/>
PREFIX x: <http://example.org/>

#CONSTRUCT { ?s ?p ?o }
SELECT ?s ?p ?o ?x # for debugging
WHERE {
  ?x a s:Person .
  { ?x ?p ?o . BIND (?x as ?s) } UNION
  { ?s ?p ?x . BIND (?x as ?o) }
}
```

The goal of the query is to extract all incoming and outgoing arcs of every `s:Person` in the data
(and ultimately to construct a graph with those, but I use `SELECT` for debugging).

As I understand, the result should be

| ?s | ?p | ?o | ?x |
| --- | --- | --- | --- |
| x:john | rdf:type | s:Person | x:john |
| x:john | s:name | "John" | x:john |
| x:jane | s:knows | x:john | x:john |
| x:jane | rdf:type | s:Person | x:jane |
| x:jane | s:name | "Jane" | x:jane |
| x:jane | s:knows | x:john | x:jane |
| x:ecorp | s:employee | x:jane | x:jane |

At least, this is what I get with Corese and Virtuoso.

But instead I get

| ?s | ?p | ?o | ?x |
| --- | --- | --- | --- |
|  | rdf:type | s:Person | x:john |
|  | s:name | "John" | x:john |
| x:jane | s:knows |  | x:john |
|  | rdf:type | s:Person | x:jane |
|  | s:name | "Jane" | x:jane |
|  | s:knows | x:john | x:jane |
| x:ecorp | s:employee |  | x:jane |

so the `BIND` (in this configuration) does not seem to work.

NB: the `UNION` is not the cause of the problem. I have the same problem without any `UNION`.

joernhees:
until now mainly test-cases in order to fix #580

### DIFF
--- a/test/test_issue580.py
+++ b/test/test_issue580.py
@@ -1,0 +1,95 @@
+# test for https://github.com/RDFLib/rdflib/issues/580
+
+from rdflib import Graph, URIRef, Literal, Namespace
+from rdflib.namespace import FOAF, RDF
+
+
+sample_data = '''
+@prefix s: <http://schema.org/> .
+@prefix x: <http://example.org/> .
+
+x:john a s:Person ; s:name "John" .
+x:jane a s:Person ; s:name "Jane" ; s:knows x:john.
+x:ecorp a s:Organization ; s:name "Evil Corp" ; s:employee x:jane .
+'''
+
+g = Graph().parse(data=sample_data, format='n3')
+assert len(g) == 8
+
+simple_query = '''
+select ?s ?o ?x where {
+  ?x s:knows ?o .
+  BIND(?x as ?s).
+}
+'''
+expected = [
+    (
+        URIRef(u'http://example.org/jane'),
+        URIRef(u'http://example.org/john'),
+        URIRef(u'http://example.org/jane'),
+    ),
+]
+res = sorted(g.query(simple_query))
+assert res == expected, '\nexpected: %s\ngot     : %s' % (expected, res)
+
+
+simple_problematic_query = '''
+select ?s ?o ?x where {
+  ?x s:knows ?o .
+  { BIND(?x as ?s). }
+}
+'''
+res = sorted(g.query(simple_problematic_query))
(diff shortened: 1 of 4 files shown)


test/test_issue580.py
```diff
@@ -0,0 +1,95 @@
...
+    ),
+]
+res = sorted(g.query(simple_query))
+assert res == expected, '\nexpected: %s\ngot     : %s' % (expected, res)
+
+
+simple_problematic_query = '''
+select ?s ?o ?x where {
+  ?x s:knows ?o .
+  { BIND(?x as ?s). }
```
joernhees: @gromgull without `{ }` this seems to work...

joernhees: actually i just found out that jena also seems to block this...
```
sparql --data=https://www.w3.org/2009/sparql/docs/tests/data-sparql11/bind/data.ttl --query=<(echo 'select * { ?s ?p ?o . { BIND(?o as ?z) } }')
------------------------------------------------------------
| s                       | p                      | o | z |
============================================================
| <http://example.org/s1> | <http://example.org/p> | 1 |   |
| <http://example.org/s4> | <http://example.org/p> | 4 |   |
| <http://example.org/s3> | <http://example.org/p> | 3 |   |
| <http://example.org/s2> | <http://example.org/p> | 2 |   |
------------------------------------------------------------
```

joernhees:
fixed in #688, partially rebased, closing...

Issue #404, opened 2014-06-17
SPARQL: COUNT DISTINCT not working properly


pchampin:
The combination of COUNT with DISTINCT is not working propetry in the RDFLib implementation of SPARQL.
Consider the following code:

``` python
from rdflib import *

g = Graph()
g.parse(format="turtle", publicID="http://example.org/", data="""
@prefix : <> .

<#a>
  :knows <#b>, <#c> ;
  :age 42 .

<#b>
  :knows <#a>, <#c> ;
  :age 36 .

<#c>
  :knows <#b>, <#c> ;
  :age 20 .

""")


print "Query 1: people knowing someone younger"
results = g.query("""
PREFIX : <http://example.org/>

SELECT DISTINCT ?x {
  ?x :age ?ax ; :knows [ :age ?ay ].
  FILTER( ?ax > ?ay )
}
""")
for i in results:
    print str(i[0])


print "\nQuery 2: count people knowing someone younger"
results = g.query("""
PREFIX : <http://example.org/>

SELECT (COUNT(DISTINCT ?x) as ?cx) {
  ?x :age ?ax ; :knows [ :age ?ay ].
  FILTER( ?ax > ?ay )
}
""")


for i in results:
    print str(i[0])
```

It produces the following output:

```
Query 1: people knowing someone younger
http://example.org/#a
http://example.org/#b

Query 2: count people knowing someone younger
3
```

while the result of Query 2 should obvioulsy be 2. It seems that the DISTINCT in Query 2 has no effect (indeed, <#a> matches the query twice, as she knows 2 people younger than herself).

pchampin:
Actually, the same problem exists with SUM(DISTINCT ?x), AVG(DISTINCT ?x) or GROUP_CONCAT(DISTINCT ?x).

gromgull:
This works in master now, presumably fixed by the refactoring of the aggregation code here: https://github.com/RDFLib/rdflib/pull/678

PR #611, opened 2016-03-19
DISTINCT was ignore inside aggregate functions: added unit tests and patch for the problem

pchampin:
This solves issue #404 which I raised some time ago.

Actually, at the time, I underestimated the problem, as DISTINCT in *any* aggregation function was simply ignored.

### DIFF
--- a/rdflib/plugins/sparql/aggregates.py
+++ b/rdflib/plugins/sparql/aggregates.py
@@ -13,10 +13,14 @@ Aggregation functions
 """
 
 
-def _eval_rows(expr, group):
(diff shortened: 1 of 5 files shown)


pchampin:
Some tests failed in Travil, but it seems to be that they are unrelated to my change :-/
There is a 403 error on URL <http://www.w3.org/2000/10/swap/test/meet/blue.rdf> ??
(which works fine for me, btw)

joernhees:
thanks for this contribution and sorry for letting it sit that long, i just didn't find the time to properly review this, yet...

Issue #609, opened 2016-03-17
Collection constructor is very slow with large lists


hsolbrig:
The collection constructor uses the append method, which iterates to the end of the list for every addition.  This results in a n! performance.  Propose replacing:

```
    for item in seq:
         self.append(item)
```

with:

```
    if seq:
        nxt = self.uri
        for item in seq:
            graph.add((nxt, RDF.first, item))
            nxt2 = BNode()
            graph.add((nxt, RDF.rest, nxt2))
            nxt = nxt2
        graph.add((nxt, RDF.first, RDF.nil))
```

in the constructor

joernhees:
it's O(n^2), but i agree that it's unnecessary... if you make a pull request, i'll accept it.

gromgull:
I went for a fix without the _end caching issues. Now there is a `__iadd__` method that lets you do `my_collection += [ Literal(1), Literal(2) ]` and this will be efficient. Repeat calls to append remain inefficient, but this is safer!

PR #612, opened 2016-03-19
Updated append method to keep an "_end" hint.  Purpose is to address …

hsolbrig:
…performance problems for large lists

### DIFF
--- a/rdflib/collection.py
+++ b/rdflib/collection.py
@@ -43,6 +43,7 @@ class Collection(object):
     def __init__(self, graph, uri, seq=[]):
         self.graph = graph
         self.uri = uri or BNode()
+        self._end = self.uri
         for item in seq:
             self.append(item)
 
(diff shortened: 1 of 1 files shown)


rdflib/collection.py
```diff
@@ -203,30 +219,58 @@ def append(self, item):
...
-                break
-            else:
-                # move down one link
-                if container != self.uri:
-                    rest = graph.value(rest, RDF.rest)
-                if not rest == RDF.nil:
-                    container = rest
-        graph.add((container, RDF.first, item))
-        graph.add((container, RDF.rest, RDF.nil))
+        if graph.value(self._end, RDF.rest) == RDF.nil:
```
gromgull: I prefer `(self._end, RDF.rest, RDF.nil) in graph`, which may also be slightly faster?

gromgull:
My gut feeling tells me there are some "cache invalidation" issues here, but since the code actually checks if `(_end, rest, NIL)` is in the graph the only thing I can think of is: 

* Make a list, the Collection's 'end' points to the end. 
* Make another list, move (outside the collection api) the tail of the first list to the end of this. 
* Now the first Collection has `end` pointing to the end of the wrong list. 

BUT - anyone who does this probably deserves everything coming to them :) 

Still, it would be a nightmarish bug to debug.

gromgull:
Actually, thinking about it - another, easier way to mess up is to create two objects for managing the same collection. 

This raises the question - why don't we have a utility function on the graph objects for creating collections? Like we do for `Resource`? Then that could have had an internal registry of the collection objects and returned the same if requested again.

joernhees:
PR for #609

joernhees:
yupp, i see the same issues... #609 was about the constructor, while this PR in general tries to improve subsequent `append` calls by giving `Collection` a state, which might cause problems later on.

What about giving `append` an optional arg `_end`? It would be a private arg to indicate that its use is potentially dangerous if you don't know what you're doing. If set it, `append` would trust it to be the end-pointer and we could use it to provide an O(n) `extend` method, which we could also use in the ctor?

gromgull:
As detailed in the ticket, I fixed this with a safer addition of an `__iadd__` method.

PR #616, opened 2016-04-03
Fix ImportError when __main__ already loaded

Dataliberate:
In rdflib/\_\_init\_\_ - place import of \_\_main\_\_ into a try catch block to catch ImportError that you get when already imported such as in a Google App Engine instance running webapp2.

Enables the use of rdflib unchanged in this configuration of Google App Engine and potentially other environments that claim \_\_main\_\_ before rdflib gets loaded.

This has been tested as functional in the latest version of the [Schema.org](https://github.com/schemaorg/schemaorg) code base.


joernhees:
thanks

PR #634, opened 2016-05-15
attempt to fix #633

### LINKED ISSUE #633: Unexpected exception with SPARQL-update
I have a problem with the following Python script, which performs a SPARQL update query.

Basically, I want to alter a set of triples : some of them must be plainly deleted (call them D), some of them must be updated (call them U).

[test-update.py.txt](https://github.com/RDFLib/rdflib/files/264936/test-update.py.txt)

As can be seen in the script, I use a SPARUL query combining DELETE, INSERT, and a WHERE with an OPTIONAL. The idea is that all concerned triples (D+U) match the "mandatory" part of the WHERE, while only the U triples match the OPTIONAL ; and the DELETE clause only uses variables of the "mandatory" part, while the INSERT clause uses a variable from the OPTIONAL clause.

My problem is that the SPARUL raises an exception (it produces an unexpected None value).

Strangely, the equivalent SELECT query works.

pchampin:
I made this fix with a _very superficial_ understanding of what was going on... It does not break any test (but this is not surprising: those tests would have failed before with an exception, as per #633), and it does give the expected result in the script that I attached to #633.

But I _really_ think it would be better if someone with a better understanding of the SPARQL engine could review it before a merge.

### DIFF
--- a/rdflib/plugins/sparql/evaluate.py
+++ b/rdflib/plugins/sparql/evaluate.py
@@ -141,8 +141,10 @@ def evalLeftJoin(ctx, join):
             # before we yield a solution without the OPTIONAL part
             # check that we would have had no OPTIONAL matches
             # even without prior bindings...
-            if not any(_ebv(join.expr, b) for b in
-                       evalPart(ctx.thaw(a.remember(join.p1._vars)), join.p2)):
+            p1_vars = join.p1._vars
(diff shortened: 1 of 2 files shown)


gromgull:
It looks fine to me, and it passes the tests!

PR #638, opened 2016-07-07
Result objects should close the sockets where the results come from

jdferreira:
The XMLResult never closed the source file where the result was coming from, resulting in "too man opened files" being raised when processing large amounts of SPARQL queries. I have added a line of code to the appropriate file, but I am not sure whether other result parsers need to have the same type of fix added to them.

### DIFF
--- a/rdflib/plugins/sparql/results/xmlresults.py
+++ b/rdflib/plugins/sparql/results/xmlresults.py
@@ -38,6 +38,7 @@ class XMLResult(Result):
     def __init__(self, source):
 
         xmlstring = source.read()
+        source.close_connection()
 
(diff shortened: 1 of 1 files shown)


joernhees:
tests fail... are you sure this behavior still exists in 4.2.2-dev (current master)?

gromgull:
Since that method doesn't open the result, it also shouldn't close it. 

However, the SPARQLStore probably should, here: 

https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/stores/sparqlstore.py#L333

We could do: 

`with contextlib.closing(SPARQLWrapper.query(self).response) as res: return Result.parse(res)`

PR #640, opened 2016-07-26
Put the original pattern in the ParseError for ntriples.eat()

justinccdev:


joernhees:
short check shows that this is totally useful:
```python
In [4]: 'this: %s' % re.compile('foo')
Out[4]: 'this: <_sre.SRE_Pattern object at 0x10537c618>'

In [5]: 'this: %s' % re.compile('foo').pattern
Out[5]: 'this: foo'

In [6]: 'this: %r' % re.compile('foo')
Out[6]: 'this: <_sre.SRE_Pattern object at 0x10537c618>'
```

justinccdev:
Yeah, from a quick look it seems travis is breaking due to dependency issues.  I ran the tests [1] locally and looks okay from here.

[1] http://rdflib.readthedocs.io/en/stable/developers.html#running-tests

joernhees:
in that case you probably still have html5lib <= 1.0b8 installed... latest seems to have some backwards incompatible changes :(

Issue #632, opened 2016-05-14
Handle namespaces that end in an underscore in turtle serializer


kshefchek:
Consider the following:

```
   from rdflib import Graph, URIRef

   graph = Graph()
   graph.bind('GENO', 'http://purl.obolibrary.org/obo/GENO_')
   graph.bind('RO', 'http://purl.obolibrary.org/obo/RO_')

   graph.add((URIRef('http://example.org'),
                     URIRef('http://purl.obolibrary.org/obo/RO_0002200'),
                     URIRef('http://purl.obolibrary.org/obo/GENO_0000385')))

   print(graph.serialize(format='turtle'))
```

Outputs:

```
    @prefix GENO: <http://purl.obolibrary.org/obo/GENO_> .
    @prefix RO: <http://purl.obolibrary.org/obo/RO_> .
    @prefix ns1: <http://purl.obolibrary.org/obo/> .
    @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
    @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
    @prefix xml: <http://www.w3.org/XML/1998/namespace> .
    @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

   <http://example.org> ns1:RO_0002200 ns1:GENO_0000385 .
```

This appears to be handled in the namespace module: https://github.com/RDFLib/rdflib/blob/master/rdflib/namespace.py#L485

Is this a bug or am I misunderstanding the turtle specs?

satra:
this is also the same issue reported here: https://github.com/trungdong/prov/issues/96

kshefchek:
this looks to be fixed in 5.0 via https://github.com/RDFLib/rdflib/pull/649

Issue #213, opened 2012-05-18
Too many calls to backend store for NamespaceManager.compute_qname()


zuotian:
While debugging some other issues, I noticed this piece of code in NamespaceManager,py

```
def compute_qname(self, uri, generate=True):
    if not uri in self.__cache:
        namespace, name = split_uri(uri)
        namespace = URIRef(namespace)
        prefix = self.store.prefix(namespace)
        ....
        self.__cache[uri] = (prefix, namespace, name)
  return self.__cache[uri]
```

Note that each self.store.prefix() call is a sql hit.

I have a triple store with 350k urls that only differ after #, this caching mechanism will seriously slow down the serialization.

I suggest to either cache prefixes in the store or have a better caching mechanism for NamespaceManager.

gjhiggins:
Superseded  by #649

PR #649, opened 2016-08-25
namespace.py fix compute_qname missing namespaces

tgbugs:
The way the compute_qname worked was to call split_uri
and then see if there was a match in self.store.prefix.
This produced incomplete behavior for namesapces that
end in LETTERS_ instead of / or #. This commit corrects
this behavior by iterating through name and testing
`namespace + name[:i]` to see if there is a matching prefix.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -353,6 +353,7 @@ class NamespaceManager(object):
     def __init__(self, graph):
         self.graph = graph
         self.__cache = {}
+        self.__cache_strict = {}
         self.__log = None
         self.__strie = {}
         self.__trie = {}
@@ -380,6 +381,13 @@ class NamespaceManager(object):
             return name
         else:
             return ":".join((prefix, name))
+
+    def qname_strict(self, uri):
+        prefix, namespace, name = self.compute_qname_strict(uri)
+        if prefix == '':
+            return name
+        else:
+            return ':'.join((prefix, name))
 
     def normalizeUri(self, rdfTerm):
         """
@@ -447,6 +455,56 @@ class NamespaceManager(object):
                 self.bind(prefix, namespace)
             self.__cache[uri] = (prefix, namespace, name)
         return self.__cache[uri]
+
+    def compute_qname_strict(self, uri, generate=True):
+        # code repeated to avoid branching on strict every time
+        # if output needs to be strict (e.g. for xml) then
+        # only the strict output should bear the overhead
+        prefix, namespace, name = self.compute_qname(uri)
+        if is_ncname(text_type(name)):
+            return prefix, namespace, name
+        else:
+            if uri not in self.__cache_strict:
+                try:
+                    namespace, name = split_uri(uri, NAME_START_CATEGORIES)
+                except ValueError as e:
+                    message = ('This graph cannot be serialized to a strict format '
+                               'because there is no valid way to shorten {}'.format(uri))
+                    raise ValueError(message)
+                    # omitted for strict since NCNames cannot be empty
+                    #namespace = URIRef(uri)
+                    #prefix = self.store.prefix(namespace)
+                    #if not prefix:
+                        #raise e
+
+                if namespace not in self.__strie:
+                    insert_strie(self.__strie, self.__trie, namespace)
+
+                # omitted for strict
+                #if self.__strie[namespace]:
+                    #pl_namespace = get_longest_namespace(self.__strie[namespace], uri)
+                    #if pl_namespace is not None:
+                        #namespace = pl_namespace
+                        #name = uri[len(namespace):]
+
+                namespace = URIRef(namespace)
+                prefix = self.store.prefix(namespace)  # warning multiple prefixes problem
+
+                if prefix is None:
+                    if not generate:
+                        raise KeyError(
+                            "No known prefix for {} and generate=False".format(namespace)
+                        )
+                    num = 1
+                    while 1:
+                        prefix = "ns%s" % num
+                        if not self.store.namespace(prefix):
+                            break
+                        num += 1
+                    self.bind(prefix, namespace)
+                self.__cache_strict[uri] = (prefix, namespace, name)
+
+            return self.__cache_strict[uri]
 
     def bind(self, prefix, namespace, override=True, replace=False):
         """bind a given namespace to the prefix
@@ -568,27 +626,28 @@ ALLOWED_NAME_CHARS = [u"\u00B7", u"\u0387", u"-", u".", u"_", u":"]
 
 
 def is_ncname(name):
-    first = name[0]
-    if first == "_" or category(first) in NAME_START_CATEGORIES:
-        for i in range(1, len(name)):
-            c = name[i]
-            if not category(c) in NAME_CATEGORIES:
-                if c != ':' and c in ALLOWED_NAME_CHARS:
-                    continue
-                return 0
-            # if in compatibility area
-            # if decomposition(c)!='':
-            #    return 0
+    if name:
+        first = name[0]
+        if first == "_" or category(first) in NAME_START_CATEGORIES:
+            for i in range(1, len(name)):
+                c = name[i]
+                if not category(c) in NAME_CATEGORIES:
+                    if c != ':' and c in ALLOWED_NAME_CHARS:
+                        continue
+                    return 0
+                # if in compatibility area
+                # if decomposition(c)!='':
+                #    return 0
 
-        return 1
-    else:
-        return 0
+            return 1
+
+    return 0
 
 
 XMLNS = "http://www.w3.org/XML/1998/namespace"
 
 
-def split_uri(uri):
+def split_uri(uri, split_start=SPLIT_START_CATEGORIES):
     if uri.startswith(XMLNS):
         return (XMLNS, uri.split(XMLNS)[1])
     length = len(uri)
@@ -598,7 +657,7 @@ def split_uri(uri):
             if c in ALLOWED_NAME_CHARS:
                 continue
             for j in range(-1 - i, length):
-                if category(uri[j]) in SPLIT_START_CATEGORIES or uri[j] == "_":
+                if category(uri[j]) in split_start or uri[j] == "_":
                     # _ prevents early split, roundtrip not generate
                     ns = uri[:j]
                     if not ns:
@@ -641,4 +700,3 @@ def get_longest_namespace(trie, value):
             else:
                 return out
     return None
-
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -606,10 +606,10 @@ def insert_trie(trie, value):  # aka get_subtrie_or_insert
     if value in trie:
         return trie[value]
     multi_check = False
-    for key in list(trie.keys()):
-        if value.startswith(key):
+    for key in tuple(trie.keys()):
+        if len(value) > len(key) and value.startswith(key):
             return insert_trie(trie[key], value)
-        elif key.startswith(value):
+        elif key.startswith(value):  # we know the value is not in the trie
             if not multi_check:
                 trie[value] = {}
                 multi_check = True  # there can be multiple longer existing prefixes
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -354,8 +354,8 @@ class NamespaceManager(object):
         self.graph = graph
         self.__cache = {}
         self.__log = None
-        self.__trie = {}
         self.__strie = {}
+        self.__trie = {}
         for p, n in self.namespaces():  # self.bind is not always called
             insert_trie(self.__trie, str(n))
         self.bind("xml", "http://www.w3.org/XML/1998/namespace")
@@ -365,6 +365,8 @@ class NamespaceManager(object):
 
     def reset(self):
         self.__cache = {}
+        self.__strie = {}
+        self.__trie = {}
 
     def __get_store(self):
         return self.graph.store
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -414,7 +414,13 @@ class NamespaceManager(object):
             )
 
         if uri not in self.__cache:
-            namespace, name = split_uri(uri)
+            try:
+                namespace, name = split_uri(uri)
+            except ValueError as e:
+                namespace = URIRef(uri)
+                prefix = self.store.prefix(namespace)
+                if not prefix:
+                    raise e
             if namespace not in self.__strie:
                 insert_strie(self.__strie, self.__trie, namespace)
 
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -357,7 +357,7 @@ class NamespaceManager(object):
         self.__trie = {}
         self.__strie = {}
         for p, n in self.namespaces():  # self.bind is not always called
-            insert_trie(self.__trie, n)
+            insert_trie(self.__trie, str(n))
         self.bind("xml", "http://www.w3.org/XML/1998/namespace")
         self.bind("rdf", RDF)
         self.bind("rdfs", RDFS)
@@ -386,7 +386,7 @@ class NamespaceManager(object):
         try:
             namespace, name = split_uri(rdfTerm)
             if namespace not in self.__strie:
-                insert_strie(self.__strie, self.__trie, namespace)
+                insert_strie(self.__strie, self.__trie, str(namespace))
             namespace = URIRef(text_type(namespace))
         except:
             if isinstance(rdfTerm, Variable):
@@ -415,12 +415,10 @@ class NamespaceManager(object):
                 insert_strie(self.__strie, self.__trie, namespace)
 
             if self.__strie[namespace]:
-                #print('BEFORE', namespace, uri, self.__strie[namespace])
                 pl_namespace = get_longest_namespace(self.__strie[namespace], uri)
                 if pl_namespace is not None:
                     namespace = pl_namespace
                     name = uri[len(namespace):]
-                #print('AFTER', namespace)
 
             namespace = URIRef(namespace)
             prefix = self.store.prefix(namespace)  # warning multiple prefixes problem
@@ -436,7 +434,6 @@ class NamespaceManager(object):
                     if not self.store.namespace(prefix):
                         break
                     num += 1
-                print(prefix, namespace)
                 self.bind(prefix, namespace)
             self.__cache[uri] = (prefix, namespace, name)
         return self.__cache[uri]
@@ -466,7 +463,7 @@ class NamespaceManager(object):
 
             if replace:
                 self.store.bind(prefix, namespace)
-                insert_trie(self.__trie, namespace)
+                insert_trie(self.__trie, str(namespace))
                 return
 
             # prefix already in use for different namespace
@@ -496,7 +493,7 @@ class NamespaceManager(object):
             else:
                 if override or bound_prefix.startswith("_"):  # or a generated prefix
                     self.store.bind(prefix, namespace)
-        insert_trie(self.__trie, namespace)
+        insert_trie(self.__trie, str(namespace))
 
     def namespaces(self):
         for prefix, namespace in self.store.namespaces():
@@ -605,14 +602,18 @@ def insert_trie(trie, value):  # aka get_subtrie_or_insert
         or not. """
     if value in trie:
         return trie[value]
-    for key in trie:
+    multi_check = False
+    for key in list(trie.keys()):
         if value.startswith(key):
             return insert_trie(trie[key], value)
         elif key.startswith(value):
+            if not multi_check:
+                trie[value] = {}
+                multi_check = True  # there can be multiple longer existing prefixes
             dict_ = trie.pop(key)  # does not break strie since key<->dict_ remains unchanged
-            trie[value] = {key:dict_}
-            return trie[value]
-    trie[value] = {}
+            trie[value][key] = dict_
+    if value not in trie:
+        trie[value] = {}
     return trie[value]
 
 def insert_strie(strie, trie, value):
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -581,13 +581,14 @@ def split_uri(uri):
     if uri.startswith(XMLNS):
         return (XMLNS, uri.split(XMLNS)[1])
     length = len(uri)
+    name_start_plus_nd = NAME_START_CATEGORIES + ["Nd"]
     for i in range(0, length):
         c = uri[-i - 1]
         if not category(c) in NAME_CATEGORIES:
             if c in ALLOWED_NAME_CHARS:
                 continue
             for j in range(-1 - i, length):
-                if category(uri[j]) in NAME_START_CATEGORIES or uri[j] == "_":
+                if category(uri[j]) in name_start_plus_nd or uri[j] == "_":
                     ns = uri[:j]
                     if not ns:
                         break
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -355,6 +355,7 @@ class NamespaceManager(object):
         self.__cache = {}
         self.__log = None
         self.__trie = {}
+        self.__strie = {}
         for p, n in self.namespaces():  # self.bind is not always called
             insert_trie(self.__trie, n)
         self.bind("xml", "http://www.w3.org/XML/1998/namespace")
@@ -384,6 +385,8 @@ class NamespaceManager(object):
         """
         try:
             namespace, name = split_uri(rdfTerm)
+            if namespace not in self.__strie:
+                insert_strie(self.__strie, self.__trie, namespace)
             namespace = URIRef(text_type(namespace))
         except:
             if isinstance(rdfTerm, Variable):
@@ -407,19 +410,22 @@ class NamespaceManager(object):
             )
 
         if uri not in self.__cache:
-            namespace = get_longest_namespace(self.__trie, uri)
-            if namespace is not None:
-                name = uri[len(namespace):]
-                if '/' in name or '#' in name:
-                    name = None
-                else:
-                    prefix = self.store.prefix(namespace)  # warning multiple prefixes problem
-            else:
-                name = None
+            namespace, name = split_uri(uri)
+            if namespace not in self.__strie:
+                insert_strie(self.__strie, self.__trie, namespace)
 
-            if name is None:
-                namespace, name = split_uri(uri)
-                namespace = URIRef(namespace)
+            if self.__strie[namespace]:
+                #print('BEFORE', namespace, uri, self.__strie[namespace])
+                pl_namespace = get_longest_namespace(self.__strie[namespace], uri)
+                if pl_namespace is not None:
+                    namespace = pl_namespace
+                    name = uri[len(namespace):]
+                #print('AFTER', namespace)
+
+            namespace = URIRef(namespace)
+            prefix = self.store.prefix(namespace)  # warning multiple prefixes problem
+
+            if prefix is None:
                 if not generate:
                     raise KeyError(
                         "No known prefix for {} and generate=False".format(namespace)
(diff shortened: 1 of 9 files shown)


joernhees:
tests fail... are you sure this behavior still exists in 4.2.2-dev (current master)?

tgbugs:
Maybe I have incorrectly assumed that OWLAPI implements prefix handling according to the spec. This change was intended to match its behavior. To clarify about / and # endings: these are correct, and they remain correct with my changes, essentially this adds the ability to use arbitrary prefixes.

I will check again to see if this is in 4.2.2 and figure out why the tests are failing.

satra:
@tgbugs - i decided to fix this together with #632 in the turtle serializer itself. see here: #661 

and for the full discussion between @gromgull and myself see #660

tgbugs:
@satra the solution in #661 looks good to me. Closing this. Thanks!

tgbugs:
I have reverted the change to NAME_START_CATEGORIES that was causing the build failures. Tests should pass now.

satra:
@tgbugs - with this change does the following work for you:

```python
    graph = Graph()
    graph.bind('GENO', 'http://purl.obolibrary.org/obo/GENO_')
    graph.bind('RO_has_phenotype',
                    'http://purl.obolibrary.org/obo/RO_0002200')
    graph.add((URIRef('http://example.org'),
               URIRef('http://purl.obolibrary.org/obo/RO_0002200'),
               URIRef('http://purl.obolibrary.org/obo/GENO_0000385')))
    output = [val for val in
              graph.serialize(format='turtle').decode().splitlines()
              if not val.startswith('@prefix')]
    output = ' '.join(output)
    assert 'RO_has_phenotype: ' in output
    assert 'GENO:0000385' in output
```

tgbugs:
@satra No, that example fails, producing `' <http://example.org> ns1:RO_0002200 ns1:GENO_0000385 . '`. From my experiments the issues arrises because a call to `graph.namespace_manager.qname(URIRef('http://purl.obolibrary.org/obo/RO_0002200'))` (which is buried in serialize) forces the creation of `('ns1', rdflib.term.URIRef('http://purl.obolibrary.org/obo/'))`. A call to `graph.namespace_manager.qname(URIRef('http://purl.obolibrary.org/obo/GENO_0000385'))` does not cause this behavior, which leads me to believe it is because my code stops just short of the end (it assumes that users would not prefix an entire iri) I think I can fix this and will look into it.

tgbugs:
@satra Fixed. Was stopping at `len(name)` instead of `len(name) + 1`. Your tests now pass. If we care about efficiency the slowest part of this seems to be the fact that I have to convert the concatenated name to a URIRef every time in order to check the prefixes.

satra:
@tgbugs - great. could you please add the test below to: `test_turtle_serialize.py`

@gromgull and @joernhees - let's focus on this PR and i'll close the other one. any questions or considerations we are overlooking?

```
def test_turtle_namespace():
    graph = Graph()
    graph.bind('GENO', 'http://purl.obolibrary.org/obo/GENO_')
    graph.bind('RO_has_phenotype',
                    'http://purl.obolibrary.org/obo/RO_0002200')
    graph.add((URIRef('http://example.org'),
               URIRef('http://purl.obolibrary.org/obo/RO_0002200'),
               URIRef('http://purl.obolibrary.org/obo/GENO_0000385')))
    output = [val for val in
              graph.serialize(format='turtle').decode().splitlines()
              if not val.startswith('@prefix')]
    output = ' '.join(output)
    assert 'RO_has_phenotype: ' in output
    assert 'GENO:0000385' in output
```

tgbugs:
I slightly modified the test case for this so that it forces us to match the longest prefix which is contained in an iri instead of the shortest. I'm pretty sure we can improve significantly on the performance here if needs be.

tgbugs:
One more issue here that I have noticed is that if something got stuck in the cache and a new longer prefix is added the old shorter prefix will hit the cache and not behave as expected.

satra:
@gromgull and @joernhees - anything else needed here?

satra:
@gromgull - just a ping here to see if this can be merged. also is there a timeline for a release?

gromgull:
I hope to get a 4.2.2 out "quite soon", i.e. maybe before end of february. I'll review this before then!

rdflib/namespace.py
```diff
@@ -330,6 +330,21 @@ def compute_qname(self, uri, generate=True):
             namespace, name = split_uri(uri)
             namespace = URIRef(namespace)
             prefix = self.store.prefix(namespace)
+            # check to see if there are longer prefixes defined
+            temp_namespace = namespace.toPython()
+            output_prefix = prefix
+            output_namespace = namespace
+            output_name = name
+            for i in xrange(1, len(name) + 1):
```
gromgull: we really cannot have this loop here. This function gets called for every single URI in the store on serialisation. 

You can have a loong URI and this will call the store for each substring in that URL. For some stores, this means network activity (sqlalchemy, etc). 

If we really want to fix this, the namespace manager needs to index the available prefixes by length in __init__ and then try the longest first. But even this means many "if uri starts with x" calls. Some sort of trie-data-structure may be useful here.

gromgull: see #213

tgbugs: Ok. This is what I expected we would have to do because this implementation produces worst case behavior all the time. Will take a look and see how it can be improved. Do we have any nice pathological cases that I could use for performance testing?

tgbugs: So my first round of performance testing on the 3 different implementations produced some weird results which I think are not representative of the spectrum of workloads. I serialized https://github.com/SciCrunch/NIF-Ontology/blob/master/ttl/NIF-Molecule.ttl 10 times using 3 different implementations and average the time spend calling compute_qname in namespace.py as measured by cProfile in python3.5, there were 246680 calls to compute_qname for each run.
current mainline rdflib implementation of compute_qname: 1.2658s
loop over uri implementation (current for this pull request) of compute_qname: 1.1765s
loop on namespaces implementation of compute_qname: 1.2025s

This was a quick and dirty test and I am going to set it up a more robust suite since I don't believe that the implementation in question here is actually the fastest for all workloads as the numbers above would seem to suggest.

tgbugs: Turns out string.split is very, very slow. New implementation using len and slicing gives 1.162s using the same method as above.

rdflib/namespace.py
```diff
@@ -325,12 +325,16 @@ def compute_qname(self, uri, generate=True):
...
 
-
         if not uri in self.__cache:
-            namespace, name = split_uri(uri)
-            namespace = URIRef(namespace)
-            prefix = self.store.prefix(namespace)
-            if prefix is None:
+            name = None
+            for lennamespace, prefix, namespace in self.ordered_prefixnames:
+                if namespace in uri:
```
joernhees: this should be `if uri.startswith(namespace):`

joernhees:
i'd leave the old method in as_is and only if that doesn't find anything try your approach...

anyhow, we should probably overhaul the whole thing and use some trie / KMP

rdflib/namespace.py
tgbugs: My testing showed that startswith is slower, however if there are cases where a namespace might somehow appear somewhere other than at the start of a URI then we would need to use startswith to prevent false positives. Personally I can't come up with any cases where this might happen.

tgbugs:
Do you have any examples where the old method is faster than the new one?

rdflib/namespace.py
joernhees: yupp... correctness > speed

joernhees:
The old method uses a dict for lookup (so O(1) per lookup), based on splits by URI parts, so (O(m) for m parts per URI). Overall O(m).

Your method goes through all namespaces (on avg. n/2), so uses O(n) for n registered namespaces.

The old method will scale better in light of many registered namespaces. Your method is likely to be faster if only few namespaces are registered.

tgbugs:
I'm happy to try the old implementation first, but it can produce incorrect results in cases where a valid namespace is defined that does not terminate in members of NAME_START_CATEGORIES (eg https://github.com/tgbugs/rdflib/blob/master/test/test_turtle_serialize.py#L78 will fail) so it will take more changes to make sure we are actually getting the longest matching namespace. In addition, looking at the [existing split_uri method](https://github.com/RDFLib/rdflib/blob/master/rdflib/namespace.py#L497) I wonder whether on average URIs are longer (`m`) or there are more prefixes defined (`n`). Clearly without knowing the empirical distributions of `m` and `n` we cant know for certain, but I suspect that on average `m` is going to be larger than `n`. Thoughts?

tgbugs:
I've updated to use a trie since it is much easier and cleaner to implement than trying to get correct behavior out of split_uri. For reasons I don't entirely understand run_tests produces a single error (which can be seen in the [travis output](https://travis-ci.org/RDFLib/rdflib/jobs/223633917#L1417)).

tgbugs:
The source of the error is that I am not checking to see if the name produced by an existing prefix is valid. I think I have an easy fix.

Correct
```
@prefix : <http://example.org/here#> .
@prefix ex: <http://example.org/> .
@prefix ns: <http://example.org/ns#> .
@prefix ns1: <http://example.org/ns/p1/> .
@prefix ns2: <http://example.org/ns/> .

:x <http://example.org/1> "2" ;
    <http://example.org/1#1> "1" ;
    ex:_1 "rdf:_1 test" ;
    ns:_1 "1" ;
    ns1:p2 "1" .
```
Incorrect?
```
@prefix : <http://example.org/here#> .
@prefix ex: <http://example.org/> .
@prefix ns: <http://example.org/ns#> .
@prefix ns2: <http://example.org/ns/> .

:x ex:1 "2" ;
    ex:1#1 "1" ;
    ex:_1 "rdf:_1 test" ;
    ns:_1 "1" ;
    ns2:p1/p2 "1" .
```

tgbugs:
The way to make this work nicely with split_uri is to insert all split_uri values into another trie-like structure which sits on top of the trie so that split_uri can simply check if there is any longer namespace without having to traverse the the full set of keys. Will implement today or tomorrow.

tgbugs:
@joernhees @gromgull I've completed the changes we discussed. The travis error is a `ConnectionResetError: [Errno 104] Connection reset by peer` so not an issue with the build (I tested against 3.5 locally as well).

satra:
@tgbugs - thanks for these updates. it would be nice to get this PR merged.

regarding the travis error, can you restart that specific build?

gromgull:
i've restarted the travis job

gromgull:
And now it passes! I'll try to find some time to review this soon!

satra:
@gromgull - any chance of reviewing this sometime soon?

tgbugs:
I'm not entirely sure why the tests don't catch this but prefixes to full iris break the xml serializer parser roundtrip. I will look into it in more detail.

tgbugs:
I added roundtrip tests for n3 in addition to nt so that I could figure out what was going wrong with the xml serialization of qnames and found two issues (the bad serializations are below). I also have a question about the behavior of `QuotedGraph` which evokes errors in nearly every serializer, should they all be able to serialize quoted graphs? See comment at https://github.com/tgbugs/rdflib/blob/5692b9bbc375df2b01b49e5e3fdef7d8d60bfd13/test/test_n3_suite.py#L26.

From https://github.com/tgbugs/rdflib/blob/master/test/n3/n3-writer-test-29.n3
``` xml
<?xml version="1.0" encoding="UTF-8"?>
<rdf:RDF
   xmlns:ex="http://example.org/"
   xmlns:ns="http://example.org/ns#"
   xmlns:ns1="http://example.org/ns/p1/"
   xmlns:ns3="http://example.org/1#"
>
  <rdf:Description rdf:about="http://example.org/here#x">
    <ns1:p2>1</ns1:p2>
    <ns:_1>1</ns:_1>
    <ns3:1>1</ns3:1>
    <ex:1>2</ex:1>                                        <!-- issue here -->
    <ex:_1>rdf:_1 test</ex:_1>
  </rdf:Description>
</rdf:RDF>
```

From https://github.com/tgbugs/rdflib/blob/master/test/n3/n3-writer-test-30.n3
``` xml
<?xml version="1.0" encoding="UTF-8"?>
<rdf:RDF
   xmlns="http://example.org/here#"
   xmlns:full="http://example.org/full"
>
  <rdf:Description rdf:about="http://example.org/full">
    <x rdf:resource="http://example.org/here#y"/>
    <full: rdf:resource="http://example.org/full"/>
  </rdf:Description>
  <rdf:Description rdf:about="http://example.org/here#x">
    <full: rdf:resource="http://example.org/here#y"/>       <!-- issue here -->
    <y rdf:resource="http://example.org/full"/>
  </rdf:Description>
</rdf:RDF>
```

satra:
@gromgull and @joernhees - is there any timeline in place for both release of rdflib 5.0 and/or the merge of this PR? 

we are currently forced to rely on rdflib whl that @tgbugs released and that's making deployment through pypi and conda-forge more difficult. we would rather just have a standard release to rely on, but this PR is quite important for our use cases.

if it is of any help, my team would be happy to take on some of the rdflib maintenance responsibilities with more frequent release cycles.

nicholascar:
We (CSIRO, Australia's science agency) have a few users of rdflib and are happy to start getting involved with rdflib maintenance. We haven't looked deeply into things yet but use rdflib every day so would like to see some of the many PRs closed. Count us in for a v5.0 release.

test/test_n3_suite.py
```diff
@@ -18,6 +21,17 @@ def _get_test_files_formats():
             elif f.endswith('.n3'):
                 yield fpath, 'n3'
 
+def all_n3_files():
+    skiptests = [
+        'test/n3/example-lots_of_graphs.n3',  # the issue here seems to be with QuotedGraph, should this work?
```
joernhees: huh?

test/test_roundtrip.py
```diff
@@ -25,6 +28,8 @@
 
 
 SKIP = [
+    ('xml', 'test/n3/n3-writer-test-29.n3'),  # has predicates that cannot be shortened to strict qnames
+    ('application/rdf+xml', 'test/n3/n3-writer-test-29.n3'),  # has predicates that cannot be shortened to strict qnames
```
joernhees: huh?

rdflib/namespace.py
```diff
@@ -284,14 +284,23 @@ class NamespaceManager(object):
...
         self.bind("rdf", RDF)
         self.bind("rdfs", RDFS)
         self.bind("xsd", XSD)
 
     def reset(self):
         self.__cache = {}
+        self.__strie = {}
+        self.__trie = {}
+        for p, n in self.namespaces():  # repopulate the trie
+            insert_trie(self.__trie, str(n))
```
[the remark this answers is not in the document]
tgbugs: Possibly, can't tell without a few tests. If the trie is not repopulated here then many prefixes will not round trip correctly becuase serializer is essentially ignoring any user defined prefixes.

test/test_roundtrip.py
tgbugs: For example there is no valid way to turn `http://example.org/1#1` into a qname.
`ValueError: Can't split 'http://example.org/1#1'` is the error producing when running the test.
https://github.com/RDFLib/rdflib/blob/7abed6165d4ad235d41d1ddfb9dbbc2c64d6df0b/test/n3/n3-writer-test-29.n3#L17

Working on this I did find a new bug though!

tgbugs:
Not sure if entirely relevant, but I have some code for profiling the performance changes [here](https://github.com/tgbugs/pyontutils/blob/master/pyontutils/rdflib_profile.py). May not be worth the time to fiddle with it since I can't remember whether I got it working an an environment independent way.

test/test_roundtrip.py
tgbugs: Test output when not skipped
```
ERROR: test.test_roundtrip.test_cases(('nt', 'xml', 'test/nt/qname-02.nt'),)
  File "/home/tom/git/rdflib/rdflib/namespace.py", line 400, in compute_qname_strict
    namespace, name = split_uri(uri, NAME_START_CATEGORIES)
  File "/home/tom/git/rdflib/rdflib/namespace.py", line 600, in split_uri
    raise ValueError("Can't split '{}'".format(uri))
ValueError: Can't split 'http://example.org/#'

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/tom/git/rdflib/test/test_roundtrip.py", line 45, in roundtrip
    s = g1.serialize(format=testfmt)
  File "/home/tom/git/rdflib/rdflib/graph.py", line 932, in serialize
    serializer.serialize(stream, base=base, encoding=encoding, **args)
  File "/home/tom/git/rdflib/rdflib/plugins/serializers/rdfxml.py", line 68, in serialize
    bindings = list(self.__bindings())
  File "/home/tom/git/rdflib/rdflib/plugins/serializers/rdfxml.py", line 35, in __bindings
    prefix, namespace, name = nm.compute_qname_strict(predicate)
  File "/home/tom/git/rdflib/rdflib/namespace.py", line 404, in compute_qname_strict
    raise ValueError(message)
ValueError: This graph cannot be serialized to a strict format because there is no valid way to shorten http://example.org/#

ERROR: test.test_roundtrip.test_n3(('n3', 'xml', 'test/n3/n3-writer-test-29.n3'),)
  File "/home/tom/git/rdflib/rdflib/namespace.py", line 400, in compute_qname_strict
    namespace, name = split_uri(uri, NAME_START_CATEGORIES)
  File "/home/tom/git/rdflib/rdflib/namespace.py", line 600, in split_uri
    raise ValueError("Can't split '{}'".format(uri))
ValueError: Can't split 'http://example.org/1'

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/tom/git/rdflib/test/test_roundtrip.py", line 45, in roundtrip
    s = g1.serialize(format=testfmt)
  File "/home/tom/git/rdflib/rdflib/graph.py", line 932, in serialize
    serializer.serialize(stream, base=base, encoding=encoding, **args)
  File "/home/tom/git/rdflib/rdflib/plugins/serializers/rdfxml.py", line 68, in serialize
    bindings = list(self.__bindings())
  File "/home/tom/git/rdflib/rdflib/plugins/serializers/rdfxml.py", line 35, in __bindings
    prefix, namespace, name = nm.compute_qname_strict(predicate)
  File "/home/tom/git/rdflib/rdflib/namespace.py", line 404, in compute_qname_strict
    raise ValueError(message)
ValueError: This graph cannot be serialized to a strict format because there is no valid way to shorten http://example.org/1


```

gromgull:
Hmm: 

```py
In [23]: g=rdflib.Graph()

In [24]: g.namespace_manager.compute_qname(rdflib.URIRef('http://example.org/'))
[...]
ValueError: Can't split 'http://example.org/'

In [26]: g.namespace_manager.compute_qname(rdflib.URIRef('http://example.org/asd'))
Out[26]: ('ns1', rdflib.term.URIRef('http://example.org/'), 'asd')

In [27]: g.namespace_manager.compute_qname(rdflib.URIRef('http://example.org/'))
Out[27]: ('ns1', rdflib.term.URIRef('http://example.org/'), '')
```

I wish empty localnames wasn't allowed :D - but seeing as it is - this should probably work the first time?

gromgull:
I did a very scientific test of this: 

```py

import resource

def using(point=""):
    usage=resource.getrusage(resource.RUSAGE_SELF)
    return '''%s: usertime=%s systime=%s mem=%s mb
           '''%(point,usage[0],usage[1],
                (usage[2])/1000000.0 )

print(using('start'))

import rdflib

print(using('post import'))

g = rdflib.ConjunctiveGraph()
g.load('<some big nq file>', format='nquads')

print(using('post parse'))

g.serialize('out.n3', format='ttl')

print(using('end'))
```

and got for a file with 950k triples: 

```
(env37) gunnar@trio:~/projects/rdflib/rdflib (tgbugs-master)$ python speed.py
start: usertime=0.015456 systime=0.0054859999999999996 mem=5.328896 mb

post import: usertime=0.18648199999999998 systime=0.035741999999999996 mem=19.042304 mb

post parse: usertime=79.981722 systime=1.170648 mem=768.356352 mb

end: usertime=182.797671 systime=1.830292 mem=818.016256 mb

(env37) gunnar@trio:~/projects/rdflib/rdflib (tgbugs-master)$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
(env37) gunnar@trio:~/projects/rdflib/rdflib (master)$ python speed.py
start: usertime=0.013701999999999999 systime=0.0050539999999999995 mem=5.234688 mb

post import: usertime=0.164597 systime=0.031455 mem=19.304448 mb

post parse: usertime=84.008706 systime=1.563531 mem=768.69632 mb

end: usertime=194.934963 systime=2.739493 mem=818.282496 mb
```

i.e. only a very modest increase in runtime (6%) and almost no increase in memory - I would say this si ok.

tgbugs:
If you want to see the horror of shortening every url you can take a look at https://github.com/tgbugs/pyontutils/blob/39a529d48774af426cf56e7efb7a6a502c8afb1a/pyontutils/ttlser.py#L751. I do it at serialization time, but you have already imagined what could happen if it was done by default.

tgbugs:
Here are some results from cprofile.
When testing the master branches against each other I see at most a 2% slowdown on a 3.6 interpreter.
```
tgbugs-master
PYTHONPATH=~/git/rdflib:~/git:~/git/pyontutils ttlfmt --profile --vanilla uberon.ttl
parse          139563971 function calls (135718976 primitive calls) in 75.614 seconds
/home/tom/git/rdflib/rdflib/plugins/parsers/notation3.py:1326(qname)
triple count for uberon.ttl: 657459
serialize          116357808 function calls (115835190 primitive calls) in 67.951 seconds

rdflib-master
PYTHONPATH=~/git/NOFORK/rdflib:~/git:~/git/pyontutils ttlfmt --profile --vanilla uberon.ttl
parse          139564723 function calls (135719728 primitive calls) in 75.421 seconds
/home/tom/git/NOFORK/rdflib/rdflib/plugins/parsers/notation3.py:1326(qname)
triple count for uberon.ttl: 657459
serialize          112256369 function calls (111735149 primitive calls) in 66.740 seconds
```

On 3.7 I also see at most a 3% slowdown.
```
PYTHONPATH=~/git/rdflib:~/git:~/git/pyontutils ttlfmt --profile --vanilla uberon.ttl
parse          139501367 function calls (138294026 primitive calls) in 72.776 seconds
triple count for uberon.ttl: 657459
serialize          116414101 function calls (115891566 primitive calls) in 64.873 seconds

PYTHONPATH=~/git/NOFORK/rdflib:~/git:~/git/pyontutils ttlfmt --profile --vanilla uberon.ttl
parse          139500563 function calls (138293222 primitive calls) in 71.756 seconds
triple count for uberon.ttl: 657459
serialize          112326867 function calls (111805647 primitive calls) in 62.816 seconds
```

On pypy3 cprofile doesn't work, but here are some even less scientific times which show a 7% ish slowdown more in line with what @gromgull sees.
```
time PYTHONPATH=~/git/rdflib:~/git:~/git/pyontutils ttlfmt --vanilla uberon.ttl
uberon.ttl

real	1m2.768s
user	1m2.018s
sys	0m0.696s

time PYTHONPATH=~/git/NOFORK/rdflib:~/git:~/git/pyontutils ttlfmt --vanilla uberon.ttl
uberon.ttl

real	0m58.373s
user	0m58.038s
sys	0m0.323s
```

Testing 4.2.2 vs tgbugs-master in different venvs I see a significant speed up. But that speed up is also already present in rdflib-master.
```
(neurdflib-5.0.0) tom@athena $ ttlfmt --profile --vanilla uberon.ttl
parse          139563871 function calls (135718876 primitive calls) in 77.179 seconds
triple count for uberon.ttl: 657459
serialize          116417247 function calls (115894672 primitive calls) in 67.372 seconds

(rdflib-4.2.2) tom@athena $ ttlfmt --profile --vanilla uberon.ttl
parse          169185322 function calls (159852788 primitive calls) in 90.640 seconds
/home/tom/.local/share/virtualenvs/rdflib-4.2.2-f6A_7WrF/lib/python3.6/site-packages/rdflib/term.py:202(__hash__)
triple count for uberon.ttl: 657459
serialize          158751444 function calls (155606478 primitive calls) in 90.051 seconds
```

joernhees:
Ok, so let's zoom out a bit, try to summarize this and maybe make a suggestion how to move forward and finally integrate this.

This PR is now in a state where it is as efficient as the old implementation. It however computes "QNames" in a much more liberal way: they're now actually a lot closer to CURIEs, which (as already discovered, e.g., #867) has high potential to cause problems in other systems that still expect QNames.

IMO RDFLib should be able to deal with such prefixes, but by default try not to cause too many problems. This means: we should always be able to parse them (and keep them when defined explicitly), but when (auto-)generating new prefixes or shortening URIs, we should by default generate QNames and not use the full range of CURIEous possibilities. We should however offer some way to users, to explicitly say "go full CURIE on me". 

To me this would also mean: no old tests are skipped / fail, new tests for non QName CURIEs should parse. On round-trip serialize, by default they should fall back to safer QNames, unless explicitly set to CURIEous mode.

Before talking about how to achieve this, i'd ask: does everyone agree / did i miss something? @gromgull @tgbugs @satra

tgbugs:
I think that covers everything.

It seems more user friendly to me to always return the prefixes that the user provided and fail loudly if there is an issue with converting between formats. This seems to adhere more closely to the principle of least surprise, especially when the user does not fully know or understand the spec. This way the errors are also aligned with action that would need to be taken regardless. Essentially I'm suggesting that if the user wants their prefixes changed, they should tell us, except in cases where we tell them that if they do not changes their prefixes then they will be producing malformed output.

Other than that the proposed default behavior seems sane.

I personally have very limited use cases for auto generated qnames, much less auto generated curies, and would prefer the qname spec because it is more consistent (aside from cases where a valid uri is invalid in xml and also cannot be shortened) and compatible more widely.

gromgull:
Sounds good to me! 

The most important part is not failing any tests - so we don't surprise break things that used to work (at least not on purpose! :D )

satra:
@joernhees and @tgbugs - for me these things capture the intent of things:

> we should by default generate QNames and not use the full range of CURIEous possibilities. We should however offer some way to users, to explicitly say "go full CURIE on me".

> It seems more user friendly to me to always return the prefixes that the user provided and fail loudly if there is an issue with converting between formats.

> The most important part is not failing any tests - so we don't surprise break things that used to work (at least not on purpose! :D )

as long as we support these principles, i'm good with the outcome.

tgbugs:
Ok, further updates on this. The skipped tests are because I have added the entire n3 suite to the round trip tests. None of the tests that I am skipping are run at all on the master branch. There are four tests that are being skipped.

1. Round tripping of `test/n3/example-lots_of_graphs.n3` which the current master branch fails if run against. This is because as far as I can tell every serializer besides n3 does not support correct serialization of quoted graphs. This issue is not related to this pull request.
2. Round tripping of `test/n3/strquot.n3` through any xml parser (xml, trix) because serializes to characters that are explicitly forbidden by the xml spec. These will never be able to round trip as long as an xml 1.0 parser is used. Both examples below cause the sax parser to barf.
```xml
<plainLiteral>\"&#13;	</plainLiteral>
```
```xml
<plainLiteral>&quot;&#x07;&x08;&#x0c;&#13;&#x9;&#x0b;</plainLiteral>
```
3. `test/n3/n3-writer-test-29.n3` round tripping through rdf+xml because it contains predicates that have no valid qname representation and whose full form can't be used directly as tags because they contain `/` in them. I will look into whether it is possible to escape the `/`s.
4. `test/nt/qname-02.nt` same as 3.

satra:
@gromgull - any progress on merging this? are there any current blockers?

effigies:
@nicholascar appears to be the most active maintainer on this project. Pinging him.

nicholascar:
OK, we're going through an rdflib management change and when that settles down and all roles are known - only a few days off - we'll be reviewing all open PRs and dealing with as many as we can. Sorry this has waited for so long but just give it one week more please! Nick

tgbugs:
I'm working on rebasing this PR so that it is easier to review and merge.

nicholascar:
Opening and closing to trigger Travis

Issue #642, opened 2016-07-27
investigate incompatiblity with html5lib > 1.0b8


joernhees:
https://travis-ci.org/RDFLib/rdflib/jobs/147619766

PR #650, opened 2016-08-30
Support newer transport_encoding parameter in html5lib parser call

redapple:
Should fix https://github.com/RDFLib/rdflib/issues/642


joernhees:
thanks, ping to @iherman in case this needs to go anywhere else...

iherman:
That is indeed the only place where html5lib is used. That being said, I have not looked at html5lib and its usage for many years now, I am not sure what that flag does... But if passes all the tests, it should be fine:-)

PR #654, opened 2016-10-11
docs: clarify the use of an identifier when persisting a triplestore

wimmuskee:


joernhees:
i like improving the docs, but this entirely removes the "default" case... could you maybe repeat your example underneath?

wimmuskee:
Well, simple using the Graph to persist and open does not work without an identifier (that would be the default use case for me).

So, 
```python
from rdflib import Graph
from rdflib import URIRef, Literal
from rdflib.namespace import RDFS

g = Graph("Sleepycat")
g.open("mystore", create=True)
g.add((URIRef("https://github.com/RDFLib"),RDFS.label,Literal("RDFlib Github homepage", lang="en")))
print("triples added: " + str(len(g)))
# prints 1
g.close()

g = None

g = Graph("Sleepycat")
g.open("mystore", create=False)
print("triples added: " + str(len(g)))
# prints 0
g.close()
```

gromgull:
It works just fine, it just creates a new bnode as identifier each time. There are valid use-cases for that! 

(but really, use a conjunctive graph, then the default is to read all graphs)

wimmuskee:
I'm sure it works as intended (and there are use cases for that), but I did not find the current docs explaining this unambiguously. 
In my perspective as a user, my example is the default use case (create-persist-open). If this is not the intended case, disregard my pull request.

gromgull:
reading it again, I think this is fine! Thanks for the contribution!

PR #661, opened 2016-10-20
fix: namespace use for turtle-based serializers

satra:
this time i do think it addresses #632 and #660 better

@gromgull - could you please take a look? the changes are now restricted to the turtle serializer.

### DIFF
--- a/rdflib/plugins/serializers/turtle.py
+++ b/rdflib/plugins/serializers/turtle.py
@@ -228,19 +228,26 @@ class TurtleSerializer(RecursiveSerializer):
         if not isinstance(uri, URIRef):
             return None
 
-        parts = None
+        pfx = self.store.store.prefix(uri)
 
-        try:
-            parts = self.store.compute_qname(uri, generate=gen_prefix)
-        except:
+        if pfx is not None:
+            parts = (pfx, uri, '')
+        else:
+            ns_list = sorted(self.namespaces.items())
+            old_length = 0
+            for prefix, ns_uri in ns_list:
+                if uri.startswith(ns_uri) and len(ns_uri) > old_length:
+                    local_name = uri.replace(ns_uri, '')
+                    if local_name.count('/') > 0 or local_name.count('#') > 0:
+                        continue
+                    parts = (prefix, ns_uri, local_name)
+                    old_length = len(ns_uri)
+                    pfx = prefix
 
-            # is the uri a namespace in itself?
-            pfx = self.store.store.prefix(uri)
-
-            if pfx is not None:
-                parts = (pfx, uri, '')
-            else:
-                # nothing worked
+        if pfx is None:
+            try:
+                parts = self.store.compute_qname(uri, generate=gen_prefix)
+            except:
                 return None
 
         prefix, namespace, local = parts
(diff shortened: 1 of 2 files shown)


rdflib/plugins/serializers/turtle.py
```diff
@@ -228,19 +228,26 @@ def getQName(self, uri, gen_prefix=True):
...
+            ns_list = sorted(self.namespaces.items())
+            old_length = 0
+            for prefix, ns_uri in ns_list:
+                if uri.startswith(ns_uri) and len(ns_uri) > old_length:
+                    local_name = uri.replace(ns_uri, '')
+                    if local_name.count('/') > 0 or local_name.count('#') > 0:
+                        continue
+                    parts = (prefix, ns_uri, local_name)
+                    old_length = len(ns_uri)
+                    pfx = prefix
```
joernhees: this looks like a O(n) algorithm (n is the number of namespaces), while the previous one was essentially a lookup, so a O(1)!?!

satra: i can implement a cache to reduce this complexity a bit, but in general the longest substring match is not going to be O(1). do you have any suggestions?

tgbugs:
On further review this does not fix cases where there is a call to `graph.namespace_manager.qname` (for example I use such a call [here](https://github.com/tgbugs/pyontutils/blob/master/utils.py#L197)). The changes in https://github.com/RDFLib/rdflib/pull/649 also allow `namespace_manager.qname` to return consistent results. If we don't want to support arbitrary prefix matching inside the namespace manager then I'm happy to maintain my changes from https://github.com/RDFLib/rdflib/pull/649 as a patch.

satra:
@tgbugs - i have a local version where i had implemented the change in `namespace.py` by modifying the `bind` function and then decided to do it during serialization as i wasn't sure of all the things that the namespace manager would touch. but yes, if you are calling `namespace_manager.qname` directly, this will not work. 

it would be good to figure out a robust solution that addresses all these use cases, while still being efficient. can you check if the tests that i added here pass with your changes?

tgbugs:
@satra I'm still seeing the same error I encountered before (log below). Let me take a closer look and see if I can figure out what is going on.

```
FAIL: testNegative (test.test_rdfxml.ParserTestCase)
  File "/home/tom/git/downstream/rdflib/build/py3_testing/test/test_rdfxml.py", line 190, in testNegative
    num_failed, 0, "Failed: %s of %s." % (num_failed, total))
AssertionError: 2 != 0 : Failed: 2 of 41.
-------------------- >> begin captured logging << --------------------
parser_rdfcore: INFO: Failed: 'http://www.w3.org/2000/10/rdf-tests/rdfcore/rdfms-rdf-id/Manifest.rdf#error001'

parser_rdfcore: INFO: Failed: 'http://www.w3.org/2000/10/rdf-tests/rdfcore/rdfms-syntax-incomplete/Manifest.rdf#error001'

--------------------- >> end captured logging << ---------------------
```

tgbugs:
Looks like the test failure was caused by the unrelated change that I made to `NAME_START_CATEGORIES` which I think is no longer needed based on the changes @satra made to the turtle serializer. I have removed those changes and reopened by [original pull request](https://github.com/RDFLib/rdflib/pull/649), if the new commit doesn't show up I will create a new version.

satra:
closing in favor of the more general #649

Issue #413, opened 2014-07-28
Think about logging in rdflib


gromgull:
Currently logging in rdflib is a bit random: 

```
rdflib/__init__.py:_LOGGER = logging.getLogger("rdflib")
rdflib/extras/infixowl.py:    logger = logging.getLogger(__name__)
rdflib/graph.py:_logger = logging.getLogger(__name__)
rdflib/namespace.py:_logger = logging.getLogger(__name__)
rdflib/plugins/sleepycat.py:_logger = logging.getLogger(__name__)
rdflib/term.py:_LOGGER = logging.getLogger(__name__)
```

We should: 
- Make a list of loggers we should use. 
- DOCUMENT this list :) 
- Configure logging to be output to stderr in commandline tools, rdfpipe and the like.

PR #662, opened 2016-10-25
Remove logger output

david4096:
Considering the version is accessible on the module, we would like to remove this logging.

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -83,9 +83,6 @@ except ImportError:
     import warnings
     warnings.warn('__main__ already imported', ImportWarning)
     del warnings
-    
-logger = logging.getLogger(__name__)
-logger.info("RDFLib Version: %s" % __version__)
 
 
 try:


david4096:
Hey, I got a +1, thanks! Anything I might need to do to get this very minor change merged?

joernhees:
:( i somehow liked the output, especially in interactive sessions as it saved me tons of times already... maybe it's a bit too close to an rdflib dev workflow though...

see #413

joernhees:
ok, so i thought about this again: i agree that always putting that message there might be annoying for some people, but in interactive sessions i have the feeling that it in the end helps us a lot to make version info very easily accessible to non rdflib experts. More people will realize if they have an old version, and they might even just write it into bug reports here.

So what i'll do (see 4bb4362):
- in interactive sessions i weigh the advantages higher than the minor annoyance of people. I'll make the check harder (it will also check if stdout is a tty, if not no INFO message)
- in productive code i weigh the advantage lower than the bigger annoyance of people who can't correctly configure python standard logging. That's why i'll make it a debug statement in those cases (they're hidden by default, unless you say you want debug output... if you want that, you'll get a version number)

Please open another issue in case that's not ok for you...

PR #663, opened 2016-10-30
namespace.py include colon in ALLOWED_NAME_CHARS

tgbugs:
This commit updates the behavior of is_ncname and
split_uri to match the behavior described here:
https://www.w3.org/TR/html4/types.html#type-name
(found via
http://stackoverflow.com/questions/2053132/is-a-colon-safe-for-friendly-url-use).
This results in the correct splitting of uris such as
http://neurolex.org/wiki/FMA:7191. Extra note that not
all user agents convert : -> %3A since they are not
required to by the spec (RFC3986).


gromgull:
This makes four of the rdfcore tests fail :(

gromgull:
This was also changed in turtle at some point, after we got out turtle parser. We would also need to update the turtle parser to be able to parse those localnames with `:`s

tgbugs:
When I originally created this I had not chased down the sources of the failures, but as you point it it does break the behvior of the core rdf specification (I have tracked this down for changes I make to NAME_START_CATEGORIES += ["Nd"] which break similar things). I'm not entirely sure what to do about it, but I'm fairly certain that other parsers like the ones used by OWLAPI do have this behavior.

tgbugs:
Removing the `:` when inside `is_ncname` allows the tests to pass and produces the desired results when serializing. The parser doesn't seem to have any issues reading something like `NLXWIKI:FMA:1234567` -> `http://neurolex.org/wiki/FMA:1234567` so I think this might be a reasonable solution.

gromgull:
Looks good! (And apologies for the slow turn around here)

joernhees:
hmmpf, this breaks using our qnames on virtuoso sparql endpoints...

Issue #648, opened 2016-08-24
date serialization failure


hsolbrig:
The following code:

``` python
import rdflib
g = rdflib.Graph()
y = ' <http://a.example/X> <http://a.example/p> "2016-01-01T00:00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime>,' \
    '                                           "2016-01-01T00:00:00Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .'

p=g.parse(data=y, format="turtle")
print(p.serialize(format="turtle").decode('utf-8'))
```

Causes an error:

```
  File "...lib/python3.5/site-packages/rdflib/term.py", line 822, in __gt__
    return self.value > other.value
TypeError: can't compare offset-naive and offset-aware datetimes
```

This is because the serializer is trying to sort the output tuples and the two different forms of date/time can't be compared.

Recommend the following change - line 822, term.py

``` python
            if self.value != None and other.value != None:
                try:
                    return self.value > other.value
                except TypeError:
                    return str(self.value) > str(other.value)
```

The idea being that, if the native comparison doesn't work, fall back on the string comparison.

gromgull:
I agree that not being able to serialise this things is very bad, but this fix introduces weird unexpected behaviour for everyone who wants semantically meaningful comparisons of literals (i.e. SPARQL) 

A fix must be to pass a custom comparator into the sort call inside the serialiser, and do the fall back there.

gromgull:
same issue: https://github.com/RDFLib/rdflib/issues/613

PR #666, opened 2016-11-14
Fix date serialization failure

alexisbergeron22:
If the first comparison doesn't work, it goes to string comparison.
For the issue #648


gromgull:
Thanks a lot, but I fixed this in the serialiser instead!

Issue #672, opened 2016-12-02
Long SPARQL queries need POST instead of GET


SeguinBe:
Hello,

Getting into RDF and this lib is awesome!

So right now, when you instantiate a `SPARQLStore` as backend, it's not possible to perform long queries (like retrieving information about 100 long URIs at the same time), because of the size limitation of GET requests.

Trying something like 
```python
import SPARQLWrapper
graph = Graph('SPARQLUpdateStore', identifier='http://localhost:8890/WGA') 
graph.store.setMethod(SPARQLWrapper.POST)
```
does not work because the HTTP method is reset (with `resetQuery`) before every query.

That make it impossible to be able to perform a query involving an important set of URI :-(

joernhees:
closed in #673

PR #673, opened 2016-12-02
Adding the option of using POST for queries in SPARQLStore

SeguinBe:
Solution to #672

Just adding a small parameter to be able to use POST for queries in order to to perform long requests.

Previously :
```python
graph = Graph('SPARQLUpdateStore', identifier='http://localhost:8890/WGA') 
graph.query('<VERY LONG STRING>')
# Query get cropped because it's done with HTTP GET
```

Now :
```python
from rdflib.plugins.stores.sparqlstore import SPARQLUpdateStore
store = sparqlstore.SPARQLUpdateStore(query_as_post=True)
graph = Graph(store, identifier='http://localhost:8890/WGA') 
graph.query('<VERY LONG STRING>')
# Works!
```

Tiny changes based on [postAsEncoded param in SPARQLUpdateStore](https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/stores/sparqlstore.py#L763)

### DIFF
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -239,7 +239,7 @@ class SPARQLStore(NSSPARQLWrapper, Store):
                  sparql11=True, context_aware=True,
                  node_to_sparql=_node_to_sparql,
                  node_from_result=_node_from_result,
-                 query_as_post=False,
+                 default_query_method=GET,
                  **sparqlwrapper_kwargs):
         """
         """
@@ -260,7 +260,7 @@ class SPARQLStore(NSSPARQLWrapper, Store):
         self.context_aware = context_aware
         self.graph_aware = context_aware
         self._timeout = None
-        self.query_as_post = query_as_post
+        self.default_query_method = default_query_method
 
     # Database Management Methods
     def create(self, configuration):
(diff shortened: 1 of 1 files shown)


rdflib/plugins/stores/sparqlstore.py
```diff
@@ -239,6 +239,7 @@ def __init__(self,
                  sparql11=True, context_aware=True,
                  node_to_sparql=_node_to_sparql,
                  node_from_result=_node_from_result,
+                 query_as_post=False,
```
joernhees: could you rename this to `default_query_method=GET`?

rdflib/plugins/stores/sparqlstore.py
```diff
@@ -259,6 +260,7 @@ def __init__(self,
         self.context_aware = context_aware
         self.graph_aware = context_aware
         self._timeout = None
+        self.query_as_post = query_as_post
```
joernhees: `self.query_method = default_query_method`

rdflib/plugins/stores/sparqlstore.py
```diff
@@ -322,6 +324,7 @@ def query(self, query,
                    " ".join(self.node_to_sparql(initBindings[x]) for x in v))
 
         self.resetQuery()
+        self.setMethod(POST if self.query_as_post else GET)
```
joernhees: `self.setMethod(self.query_method)`

SeguinBe:
Yeah having something automatic might be good for users, because the current error message is really not hepful right now (complains about a weird character).

Also, it should be noted that this extension doesn't allow to make a POST query with the sparql query as POST data directly. By default it uses an URL_ENCODED scheme. Not sure if it might be useful at one point though.

Issue #671, opened 2016-11-25
Put html5lib and/or sparqlwrapper in extra_requires


arthurlogilab:
I noticed using the debian package https://packages.debian.org/jessie/python-rdflib with --no-install-recommends that python-pkg-ressources does not find html5lib and sparqlwrapper and thus fails. Are these really required for rdflib to be used. If not they could probably be put in extra_requires (cf docutils example in http://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=requires#overview & http://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=requires#requirements-parsing)

Starting the same discussion on debian packaging, will add URL here.

arthurlogilab:
cf http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845643

chrysn:
the referenced debian bug has sparked a follow-up in keepalive (https://github.com/RDFLib/sparqlwrapper/pull/79).

what's interesting here is that in python3, html5lib is already only in `tests_require` (and thus even weaker than extras_require). the correct way to declare the dependencies in my opinion is that both sparqlwrapper and html5lib should be in tests_require, and there should be an `extras_require={'sparql': ['SPARQLWrapper'], 'html': ['html5lib']}` ('html' could be split into 'rdfa' and 'microformats' both depending on 'html5lib', but that's probably too fine grained).

PR #675, opened 2016-12-07
Relax dependencies on html5lib and SPARQLWrapper

chrysn:
Demote html5lib and SPARQLWrapper to from install_requires to tests_require and extras_require

Closes: https://github.com/RDFLib/rdflib/issues/671

----

i tried to balance code changes against code duplication here; depending on your style preferences, there are sure to be other ways of phrasing this.

this pull-request is similar to https://github.com/RDFLib/sparqlwrapper/pull/79.

(side note: i thought that the python-version-dependent sections could be simplified [as showin in a bug report example](https://bugs.debian.org/834133), but that [still seems to be difficult to get right in presence of old versions](https://hynek.me/articles/conditional-python-dependencies/)).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,8 @@ if sys.version_info[0] >= 3:
     kwargs['tests_require'] = ['html5lib', 'networkx']
     kwargs['requires'] = [
         'isodate', 'pyparsing',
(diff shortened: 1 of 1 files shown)


gromgull:
Thanks!

Issue #223, opened 2012-07-06
rdflib.collections issues


gromgull:
Adding the same element twice to a collection triggers an exception when serializing the graph, but this shouldn't be a problem:

```
/usr/local/lib/python2.7/dist-packages/rdflib/collection.pyc in __len__(self)
     77         links=set()
     78         for item in self.graph.items(self.uri):
---> 79             assert item not in links,"There is a loop in the RDF list! (%s has been processed before)"%item
     80             links.add(item)
     81             count += 1

AssertionError: There is a loop in the RDF list! (http://www.w3.org/2000/01/rdf-schema#Class has been processed before)
```

There is also no way to close the list, appending RDF.nil adds another bnode with first->nil and rest->nil.

wwaites:
This one just bit me

gromgull:
@wwaites if you find a fix in the next weeks we can get it into 4.2.2!

wwaites:
Near as I can tell, it is safe to remove this assertion. Graph.items [checks for recursion](https://github.com/RDFLib/rdflib/blob/53258d3a57e9c4b22fac9372ff3dbaa6296b8df7/rdflib/graph.py#L801) on the list structure itself . This check is simply incorrect for [rdf:List](https://www.w3.org/TR/2004/REC-rdf-schema-20040210/#ch_collectionvocab) that is quite unrestricted. There is no type that I know of with set semantics so I can't see where the assertion can ever be correct even for containers.

PR #690, opened 2017-01-19
Allow duplicates in rdf:List

wwaites:
The check causing the problem in #223 is incorrect. This addresses the first half of the issue, allowing duplicates in lists. It does not address closing the list.

### DIFF
--- a/rdflib/collection.py
+++ b/rdflib/collection.py
@@ -82,12 +82,7 @@ class Collection(object):
 
     def __len__(self):
         """length of items in collection."""
-        count = 0
-        links = set()
-        for item in self.graph.items(self.uri):
-            links.add(item)
-            count += 1
-        return count
+        return len(list(self.graph.items(self.uri)))
 
     def index(self, item):
         """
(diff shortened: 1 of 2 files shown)


gromgull:
The closing the list issue is less important. It's now always closed. Leaving an open list is a very special case, people who want that can probably fix it themselves.

wwaites:
Coveralls seems a bit confused. Removing code and adding a test should only increase test coverage!

gromgull:
Actually - I find coveralls almost completely useless - we should just remove it :)

wwaites:
Seems so. Anyhow, are you satisfied that this is the right thing to do? Should I merge it?

gromgull:
The check was added to avoid an infinite loop in cases of degenerate lists I guess, but it's totally misguided, since it checks for duplicate values, not duplicate list elements. 

The correct (?) test is here: https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L801-L802

The whole __len__ method could then just be `len(list(self.graph.items(self.uri)))` ?

wwaites:
Yes, you're correct of course. Would have been obvious if I had changed the test to do

    assert len(c) == 3

now done.

gromgull:
👍 Thanks!

PR #696, opened 2017-01-23
removing pyparsing version requirement

nateprewitt:
The versioning requirement was put in place 4 years ago to handle older
versions of Python. setuptools no longer supports pyparsing < 2.1, so
we'll need to adhere to that to continue using pip or put a hard version
on pip. This seems like the more reasonable approach.

### DIFF
--- a/requirements.py2.txt
+++ b/requirements.py2.txt
@@ -1,5 +1,5 @@
 flake8
 html5lib
 isodate
-pyparsing<=1.5.7
+pyparsing
 SPARQLWrapper


nateprewitt:
While we're here, there's still code for this 2.5 requirement in [setup.py](https://github.com/RDFLib/rdflib/blob/14243b6a48fa22192ebdb55702f9c4031252821c/setup.py#L61-L64). It can probably be left there for any users trying to install via setup.py, but 2.5 appears to not have been supported for almost 2 years now.

gromgull:
thanks!

gromgull:
I would say we should clean up setup.py as well - it's not super urgent though.

PR #697, opened 2017-01-24
updating deprecated testing syntax

nateprewitt:
While trying to wade through Travis' logs yesterday, I noticed there's 700+ lines of DeprecationWarnings from the test suite. This makes it fairly difficult to find things, or easily scroll between sections of the log. I updated the warnings to the 2.5+ compatible alternatives, which should make things more readable.


gromgull:
Thanks a lot! I really wanted to do this - but it was too much manual labour :)

nateprewitt:
No problem, I have to give `sed` the credit here, it did all the heavy lifting  :)

PR #700, opened 2017-01-24
add a NTSerializer sub-class for nt11

### LINKED ISSUE #695: Add a RDF 1.1 ntriples serializer/parser
They redefined NT to be utf-8, not ascii with escaping between the RDF Testcase doc, and the RDF 1.1 version.

gromgull:
NTriples in RDF1.1 are utf8 encoded.
This is backwards compatible, but you can have nicer output if you
know the consumer will support nt11/utf-8

Fixes #695


joernhees:
;)

PR #701, opened 2017-01-25
More aliases for common formats

joernhees:
`format='ttl'` and `'ntriples'` has bitten me too often, especially considering that `'turtle'` and `'nt'` work


joernhees:
yeah, coveralls now seems to have scientific notation support (mixing e and %, that's a first for me) in case the coverage drop can't be small enough :D

Issue #436, opened 2014-11-22
ConjunctiveGraph doesn't handle parsing datasets with default graphs properly


niklasl:
When ConjunctiveGraph.parse is called, it wraps its underlying store in a regular Graph instance. This causes problems for parsers of datasets, e.g. NQuads, TriG and JSON-LD.

Specifically, the triples in the default graph of a dataset haphazardly end up in bnode-named contexts.

Example:

``` python
import sys
from rdflib import *

cg = ConjunctiveGraph()
cg.parse(format="nquads", data=u"""
<http://example.org/a> <http://example.org/ns#label> "A" .
<http://example.org/b> <http://example.org/ns#label> "B" <http://example.org/b/> .
""")
assert len(cg.default_context) == 1 # fails
```

While I've attempted to overcome this by using the underlying `graph.store` in these parsers, they cannot access the `default_context` of ConjunctiveGraph through this store. It is there in the underlying store, but its identifier is inaccessible to the parser without further changes to the parse method of ConjunctiveGraph.

This becomes tricky because the contract for ConjunctiveGraph:s parse method is:

```
    Parse source adding the resulting triples to its own context
    (sub graph of this graph).

    See :meth:`rdflib.graph.Graph.parse` for documentation on arguments.

    :Returns:

    The graph into which the source was parsed. In the case of n3
    it returns the root context.
```

I am not sure how we can change this behaviour, since client code may rely on this. We could either add a new method, e.g. `parse_dataset`, or a flag. That would not be obvious to all users though, and somehow I would like to change the behaviour to handle datasets as well. It is always possible to get/create a named graph from a conjunctive graph and parse data into that.

I have gotten further by adding `publicID=cg.default_context.identifier` to the parse invocation. This causes the TriG parser to behave properly (and it is easy to adapt the nquads parser to work from there on). But I am not sure if this is a wise solution to the problem.

I'll mull more on this given time, but it would be good to have more people consider a proper revision of the parsing mechanism for datasets.

This underlies the problems described in #432 and #433 (and is related #428).

(Obviously, this in turn causes the serializers for the same formats to emit unexpected bnode-named graphs when data has been read through these parsers.)

niklasl:
It might make sense that one should simply parse into the `default_context` of a `ConjunctiveGraph` or `Dataset`, like:

```python
cg = rdflib.ConjunctiveGraph()
cg.default_context.parse(data=data, format='trig')
print cg.serialize(format='trig')
```

[...]

PR #702, opened 2017-01-26
Revert "skip round-trip test, unfixable until 5.0"

joernhees:
This reverts commit 7bec219df4ab78afcd17f3611bba10aa9ee05b37, see #431 and #699

depends on #436

This should remind us to not forget this in 5.0.0


nicholascar:
This wasn't merged into 5.0.0 as it's potentially still an issue as of July 2020. The test still fails on 5.0.0.

nicholascar:
The test that this PR is aiming to re-enable is wrong.

Each time the `ConjunctiveGraph` is re-fed the default and Blank Node-identified graphs [in the loop](https://github.com/RDFLib/rdflib/blob/master/test/test_trig.py#L136-L138) it correctly assigns new BNs to them. So this means that indeed the count of `thing_b` etc. will be > 1.

So, I think the test should be changed as the behavior is correct and, if so, this PR can be merged with the updated test.

aucampia:
I'm going to close this PR after 2022-08-22 as I don't think it is relevant anymore:

- We are using xfails now, and any test that has xfail marker that passes will be reported as xpass.
- This PR still references nosetest which we no longer use.

PR #711, opened 2017-02-06
make namespace module flake8-compliant, change exceptions in that mod…

adamhadani:
- Updates Exceptions raised in this module to specific ones so possible to catch them without a "except Exception" catch-all in application code that uses this library
- Some formatting updates to make this module [flake8](http://flake8.pycqa.org/en/latest/) compliant! 
- Mention `doctest-ignore-unicode` dependency in run_tests.py docs for running tests, update it to suggest using pip (without sudo - use virtualenv!)

I Opted to use built in exception types that seemed most appropriate, ideally we should define our own custom error hierarchy, but for now this seems a good start.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -2,6 +2,20 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import logging
+
+import os
+from unicodedata import category
+
+from six import string_types
+from six import text_type
+
+from six.moves.urllib.request import pathname2url
+from six.moves.urllib.parse import urldefrag
+from six.moves.urllib.parse import urljoin
+
+from rdflib.term import URIRef, Variable, _XSD_PFX, _is_valid_uri
+
 __doc__ = """
 ===================
 Namespace Utilities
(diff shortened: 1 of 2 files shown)


adamhadani:
ping @gromgull for your kind consideration

rdflib/namespace.py
```diff
@@ -120,7 +120,7 @@ def __getattr__(self, name):
             return self.term(name)
 
     def __repr__(self):
-        return "Namespace(%s)"%text_type.__repr__(self)
+        return "Namespace(%s)" % text_type.__repr__(self)
```
joernhees: can't we just replace this with `"Namespace(%r)" % text_type(self)`?

rdflib/namespace.py
```diff
@@ -150,8 +150,7 @@ def format(self, *args, **kwargs):
         return URIRef(text_type.format(self, *args, **kwargs))
 
     def __repr__(self):
-        return "URIPattern(%r)"%text_type.__repr__(self)
-
+        return "URIPattern(%r)" % text_type.__repr__(self)
```
joernhees: this is double `repr`, should probably be `"URIPattern(%r)" % text_type(self)`

rdflib/namespace.py
```diff
@@ -329,17 +330,19 @@ def normalizeUri(self, rdfTerm):
...
 
-        if not uri in self.__cache:
+        if uri not in self.__cache:
             namespace, name = split_uri(uri)
             namespace = URIRef(namespace)
             prefix = self.store.prefix(namespace)
             if prefix is None:
                 if not generate:
-                    raise Exception(
-                        "No known prefix for %s and generate=False")
```
joernhees: this was a great exception :D

rdflib/namespace.py
adamhadani: @joernhees there's a lot of code here that looks suspicious to me :) as first pass at this module I just took care of some flake8 style fixes which are ancillary to main purpose of this PR (making exception types raised here more specific). I think we should have a separate effort around standardizing code a bit more (switch to use .format() rather than % for string interpolation, etc. etc.). Probably not in scope for this PR though

joernhees:
ok, i just overrode the very weird stuff (wrt. `repr` and `str`)... if anything goes wrong, i'm to blame ;)

wrt. `%` vs. `.format()` cleanup... i'm not decided about this... `format` was meant to be the future, but then somehow got stuck... i think initially (py 3.0 or so) `%` was deprecated, but in the current docs don't deprecate it anymore... the logging module still heavily relies on it (`logging.debug('foo %r bar', huge_list)` won't actually do any formatting for non debug log-levels, etc.). I guess the smartest thing we can do there is wait and allow both for now ;)

joernhees:
merged... thanks :)

Issue #682, opened 2017-01-12
Signing named graphs


LibrEars:
Hi all,

I'm would like to store my experiment data with RDFLib. Moreover I want to sign the data of every single experiment to make it sharable for the future. My approach is to use one named graph for each experiment, hash this graph and sign the hash.

It seems that I don't really understand the concept of isomorphic graphs. Why is the hash of the isomorphic named graph the same as the hash of its isomorphic conjunctive graph?

Should I sign the `_TripleCanonicalizer(gmary).to_hash()` instead or do I run in problems with blank nodes with this approach?

Here is some code to clarify what I want to do:

```python
from rdflib import Namespace, Literal, URIRef, BNode
from rdflib.graph import Graph, ConjunctiveGraph
from rdflib.plugins.memory import IOMemory

from rdflib.compare import to_isomorphic, _TripleCanonicalizer


ns = Namespace("http://love.com#")

mary = BNode()
john = URIRef("http://love.com/lovers/john#")

cmary=URIRef("http://love.com/lovers/mary#")
cjohn=URIRef("http://love.com/lovers/john#")

store = IOMemory()

g = ConjunctiveGraph(store=store)
g.bind("love",ns)

gmary = Graph(store=store, identifier=cmary)

gmary.add((mary, ns['hasName'], Literal("Mary")))
gmary.add((mary, ns['loves'], john))

gjohn = Graph(store=store, identifier=cjohn)
gjohn.add((john, ns['hasName'], Literal("John")))

print("The internal hash of an named graph is the same as the internal hash of the Conjunctive graph: " +
      str(to_isomorphic(g).internal_hash() == to_isomorphic(gmary).internal_hash()) + "\n")

# Prints to 'True'

print("The internal hash of an named graph is the same as the internal hash of the Conjunctive graph: " +
      str(_TripleCanonicalizer(g).to_hash() == _TripleCanonicalizer(gmary).to_hash()))

# Prints to 'False'


# Example how I think to proove the signature of signed graphs:
for h in g.objects(mary_public_keys, wot.signed):
    # First verify the signature
    if gpg.verify(str(h)):
        # Second compare hash
        sigHash = gpg.decrypt(h).data.decode("utf-8").strip() # gpg.decrypt(h): bytes --> string
        
        identifier = str(g.value(h, RDFS.label))
        signedG = g.get_context(identifier)
        realHash = str(to_isomorphic(signedG).internal_hash())  # Gives the wrong hash?
[...]

PR #718, opened 2017-02-16
Added test for Issue #682 and fixed.

jimmccusker:
Fixed issue #682.

### DIFF
--- a/rdflib/compare.py
+++ b/rdflib/compare.py
@@ -469,8 +469,11 @@ class _TripleCanonicalizer(object):
 def to_isomorphic(graph):
     if isinstance(graph, IsomorphicGraph):
         return graph
-    return IsomorphicGraph(store=graph.store)
-
+    result = IsomorphicGraph()
+    if hasattr(graph, "identifier"):
+        result = IsomorphicGraph(identifier=graph.identifier)
+    result += graph
+    return result
 
 def isomorphic(graph1, graph2):
     """Compare graph for equality.
--- a/rdflib/compare.py
+++ b/rdflib/compare.py
@@ -168,6 +168,9 @@ class IsomorphicGraph(ConjunctiveGraph):
         """Negative graph isomorphism testing."""
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return super(IsomorphicGraph, self).__hash__()
+
     def graph_digest(self, stats=None):
         """Synonym for IsomorphicGraph.internal_hash."""
         return self.internal_hash(stats=stats)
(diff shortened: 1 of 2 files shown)


jimmccusker:
@joernhees, can I get some help from someone who knows python 3 better than I on these failures?

jimmccusker:
The current error (in Py3) seems to expect IsomorphicGraph to be hashable:

```
...
  File "/home/travis/build/RDFLib/rdflib/rdflib/plugins/memory.py", line 254, in add
    self.__all_contexts.add(context)
TypeError: unhashable type: 'IsomorphicGraph'
```

joernhees:
:-/ sad, i thought maybe just removing the py2 print statements already fixes it...

joernhees:
aaaahaaa:

https://docs.python.org/3/reference/datamodel.html#object.__hash__
> ```A class that overrides __eq__() and does not define __hash__() will have its __hash__() implicitly set to None. When the __hash__() method of a class is None, instances of the class will raise an appropriate TypeError when a program attempts to retrieve their hash value, and will also be correctly identified as unhashable when checking isinstance(obj, collections.Hashable).```

joernhees:
@jimmccusker tests pass, but i'm not entirely sure f72b51a is the way to go... the store uses the hash as context (quads), so changing it as new triples are added is bad bad bad... however, hashing the full `IsomorphicGraph` should probably have something to do with its content?!? luckily we also have an `IsomorphicGraph.internal_hash` 🙄

jimmccusker:
Right, that's the thing. __hash__ is supposed to be invariant. What does `set` do? It's hashable, but variant. Since RDF graphs are supposed to be sets of triples (or quads), maybe we can look there for guidance.

gromgull:
Python sets are NOT hashable, precisely because they are mutable. That's
why there is "FrozenSet", this thing has to copy all the triples into a new
object that is unchangable. OR make a "FrozenGraph" subclass, that does not
allow writes, and hopes the underlying store is not changed during the
operation of checking isomorphism.




-- 
http://gromgull.net

jimmccusker:
Then that argues that Graphs shouldn't be hashable either. For some reason, there seems to be a use case in Python 3 where graphs are expected to be hashable, but not in Python 2.7.

joernhees:
hmm, not sure we're not getting hung up here on the alternative fact that IOMemory internally hashes the graph's identifier to transform triples into quads... i'm leaning toward hitting that the merge button and be done with this for now, but maybe we should add a cleanup issue to remove the `__hash__` from all mutable things and make IOMemory explicitly hash the identifier?

PR #720, opened 2017-02-24
Accept header should use a q parameter

acoburn:
[RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.3.2) defines the HTTP Accept header to use a `q` parameter. The use of a `p` parameter in the code appears to be a typo and probably would not be understood by a conforming HTTP server.


niklasl:
That certainly looked like a typo. Thanks!

Issue #274, opened 2013-05-04
Custom FILTER function for SPARQL engine


gromgull:
Like Jena :)

pchampin:
I'm assuming that, by "FILTER function", you mean any function that can be used in [FunctionCall](https://www.w3.org/TR/sparql11-query/#rFunctionCall)...
If that's the case, +1 :)

Actually, I'm working on it...

white-gecko:
Closed by #723 via 2182c0e08e540daa4708cee619fe6588dff0b83a.

PR #723, opened 2017-03-08
added a hook to add custom functions to SPARQL

pchampin:
might be an answer to #274 .

My next plan is to provide an implementation of Virtuoso's ``bif:`` functions (sqrt, cos, sin...).

### DIFF
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -624,7 +624,7 @@ def Function(e, ctx):
         else:
             try:
                 return func(*e.expr)
-            except TypeError, ex:
+            except TypeError as ex:
                 raise SPARQLError(*ex.args)
     else:
          raise SPARQLError('Unknown function %r"%e.iri')
@@ -808,7 +808,7 @@ def RelationalExpression(e, ctx):
             try:
                 if x == expr:
                     return Literal(True ^ res)
-            except SPARQLError, e:
+            except SPARQLError as e:
                 error = e
         if not error:
             return Literal(False ^ res)
@@ -842,7 +842,7 @@ def RelationalExpression(e, ctx):
         r = ops[op](expr, other)
         if r == NotImplemented:
             raise SPARQLError('Error when comparing')
-    except TypeError, te:
+    except TypeError as te:
         raise SPARQLError(*te.args)
     return Literal(r)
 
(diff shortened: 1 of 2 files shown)


rdflib/plugins/sparql/operators.py
```diff
@@ -582,14 +582,62 @@ def Builtin_EXISTS(e, ctx):
...
     """
+    pair =_CUSTOM_FUNCTIONS.get(e.iri)
+    if pair is not None:
+        func, raw = pair
+        if raw:
+            return func(e, ctx)
+        else:
+            try:
+                return func(*e.expr)
+            except TypeError, ex:
```
joernhees: this exception syntax is considered evil and hence was deprecated in py2, doesn't work in py3 anymore...
`TypeError as ex`

rdflib/plugins/sparql/operators.py
```diff
@@ -582,14 +582,62 @@ def Builtin_EXISTS(e, ctx):
...
+                raise SPARQLError(*ex.args)
+    else:
+         raise SPARQLError('Unknown function %r"%e.iri')
```
joernhees: something is weird here

rdflib/plugins/sparql/operators.py
pchampin: I can fix it, but the same syntax was used in other places, that I didn't change, so I was not sure.

rdflib/plugins/sparql/operators.py
pchampin: Granted, it is a bit cryptic. I rewrote it, hopefully it is less weird now...

rdflib/plugins/sparql/operators.py
```diff
@@ -617,17 +617,21 @@ def Function(e, ctx):
...
-        if raw:
-            return func(e, ctx)
-        else:
-            try:
-                return func(*e.expr)
-            except TypeError as ex:
-                raise SPARQLError(*ex.args)
+    if pair is None:
+        # no such function is registered
+        raise SPARQLError('Unknown function %r"%e.iri')
```
joernhees: i think what you actually want is `raise SPARQLError('Unknown function %r' % e.iri)`

rdflib/plugins/sparql/operators.py
joernhees: might've been before 5.0.0, but py3 would always err in those cases as 5.0.0 doesn't use 2to3 anymore

joernhees:
nice change, tests now work 👍

rdflib/plugins/sparql/operators.py
pchampin: what the... :-/ well spotted !

gromgull:
Sorry somehow I missed this! Looks very good - thanks!

PR #731, opened 2017-04-12
Fix logging in interactive mode

amitbeka:
Fix some thorny problems with interactive mode logging in rdflib:
- When importing rdflib from the command line, it opens the loggers for every possible package, which is a very bad practice
- More over, when using jupyter console and importing anything that imports rdflib somewhere along the chain, all loggers in the system pour their logs. This is because the condition (`not hasattr(__main__, '__file__') and sys.stderr.isatty()`) is True but it's of course what we want here
- Another small fix is checking the stderr is a TTY, not stdout which isn't used

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -72,13 +72,15 @@ import sys
 assert sys.version_info >= (2, 7, 0), "rdflib requires Python 2.7 or higher"
 
 import logging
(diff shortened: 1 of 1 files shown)


gromgull:
this looks good! the travis build timed out, I've restarted it.

joernhees:
thanks :)

Issue #732, opened 2017-04-15
N3 parse error on single quote within single quotes 


hsolbrig:
Escaping an interior single quote within a single quoted string:

```python
from rdflib import Graph

rdf = """PREFIX ex: <http://a.example>
             ex:test ex:pattern ' \\'' .
"""
Graph().parse(data=rdf, format="turtle")
```
produces the following error:

```
  File "/Users/mrf7578/Development/venv/parsers/lib/python3.6/site-packages/rdflib/plugins/parsers/notation3.py", line 1591, in strconst
    "bad escape")
  File "/Users/mrf7578/Development/venv/parsers/lib/python3.6/site-packages/rdflib/plugins/parsers/notation3.py", line 1615, in BadSyntax
    raise BadSyntax(self._thisDoc, self.lines, argstr, i, msg)
rdflib.plugins.parsers.notation3.BadSyntax: at line 2 of <>:
Bad syntax (bad escape) at ^ in:
"b"PREFIX ex: <http://a.example>\nex:test ex:pattern ' "^b"\\'' .\n""
```

Changing lines 1578 and 1580 in notation3.py (rdflib 4.2.2) and adding a "\\'" on the end of the pattern strings - current:
```python
1588                k = 'abfrtvn\\"'.find(ch)
1589                if k >= 0:
1590                   uch = '\a\b\f\r\t\v\n\\"'[k]
```

new:
```python
                k = 'abfrtvn\\"\''.find(ch)
                if k >= 0:
                    uch = '\a\b\f\r\t\v\n\\"\''[k]
```
Fixes this problem

PR #736, opened 2017-04-19
Adds escaped single quote to literal parser

hsolbrig:
Addresses issue #732.  Literals with double quotes ("....") include an escaped interior quote ("...\"..."), but literals with single quotes ('...\'...') currently produce a parsing error.  This adds a single quote to the escaped character list


gromgull:
Thanks!

Issue #391, opened 2014-05-08
 RDFLib core vs dependencies


niklasl:
A long time ago, the intent to split RDFLib into a core and separate packages for complex parsers, SPARQL etc. was cancelled, since there was a desire to have "the batteries included". Since then (and mainly because of the advanced parsing and query features), RDFLib has grown dependencies upon external packages, making it non-self-contained anyway.

One very strange effect of this is the dependency on SPARQLWrapper, which depends on RDFLib (which depends on SPARQLWrapper..). For one,  this is an odd circular dependency, which e.g. causes an older RDFLib to be installed before running tests.

We should reconsider this situation. It is good and well to depend on packages. Therefore, we should seriously consider breaking out advanced parts (again), instead of this odd situation with a fat, complex core, and dependencies which sometimes depends on parts of those innards. We could e.g. make an rdflib-core, and reshape the 'rdflib' package into an umbrella package, which upon installation would pull in all these parts into the nice full suite. (Although some more complex parts should probably be considered optional anyway, such as non-memory backend stores.)

(In any case, the intended integration of rdflib-jsonld could just as easily be done by adding a dependency to RDFLib's setup.py and call it a day. Apart from the strange circularity if we keep the current situation, we wouldn't have to copy the separate repo code into RDFLib, tests and all, and then either mark the separate package "frozen", or have to maintain a copy manually.)

gromgull:
I would make the following suggestion: 

* keep this repo for a central rdflib package which pulls in everything else. This could also contain the tools/extras (unless we retire some of them)
* make a new repository + package rdflib-core with everything directly in the rdflib package, the memory store, the proxy-stores (auditable, regex, concurrent), and also the main serialisers/parsers (n3/ttl/nquads/trig/rdfxml)
* move sparql to rdflib-sparql
* move sleepycat to rdflib-sleepycat (it's only one file, but no bsddb built into py3, ? Could also be core.)
* remove rdfa/mdata from the core, and let them only be in their own repo (where the actual development already happens today)
* talk to the sparqlwrapper guys and see if we can have the sparqlstore in their repo (which in turn only depends on rdflib-core)

The tests go with each repo, some could stay here for testing the full package. 

I don't know about the docs. Maybe also here, if you install all packages sphinx can find them for generating API docs.

joernhees:
Let me put this first: any change wrt. this is a step forward.

I'm however a bit worried about splitting too much:
- splits the (already small) developer community
[...]

PR #744, opened 2017-05-18
Remove SPARQLWrapper dependency

gromgull:
Replace it with "SPARQLWrapper-Lite", which is based on the `requests` library. 

This gains us lots of nice modern HTTP library features, such as keepalive, connection pooling, and by exposing the `kwargs` to be sent to requests, we can do http basic auth etc. easily. 

SPARQLWrapper also has lots of legacy code from supporting various old RDF stores from the days before the SPARQL Protocol was a thing. The new code only supports [SPARQL 1.1 Protocol](https://www.w3.org/TR/sparql11-protocol/) standard, and can therefore be much shorter!

This is not quite done - PR for testing + feedback! (but I run this in production, so it's also not completely pre-beta :) )

### DIFF
--- a/rdflib/plugins/sparql/results/xmlresults.py
+++ b/rdflib/plugins/sparql/results/xmlresults.py
@@ -37,25 +37,18 @@ Authors: Drew Perttula, Gunnar Aastrand Grimnes
 
 class XMLResultParser(ResultParser):
 
-    def parse(self, source):
+    def parse(self, source, content_type=None):
         return XMLResult(source)
 
 
 class XMLResult(Result):
-    def __init__(self, source):
+    def __init__(self, source, content_type=None):
 
-        xmlstring = source.read()
-
-        if isinstance(xmlstring, text_type):
-            xmlstring = xmlstring.encode('utf-8')
         try:
             parser = etree.XMLParser(huge_tree=True)
-            tree = etree.parse(BytesIO(xmlstring), parser)
+            tree = etree.parse(source, parser)
         except TypeError:
-            tree = etree.fromstring(xmlstring)
-        except Exception as e:
-            log.exception("Error parsing XML results: %s"%xmlstring)
-            raise e
+            tree = etree.parse(source)
 
         boolean = tree.find(RESULTS_NS_ET + 'boolean')
         results = tree.find(RESULTS_NS_ET + 'results')
@@ -65,18 +58,11 @@ class XMLResult(Result):
         elif results is not None:
             type_ = 'SELECT'
         else:
-            g = Graph()
-            try:
-                g.parse(data=xmlstring)
-                if len(g) == 0:
-                    raise
-                type_ = 'CONSTRUCT'
-
-            except:
-                raise ResultException(
-                    "No RDF Graph, result-bindings or boolean answer found!")
+            raise ResultException(
+                "No RDF result-bindings or boolean answer found!")
 
         Result.__init__(self, type_)
+
         if type_ == 'SELECT':
             self.bindings = []
             for result in results:
@@ -90,10 +76,11 @@ class XMLResult(Result):
                          './%shead/%svariable' % (
                              RESULTS_NS_ET, RESULTS_NS_ET))]
 
-        elif type_ == 'ASK':
+        else:
             self.askAnswer = boolean.text.lower().strip() == "true"
-        elif type_ == 'CONSTRUCT':
-            self.graph = g
+
+
+
 
 
 def parseTerm(element):
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -635,8 +635,7 @@ class SPARQLUpdateStore(SPARQLStore):
 
         self.debug = DEBUG
         assert isinstance(query, string_types)
-        self.setNamespaceBindings(initNs)
-        query = self.injectPrefixes(query)
+        query = self._inject_prefixes(query, initNs)
 
         if self._is_contextual(queryGraph):
             query = self._insert_named_graph(query, queryGraph)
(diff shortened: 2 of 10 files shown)


gromgull:
I have used this in production since May - it's pretty good. Lets merge? @joernhees ?

joernhees:
sure, but shouldn't this just go upstream and replace the SPARQLWrapper default? i somehow fear that we're creating yet another WTF for the future ;) especially with colliding names won't we confuse everyone and in future talk about SPARQLWrapper's SPARQLWrapper and RDFLib's SPARQLWrapper ?

gromgull:
No - it's only 120 lines of code, and it removes the SPARQLWrapper dependency all together. But it doesn't try to replace everything SPARQLWrapper does - people also use that for getting stuff out of a SPARQL endpoint, without wrapping it as a RDFLib store. This code does nothing for them.

scossu:
@gromgull @joernhees Any decision on merging this PR?
Also, since it is based on requests, does this allow for connection pooling without depending on the keepalive package?

gromgull:
I use it in production - so I am happy to have it merged :) 

And yes, it should support connection pool without keepalive.

scossu:
@gromgull I merged this on a separate repo for testing but I get an error: 

~~~
>>> from rdflib.plugins.stores.sparqlstore import SPARQLUpdateStore
RDFLib Version: 5.0.0-dev
>>> from rdflib.term import URIRef
>>> store = SPARQLUpdateStore(queryEndpoint='http://localhost:3030/lsup-dev/query', update_endpoint='http://localhost:3030/lsup-dev/update')
>>> fds = Dataset(store)
>>> fgr = fds.graph(URIRef('urn:bogus:1'))
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/graph.py", line 1623, in graph
    self.store.add_graph(g)
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/plugins/stores/sparqlstore.py", line 721, in add_graph
    "CREATE GRAPH %s" % self.node_to_sparql(graph.identifier))
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/plugins/stores/sparqlstore.py", line 638, in update
    self.setNamespaceBindings(initNs)
AttributeError: 'SPARQLUpdateStore' object has no attribute 'setNamespaceBindings'
~~~

Did the API change?

gromgull:
@scossu yes - that method is gone. I wonder why the tests didn't catch that the first time around. Will fix!

setup.py
```diff
@@ -6,9 +6,9 @@
 
 kwargs = {}
 kwargs['install_requires'] = [ 'six', 'isodate', 'pyparsing']
```
joernhees: requests is a new dependency, no?

rdflib/plugins/sparql/results/xmlresults.py
```diff
@@ -65,18 +58,11 @@ def __init__(self, source):
         elif results is not None:
             type_ = 'SELECT'
         else:
-            g = Graph()
-            try:
-                g.parse(data=xmlstring)
```
joernhees: what happens to CONSTRUCT queries now?

rdflib/plugins/stores/sparqlstore.py
```diff
@@ -13,67 +13,20 @@
...
-import warnings
-import contextlib
 
-try:
-    from SPARQLWrapper import SPARQLWrapper, XML, JSON, POST, GET, URLENCODED, POSTDIRECTLY
-except ImportError:
-    raise Exception(
-        "SPARQLWrapper not found! SPARQL Store will not work." +
-        "Install with 'pip install SPARQLWrapper'")
+from .sparqlwrapper import SPARQLWrapper
```
joernhees: can we somehow avoid using SPARQLWrapper as a name (any caps)? Maybe SPARQLConnector? Or not expose it in its own module and internalize it here as `_SPARQLWrapper`?

I really would like to avoid the confusion

rdflib/query.py
```diff
@@ -195,10 +195,11 @@ def _set_bindings(self, b):
         _get_bindings, _set_bindings, doc="a list of variable bindings as dicts")
 
     @staticmethod
-    def parse(source, format='xml', **kwargs):
+    def parse(source=None, format=None, content_type=None, **kwargs):
```
joernhees: hmm, what happens if `source is None`?

i'd somehow find the following slightly clearer (if it's what is intended)
```python
def parse(source, format='xml', content_type=None, **kwargs):
    parser = plugin.get(content_type or format, ResultParser()
    return parser.parse(source, content_type=content_type, **kwargs)
```

joernhees:
re #391

wikier:
In addition, if you consider the underline implementation could be useful standalone, happy to make it part of `SPARQLWrapper` 2.x ;-)

jimmccusker:
This broke our ability to configure POST requests against SPARQL endpoints, and it looks like the setReturnFormat() function went missing. How should I address switching over?

gromgull:
@jimmccusker this builds on the requests and you can configure requests as much as requests lets you, any additional kwargs you pass to the store constructor will be passed to requests, i.e. for basic auth: 

```py
import rdflib
from rdflib.plugins.stores.sparqlstore import SPARQLStore

store = SPARQLStore('http://localhost:8000', auth=('user', 'pass'))
g = rdflib.Graph(store)
list(g[rdflib.RDFS.Class])
```

Return format is also set for the constructor - i.e. pass `returnFormat`

I will update the doc strings in #872 to make this clear!

gromgull:
@jimmccusker For max backward-compatability, hopefully this will be merged and you can use your original code: https://github.com/RDFLib/sparqlwrapper/pull/126

PR #745, opened 2017-05-23
Added the changes in the initial context

iherman:
See [Official Announcement](https://lists.w3.org/Archives/Public/semantic-web/2017May/0114.html) for the details

### DIFF
--- a/rdflib/plugins/parsers/pyRdfa/initialcontext.py
+++ b/rdflib/plugins/parsers/pyRdfa/initialcontext.py
@@ -21,15 +21,20 @@ initial_context = {
 }
 
 initial_context["http://www.w3.org/2011/rdfa-context/rdfa-1.1"].ns = {
+	'as'		: 'https://www.w3.org/ns/activitystreams#',
 	'csvw'		: 'http://www.w3.org/ns/csvw#',
 	'dcat'		: 'http://www.w3.org/ns/dcat#',
-    'qb'		: 'http://purl.org/linked-data/cube#',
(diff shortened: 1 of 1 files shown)


gromgull:
The current state of the RDFa and mData parser is unfortunately that it has drifted out of sync with the stand-alone versions you maintain (as discussed in https://github.com/RDFLib/rdflib/pull/443) 

No one has time to sort this mess out. I would vote for removing RDFa and mData from core rdflib for the 5.0 release.
 
They will still be a simple pip install away, and with the (now many years old!) plugin architecture, people do not have to change their code for this to work. 

This also matches the way we deal with json-ld and lots of other extra modules.

iherman:
@gromgull I understand and see that it is a bit messy but, unfortunately, the separate mData and RDFa repositories are not equivalent. Those are the original programs that are built _on top_ of RDFLib: their usage pattern is that the user runs some dedicated function parsing, say, and HTML file and returning an RDF(Lib) Graph. They are not based on the plugin architecture of RDFLib, ie, they will not be usable from the standard `parse` function of RDFLib. In this sense, it is not equivalent to the way json-ld is handled.

I agree it would be better to remove that part from the RDFLib code and put both of them in their respective repository, but that is an extra work that I simply do not know if and when I could/would do. In other words, we are a bit stuck:-(

What is the best way of moving forward?

gromgull:
We've removed RDFa now - I will open PRs for rdfa and microdata to keep them working as rdflib plugins!

Issue #749, opened 2017-06-07
Parsing large sparql responses leads to lxml error: lxml.etree.XMLSyntaxError: internal error: Huge input lookup


artreven:
When I connect to an external SPARQL store and make a query that returns many results (more than 9.5 MB https://github.com/KimiNewt/pyshark/issues/75) the xmlparser throws the error: "lxml.etree.XMLSyntaxError: internal error: Huge input lookup"

gjhiggins:
Addressed in https://github.com/RDFLib/rdflib/commit/6cb0dabce15a0d5dc81954ecb833e31b47a68e28

PR #750, opened 2017-06-07
Add support to parsing large xml inputs (https://github.com/RDFLib/rdflib/issues/749)

artreven:


gromgull:
This fails with `huge_tree` is not a valid keyword argument on all python versions. 

We are quite lax with where we get elementtree from: https://github.com/RDFLib/rdflib/blob/master/rdflib/compat.py#L16-L36

I guess the default version doesn't support this?

artreven:
I just checked what happens at my end: I use Python 3.5.2 and lxml 3.5.0. I get `etree` from `lxml` and the line `etree.XMLParser(huge_tree=True)` returns my the parser without any exceptions.
What could be the issue?

niklasl:
As @gromgull says, the default etree from the standard library does not support this option:

```python

etree.XMLParser(huge_tree=True)
```

On Python 3, this fails with:

```
  File "<stdin>", line 3, in <module>
TypeError: 'huge_tree' is an invalid keyword argument for this function
```

Since lxml is an optional package for RDFLib, you cannot rely on it being available to support the non-standard keyword (`huge_tree`).

artreven:
Sorry, I did not get the original comment from @gromgull at first, I thought `lxml` is the default version. I have added an `except` clause to provide a fallback in case of default version.

gromgull:
Thanks! (and sorry about the delay)

PR #764, opened 2017-08-24
Update README.md

diegoquintanav:
fixes a missing reference to `rdflib` when using `Namespace()` in the *getting started* section.


joernhees:
thanks

Issue #773, opened 2017-10-09
URIRef comparison with Python `str` fails using `==`


luav:
URIRef comparison with Python `str` fails in both Python v2 and v3.  
Per-char comparison of URIRef with str and `==` yield distinct values:
```python
import rdflib
from rdflib.namespace import RDF
typeProp = '<' + RDF.type + '>'  # N3/quad type property
prop = '<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>'

eq1 = typeProp == prop  # False
eq2 = len(prop) == len(typeProp) and not prop.find(typeProp)
assert eq1 == eq2, '{} vs {}'.format(eq1, eq2)
```
yields:
```sh
>>> assert eq1 == eq2, '{} vs {}'.format(eq1, eq2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError: False vs True
```
Per-char comparison yields the same as inclusion search, but distinct from the equality check, which is confusing:
```python
class Entity(object):
	"""Entity to load RDF tripples"""
	# ATTENTION: type(RDF.type + 'x') is <class 'rdflib.term.URIRef'>, which breaks the str comparison
	#typeProp = '<' + str(RDF.type) + '>'  # N3/quad type property          OK
	typeProp = '<' + RDF.type + '>'  # FAILS str comparison
...
...
            prop = '<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>'
            print('extend(), props: \n{}\n{}\neq: {}'.format(prop, Entity.typeProp, prop == Entity.typeProp))
            for i in range(min(len(prop), len(Entity.typeProp))):
                if prop[i] != Entity.typeProp[i]:
                    print('Unequal pos: {}, val: {} ({}) vs {} ({})', i, prop[i], ord(prop[i]), Entity.typeProp[i], ord(Entity.typeProp[i]))
                    break
            else:
                print('len: {} vs {}, eq: {}, index: {}'.format(len(prop), len(Entity.typeProp), prop == Entity.typeProp,
                    prop.find(Entity.typeProp)))
```
yields:
```sh
extend(), props: 
<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>
<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>
eq: False
len: 49 vs 49, eq: False, index: 0
```
Investigating the cause:
```sh
>>> type(prop)
<type 'str'>
>>> type(typeProp)
[...]

PR #774, opened 2017-10-15
URIRef comparison with other types fixed, see #773

luav:
URIRef comparison with other types fixed, see #773 (caused invalid comparison of URIRef with str on Python 3)

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -157,8 +157,6 @@ class Identifier(Node, text_type):  # allow Identifiers to be Nodes in the Graph
         True
         >>> Identifier("foo")=="foo"
         True
-        >>> Literal("foo")=="foo"
-        True
         >>> BNode("foo")!=URIRef("foo")
         True
         >>> URIRef("foo")!=BNode("foo")
@@ -953,7 +951,7 @@ class Literal(Identifier):
 
     def __eq__(self, other):
         """
-        Literals are only equal to other literals.
+        Literals are only equal to other literals, pure Identifiers and str.
 
         "Two literals are equal if and only if all of the following hold:
         * The strings of the two lexical forms compare equal, character by character.
@@ -972,6 +970,10 @@ class Literal(Identifier):
         False
         >>> Literal("1", datatype=URIRef("foo")) == "asdf"
         False
+        >>> Literal("1", datatype=URIRef("foo")) == "1"
+        True
+        >>> Literal("1", datatype=URIRef("foo")) == Identifier("1")
+        True
         >>> from rdflib import XSD
         >>> Literal('2007-01-01', datatype=XSD.date) == Literal('2007-01-01', datatype=XSD.date)
         True
@@ -996,7 +998,7 @@ class Literal(Identifier):
                 and (self.language.lower() if self.language else None) == (other.language.lower() if other.language else None) \
                 and text_type.__eq__(self, other)
 
-        return False
+        return super(self) == other
 
     def eq(self, other):
         """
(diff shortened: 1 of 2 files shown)


luav:
Implemented matching of `Identifier` subclasses to the same types, to the base type `Identifier` and to the raw python strings. Note that different subclasses of `Identifier` (`Literal` vs `BNode`, etc.) remained incomparable between each other.
Initial version where `Identifier` subclasses  like `URIRef` could be extended with python `str`, but silently failed to compare to `str` was confusing and unexpected.

Basic tests are passed now, except tests in `test_dawg.py`. I have no idea about that tests and don't want to spend my time for it. Feel free to complete the integration!

gromgull:
Hi @luav - Thanks for you input!

Unfortunately, `test_dawg` is sort of "the most important test". It tests the SPARQL unit tests from the data-access group. RDFLib implements SPARQL semantics by directly comparing the RDFLib terms, for SPARQL it is crucial that the URI `<http://example.org/a>` is NOT equal to the string "<http://example.org/a>"`

In fact, this is pretty much crucial for all of RDFLib - I don't see why you would want it to be otherwise? 

The real issue here is probably that we inherit from unicode/str - leading people like you to believe they are similar, but then they are not. We have discussed removing unicode from the class hierarchy before, but I can't find the pointer right now. 

So unless there are some VERY good arguments I am missing, I am "wont fix" on this one. Sorry!

luav:
Hi @gromgull ,
you are right that `http://example.org/a` != URI `<http://example.org/a>` but I'm talking about another comparison, comparison of the str `<http://example.org/a>` vs URI `<http://example.org/a>`.
Please, consider the following example:
```python
import rdflib
from rdflib.namespace import RDF
# print(RDF.type) yields: http://www.w3.org/1999/02/22-rdf-syntax-ns#type
typeProp = '<' + RDF.type + '>'  # N3/quad type property
prop = '<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>'

eq1 = typeProp == prop  # False
eq2 = len(prop) == len(typeProp) and not prop.find(typeProp)  # find returns 0 for the match
assert eq1 == eq2, '{} vs {}'.format(eq1, eq2)
```
yields
```sh
>>> assert eq1 == eq2, '{} vs {}'.format(eq1, eq2)
AssertionError: False vs True
```
The representation of URI was ambiguous and misleading before the suggested fix. If your show the content of URI without the `<`, `>`  but semantically assume presence of the URI brackets then the modifying operations ('+', etc.) should behave differently, which would not be a Pythonic way of doing things and extremely hard to be properly controlled and efficiently implemented.

Basically, to avoid any concerns and be consistent, the URI either should return str URI (including `<` '>') or should support consistently all possible operations (assignment, comparison, modification, etc.).

white-gecko:
Hi @luav ,
thank you very much for all the work you put into the code and also in making an argument. I can understand that the current behavior of the rdflib in combination with the python datatypes is not always intuitive. As changing this behavior would break the current API we will definitely not introduce these changes in the 5.0.0 release. For the 6.0.0 release we want to drop Python 2 support and have some major rework. We also think about doing something to the relation of RDFLib objects and python data types (@nicholascar I hope I got this one right). I sure this will involve some discussion were I would like to welcome you.
Adding the angle brackets around the IRI would follow the Turtle, NTriples, Trig, NQuads, and SPARQL syntax, which is ok in my eyes, as it would not be easy to serialize individual nodes following JSON-LD or RDF/XML. In addition we would also need to see if we deal with the other RDF terms in the same way and print Literals as `"Literal"@en` instead of `Literal` which would definitely break the current API.
I hope this as well as the earlier comment as well as the comment in #773 depict how complicate it is and that we can not accept this PR at the moment.
I hope it is ok for you that I close this PR and #773 and put them on "out of scope".

PR #789, opened 2017-11-06
setup.py: Make the license field use an SPDX identifier

sschuberth:
According to [1] this should rather be the license name than a link to the
LICENSE file, so use the SPDX identifier "BSD-3-Clause".

[1] https://docs.python.org/2/distutils/setupscript.html#additional-meta-data


gromgull:
Thanks!

Issue #630, opened 2016-05-11
xsd:dateTime less than operator issue


wrobell:
Filters on `xsd:dateTime` work with literals, i.e.

```
FILTER(?start > "2016-01-01 20:01"^^xsd:dateTime)
```

But if there are two variables of `xsd:dateTime` then filter does not work properly, i.e.

```
FILTER(?start < ?end)
```

Script illustrating the issue below. It should show `c1` and `c2` entities, but nothing is shown instead.

```
from pprint import pprint
import rdflib
import io

data = """
@prefix : <#>.
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.

:C a rdfs:Class.

:start a rdf:Property;
    rdfs:domain :C; 
    rdfs:range xsd:dateTime.

:end a rdf:Property;
    rdfs:domain :C; 
    rdfs:range xsd:dateTime.

:c1 a :C;
    :start "2016-01-01 20:00:00"^^xsd:dateTime;
    :end "2016-01-02 20:01:00"^^xsd:dateTime.

:c2 a :C;
    :start "2016-01-01 20:05:00"^^xsd:dateTime;
    :end "2016-01-01 20:30:00"^^xsd:dateTime.
"""

graph = rdflib.Graph()

f = io.StringIO(data)
graph.parse(f, format='n3')

result = graph.query("""
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

SELECT ?c
WHERE {
    ?c :start ?start;
        :end ?end.
#    FILTER(?start > "2016-01-01 20:01"^^xsd:dateTime)   # this works!
    FILTER(?start < ?end)
}
""")

pprint(list(result))
```

colinfang:
Equality operator doesn't work either.

```
g2 = rdflib.Graph()
literal_date = Literal(datetime.date(2016,1,1), datatype=XSD.date)
print g2.serialize(format="n3")
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

[...]

Issue #613, opened 2016-03-20
sorting in turtle serializer breaks with datetime


satra:
if one represents a graph like the one below

```
    ex:gen1 a prov:End ;
        prov:time "2016-03-20T18:11:44.996521"^^xsd:dateTime,
            "2012-12-03T21:08:16.686000+00:00"^^xsd:dateTime .
```

the following line:
https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/serializers/turtle.py#L106

raises the following error:

`TypeError: can't compare offset-naive and offset-aware datetimes`

this should do a lexical sort rather than a numeric or datatype sort.

satra:
from a turtle perspective a sort isn't necessary at all.

gromgull:
fixed in https://github.com/RDFLib/rdflib/commit/13ce988925b4dd919c8713f9a890acb8d962370c

PR #793, opened 2017-11-30
Literal total ordering

tgbugs:
This pull request provides basic infrastructure for sorting Literals by value where the underlying native type has no total ordering. This provides a more consistent solution to issues like: https://github.com/RDFLib/rdflib/issues/648, https://github.com/RDFLib/rdflib/issues/630, and https://github.com/RDFLib/rdflib/issues/613, where workarounds are implemented in the serializer. This leads to massively increased code complexity in the serializers to compensate for the fact that Literal does not support a total ordering because some of the underlying python types do not.

The two types that currently cause issues are datetime and time, but the proposed solution is easily extensible. If other types are found to have this issue the solution is to add an entry to _NO_TOTAL_ORDER_TYPES that includes a function that partitions the type into subtypes that do have total orders.

This pull request also adds test to ensure that types with known issues sort as expected.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -1406,6 +1406,7 @@ _NUMERIC_INF_NAN_LITERAL_TYPES = (
 _NO_TOTAL_ORDER_TYPES = {
     datetime:lambda value:bool(value.tzinfo),
     time:lambda value:bool(value.tzinfo),
+    xml.dom.minidom.Document:lambda value:value.toxml(),
 }
 
 def _castPythonToLiteral(obj):
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -812,6 +812,9 @@ class Literal(Identifier):
                     return self.language > other.language
 
             if self.value != None and other.value != None:
+                if dtself in _NO_TOTAL_ORDER_TYPES:
+                    comparator = _NO_TOTAL_ORDER_TYPES[dtself]
+                    return comparator(self.value) > comparator(other.value)
                 return self.value > other.value
 
             if text_type(self) != text_type(other):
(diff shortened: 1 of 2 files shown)


gromgull:
Thanks - it would be nice to clean-up any now unnecessary  code in serializers, but that can be another PR!

rdflib/term.py
```diff
@@ -1397,6 +1400,14 @@ def _writeXML(xmlnode):
     _XSD_DECIMAL,
 )
 
+# these are not guranteed to sort because it is not possible
+# to calculate a total order over all valid members of the type
+# the function must partition the type into subtypes that do have total orders
+_NO_TOTAL_ORDER_TYPES = {
+    datetime:lambda value:bool(value.tzinfo),
+    time:lambda value:bool(value.tzinfo),
```
joernhees: this is actually quite bad, as it will never sort (date)time objects anymore... fix is incoming...

tgbugs: Are you sure? I have a test for this which produces the expected and correct reordering.

https://github.com/tgbugs/pyontutils/blob/master/test/nasty.ttl#L203
https://github.com/tgbugs/pyontutils/blob/master/test/good.ttl#L234

tgbugs: Remember that this is a partition function not the sort key.

tgbugs: Hrm... checking again, it does seem that I actually did not implement this correctly, which makes it all the more curious why my tests pass.

joernhees: already fixed, just see #862

joernhees: but thanks for the reply ;)

tgbugs: That is the right solution and I think ultimately much clearer than what I was thinking about, which was to turn the caster into just the partition, which I think is where my brain fart came from. Many thanks!

PR #796, opened 2017-12-04
weisfeiler lehman kernel and example

GillesVandewiele:
A working example for the weisfeiler-lehman kernel, which is one of the more 'advanced' kernels. Other kernels are possible as well and will probably require less effort.


GillesVandewiele:
So, any status update on this? @joernhees ?

gromgull:
if you want a project for kernels in the rdflib namespace that you can admin just let me know! (like Jörn has https://github.com/RDFLib/graph-pattern-learner )

GillesVandewiele:
Could be very interesting! Perhaps even a bit broader, and more of a library/module for feature extraction from knowledge graphs (under which kernel methods can be classified). We are doing quite some work around that currently

GillesVandewiele:
Good argument @joernhees . On the other hand, we also have an implementation of RDF2Vec laying around and are working on our own feature extraction technique, so that could potentially be added as well (making the complexity already a bit larger)

gromgull:
I disagree with @joernhees - "today is the first day of the rest of our lives" :D 

Lets keep a lean rdflib-core - and put this in it's own project!

nicholascar:
Hi @GillesVandewiele, Are you still interested in creating a place within the RDFlib family of repositories? Since it's been decided not to include it in the core, we'd like to resolve the PR some way which could mean closing it. We could open another one then to put your code in it's own home.

GillesVandewiele:
Hi @nicholascar,

Sure! You can definitely close this one here :). This implementation is also a bit outdated. I can update it in a different repo!

nicholascar:
Ok, thanks @GillesVandewiele. I'll close this but please let me know if/when you want me to create a separate repo within RDFlib for this kernel. If you want one, what should it be called, just `wl-kernel` or similar?

PR #800, opened 2017-12-14
Suppress warning for not using custom encoding.

white-gecko:
Suppress the warning "NQuadsSerializer/NTSerializer does not use custom encoding." in the case that the `encoding` argument (custom encoding) is set to the same value as the default encoding (`self.encoding`).

This is especially relevant, if one is selecting the NQuadsSerializer/NTSerializer for serializing query results, since it is setting `utf-8` as `encoding`: https://github.com/RDFLib/rdflib/blob/0469adb01a4dc3dd7b153496ad5e1c0f09a66465/rdflib/query.py#L203-L230


joernhees:
👍 thanks

Issue #804, opened 2018-01-04
Parsing of long lists causes python recursion problems


hsolbrig:
`newList` in notation3.py uses recursion to load lists.   Sufficiently long lists cause the interpreter to raise `RecursionError: maximum recursion depth exceeded while getting the str of an object`. In the example below, the list has well over 1000 entries:

```python
from rdflib import Graph
g = Graph()
g.load('https://raw.githubusercontent.com/shexSpec/shexTest/master/validation/manifest.ttl', format='turtle')
```

Output:

```text
      ...
  File "lib/python3.6/site-packages/rdflib/plugins/parsers/notation3.py", line 1743, in newList
    self.makeStatement((f, rest, a, self.newList(n[1:], f)))
     ...
RecursionError: maximum recursion depth exceeded while getting the str of an object
```

gromgull:
Fixed in https://github.com/RDFLib/rdflib/pull/805

PR #805, opened 2018-01-04
Change notation3 list realization to non-recursive function.

hsolbrig:
Proposed fix for issue #804 

Refactored notation3.py `newList` function to use a loop rather than recursion


hsolbrig:
Apologies - wrestling with the test system -- not sure which issues are legit, but I think that this one has (finally) got it right.

gromgull:
Thanks! Digging into the n3 parser is not for the faint-hearted!

PR #807, opened 2018-01-16
Add possibility to set authority and basepath to skolemize graph

white-gecko:
When calling `Graph.skolemize()` it is not possible to set the authority, while the parameter exists on the `BNode.skolemize(authority)` method.

With this contribution it is possible to set the `authority` also when calling `skolemize()` on a graph.
Further an additional optional parameter `basepath` is added to allow the configuration of the path which is perpended to the generated ID.


gromgull:
Looks good to me!

PR #808, opened 2018-02-11
Adds basic support of xsd:duration

Tpt:
Uses `isodate.parse_duration` that converts a duration ISO string to timedelta if possible and if not fallbacks to the `Duration` type of the `isodate` library.

Supports also xsd:yearMonthDuration and xsd:dayTimeDuration.
xsd:dayTimeDuration is the exact match of the set of durations supported by `timedelta`.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -47,12 +47,11 @@ import math
 import base64
 import xml.dom.minidom
 
-from datetime import date, time, datetime
+from datetime import date, time, datetime, timedelta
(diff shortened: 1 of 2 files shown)


gromgull:
Hmm - somewhere in the back of my head I remember that the SPARQL test-suite actually tests that you don't support some data-types not defined in SPARQL. Clearly this isn't one of them though - thanks!

Tpt:
Thank you!

Issue #715, opened 2017-02-13
Path query chaining issue


bgyori:
Consider the following RDF file `example.rdf`
```
<http://example.org/A> <http://example.org/relations/isa> <http://example.org/X> .
<http://example.org/A> <http://example.org/relations/isa> <http://example.org/Y> .
<http://example.org/B> <http://example.org/relations/isa> <http://example.org/X> .
```
With rdflib 4.2.2 I do
```
import rdflib
g = rdflib.Graph()
g.parse('example.rdf', format='nt')
g.query('SELECT ?child ?parent WHERE {?child <http://example.org/relations/isa> ?parent .}')
```
and print the results, and I get the correct relationships:
```
http://example.org/A http://example.org/Y
http://example.org/B http://example.org/X
http://example.org/A http://example.org/X
```
However, when changing the query to
```
g.query('SELECT ?child ?parent WHERE {?child <http://example.org/relations/isa>+ ?parent .}')
```
I get 
```
http://example.org/A http://example.org/Y
http://example.org/A http://example.org/X
http://example.org/B http://example.org/X
http://example.org/B http://example.org/Y
```
where the relationship `B isa+ Y` was incorrectly inferred. Is this a bug or am I misunderstanding the meaning of `+` in path chaining?

joernhees:
reproducible in 5.0.0-dev:
```
In [1]: from rdflib import URIRef, Graph
INFO:rdflib:RDFLib Version: 5.0.0-dev

In [2]: g = Graph()

In [3]: a, b, x, y = [URIRef(s) for s in "abxy"]

In [4]: isa = URIRef('isa')

In [5]: g.add((a, isa, x))

In [6]: g.add((a, isa, y))

In [7]: g.add((b, isa, x))

In [8]: list(g)
Out[8]:
[(rdflib.term.URIRef(u'a'),
  rdflib.term.URIRef(u'isa'),
  rdflib.term.URIRef(u'y')),
 (rdflib.term.URIRef(u'b'),
  rdflib.term.URIRef(u'isa'),
  rdflib.term.URIRef(u'x')),
 (rdflib.term.URIRef(u'a'),
  rdflib.term.URIRef(u'isa'),
  rdflib.term.URIRef(u'x'))]

In [9]: list(g.query('SELECT ?child ?parent WHERE {?child <isa> ?parent .}'))
Out[9]:
[(rdflib.term.URIRef(u'a'), rdflib.term.URIRef(u'y')),
[...]

PR #822, opened 2018-04-12
fix property paths bug: issue #715

jeromedockes:
As mentioned in #715, when trying to match an arbitrary length property path
between two variables, e.g. "?a rdfs:subClassOf+ ?b", some bad solutions were
returned. I tried to fix this while changing as little as possible the logic of
the property paths implementation.


gromgull:
Thanks!

PR #827, opened 2018-05-14
ClosedNamespace KeyError -> AttributeError

tgbugs:
The current implementation of ClosedNamespace raises a KeyError via `__getattr__` which breaks hasattr. This patch fixes it by raising an AttributeError instead.


gromgull:
Nice and easy! Thanks!

joernhees:
actually on second thought will change so that the appropriate exception is returned for both...
- `RDFS['xyz']` --> KeyError
- `RDFS.xyz` --> AttributeError

PR #828, opened 2018-05-15
remove rdfa and microdata parsers from core RDFLib

### LINKED ISSUE #582: code duplication issue between rdflib and pymicrodata
There seems to be a code duplication problem between https://github.com/RDFLib/rdflib and https://github.com/RDFLib/pymicrodata which lead to problems around #443 before. There are several ways to solve this:
- [ ] remove the pymicrodata part from rdflib, put it in an own package and make rdflib depend on that...
- [ ] make a git submodule
- [ ] remove this repository and just have the one in rdflib
- [x] keep both and maybe have this problem again

Once #443 is solved, i'd love to move that tick mark to some other box...

gromgull:
They were unteasted and falling behind their individual repos on
github.

They remain installable as plugins.

Fixes #582


gromgull:
re: #391

joernhees:
hmm, before doing this, we should at least update the docs and provide an immediate easy to understand mitigation strategy to current users i think.

jimmccusker:
What are the issues with RDFa?

gromgull:
Pulled the trigger on this one - I will make PRs for pyRDFa and pyMicrodata to ensure they still work as rdflib plugins. 

We will also add a warning if people try to do `graph.parse(..., format='rdfa')` without the plugin installed and let them know to install the plugin projects.

gromgull:
https://github.com/RDFLib/pyrdfa3/pull/26

gromgull:
https://github.com/RDFLib/pymicrodata/pull/7

PR #832, opened 2018-06-09
Not adding all namespaces for n3 serializer

mwatts15:
- It looks like the serialzer does that anyway, but the removed code adds more than
  are needed


gromgull:
nice and easy! Thanks!

Issue #833, opened 2018-06-18
Allow to return the number of triples added from Graph.addN


mwatts15:
Some triple stores can tell you whether the triple you requested to add actually resulted in a new entry in the store. Knowing this information is helpful for debugging, informing users, and optimization (e.g., sizing buffers). It would be helpful to return such information from the `addN` method especially since, as a bulk operation, the only way to determine this information otherwise would be iterating over the entries twice, once to get the number that would be added and then to actually add.

nilay77:
Hi, I would like to solve this issue. Please guide me on what changes need to be made. Thanks!

mwatts15:
Probably, just return the values from stores for add() and addN() in Graph
and sub-classes. It would also be good to have at least one of the built-in
stores (e.g., IOMemory) return the number of triples added.

To rdflib maintainers (@nicholascar, @joernhees, etc): there may be scope
for a generic 'AddResult' or 'AddNResult' rather than just the number, in
case there's more info a store can provide: any thoughts on that?

nicholascar:
> here may be scope for a generic 'AddResult' or 'AddNResult' rather than just the number

What other info might be interesting? Counts of unique subjects, counts of URIs v. Literals etc? I wonder what precedence in the toolkit is for complex returns like this? It could be a small graph...

Also, regarding implementation, in the description you said: "the only way to determine this information otherwise would be iterating over the entries twice" but how would an implementation here be any different? Wouldn't the implementation have to do something like this:

```python
def add(...)
    count_before = len(g)
    # add the triples
    count_after = len(g)

    return count_after - count_before
```

So this is still a double count of `g`.

Unless if you mean the back-end store might have some native way of doing the `len(g)` work and this function passes back to them, so the above implementation would be in-memory only but able to be overridden by other stores.

mwatts15:
On Wed, May 27, 2020, 00:39 Nicholas Car <notifications@github.com> wrote:

> Unless if you mean the back-end store might have some native way of doing
> the len(g) work and this function passes back to them, so the above
> implementation would be in-memory only but able to be overridden by other
> stores
>
> I meant this. rdflib-zodb, for instance, uses a set data structure that
returns this info from its 'insert' method. I definitely want this to be an
optional interface that depends entirely on the store implementation rather
than providing a default in Graph just because I know some stores don't
take care to or can't get this statistic efficiently. Also, I had been
ambivalent about this but a marker like 'supports_modification_stats' on
[...]

PR #834, opened 2018-06-18
Returning a count from addN (#833)

mwatts15:


rdflib/graph.py
```diff
@@ -236,12 +236,7 @@
 import logging
 logger = logging.getLogger(__name__)
 
-# import md5
 import random
-import warnings
-
-from hashlib import md5
-
```
joernhees: please keep commits concise, don't change random other things...

rdflib/graph.py
```diff
@@ -394,11 +388,14 @@ def add(self, triple):
...
-                          and _assertnode(s,p,o)
-                          )
+        triples_stored = self.__store.addN((s, p, o, c) for s, p, o, c in quads
+                                           if isinstance(c, Graph)
+                                           and c.identifier is self.identifier
+                                           and _assertnode(s, p, o)
+                                           )
+        if triples_stored is None:
+            return -1
+        return triples_stored
```
joernhees: hmm, this won't really work without also at least updating the store interface in `store.py`

also i'd prefer a simple `return` in front of the dispatch to store... i don't see how `-1` is better than simply returning `None`

mwatts15: You're correct. I don't actually recall why I went with -1.
The intention is to make it optional. I have added a flag member variable to indicate whether addN returns a count on a store I maintain, but I didn't see a reason to make it official interface.

nicholascar:
Open/Close Travis

PR #875, opened 2018-11-09
assign members to subjects instead of adding each member individually

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

### DIFF
--- a/rdflib/plugins/serializers/turtle.py
+++ b/rdflib/plugins/serializers/turtle.py
@@ -77,7 +77,7 @@ class RecursiveSerializer(Serializer):
             members = list(self.store.subjects(RDF.type, classURI))
             members.sort()
 
-            subjects = members
+            subjects.extend(members)
             for member in members:
                 self._topLevels[member] = True
                 seen[member] = True
--- a/rdflib/plugins/serializers/turtle.py
+++ b/rdflib/plugins/serializers/turtle.py
@@ -77,8 +77,8 @@ class RecursiveSerializer(Serializer):
             members = list(self.store.subjects(RDF.type, classURI))
             members.sort()
 
+            subjects = members
             for member in members:
-                subjects.append(member)
                 self._topLevels[member] = True
                 seen[member] = True
 


rdflib/plugins/serializers/turtle.py
```diff
@@ -77,8 +77,8 @@ def orderSubjects(self):
             members = list(self.store.subjects(RDF.type, classURI))
             members.sort()
 
+            subjects = members
             for member in members:
-                subjects.append(member)
```
joernhees: this is not equivalent: if there are multiple `classURI` then before `subjects` contained all members, now it will only contain those of the last...

joernhees:
may i ask what the intention of this PR is?

is it performance only or is the current implementation wrong?

if it's the former, then instead of your `subjects = members` line, one could simply write `subjects.extend(members)`

tgbugs:
It seems that the current (master branch) implementation is correct to me. In cases where there are multiple top classes this change seems like it should result in dangling bnodes or something like that, which is strange, given that all the checks pass.

rdflib/plugins/serializers/turtle.py
AxelNennker: line 74 is `subjects = []` so I think this is equivalent

AxelNennker:
Hm, I think I was wrong. 
subjects.extend(members) is the right way to go, I think (now)

rdflib/plugins/serializers/turtle.py
AxelNennker: and because topClasses has one element

AxelNennker: assuming that selfClasses is always [RDFS.Class] the outer loop can be removed and subject = members is correct

AxelNennker:
@joernhees regarding the intention of this PR: I have a graph with 272029 triples and serializing it with this serializer never finished before I ran out of patience.
So I looked at the code and searched for a reason.
This PR does not fix this but it just looks wrongs to add the elements one by one when there is a methods to extend the list in one go.

I think that the serializer spends too much time here:
https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/serializers/turtle.py#L101
But I have no proof for that.

AxelNennker:
In this repository there is no class deriving from RecursiveSerializer that changes topClasses.
topClasses has one element in the tests, so the tests didn't fail

Dell Precision 7710, Ubuntu16.04, python 3.5.2

Meanwhile I switch to Redland librdf because it managed to serialized that graph. Although now I don't create that particular graph anymore but create another graph in that project and will give rdflib another try.

If you think this PR is OK would you review and merge?

AxelNennker:
Can I do something to get this merged? What is the timeline? No pressure, thanks for providing and maintaining rdflib, just curious.

AxelNennker:
Any chance this gets merged some time?

JervenBolleman:
@nicholascar I think this is good to be merged. Python wise it is a small improvement, but still worth having. Could be worth squashing berfore merging.

nicholascar:
This PR seems now to only improve things and not create any issues (from my reading of it) so let's merge.

PR #877, opened 2018-11-11
Update README.md

jodischneider:
update syntax of Python snippet; print expects parentheses


white-gecko:
Shouldn't it be `print(s,p,o)` for python3?

nicholascar:
I've made the change to print(s, p, o)

PR #886, opened 2018-12-10
Fix blank node label to not contain '_:' during parsing

white-gecko:
When parsing a query, e.g.:

```
PREFIX ex: <http://example.org/>
INSERT DATA { GRAPH ex: { _:blankA a ex:Bla }}
```

The generated `BNode` would be `BNode("_:blankA")` serializing this blank node with `BNode("_:blankA").n3()` results in `_:_:blankA` which the parser can not understand.


white-gecko:
To reproduce the whole round trip situation:

```
import rdflib

graph = rdflib.Graph()
graph.update("INSERT DATA { _:blankA a <urn:Bla> }")
string = graph.serialize(format='ntriples').decode('utf-8')

# Does not work:
rdflib.Graph().parse(data=string, format="ntriples")

# This would be correct:
rdflib.Graph().parse(data="_:blankA <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <urn:Bla> .", format="ntriples")
```

white-gecko:
I will squash the test commits once it works.

white-gecko:
Now this is even tested :-)

white-gecko:
Is it possible to accept this PR? What is missing?

joernhees:
sorry, time was/is missing :-/

PR #905, opened 2019-03-28
Remove previous prefix when calling bind

sleak-lbl:
In Memory store plugin, bind() adds the new binding to the dicts but
doesn't remove any old one. The Sleepycat store does better, but misses
when the prefix is empty (as often happens in turtle files). This fix
applies the Sleepycat solution modified to delete re-bound empty
prefixes too

### DIFF
--- a/rdflib/plugins/memory.py
+++ b/rdflib/plugins/memory.py
@@ -158,6 +158,9 @@ class Memory(Store):
 
     def bind(self, prefix, namespace):
         self.__prefix[namespace] = prefix
+        bound_prefix = self.__prefix.get(namespace)
+        if bound_prefix is not None:
+            self.__namespace.delete(bound_prefix)
         self.__namespace[prefix] = namespace
 
     def namespace(self, prefix):
@@ -238,6 +241,9 @@ class IOMemory(Store):
 
     def bind(self, prefix, namespace):
         self.__prefix[namespace] = prefix
+        bound_prefix = self.__prefix.get(namespace)
+        if bound_prefix is not None:
+            self.__namespace.delete(bound_prefix)
         self.__namespace[prefix] = namespace
 
     def namespace(self, prefix):
--- a/rdflib/plugins/sleepycat.py
+++ b/rdflib/plugins/sleepycat.py
@@ -436,7 +436,7 @@ class Sleepycat(Store):
         prefix = prefix.encode("utf-8")
         namespace = namespace.encode("utf-8")
         bound_prefix = self.__prefix.get(namespace)
-        if bound_prefix:
+        if bound_prefix is not None:
             self.__namespace.delete(bound_prefix)
         self.__prefix[namespace] = prefix
         self.__namespace[prefix] = namespace


nicholascar:
Simple enough, merging

akimul:
May be after this commit, we are having the following error:
Can you please check:

```
File "/app/env/local/lib/python2.7/site-packages/rdflib/plugins/memory.py", line 246, in bind
self.__namespace.delete(bound_prefix)
AttributeError: 'dict' object has no attribute 'delete'
```
Thanks

tgbugs:
Hopefully this will also resolve one of the final sources of non-determinism during serialization, however there is also a chance that it will break some user code that relied on the namespace manager retaining multiple namespace/prefix pairings, so it should probably be included in any release notes.

nicholascar:
Tagged with "release note" as a reminder of above

gromgull:
Did the tests actually pass after this was merged? @akimul's comment above suggests not?

nicholascar:
I thought I saw that Travis had passed everything - I would usually require that as a prerequisite before merging - but I can’t see any Travis notes above. I’ll run test locally and log a bug if I see any issue.

white-gecko:
Yes this causes the error highlighted by @akimul 
https://travis-ci.org/github/AKSW/QuitStore/jobs/661585396?utm_medium=notification&utm_source=github_status

nicholascar:
@sleak-lbl despite passing all the tests, when we merged this, it caused problems, probably because changes in master since you raised this PR did something.

When I ran some tests with master (which we will shortly release as rdflib 5.0.0), the redundant namespace issue goes away.

For this code:

```
from rdflib import Graph, Namespace, URIRef

g = Graph()
g.bind('', Namespace('http://empty.com/'))
g.add((URIRef("http://example1.com/foo"),
       URIRef("http://example2.com/bar"),
       URIRef("http://example3.com/baz")))
n3 = g.serialize(format="n3").decode('utf-8')

print(n3)
```

serialized with rdflib 4.2.2:

```
@prefix ns1: <http://example2.com/> .

<http://example1.com/foo> ns1:bar <http://example3.com/baz> .
```

serialized with rdflib master:

```
@prefix ns1: <http://example2.com/> .

<http://example1.com/foo> ns1:bar <http://example3.com/baz> .
```

So I think the error you wanted fixed is now fixed.

tgbugs:
Did you test this on a version of python that does not have insertion order preserving dictionaries? I'm betting that the problem is masked but not solved.

sleak-lbl:
sorry for the silence (buried in another thing all week) .. now that I look at it again, I don't recall why I used:
`self.__namespace.delete(bound_prefix)`
and not:
`del self.__namespace[bound_prefix]`
(and why that patch didn't fail for me..)

white-gecko:
I also did not find a way to produce a test that made your changes fail, I just had it in my more complex tests for the QuitStore.

In addition to the pullrequest by @ashleysommer we should also add a test to prevent regression.

nicholascar:
> Did you test this on a version of python that does not have insertion order preserving dictionaries? I'm betting that the problem is masked but not solved.

No! How far back should I be going for this, Python 3.4 perhaps?

white-gecko:
For me it failed for 3.5, 3.7, and 3.8: https://travis-ci.org/github/AKSW/QuitStore/builds/661585394?utm_medium=notification&utm_source=github_status

tgbugs:
Ya, you have to go back to 3.4 (the insertion order preserving behavior was introduced in 3.5), and you may have to run a few times to hit a RNG state to see the prefix switch between serializations. My tests for that are [here](https://github.com/tgbugs/pyontutils/blob/d6b414a24528d9237f59a98a4350a989ca429f35/ttlser/test/test_ttlser.py#L129), they might not work on 3.4 due to the use of fstrings elsewhere in the code base.

PR #906, opened 2019-04-03
infixowl.manchesterSyntax do not encode strings

tgbugs:
The manchesterSyntax function in infixowl was encoding some but not
all string typed objects to bytes, this commit removes the remaining
cases where the encoding was being done. This caused internal errors
like the one below whenever one tried to repr infixowl classes.

File "rdflib/extras/infixowl.py", line 1291, in __repr__
    for s in dc]))
TypeError: sequence item 0: expected str instance, bytes found


nicholascar:
Close/open for Travis

PR #916, opened 2019-05-24
Added replace param to Graph.bind

trueg:
Pretty self-explanatory: Expose the "replace" parameter from the namespace manager's bind method in Graph.bind.


gromgull:
✨

PR #918, opened 2019-06-13
IPython tab-completions for ClosedNamespace

jochemb:
For a ClosedNamespace all allowed keys are known in advance, so we might as well allow IPython users to discover them via tab-completions: https://ipython.readthedocs.io/en/stable/config/integrating.html 

This change will enable tab-completion for ClosedNamespace objects when using dot notation (`RDF.Pro\t -> RDF.Property`) and for square bracket notation (`RDF['Pro\t -> RDF['Property']`).

This really adds convenience for only a subset of users, so I'd understand if the additional logic is not welcome.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -193,6 +193,14 @@ class ClosedNamespace(object):
     def __repr__(self):
         return "rdf.namespace.ClosedNamespace(%r)" % text_type(self.uri)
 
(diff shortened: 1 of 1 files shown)


nicholascar:
@jochemb we're adding this for the 6.0.0 release as the immediate 5.0.0 release is bug fixes only so we can get it out ASAP (a fortnight).

Issue #357, opened 2014-02-20
Clarify respective use cases of `Store.add` vs. `Store.addN`, and correct use of `Graph.parse`


pchampin:
The respective use cases between `Store.add` and `Store.addN` are not really clear. More precisely, when some one has several triples to write to a store (not necessarily many, nor necessarily available as an iterable), is it
- _(a)_ acceptable to use multiple calls to `Store.add`, or
- _(b)_ should one avoid that and absolutely use `Store.addN`?

This needs clarification because currently the answer is "it depends on the store"; more precisely "IOMemory" and "Sleepycat" answer (a), as they have a rather efficient `add` method, while `SPARQLUpdateStore` or the SqlAlchemy plugin answer (b), since they make a single query to the underlying store for each call to `add`.

There is a rather serious consequence to this lack of clarity, in the use of `Graph.parse`. All parsers in `rdflib.plugins` use `add` rather than `addN`. As a consequence, it _might_ be very inefficient to write:

``` python
# g is a Graph
g.parse(filename, format=f)
```

depending on the underlying store of `g`. For "add-inefficient" stores, one has to write:

``` python
data = rdflib.Graph() # in memory graph
data.parse(filename, format=f)
g += data  # uses addN
```

This breaks object encapsulation, which is bad.

I see two solutions :
1. Make it explicit in the documentation of `Store` that multiple uses of `Store.add` should be considered an exceptional case, `Store.addN` should always be preferred. This makes it easier for `Store` implementers, but harder for programmers using stores (as they have to handle the "buffering" of triples). In particular, this requires to fix all the parser implementation to comply with this.
2. Make it explicit in the documentation of `Store` that multiple uses of `Store.add` can be a valid use case, and thus encourage `Store` implementers to consider naive/straightforward implementations of `Store.add` as a bug rather than a feature.

mwatts15:
As an active user of RDFLib, I don't see that there's much "clarification" needed. The choice between `addN` and `add` is as @gromgull describes, but it is, ultimately, up to the application using RDFLib to decide whether batching is appropriate based on their application needs.

Batching `add` into `addN` is pretty trivial, but we can save users the effort of doing it themselves. I've made a context manager and wrapper for Graph that does this.

pchampin:
@mwatts15 
> The choice between `addN` and `add` is (...) ultimately, up to the application using RDFLib

Well, not when they are using `graph.parse`, which makes this choice for them...

[...]

PR #931, opened 2019-08-31
Adding a wrapper for batching add() calls to a Graph

mwatts15:
- Should address RDFLib/rdflib#357


nicholascar:
@mwatts15 we need to discuss this: we think the solution here to #357 should overload the `g.add()` method, so that it's seamlessly available, rather than declaring another class which may or may not be used.

mwatts15:
Can you elaborate on this?
> overload the g.add() method, so that it's seamlessly available

nicholascar:
Let me think about this a bit for a sec more: I'm not sure I've really worked out what I'm thinking!

nicholascar:
Could you add a test file for this perhaps?

nicholascar:
Close/Open Travis

mwatts15:
> Could you add a test file for this perhaps?

Can do. As I was stubbing out tests, I realized there's some input validation I could do. Also, can support quads for conjunctive graphs pretty trivially, so will also add that.

nicholascar:
Seems to be failing only Python2.7. If you want this in the 5.0.0 release (end of March), you'll need to pass 2.7. If you want for the 6.0.0 release (end of June), you can drop the 2.7 and 3.4 tests as we'll only be supporting 3.5+ then and forward.

mwatts15:
I was actually looking at a problem from an rdflib-sqlalchemy user which is similar, but I realize that this PR doesn't address their use-case since they really need a proxy so they can use the wrapped Graph with just add/addN modified. 

Any objection to adding wrapt as a dependency to rdflib to implement this?

nicholascar:
I'll merge this in for now. Not 100% sure about adding dependencies and I just raised a dependency-related Issue (#1035) so let's add this in and see the SQLAlchemy fix as a separate PR?

PR #941, opened 2019-10-30
make SPARQLConnector work with DBpedia

joernhees:
i think this was a leftover from #744 

DBpedia seems to answer with `application/sparql-results+xml; charset=UTF-8` header, which wasn't registered in plugins...

also unskips an old test on travis tha otherwise failed locally due to now unresolved SPARQLWrapper exception ref

### DIFF
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -286,6 +286,9 @@ register(
 register(
     'application/sparql-results+xml', ResultParser,
     'rdflib.plugins.sparql.results.xmlresults', 'XMLResultParser')
+register(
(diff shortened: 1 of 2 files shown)


gromgull:
My gut feel would be to strip off the `; chatset=..` bit before looking for plugins. 

Will we have one plugin for utf-8, one for utf-16? 

But whatever, I don't care that deeply :D

joernhees:
close & reopen to trigger travis...

joernhees:
yupp, actually thought about that, but didn't find the right spot quickly... also question is: what if the encoding is different, how to handle that correctly?

I'll merge for now as a quick-fix is better than none, but feel free to cleanup

Issue #484, opened 2015-05-06
 ServiceGraphPattern not implemented ?


rnd0101:
While performing the following script with SPARQL query:

``` python
# INFO:rdflib:RDFLib Version: 4.2.1-dev

from rdflib import Graph
from rdflib import Namespace
from rdflib import URIRef

OWL = Namespace("http://www.w3.org/2002/07/owl#")

g = Graph()
g.namespace_manager.bind('owl', URIRef("http://www.w3.org/2002/07/owl#"))
g.namespace_manager.bind('xsd', URIRef("http://www.w3.org/2001/XMLSchema#"))


qres = g.query("""
PREFIX foaf:    <http://xmlns.com/foaf/0.1/>
PREFIX dbpprop: <http://dbpedia.org/property/>

SELECT ?place (AVG(?rain) as ?avgRain)
WHERE{
  VALUES ?rainPred {
    dbpprop:janRainMm dbpprop:febRainMm dbpprop:marRainMm dbpprop:aprRainMm
    dbpprop:mayRainMm dbpprop:junRainMm dbpprop:julRainMm dbpprop:augRainMm
    dbpprop:sepRainMm dbpprop:octRainMm dbpprop:novRainMm dbpprop:decRainMm
  }
  SERVICE <http://dbpedia.org/sparql> {
     ?place ?rainPred ?rain .
  }
}
GROUP BY ?place
HAVING (COUNT(?rain) = 12)
ORDER BY ?avgRain
""" )

for row in qres:
    print row
```

( thanks to http://answers.semanticweb.com/questions/22080/whats-your-favorite-recent-public-sparql-endpoint for an example idea)

I am getting: Exception: ServiceGraphPattern not implemented

The same query is executed fine in, for example, Sesame workbench.

gjhiggins:
Fixed in #943

Issue #769, opened 2017-09-25
Implementing ServiceGraphPattern


marqh:
Hello RDFLib

In trying to re-implement a system using rdflib I am facing a key capability gap: I require the ability to run queries that have SERVICE calls within the Query: federated queries, as defined by W3 in:

https://www.w3.org/TR/sparql11-federated-query/

Using rdflib (v4.2.2) I receive 
`Exception: ServiceGraphPattern not implemented`
errors, from
https://github.com/RDFLib/rdflib/blob/58d6369a12791ae7c1e259c86aeea384390d36ee/rdflib/plugins/sparql/evaluate.py#L262

#484 suggests that this has been the case for some time.

I am interested in looking to implement this capability and I would like to start a discussion on how amenable RDFLib developers would be to an enhancement in this space?

I would also very much appreciate some advice on how to approach an implementation: what are the key factors to be aware of? what are the implementation and testing preferences within the developer community?  

It strikes me that if it were very simple to do, it may have been done by now, so I would like to explore some of the potential complexities in principal before putting too much time into an implementation that will never fit with the RDFLib approach or simply never work.

I will add further comments on key details as I find them and explore implementation ideas with you as this ticket develops.

Your thoughts on whether this capability is of interest to RDFLib and how to implement it effectively are hugely welcome at this stage.

many thanks
mark

marqh:
It appears to me there are 2 facets to this capability:

1. running the 'SERVICE' subquery remotely
1. providing the results into the main query

Regarding 2. the context: `ctx` is in scope at the point of identification of a SERVICE query fragment, so this should be achievable, although I'm not sure what form is required for this as yet.

However, there is something of an implementation challenge for 1.
At this point, there is a `rdflib.plugins.sparql.parserutils.CompValue` instance, a parsed representation of the SPARQL including significant rdflib implementation details.  It is not obvious to me how to obtain the raw SPARQL query string from this.
If the raw SPARQL query string were available, a remote call to the relevant SPARQL endpoint could simply be implemented via a variety of means.

I feel that addressing this aspect is the key enabling factor for this work.

I am currently considering options including:
* passing the raw sparql fragment to the `CompValue` on creation, so that it can be retrieved and used later
* writing a `regenerate_sparql` method on `CompValue` to enable the sparql string to be regenerated on request

[...]

PR #943, opened 2019-12-02
Implement Service Clause in Sparql Support

JervenBolleman:
This builds on the excellent work by @marqh.

I have tested it and it works for json replies only. 
The existing solution binding are concatenated to the query string in an ugly way, but functional.
Would love to know how to do this nicer.

What is the procedure to get this merged?
Should the testcases not depend on dbpedia, but some test server?


ab-pm:
Does this close #484 and #769?

gjhiggins:
> Does this close #484 and #769?

Yes, thanks for the references.

PR #945, opened 2019-12-03
Missing check in __init__.py prevents rdflib to be usable in QGIS plugin

situx:
When using rdflib inside our QGIS Plugin (https://plugins.qgis.org) SPARQLUnicorn (https://github.com/sparqlunicorn/sparqlunicornGoesGIS) sys.stdout is not defined when executing the Plugin in Windows.
This causes our plugin to crash.
By adding a simple check if sys.stdout is None rdflib is working perfectly when used in QGIS.

I would appreciate if you could include this very simple check.

Thank you!

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -76,7 +76,7 @@ logger = logging.getLogger(__name__)
 _interactive_mode = False
 try:
     import __main__
-    if not hasattr(__main__, '__file__') and sys.stderr.isatty():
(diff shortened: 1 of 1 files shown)


situx:
It seems the call has shifted from stdout to stderr? If so one would also need to create a check if sys.stderr is None

nicholascar:
Opening and closing for Travis reassessment

PR #950, opened 2019-12-24
Update to import directly from collections.abc

tmsincomb:
Error :: "Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working from collections import Mapping, MutableMapping."

The query function might break with python 3.8 because of this. Should update this soon to avoid that possibility.

### DIFF
--- a/rdflib/plugins/sparql/sparql.py
+++ b/rdflib/plugins/sparql/sparql.py
@@ -3,8 +3,10 @@ from __future__ import absolute_import
 import collections
 import itertools
 import datetime
-from collections.abc import Mapping, MutableMapping
-
+try:
+    from collections.abc import Mapping, MutableMapping
+except:
+    from collections import Mapping, MutableMapping
 import isodate
 from six import text_type, iteritems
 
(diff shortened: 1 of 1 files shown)


rdflib/plugins/sparql/sparql.py
```diff
@@ -3,7 +3,7 @@
 import collections
 import itertools
 import datetime
-from collections import Mapping, MutableMapping
+from collections.abc import Mapping, MutableMapping
```
tirkarthi: The removal was done in Python 3.9 . The project has Python 2 as a classifier. Perhaps trying collections.abc and falling back to collections for python 2 will be a more compatible fix.

gromgull: we generally put stuff to support multiple versions in https://github.com/RDFLib/rdflib/blob/master/rdflib/compat.py

tmsincomb: moved to #959. Closing this pull request because it is now a duplicate.

Had to start a new pull request after creating a proper fork.

PR #961, opened 2020-02-20
fixed URIRef including native unicode characters

kempei:
URIRef should be able to include native unicode like <http://ja.dbpedia.org/resource/日本> equal to <http://ja.dbpedia.org/resource/\\u65e5\\u672c>.
Adding the code as same as line 185-187.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -50,6 +50,7 @@ import xml.dom.minidom
 from datetime import date, time, datetime, timedelta
 from re import sub, compile
 from collections import defaultdict
+from unicodedata import category
 
 from isodate import parse_time, parse_date, parse_datetime, Duration, parse_duration, duration_isoformat
 
@@ -73,10 +74,11 @@ _invalid_uri_chars = '<>" {}|\\^`'
 
 
 def _is_valid_uri(uri):
-    for c in _invalid_uri_chars:
-        if c in uri:
-            return False
-    return True
+    """
+    Verify invalid chars other than Unicode Letter (Category: Lx)
+    Ref: https://www.unicode.org/reports/tr44/#General_Category
+    """
+    return all(map(lambda c: category(c).startswith('L') or not c in _invalid_uri_chars, uri))
 
 
 _lang_tag_regex = compile('^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$')
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -74,7 +74,7 @@ _invalid_uri_chars = '<>" {}|\\^`'
 
 
 def _is_valid_uri(uri):
-    return all(map(lambda c: category(c).startswith('L') or not c in _invalid_uri_chars, uri))
+    return all(map(lambda c: ord(c) > 256 or not c in _invalid_uri_chars, uri))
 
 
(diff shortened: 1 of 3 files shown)


nicholascar:
This seems very sensible to me: we really should handle IRIs, not just URIs to try and cater better for other writing systems, and the encode/decode prosed here is the same as used further down in util.py for Literal encoding (see Line 185).

But are there any issues to do with character encoding in URIs/IRIs that I'm not aware of? 

1. Can anyone with encoding experience other that the proponent of this PR (who I presume thinks it's fine!!) just assure me that this is fine?

2. @kempei please could you add a test or two to test/test_util.py that shows this addition working (probably just the example you cite in the description above) and ordinary URIs continuing to work?

nicholascar:
Before merging, we need:

1. a resolvable URL using URL escaped characters, not \u65e5 \u672c
2. a test function

This will go into the 5.0.0 release as that will drop Python 2 support and we don't think this will work with Python 2.

kempei:
I've added additional IRI string with Unicode to test_util.py/test_util_from_n3_expectpartialidempotencewithn3()
And I've also changed term.py/_is_valid_uri() for Unicode IRI.
I can run all testcases for util.

$ python -m unittest test/test_util.py 
.........................................
Ran 41 tests in 0.056s

OK
=============
CI build is failed but it seems to be other problem.

kempei:
It passed run_tests.py on my env.

kempei:
retest this please

kempei:
CI error may occur by PR#905 so that CI will succeed after merging PR#966
https://github.com/RDFLib/rdflib/pull/966

ashleysommer:
@kempei 
Thanks for your testing.
CI is still showing a failure in py2.7
That is why we are planning to merge this after v5.0.0 is release, because we are removing py2.7 support after v5.0.0 release.
```
File "/home/travis/build/RDFLib/rdflib/examples/conjunctive_graphs.py", line 13, in <module>
    from rdflib import Namespace, Literal, URIRef
  File "/home/travis/build/RDFLib/rdflib/rdflib/__init__.py", line 159, in <module>
    from rdflib.term import (
  File "/home/travis/build/RDFLib/rdflib/rdflib/term.py", line 1349, in <module>
    _RDF_XMLLITERAL = URIRef(_RDF_PFX + 'XMLLiteral')
  File "/home/travis/build/RDFLib/rdflib/rdflib/term.py", line 229, in __new__
    if not _is_valid_uri(value):
  File "/home/travis/build/RDFLib/rdflib/rdflib/term.py", line 78, in _is_valid_uri
    return all(map(lambda c: category(c).startswith('L') or not c in _invalid_uri_chars, uri))
  File "/home/travis/build/RDFLib/rdflib/rdflib/term.py", line 78, in <lambda>
    return all(map(lambda c: category(c).startswith('L') or not c in _invalid_uri_chars, uri))
TypeError: category() argument 1 must be unicode, not str
```

kempei:
Should I add any patch for py2.7 for CI build before v5.0.0 release?

nicholascar:
> Should I add any patch for py2.7 for CI build before v5.0.0 release?

If you could, that would be good, then this can be included in 5.0.0.

kempei:
I've changed term.py into universal logic using ord(), then all checks have passed including py2.7.
So please consider it adding 5.0.0.

PR #970, opened 2020-03-15
Allows RDF terms introduced by JSON-LD 1.1

Tpt:
[JSON-LD 1.1 introduces some RDF terms](https://www.w3.org/TR/json-ld11/#relationship-to-rdf), some of them being in a "work in progress" state.

It might still be convenient to allow them to be used with the `RDF.*` syntax now.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -220,7 +220,10 @@ class _RDFNamespace(ClosedNamespace):
                 "nil",
 
                 # Added in RDF 1.1
-                "XMLLiteral", "HTML", "langString"]
+                "XMLLiteral", "HTML", "langString",
+            
+                # Added in JSON-LD 1.1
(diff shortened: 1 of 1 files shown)


nicholascar:
Great addition thanks @Tpt and we will merge it but only for the 6.0.0 release, as @white-gecko has already tagged it, not for the 5.0.0 release for which we're just doing bug fixes. 

5.0.0 should only be a week or so away now, 6.0.0 a couple of months.

Tpt:
Thank you!

nicholascar:
I'm allowing this through, despite it being a "feature" as it's only implication is allowing more terms in the CloseNamepace of RDF and I just added a bunch of ClosedNamespaces recently anyway (SKOS, FOAF etc.).

ashleysommer:
As this is already merged, changing the milestone to "rdflib 5.0.0" for inclusion in the 5.0.0 changelog generation.

PR #978, opened 2020-03-19
Pr 451 redux

nicholascar:
A re-implementation of PR #451 because I don't believe the fix posed there addresses all conceivable add() scenarios.

This is failing to pass all my new tests and needs work with double/float rounding to get there.


nicholascar:
i.e. passing all numeric tests: now it's failing an XML Literal test!

ashleysommer:
@nicholascar 
I'm just doing a couple of my own tests locally against this, then I'll approve it to merge in before our release.

PR #987, opened 2020-03-26
Register additional serializer plugins for SPARQL mime types.

darrengarvey:
This mirrors the similar parser registrations below in the same file.
With this commit, an Accept header can be used directly to serialize a
response, making putting up a SPARQL compatible endpoint trivial.

### DIFF
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -271,13 +271,22 @@ register(
     'xml', ResultSerializer,
     'rdflib.plugins.sparql.results.xmlresults', 'XMLResultSerializer')
 register(
+    'application/sparql-results+xml', ResultSerializer,
+    'rdflib.plugins.sparql.results.xmlresults', 'XMLResultSerializer')
+register(
     'txt', ResultSerializer,
     'rdflib.plugins.sparql.results.txtresults', 'TXTResultSerializer')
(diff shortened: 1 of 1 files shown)


ashleysommer:
@darrengarvey 
This looks great, thanks.

Just one thing we'd like to discuss before merging.

Do we need two different entries for `application/sparql-results+xml; charset=UTF-8` and `application/sparql-results+xml`?

I understand the content-type header can and sometimes does include the charset (or other additional parameters), do you expect the user will pass that directly into the serializer? Should there be a mimetype sanitizer that splits removes anything after a semicolon when matching the mimetype to serializer (or parser)?

darrengarvey:
I was following the pattern from a few lines below on the parser side, which was added intentionally in #941.

Actually the charset isn't valid in the [`Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) header: there is [`Accept-Charset`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset) for that. The line you referenced was redundant, so I've removed it. Thanks @ashleysommer!

ashleysommer:
Thanks. Merged, just in time for the 5.0.0 release.

PR #995, opened 2020-04-05
fix #913 : Added _parseBoolean function to enforce correct Lexical-to-value mapping

### LINKED ISSUE #913: Update Lexical-to-value mapping for xsd:boolean
Hi,
I think the function for evaluating literals with datatype ```xsd:boolean``` should be updated. 

https://github.com/RDFLib/rdflib/blob/5fa18be1231a5e4dfc86ec28f2f754158c6f6f0b/rdflib/term.py#L1484

Currently, any string which is not equal to "true" or "1" is mapped to "false". However, according to [the spec](https://www.w3.org/TR/xmlschema11-2/#boolean) the lexical space for  boolean is only "0", "1", "true", "false". As a result, in case any other lexical value occurs, the parser should raise an Error and not parse it to a Literal with value "false".

arushi019:


rdflib/term.py
```diff
@@ -1383,6 +1383,14 @@ def _unhexlify(value):
         value = value.encode()
     return unhexlify(value)
 
+def _parseBoolean(value):
+    new_value = value.lower()
+    if new_value == 'true' or new_value=='1':
+        return True
+    if new_value != 'false' and new_value != '0':
+        warnings.warn('Parsing weird boolean, %r does not map to True or False'%value,DeprecationWarning)
```
white-gecko, line 1391: could you please specify the attribute name `category=` for the warning and add spaces around the `%`.

rdflib/term.py
```diff
@@ -1383,6 +1383,14 @@ def _unhexlify(value):
         value = value.encode()
     return unhexlify(value)
 
+def _parseBoolean(value):
+    new_value = value.lower()
+    if new_value == 'true' or new_value=='1':
```
white-gecko, line 1388: `new_value=='1'` -> `new_value == '1'`

Maybe you just use `new_value in ['true', '1']` resp. `new_value not in ['false', '0']`.

rdflib/term.py
arushi019: I've added these changes in the new commit in this pull request :)

rdflib/term.py
arushi019: Added these changes in the new commit 👍

arushi019:
> Would you be able to add a test as well?

Where should I add the tests? Is there a specific file?

white-gecko:
> Where should I add the tests? Is there a specific file?

https://github.com/RDFLib/rdflib/blob/master/test/test_literal.py should be the right place.

test/test_literal.py
```diff
@@ -163,6 +186,7 @@ def unlexify(s):
         self.assertEqual(specific_l.toPython(), s)
         self.assertEqual(specific_l.datatype, datatype)
 
+
```
white-gecko, line 189: Could you please remove this white space line. The rest is cool.

arushi019:
Hi! I've added the tests for the _parseBoolean function. Please take a look at the changes made. Thanks a lot :)

test/test_literal.py
arushi019: Removed the whitespace line. Please let me know if any other changes are needed.

PR #1000, opened 2020-04-10
Namespaces improvements

nicholascar:
Added QB to namespaces, fixed XMLNS (incorrectly declared), ensured all namespaces listed in `namespaces.py.__all__` so can be found, blacked `namespaces.py` & `__init__.py`.

This PR is failing the Py2 tests and this is ok: we'll likely be removing the py2 tests soon after the 5.0.0 release and this PR is aimed at the 6.0.0 release.


tgbugs:
Possibly of interest here https://github.com/tgbugs/pyontutils/blob/master/pyontutils/closed_namespaces.py. Generates the closed namespaces from their source document when main is run, otherwise just acts as a library.

nicholascar:
Thanks @tgbugs for the script! I've only timidly added ClosedNamespaces for a couple of very well known and long-term stable ontologies but we could perhaps use the script for all onts listed here (including schema.org??) since you've made it fast to re-run.

I see no reason why we wouldn't add lots of ontologies to rdflib since they aren't large and in doing so we accord with the "batteries included" Python philosophy.

This is only due for 6.0.0 (failing Python 2 tests and I won't fix that, just remove Py2 testing soon after 5.0.0).

niklasl:
FWIW, I think `ClosedNamespace` is a too restrictive for any living ontology (such as schema.org), and would prefer something like an *optional* verifying mechanism which dereferences ontologies, caches them (using or establishing a convention for that, perhaps similar to [XML Catalogs](https://en.wikipedia.org/wiki/XML_catalog)), and then checks for a definition of the term IRI to be used.

(Related would be something like a `CachingNamespace` which wouldn't check anything but lazily set attributes on the namespace to avoid any recomputation of `URIRef` (or `IRI` if that would be the name in 6.0) values after first use.)

nicholascar:
> FWIW, I think ClosedNamespace is a too restrictive for any living ontology 

Agreed, hence I've not done anything with schema.org other than just declare the namespace.

> ...would prefer something like...

I'd like to get a many common namespaces in to rdflib and then pick off those that could be `ClosedNamespace` and then consider others. With @tgbugs's script mentioned above, it might be possible to add in lots of Closed ones for ontologies that are not living.

It might be worth putting out the call to rdflib-dev to ask for namespace contributions - the declarations and the CloseNamespace declaration - for their favourite ontologies.

hsolbrig:
My take is that the ClosedNamespace was a way of detecting typos.  I would suggest that we consider some variant of `dataclass` or `enum`  for the 6.0 version that would allow IDE's to auto-fill w/o absolutely forcing answers

tgbugs:
Failing errors on typos is the primary use case for ClosedNamespace, so that functionality needs to be retained. OpenNamespace (or similar) could be created to support the non-failing case where we also want to provide values for auto-complete but don't want to fail on mismatches. In theory could also add an option to OpenNamespace to warn on mismatch rather than fail.

nicholascar:
@tgbugs @hsolbrig these comments indicate one of both of you might lead this particular (I think small) piece of work??

Looks like we are talking about a new Class - Open/Cached Namespace with a couple of methods and perhaps a one-time processing of a bunch of ontologies into either Open- or ClosedNamespace objects, using the above script.

Could one or other of you please champion this for the planning meeting (whether in personal attendance or just by adding to the [agenda](https://github.com/RDFLib/rdflib/wiki/2020-04-28-6.0.0-Planning-Meeting) with your name next to it).

Note that there is this difficult issue that is somewhat related: https://github.com/RDFLib/rdflib/pull/597. We don't have to solve this in order to progress this Open- ClosedNamespace work but since we're all looking at `Namespace` it might be worth considering this PR.

nicholascar:
New work is about to start on reevaluating `ClosedNamespace` but, for the moment, I'm going to merge this PR in to get it off the books. After all, it only adds a single `ClosedNamespace` and fixes another.

PR #1002, opened 2020-04-14
CONSTRUCT resolve with initBindings fixes #1001

### LINKED ISSUE #1001: Variables from query initBindings ignored in CONSTRUCT body
If I feed a `CONSTRUCT` query with some initial bindings, e.g.
```python
query = """
prefix ex: <http://example.com/>
CONSTRUCT {
  ?uri ex:prop1 ?val1;
       ex:prop2 ?c .
}
WHERE {
  bind(uri(concat("http://example.com/", ?a)) as ?uri)
  bind(?b as ?val1)
}
"""
graph = rdflib.Graph()
results = graph.query(query, initBindings={
  'a': graph.Literal('A'),
  'b': graph.Literal('B'),
  'c': graph.Literal('C')
})
```
then I will get a single triple back (`ex:A ex:prop1 "B"`), because it uses variables that are bound in the `WHERE` block, but I will not get back `ex:A ex:prop2 "C"`, because the `?c` binding is not propagated into the `CONSTRUCT` evaluation.
If I modify the query as follows, it works fine:
```python
query = """
prefix ex: <http://example.com/>
CONSTRUCT {
  ?uri ex:prop1 ?val1;
       ex:prop2 ?val2 .
}
WHERE {
  bind(uri(concat("http://example.com/", ?a)) as ?uri)
  bind(?b as ?val1)
  bind(?c as ?val2)
}
"""
```
Is this is not WAI, would love to help on  a fix, trying to learn this code base.

sa-bpelakh:
Addresses https://github.com/RDFLib/rdflib/issues/1001 by defaulting to initBindings from the context during template substitution.

### DIFF
--- a/rdflib/plugins/sparql/sparql.py
+++ b/rdflib/plugins/sparql/sparql.py
@@ -173,7 +173,10 @@ class FrozenBindings(FrozenDict):
         if not type(key) in (BNode, Variable):
             return key
 
-        return self._d[key]
+        if key not in self._d:
+            return self.ctx.initBindings[key]
+        else:
+            return self._d[key]
 
     def project(self, vars):
         return FrozenBindings(
--- a/test/test_sparql_construct_bindings.py
+++ b/test/test_sparql_construct_bindings.py
@@ -3,6 +3,7 @@ from rdflib.plugins.sparql import prepareQuery
 from rdflib.compare import isomorphic
 
 import unittest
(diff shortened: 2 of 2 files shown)


sa-bpelakh:
Confused about the decrease in coverage, as coverage on the one module I modified actually increased from 87 to 88%.

PR #1005, opened 2020-04-16
Issue 1003

### LINKED ISSUE #1003: @base is not set
Hello there,

I have the following issue:

the `@base`-property is not set like it is recommended, when I serialize the Graph as turtle (e.g. https://www.w3.org/TR/turtle/#sec-intro, Example 1). I hope I'm just missing something, but I couldn't find anything in the documentation.

Here is an minimal example:
```
from rdflib import Graph, Literal, Namespace, RDF, URIRef
from rdflib.namespace import SKOS, DCTERMS

g = Graph()

n = Namespace("http://example.org/")

category = URIRef(n)
title = Literal("Title", lang="en")
description = Literal("Test Description", lang="en")
creator = Literal("<https://creator.com>")

# Add triples using store's add method.
g.add( (category, RDF.type, SKOS.ConceptScheme) )
g.add( (category, DCTERMS.title, title) )
g.add( (category, DCTERMS.description, description) )
g.add( (category, DCTERMS.creator, creator) )

# Bind a few prefix, namespace pairs for more readable output
g.bind("dct", DCTERMS)
g.bind("skos", SKOS)
g.bind("category", n)

output = g.serialize(format='turtle', base=n).decode("utf-8")

print(output)
```

This is the output:
```
@prefix category: <http://example.org/> .
@prefix dct: <http://purl.org/dc/terms/> .
@prefix skos: <http://www.w3.org/2004/02/skos/core#> .

<> a skos:ConceptScheme ;
    dct:creator "<https://creator.com>" ;
    dct:description "Test Description"@en ;
    dct:title "Title"@en .
```

This is the output I would expect (and want):
```
@base <http://example.org/> .
@prefix dct: <http://purl.org/dc/terms/> .
@prefix skos: <http://www.w3.org/2004/02/skos/core#> .

<> a skos:ConceptScheme ;
    dct:creator "<https://creator.com>" ;
    dct:description "Test Description"@en ;
    dct:title "Title"@en .
```

Am I missing something or is this right and on purpose?

nicholascar:
Adds `@base` to N3 & Turtle serializations, if `base` is set in `.serialize()`

Closes #1003


nicholascar:
close/open Travis

rdflib/graph.py
```diff
@@ -952,14 +953,19 @@ def serialize(
         Format support can be extended with plugins,
         but "xml", "n3", "turtle", "nt", "pretty-xml", "trix", "trig" and "nquads" are built in.
         """
+
+        # if a base is set here, use it, else if it was set at graph creation time, use that (self.base)
+        if base is not None:
+            self.base = base
+
```
white-gecko, line 960: I don't like this. It would overwrite the base as it was set on the graph.
I would expect a base that I set on the graph object to remain like that unless I explicitly change it on the graph object. If I set a base for the call to the `serialize` method I would expect this base to be only valid for the run time of this serialization.

I would make the behavior the other way round and set the local variable to `self.base`, if `base is None`.

rdflib/plugins/serializers/turtle.py
```diff
@@ -224,7 +224,9 @@ def serialize(self, stream, base=None, encoding=None,
                   spacious=None, **args):
         self.reset()
         self.stream = stream
-        self.base = base
+        # if base is set here, override previously set base if set at graph init
+        if base is not None:
+            self.base = base
```
white-gecko, line 229: In which case is `self.base` set already? The `TurtleSerializer` inherits from `RecursiveSerializer` from `Serializer` which is an `object`. It does not come from `Graph`.
The only reason is a previous call of `serialize`, which can be considered as a side effect.

white-gecko:
We should also make sure to have the base set for TriX, RDF/XML, and TriG (https://github.com/RDFLib/rdflib/tree/master/rdflib/plugins/serializers)

rdflib/graph.py
nicholascar: > I would expect this base to be only valid for the run time of this serialization.

I agree, I think this is better. I'll make this change.

rdflib/plugins/serializers/turtle.py
nicholascar: OK, yes I got that wrong, will change for this too. As above, I agree that setting base at serialization time should be for that serialization only - no side effects. An the Graph's base should override.

nicholascar:
> We should also make sure to have the base set for TriX, RDF/XML, and TriG

OK, I'll make the same changes to all

rdflib/plugins/serializers/turtle.py
white-gecko: I think the graphs base should only override in the case that the serialize base is None.

nicholascar:
I've made the changes I said I would above and got RDF/XML & TriX to work. TriG only 1/2 works since it can't express a base per graph, only one for the whole dataset/conjunctive graph. This means the full solution, which I have no time for, is to reverse QNAME any URI in any individual graph that doesn't use the same base as the dataset/conjunctive graph but the Turtle serializer, which the TriG serializer uses, doesn't keep track of things to make this easy! 

So there is a loss of info if multiple graphs within a dataset/conjunctive graph declare base URIs different to the dataset/conjunctive graph and then serialize to TriG... future work!

rdflib/plugins/serializers/turtle.py
nicholascar: Just check, this may not be arond the way you expect it to be.

nicholascar: Just check this is now the way you expect it should be.

white-gecko:
> So there is a loss of info if multiple graphs within a dataset/conjunctive graph declare base URIs different to the dataset/conjunctive graph and then serialize to TriG... future work!
I think this is not necessary. The base in general is not relevant info. It is just a convenience for serializations. So I think we do not need to go the extra mile here.

Setting the base on a Graph object is also not necessary but I think a nice feature to give a hint to the Serializers.

An I think the currently set base as attribute on a serialize method should always win over the base set in a graph. `serialize(base=base) > graph.base > None`

white-gecko:
@nicholascar I have made the changes in the way as I would expect the behavior. Is this ok for you? Maybe @ashleysommer should now review and merge this pullrequest.

nicholascar:
Hi @white-gecko great, I was just about to make the changes you just made! I'll approve and merge as @ashleysommer's not available at the moment.

PR #1018, opened 2020-04-20
fixing non recognition of @base declaration when parsing turtle files

S0MXcC0i64zh:
Hi,

When attempting to import some example turtle [file](https://www.w3.org/2007/02/turtle/primer/#example19) with RDFLib, the following code

`from rdflib import Graph`
` `
`g = Graph()`
`g.parse("test.ttl", format="turtle")`

returns a BadSyntax exception, because it does not take into account of the following line of the test.ttl file:

`@base <http://example.org/schemas/vehicles>`

After digging into the matter, it seems that the bug originates from the Turtle/Notation3 parser, more especially rdflib/plugins/parsers/notation3.py, line 1186, in uri_ref2.

I created some test file to reproduce the bug, and then fixed it by the present pull request.

My correction does not seem to break existing nose tests.

Please consider my pull request to mitigate said issue

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -1187,7 +1187,7 @@ class SinkParser:
                     if pfx == "_":  # Magic prefix 2001/05/30, can be changed
                         res.append(self.anonymousNode(ln))
                         return j
-                    if not self.turtle and pfx == "":
+                    if pfx == "":
                         ns = join(self._baseURI or "", "#")
                     else:
                         self.BadSyntax(argstr, i,
--- a/test/test_turtle_base.py
+++ b/test/test_turtle_base.py
@@ -1,0 +1,108 @@
+import os
+from rdflib.graph import Graph
+from rdflib.term import URIRef
+
+turtle = """@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
+@base <http://example.org/schemas/vehicles>.
+
+:MotorVehicle a rdfs:Class.
+
+:PassengerVehicle a rdfs:Class;
+   rdfs:subClassOf :MotorVehicle.
+
+:Truck a rdfs:Class;
+   rdfs:subClassOf :MotorVehicle.
+    
+:Van a rdfs:Class;
+   rdfs:subClassOf :MotorVehicle.
+
+:MiniVan a rdfs:Class;
+   rdfs:subClassOf :Van;
+   rdfs:subClassOf :PassengerVehicle.
+
+:Person a rdfs:Class.
+
+xsd:integer a rdfs:Datatype.
+
+:registeredTo a rdf:Property;
+   rdfs:domain :MotorVehicle;
+   rdfs:range  :Person.
+    
+:rearSeatLegRoom a rdf:Property;
+   rdfs:domain :MotorVehicle;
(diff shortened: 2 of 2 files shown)


white-gecko:
From the first sight you are following some misconception. There is a difference between a relative IRI and the usage of a prefixed name with an empty prefix label `:localPart` (cf. https://www.w3.org/TR/turtle/#sec-iri; Example 9).

Consider the following example:
```
from rdflib import Graph

g = Graph()
g.parse(data="""
@base <http://example.org/schemas/vehicles/>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
<Bicycle> a rdfs:Class.
""", format="turtle")
for t in g:
    print(t)

h = Graph()
h.parse(data="""
@base <http://example.org/schemas/vehicles#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
<Bicycle> a rdfs:Class.
""", format="turtle")
for t in h:
    print(t)

i = Graph()
i.parse(data="""
@base <http://example.org/schemas/vehicles>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
<Bicycle> a rdfs:Class.
""", format="turtle")
for t in i:
    print(t)

j = Graph()
j.parse(data="""
@base <http://example.org/schemas/vehicles>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
<#Bicycle> a rdfs:Class.
""", format="turtle")
for t in j:
    print(t)

k = Graph()
k.parse(data="""
@base <http://example.org/schemas/vehicles>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
:Bicycle a rdfs:Class.
""", format="turtle")
for t in k:
    print(t)
```

```
(rdflib.term.URIRef('http://example.org/schemas/vehicles/Bicycle'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'), rdflib.term.URIRef('http://www.w3.org/2000/01/rdf-schema#Class'))
(rdflib.term.URIRef('http://example.org/schemas/Bicycle'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'), rdflib.term.URIRef('http://www.w3.org/2000/01/rdf-schema#Class'))
(rdflib.term.URIRef('http://example.org/schemas/Bicycle'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'), rdflib.term.URIRef('http://www.w3.org/2000/01/rdf-schema#Class'))
(rdflib.term.URIRef('http://example.org/schemas/vehicles#Bicycle'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'), rdflib.term.URIRef('http://www.w3.org/2000/01/rdf-schema#Class'))
Bad syntax (Prefix ":" not bound) at …
```

In the first four cases the relative IRIs are interpreted relative to the in-scope base, while in the fifth case correctly a Bad syntax is recognized as the Prefix is not defined.

I'm unsure if the cases `h` and `i` are correct and could not find it from briefly skim through the specs.
While I would at least expect case `h`to be `http://example.org/schemas/vehicles#Bicycle` instead of `http://example.org/schemas/Bicycle`.

S0MXcC0i64zh:
Thanks for the time you dedicated to the reply @white-gecko. It is not always easy to find our way in the w3c documentation. At least, I know which document is the good entry point for the turtle specification. I already came across it but found it hard to understand, this is probably why I went to the outdated Turtle primer...
I also learnt how to use the RDFLib in editable mode for development purpose and how to prepare a pull request. This was not totally useless :-)

PR #1022, opened 2020-04-24
Add the content type 'application/sparql-update' when preparing a SPARQL update request

### LINKED ISSUE #1021: Missing Content-Type header for SPARQLConnector update method (?)
In rdflib 4.2.2 the SPARQLWrapper added an applicable content-type header when sending queries and updates to a SPARQL endpoint.
The 5.0.0 the SPARQLConnector does not add a content-type and, since the GraphDB we use as SPARQL endpoint rejects update requests when Content-Type: application/sparql-update is not added to the header, I had to monkey patch the update method to forcebily add the content type to the header.

I also tried to add a headers parameter with the content-type to the SPARQLUpdateStore parameter when constructing a SPARQLUpdateStore instance, but this then the content type is also used for query request, which results in rejected query requests in case the store instance is used for querying.

Should the the content type be added by the SPARQLConnector like the SPARQLWrapper did or am i missing an other intended solution?

Best regards,
Arnoud

NoutieH:
PR for issue #1021  @nicholascar 

When sending a POST request to the SPARQL update endpoint, the endpoint wants to be told which type of content it'll receive in request.
Since SPARQLConnector encodes the data with UTF-8 the request library can not deduce the correct Content-Type itself. By adding it explicitly to the headers parameter, the requests library will send to request the SPARQL update endpoint understands. At least when the sparql endpoint is a GraphDB instance...

Note: Didn't find a SPARQLConnector unit test where the addition of the content type header could be validated. Skipped making one... (?)

Closes #1021

### DIFF
--- a/rdflib/plugins/stores/sparqlconnector.py
+++ b/rdflib/plugins/stores/sparqlconnector.py
@@ -73,16 +73,18 @@ class SPARQLConnector(object):
         if default_graph:
             params["default-graph-uri"] = default_graph
 
+        headers = {
+            'Accept': _response_mime_types[self.returnFormat],
+            'Content-Type': 'application/sparql-query',
+        }
+
         args = dict(self.kwargs)
         args.update(url=self.query_endpoint)
 
         # merge params/headers dicts
         args.setdefault('params', {})
 
-        headers = {
-            'Accept': _response_mime_types[self.returnFormat],
-            'Content-Type': 'application/sparql-query',
-        }
+        args.setdefault('headers', {})
         args['headers'].update(headers)
 
         if self.method == 'GET':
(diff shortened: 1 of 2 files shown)


effigies:
Going by https://w3c.github.io/rdf-tests/sparql11/data-sparql11/protocol/index.html, might be worth updating `query()` to use `application/sparql-query` as well?

nicholascar:
@effigies I've done as you suggest adding to `query()`.

@white-gecko: since I've added to this, please could you review?

NoutieH:
@nicholascar Tnx for your updates! One thing: Given the w3C query spec the Content-Type header only needs to be set when request method is POST...

nicholascar:
@white-gecko can you please review and likely merge? Just want to make sure I, as a contributor to the PR, don't merge

white-gecko:
Could we add a test for this?

white-gecko:
Now I have contributed as well. But I think the test is the code-ification of my review result so I did approve it and will merge it.

PR #1044, opened 2020-05-10
Updating namespace.py to solve issue #801

### LINKED ISSUE #801: Problem with prefixes created for URIs containing %20
The result of runnig this code
```python
from rdflib import Namespace, Graph, BNode, Literal

graph = Graph()
namespace = Namespace('http://example.org/')
graph.bind('', namespace)
node = BNode()

graph.add((node, namespace['first%20name'], Literal('John')))
print(graph.serialize(format='turtle').decode())
```
is
```
@prefix : <http://example.org/> .
@prefix ns1: <http://example.org/first%20> .

[] ns1:name "John" .
```
I expect that triple would be displayed as 
```
[] :first%20name "John" .
```

aayush17002:
Added acceptance clause for "%" in Allowed name chars.

Closes #801


white-gecko:
Were you able to check if this also results in valid results for the other serialization formats? https://github.com/RDFLib/rdflib/issues/801#issuecomment-353026657

aayush17002:
Yes on making the changes described in my code I was able to get valid results other serialization formats also.

aayush17002:
Sir,
Are there any other changes required, If not, please accept my solution.

test/test_issue801
```diff
@@ -0,0 +1,23 @@
+  """
+Issue 715 - path query chaining issue
+Some incorrect matches were found when using oneOrMore ('+') and
+zeroOrMore ('*') property paths and specifying neither the
+subject or the object.
```
darrengarvey, line 5: This comment looks out of date.

test/test_issue801
```diff
@@ -0,0 +1,23 @@
...
+
+class TestIssue801(unittest.TestCase):
+
+    def test_issue_801(self):
+        g = Graph()
+        example = Namespace('http://example.org/')
+        g.bind('', example)
+        node = BNode()
+        g.add((node, example['first%20name'], Literal('John')))
+        print(g.serialize(format="turtle").decode())
```
darrengarvey, line 19: Shouldn't this be checking the result of the serialisation?

test/test_issue801
aayush17002: @darrengarvey I have updated the comment

test/test_issue801
aayush17002: @darrengarvey I have added the assert statement for checking the result

test/test_issue801.py
```diff
@@ -13,7 +13,8 @@ def test_issue_801(self):
         g.bind('', example)
         node = BNode()
         g.add((node, example['first%20name'], Literal('John')))
-        print(g.serialize(format="turtle").decode())
+        self.assertEqual(g.serialize(format="turtle").decode().split("\n")[-3] == '[] :first%20name "John" .', True)
```
darrengarvey, line 16: Can this instead be:

```
self.assertEqual(g.serialize(format="turtle").decode().split("\n")[-3],
                '[] :first%20name "John" .')
```

That way if the test fails you get to see what both sides of the equality comparison are. Plus keeping the line length shorter and making the string easier to see.

test/test_issue801.py
```diff
@@ -13,7 +13,8 @@ def test_issue_801(self):
...
+        print (g.serialize(format="turtle").decode())
```
darrengarvey, line 17: This is going to make test run output noisy. I don't see other tests in rdflib printing output, so do you need this?

test/test_issue801.py
aayush17002: @darrengarvey I have applied the requested changes

test/test_issue801.py
aayush17002: @darrengarvey I have applied the requested changes

nicholascar:
@aayush17002 please merge in `master` branch since a whole host of changes were recently merged in to remove Python 2 code. This will trigger a re-checking of the code by Travis. Then it may be ready for merging.

aayush17002:
@nicholascar I have merged into the master branch again as suggested by you

Issue #965, opened 2020-03-10
Serialization drops unused namespaces


Abbe98:
Sterilization drops unused namespaces even when those are explicitly bound to the graph:

```python
graph = Graph()
SOCH = Namespace('http://kulturarvsdata.se/ksamsok#')
PRES = Namespace('http://kulturarvsdata.se/presentation#')
graph.bind('ksam', SOCH)
graph.bind('pres', PRES)
print(graph.serialize(format='xml'))
```

Output:
`b'<?xml version="1.0" encoding="UTF-8"?>\n<rdf:RDF\n   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"\n>\n</rdf:RDF>\n'`

tgbugs:
See https://github.com/RDFLib/rdflib/pull/914.

nicholascar:
Is this perhaps an RDF/XML issue only? When I run that code for Turtle output from rdflib 4.2.2 (are you running with a build from `master` perhaps?) , I see the namespaces:

```
from rdflib import Graph, Namespace

graph = Graph()
SOCH = Namespace('http://kulturarvsdata.se/ksamsok#')
PRES = Namespace('http://kulturarvsdata.se/presentation#')
graph.bind('ksam', SOCH)
graph.bind('pres', PRES)
print(graph.serialize(format='turtle').decode('utf-8'))
```

I get:

```
@prefix ksam: <http://kulturarvsdata.se/ksamsok#> .
@prefix pres: <http://kulturarvsdata.se/presentation#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
```

tgbugs:
4.2.2 does not have the issue. The change is cause by https://github.com/RDFLib/rdflib/pull/832

nicholascar:
@tgbugs thanks for the confirmation of version. 

As per the thoughts in Issue #948 that you've commented on, I think we need to review Turtle 1.1 parsing/serialization generally which, as you've indicated, might be dependent on deeper CURIE suffix handling.

I've labelled this Turtle 1.1 to keep those issues together.

nicholascar:
[...]

Issue #794, opened 2017-12-03
Remove unused prefixes from serialization


noorbakerally:
I have a rdflib.Graph object to which I bind several namespaces. The issue is that when serializing the graph, many of the binded namespaces are not used but are still in the serialized text. I want only the used prefixes to be serialized. How is this possible ?

For example, the graph below is generated by rdflib and many unused prefixes are serialized together:

```
@prefix child: <http://opensensingcity.emse.fr/ldpdfend/paris/d1/catalog/> .
@prefix dcat: <http://www.w3.org/ns/dcat#> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix ldp: <http://www.w3.org/ns/ldp#> .
@prefix parent: <http://opensensingcity.emse.fr/ldpdfend/paris/d1/> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

parent:catalog a ldp:BasicContainer,
        ldp:Container,
        ldp:RDFSource ;
    ldp:contains child:datasets ;
    foaf:primaryTopic <https://opendata.paris.fr/api/v2/catalog/exports/ttl> .

<https://opendata.paris.fr/api/v2/catalog/exports/ttl> a dcat:Catalog ;
    dcterms:Language <http://id.loc.gov/vocabulary/iso639-1/en>,
        <http://id.loc.gov/vocabulary/iso639-1/fr> ;
    dcterms:description "Parisdata Catalog" ;
    dcterms:publisher <http://www.opendatasoft.com> ;
    dcterms:title "Parisdata's catalog" ;
    dcat:themeTaxonomy <https://opendata.paris.fr/concept-scheme/themes> ;
    foaf:homepage <https://opendata.paris.fr> .
```

joernhees:
I think atm this is mostly done so that prefixes are listed first and as we might not know which of the bound namespaces is actually used. The current store backend might not give us this informaiton in a cheap way, requiring us to run a full pre-iteration of the data to be serialized.

That said, i'd also like options to drop unused prefixes and maybe even have inline prefix serialization...

noorbakerally:
[...]

Issue #710, opened 2017-02-05
Remove some/all default prefix bindings


niklasl:
In https://github.com/RDFLib/rdflib/blob/master/rdflib/namespace.py, there are currently some prefixes bound by default in `NamespaceManager`.

At least one of them, `xml`, should be removed, since that isn't an RDF vocabulary at all (but rather an old artifact stemming from RDF being nigh synonymous with RDF/XML in the past). This unnecessarily clutters all serialization output.

The other ones, `rdf`, `rdfs` and `xsd` are more relevant of course. It is still the case that neither `rdfs` nor `xsd` are used in every piece of serialized data though (schema.org comes to mind). And even if `rdf:type` is close to ubiquitous, it doesn't necessarily warrant the prefix being bound by default.

My thought is to remove all but `rdf`, unless anyone can argue for something else. They are relevant, AFAIK, only to serializations (and for relying on implicit bindings in SPARQL queries). In these cases, the source data and/or explicit bindings ought to give enough control over this.

I'll heartily do the removal; I just want to run it by you first. Is there any (conceivable) real usage depending on one of more of these being bound by default? Even if there are, it could be argued that such practice would be obscure, and that in any case, RDFLib 5 can mark a departure from that. Such usage might just as well expect `owl` or `skos` to be bound as well (or `dc`, `foaf`, `schema`...).

Thoughts?

joernhees:
hmm, manually binding them is cumbersome, isn't it?

As long as they're lazily serialized, i'd actually prefer fallbacks to all in http://prefix.cc ;)

niklasl:
Well, it's just a call to: `graph.namespace_manager.bind(pfx, ns)`. Or you could parse e.g. an empty turtle file with prefixes, which causes them to be bound in the namespace manager of the parsing graph.

They're not lazily serialized (at least not consistently). They appear in e.g. all turtle output. I jumped through some hoops to avoid them popping up by default in the json-ld serializer.

And then there is the issue of re-binding: #543. I guess it's not as common, but I think anyone wanting to use `xs` instead of `xsd` is hit by that.

For pretty serialization, it would be a boon to have lots of lazily bound prefixes over the generated ones, that I agree with. But I'd say that's a different feature: to use well-known prefixes over generating the `ns{i}` stuff. For one, it should not override a prefix given in the source, unless explicitly told so. In fact, that's yet another possible feature: providing control over whether to retain or discard prefixes in a source during parse time.

gromgull:
[...]

PR #1047, opened 2020-05-12
Solution for issue #794 

piyushjainiiitd:
This will provide a solution to **issue 794** with this user can remove the unused prefix by calling the function removeUnusedPrefix() in the graph class. If this function is not called then all the unused prefixes will be shown. This will provide substantial power to the user whether he/she wants to see the unused prefixes or not. It also helps with the **issue 965** and **issue 710**. #794 #965 #710

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -943,6 +943,35 @@ class Graph(Node):
         """Turn uri into an absolute URI if it's not one already"""
         return self.namespace_manager.absolutize(uri, defrag)
 
+    def removeUnusedPrefix(self):
+        """Remove unused prefix in the graph and remove it from the graph"""
+        for namespace in self.namespaces(): #Iterate for all the namespace in the document check if the namespace is present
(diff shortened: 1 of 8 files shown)


nicholascar:
@piyushjainiiitd please merge in `master` branch since a whole host of changes were recently merged in to remove Python 2 code. This will trigger a re-checking of the code by Travis. Then it may be ready for merging.

nicholascar:
Sorry but this PR is problematic in several ways. Unused prefixes should not appear in serialised graphs - this is the behaviour of RDFlib 5.0.0 but didn't work in 4.2.2 - so this work is actually redundant as of 5.0.0.

Secondly, even if this issue wasn;t already addressed in 5.0.0, this PR unbinds a bound prefix from a graph BUT this is problematic: a graph could eventually have a triple added that does use the prefix so the binding should stay there until the graph is destroyed

Closing PR without merging.

PR #1051, opened 2020-05-17
Removal of six & styling

nicholascar:
This PR removes all references to the package six which handles PY2/PY3 differences since we are no longer catering for PY2.

All tests pass locally - I even ran the SPARQLUpdateStore tests!

I've also `black`ed all Python files to ensure consistent formatting and addressed all issues from `flake8` testing except for broad exceptions & line length warnings.

No functionality changes.


ashleysommer:
How hard would it be to split these into two different PRs, one for `six` and one for black and flake8 styling?

nicholascar:
Well, they are split by commits, but why the worry, they are all tidy-up steps without any functional changes?

`flake8` used to be run semi-regularly on the codebase so that's why I did that one.

ashleysommer:
No concern really, should be fine.

nicholascar:
All right @ashleysommer & @white-gecko: please review and, if this gets past, we can then look into all the functional change PRs properly. 

There *are* lots of edits here but they are all the same thing, many times over, in many files, so you should be able to be happy with the pattern after a few checks...

rdflib/plugins/serializers/turtle.py
```diff
@@ -292,122 +294,125 @@ def getQName(self, uri, gen_prefix=True):
...
             self.predicateList(node, newline=False)
             # self.write('\n' + self.indent() + ']')
-            self.write(' ]')
+            self.write(" ]")
             self.depth -= 1
 
         return True
 
-    def isValidList(self, l):
+    def isValidList(self, l_):
```
niklasl, line 394: Just curious, why wasn't just `l` OK here? (And if it isn't, the docsstring ought to correspond.)

nicholascar: Only because `flake8` threw up a lot of warnings about possible variable ambiguity for `l` so there must be a higher-order variable `l` still within scope that this might clash with.

I suppose I should have thought o rebuild docstrings before putting in the PR. I'll try and do that and add to this.

ashleysommer: the flake8 error regarding `l` is because it can look like an uppercase i or a number 1. Its generally considered using a single letter variable named `l` is a bad thing. This solution works to suppress the error, and I think doesn't harm anything.

examples/film.py
```diff
@@ -35,7 +35,7 @@
 
 from rdflib import BNode, ConjunctiveGraph, URIRef, Literal, Namespace, RDF
 from rdflib.namespace import FOAF, DC
-from six.moves import input
+from builtins import input
```
ashleysommer, line 38: You don't need to import `input` from `builtins`. Anything in `builtins` is automatically imported into every file by default, you can just use `input()` without importing anything.

rdflib/plugins/serializers/turtle.py
niklasl: I see. In practise I think `l` is good enough (in spite of flake8), but following that general advice, I think `lst` would be better, or even if we could establish a naming pattern like `rlist` (and similarly one for types, like `rtype`). (Going for `rdf_list` and `rdf_type` may seem even better/clearer, unless it becomes unclear whether they would reference an object or the terms themselves (`RDF.type` and `RDF.List`).)

Just something to consider going further, nothing critical for this PR.

examples/film.py
nicholascar: Thanks @ashleysommer, I've removed that line.

nicholascar:
@white-gecko can you please review this?

PR #1053, opened 2020-05-19
Enhacement for graph visualization

piyushjainiiitd:
This pull request contains code for graph visualization. This code is based on graph viz library of python which is used to visualize the graph. It allows great freedom to the user to see their RDF graph in various formats like png, pdf, etc. Along with this people can also see their graphs in **short mode** which means the prefix part is removed so that graphs become small and easy to visualize. Different color codes are used to differentiate graph entities.


tgbugs:
It is unlikely that any pull request that adds graphviz or any number of other visualization/graph libraries as a core dependency will be accepted. Sandboxing the dependency so that it can be made optional would be one way around this.

niklasl:
I'd recommend making a separate package for visualizing RDFLib graphs using graphvis instead.

PR #1056, opened 2020-05-19
Graph visualization using graphviz

piyushjainiiitd:
This pull request contains code for graph visualization. This code is based on graph viz library of python which is used to visualize the graph. The prerequisite for this code is that the Graphviz python library should be installed on your device. It allows great freedom to the user to see their RDF graph in various formats like png, pdf, etc. Along with this people can also see their graphs in a short mode which means the prefix part is removed so that graphs become small and easy to visualize. Different color codes are used to differentiate graph entities.
An example of the same is given in graphvizVisualize.py file in the example folder.


nicholascar:
@piyushjainiiitd as per your other PRs, please merge in `master` branch since a whole host of changes and let's see the tests run again after that.

piyushjainiiitd:
This new commit adds the feature to visualize the path between two nodes. This helps in finding the similarity to other nodes. Trying to add more features for visualization.

nicholascar:
Hi @piyushjainiiitd, Thanks for contributing to RDFlib - grahp visualisation is an interesting and useful thing.

However, there are a couple of issues here.

* your example data is a Python file that wraps an RDF file and returns a graph - why not just provide examples as RDF?
* your actual example running code, in the mainly data file, is commented out so doesn't run
* you've created a whole top-level RDFlib folder just for your one file: graphvizVisualization/
    * this adds complexity to the project for very little - one example

Please could you re-issue this PR with a working example single python file in examples/?

nicholascar:
I'm closing this PR as there's been no reply from the raiser, @piyushjainiiitd in about a month.

PR #1057, opened 2020-05-20
fix #910: Updated evaluate.py so that union includes results of both branches, even when identical.

### LINKED ISSUE #910: SPARQL: UNION with two identical branches returns only result for one branch
This SPARQL query should return two solutions, both identical, binding variable `?a` to string `"a"`:
```sparql
SELECT * {
    { BIND ("a" AS ?a) }
    UNION
    { BIND ("a" AS ?a) }
}
```
But it returns only one of the two solutions. The same effect—only the first union returns any results—happens when the unions have more than one variable or more than one solution each. The problem is triggered by the _results_ being identical, not just by the _graph patterns_ being identical. This one also just returns one solution:
```sparql
SELECT * {
    { BIND ("a" AS ?a) }
    UNION
    { VALUES ?a { "a" } }
    UNION
    { SELECT ("a" AS ?a) {} }
}
```
The problem goes away as soon as there is any difference between the results of the two union branches.

kushagr08:
The result returned by select queries involving union should contain results of both the branches, even if they are identical, as per the w3c union cardinality documentation.
https://www.w3.org/TR/sparql11-query/#defn_algUnion

Close #910


nicholascar:
@white-gecko will this use of a list instead of a generator be problematic for large results?

kushagr08:
@nicholascar 
I have added the unit test file for this issue in the test folder, named  test_issue910.py . Thanks

PR #1058, opened 2020-05-21
Fixes #1055 added a subgraph function.

### LINKED ISSUE #1055: Like .triples() but return a graph
My use case is that I would like to chain multiple operations like this:

    graph.triples().add().remove().serialize()

The problem is that `triples` returns an iterator, not a graph object. I do understand that `triples` needs an iterator to deal with the possible situation of having too many results at once. But assuming that I'm only retrieving a small number of triples, for example a single node like this `.triples(subject, None, None)` it would be nice to be able to chain those operations. Is is possible to do something like this?

achaudhary997:
Added a subgraph function which returns a subgraph consisting of the
triples which match the given pattern.


nicholascar:
@achaudhary997 as per your other PR, please merge in `master` here and let's see what the tests look like then.

nicholascar:
Hi @achaudhary997, sorry but there are two main issues with this POR:

1. you've included the unrelated work for decimal serialisation
2. the subgraph function offers no benefit to just piping out the triples from the normal triples() generator 
   * this is what you're doing internally - your code tries to hide this but imposes an arbitrary size limit, as it has to to stop memory blowout
    * for effective use by someone, this limit must be known so they have to then look inside your code. They may as well have tested the size and then used triples() to get the triples, not subgraph

We've decided not to implement this feature.

PR #1059, opened 2020-05-21
fixedIssue896

Kshitiz-Jain:
2 functions are added in the graph.py file.
1 for adding multiple triples in graph, whose input is list of triples
1 for adding multiple quads in graph, whose input is list of quads

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -409,16 +409,6 @@ class Graph(Node):
             and _assertnode(s, p, o)
         )
 
-    def addNAll(self, quads):
-        """Add a sequence of triples with context"""
-        for i in quads:
-            self.__store.addN(
-                (s, p, o, c)
-                for s, p, o, c in i
-                if isinstance(c, Graph)
-                and c.identifier is self.identifier
-                and _assertnode(s, p, o)
-            )
 
     def remove(self, triple):
         """Remove a triple from the graph
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -389,6 +389,15 @@ class Graph(Node):
         assert isinstance(o, Node), "Object %s must be an rdflib term" % (o,)
         self.__store.add((s, p, o), self, quoted=False)
 
+    def addAll(self, triples):
+        """Add a list of triples with self as context"""
+        for i in triples:
+            s,p,o = i
+            assert isinstance(s, Node), "Subject %s must be an rdflib term" % (s,)
+            assert isinstance(p, Node), "Predicate %s must be an rdflib term" % (p,)
+            assert isinstance(o, Node), "Object %s must be an rdflib term" % (o,)
+            self.__store.add((s, p, o), self, quoted=False)
+
     def addN(self, quads):
         """Add a sequence of triple with context"""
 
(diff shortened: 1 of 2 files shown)


nicholascar:
Hi @Kshitiz-Jain, This looks like it's a straightforward addition and likely a useful one so thanks for the work. Could you please also add a test in the test/folder (either create a new file or add to an existing one testing similar functions) so that this work is both covered by a test to maintain test coverage and also to ensure that we can test future changes against this code. 

You should add tests for:

* adding correctly formed triples
* broken triples (to trigger the `assert` statements you added)
* adding correctly formed quads
* broken quads

Thanks!

tgbugs:
@nicholascar no other method on Graph currently checks the validity of triples on the way in, why should this one? Further, checking using assert statements is useless in many production deployments because they are disabled.

Kshitiz-Jain:
Added test for addAll() API. It includes the required tests for adding triples and assertions statements.

nicholascar:
@tgbugs yes, you are right. I asked @Kshitiz-Jain to add tests for all aspects of the new code but I hadn't really reviewed the appropriateness of the code. The `assert` statements in there are definitely a new thing.

If @Kshitiz-Jain adds a couple more tests, perhaps this PR can be passed after stripping those statements out.

ashleysommer:
@tgbugs  @nicholascar 
>no other method on Graph currently checks the validity of triples on the way in, why should this one?

See graph.py fn `add()` at line 387. It _does_ currently have those same assert statements, that is why they are reproduced in the new `addAll` function.
PR https://github.com/RDFLib/rdflib/pull/1065 was also created to fix this same issue. It also includes identical assertion checks.

rdflib/graph.py
```diff
@@ -400,6 +409,17 @@ def addN(self, quads):
             and _assertnode(s, p, o)
         )
 
+    def addNAll(self, quads):
```
ashleysommer, line 412: the existing addN function already accepts a list of multiple quads. Adding `addNAll` is redundant.

ashleysommer, line 412: the existing `addN` function already accepts a list of multiple quads. Adding `addNAll` is redundant.

rdflib/graph.py
```diff
@@ -400,6 +409,17 @@ def addN(self, quads):
...
+        """Add a sequence of triples with context"""
+        for i in quads:
+            self.__store.addN(
+                (s, p, o, c)
+                for s, p, o, c in i
```
ashleysommer, line 417: I think this is a bug. Was this method tested at all? `i` is already a quad item from the list of quads, you cannot unpack it futher into more quads using this `for .. in` statement.

Kshitiz-Jain:
As rightly suggested by @ashleysommer the `addN` API is already adding multiple quads in the graph therefore the API `addNAll` is removed in the latest commit. 
New commit "addAll+testIssue896" contains only `addAll` API for adding multiple triples in the graph, along with that assertion tests suggested by @nicholascar are also added in the ./test/test_graph.py module with name `testAddAll`.

nicholascar:
This PR is essentially the same as PR #1065 but 1065 has an additional option and more tests, so we (maintainers) are closing this in favour of 1065, This doesn't mean we will necessarily accept 1065.

PR #1064, opened 2020-05-24
Fix Issue #981

### LINKED ISSUE #981: Auto-detect RDF type from file extension in parse()
`Graph().parse("some-file.ttl", format="turtle")` is a common way to load RDF into an rdflib Graph. We have methods such as `rdflib.util.guess_format()` to guess `format` from the file extensions so what we need now is for `guess_format()` to be triggered automatically if `format` is not given so `parse()` can be used like this:

```
Graph().parse("some-file.ttl")
```

mehak16163:
Fix for issue #981
I have added the functionality to predict the RDF type if either
1. Location of the file is specified
2. Under source a string representing the location of the file is passed. 

Along with that I have also added a test to verify that it correctly identifies a .n3 file without specifying the format keyword. 

Close #981


rdflib/parser.py
```diff
@@ -214,3 +219,75 @@ def create_input_source(source=None, publicID=None,
         elif input_source.getPublicId() is None:
             input_source.setPublicId(absolute_location or "")
         return input_source
+
+SUFFIX_FORMAT_MAP = {
```
niklasl, line 223: This is just a copy of the code in https://github.com/RDFLib/rdflib/blob/master/rdflib/util.py#L345? Why not import it instead?

mehak16163: I tried importing, but I was not able to. It led to many test cases failing.

mehak16163:
Imported the code for guess_format function. Importing just the function later in the code worked fine.

nicholascar:
Hi @mehak16163, thanks for this work. Could you please:

1. Merge in the `master` branch and retest - we're removed lots of Python 2 code
2. Add in at least one other file format test, e.g. RDF/XML
3. Add in a breaking test (perhaps for a .doc MS Word file or something irrelevant)

Thanks!

mehak16163:
Hey @nicholascar, 

I've have made all the changes. Can you please go through them and let me know if there are any more changes to be done.

ashleysommer:
@mehak16163 
There is an existing PR for a different fix for #981:
https://github.com/RDFLib/rdflib/pull/1046
and seems another just created 2 hours ago:
https://github.com/RDFLib/rdflib/pull/1075
Perhaps we can try to pick the best parts of all of these?

mehak16163:
> @mehak16163
> There is an existing PR for a different fix for #981:
> #1046
> and seems another just created 2 hours ago:
> #1075
> Perhaps we can try to pick the best parts of all of these?

Yes sure, how should we go about it?

mehak16163:
Hey @ashleysommer  , @nicholascar 
Is the current solution fit for merging or should I make some changes to it?

rdflib/parser.py
```diff
@@ -189,6 +189,9 @@ def create_input_source(
     auto_close = False  # make sure we close all file handles we open
     if location is not None:
         # Fix for Windows problem https://github.com/RDFLib/rdflib/issues/145
```
mwatts15, line 191: Please keep comments with the code they're commenting on.

rdflib/graph.py
```diff
@@ -1762,7 +1762,7 @@ def add(self, triple):
 
     def addN(self, quads):
         """Add a sequence of triple with context"""
-
+
```
mwatts15, line 1765: Why did you add a newline here?

tgbugs:
https://github.com/RDFLib/rdflib/pull/1046 seems like a better solution to me.

mehak16163:
> Please, delete rdflib/tc5.in

Done.

rdflib/graph.py
mehak16163: Fixed this. It wasn't required. I was just going through the code.

rdflib/parser.py
mehak16163: Fixed this.

nicholascar:
@mehak16163 I've gone with PR #1046 to address this Issue. You PR was close to that one but I slightly preferred the that implementation. You might like to compare your work and that work.

Thanks for the PR all the same: it's a real pleasure to get more than one to improve the code!

PR #1065, opened 2020-05-24
Fix Issue #896

mehak16163:
I have added a function Graph.add_all(self, triples) which can add all the triples in the list provided as an input to this function. 
Fix for issue #896


nilay77:
- New function `add_all` added to graph.py
- Added test case for adding correctly formed triples
- Added test case for broken triples

test/test_issue896.py
```diff
@@ -0,0 +1,72 @@
...
+
+        with self.assertRaises(AssertionError):
+            # Changing function parameter all_or_none to False
+            self.graph.add_all(self.triples, all_or_none=False)
+
+        # Triples before the assertion should be added to the graph
+        asserte(len(self.graph),3)
+
+if __name__ == "__main__":
+    unittest.main()
```
ashleysommer, line 72: Please add a newline at the end of this file

rdflib/graph.py
```diff
@@ -389,6 +389,27 @@ def add(self, triple):
...
+        """
+
+        if(all_or_none):
+            for i, (s, p, o) in enumerate(triples,1):
+                assert isinstance(s, Node), "Subject %s in Triple %d must be an rdflib term" % (s,i,)
+                assert isinstance(p, Node), "Predicate %s in Triple %d must be an rdflib term" % (p,i,)
+                assert isinstance(o, Node), "Object %s in Triple %d must be an rdflib term" % (o,i,)
+
+        for i, (s, p, o) in enumerate(triples,1):
+            if(all_or_none==False):
```
ashleysommer, line 407: I wonder if there might be a way to optimise this when `all_or_none` is True, to avoid this `if` statement on every iteration of this loop (answer is always going to be the same every iteration).

ashleysommer:
Even though the contributor in https://github.com/RDFLib/rdflib/pull/1059 has now removed the redundant `addNAll` method, I think I still prefer this implementation over #1059

rdflib/graph.py
```diff
@@ -389,6 +389,27 @@ def add(self, triple):
         assert isinstance(o, Node), "Object %s must be an rdflib term" % (o,)
         self.__store.add((s, p, o), self, quoted=False)
 
+    def add_all(self , triples, all_or_none=True):
```
ashleysommer, line 392: I know the name `add_all` is more pythonic and PEP8-compliant than `addAll`, but the rest of the methods on `Graph` were written a very long time before any sense of style existed in python. One day we might break everyones code by renaming them all to PEP8-compant names, but until then we'd better stick with the existing naming conventions in this file, rather than introduce inconsistencies.

test/test_issue896.py
nilay77: Sure.

rdflib/graph.py
nilay77: Yes I agree. I'll think of a better solution.

rdflib/graph.py
nilay77: I gave the name following PEP8 standards and also the current method names in `graph.py` namely `subject_predicates`, `subject_objects`, `all_nodes`. Should I still change the name to `addAll`?

rdflib/graph.py
nilay77: @ashleysommer I have optimized the code and I feel it has better readability too. Is it fine now?

rdflib/graph.py
ashleysommer: I'll have to get opinion of the other rdflib maintainers on that.

rdflib/graph.py
ashleysommer: Yes. Much better now. Thanks.

rdflib/graph.py
nilay77: Okay sure, do let me know.

nicholascar:
Dear @mehak16163, this is the best of the set of PRs all addressing this issue and there's nothing wrong with your code (passes tests, documented etc.) however we've decided not to merge this PR in for the following reason:

* we would like to consolidate the API of `Graph()` down to a few, well documented functions 

Fewer functions means less documentation maintenance, fewer tests required and people will tend to write more similar application code.

So, thanks for this PR but we will close it un-merged.

PR #1066, opened 2020-05-25
Fix #913: Return None in parseBoolean in case of weird boolean

### LINKED ISSUE #913: Update Lexical-to-value mapping for xsd:boolean

arushi019:


nicholascar:
Hi @arushi019, thanks for the PR. We (maintainers) see that you're allowing `None` values but we prefer RDFlib to always return either a `True` ore `False` for RDF boolean parsing into Python so we're going to reject this PR.

As a matter of RDFlib policy, we think, even though Turtle and other specs allow things like `"0"^^xsd:boolean`, people should use `true` & `false` as much as possible and errors from parsing are really user input errors so we're not going to try to hard to cater for every possibility.

Issue #771, opened 2017-10-05
ttl serialization incorrectly allows spaces in @prefix


tgbugs:
If a prefix has a space in it there is no warning and we end up with `@prefix This has a space. This may be an issue with namespace_manager.

abhishekvickyvyas:
May I contribute to solve this issue.?
In my opinion we need to add warning in  function   "def bind(self, prefix, namespace, override=True, replace=False):"  of "NamespaceManager" class of file  \rdflib\namespace.py . 

In case if prefix have spaces in it we can show warning and along with it if it is allowed we may replace space with underscore or may do necessary changes as per requirements.
Seeking for positive  reply from maintainers and contributors.

white-gecko:
You are welcome to contribute a pull request solving this issue. But I don't think replacing spaces with underscore is a good solution in this case.

tgbugs:
Spaces in a prefix should be an error that is detected and raised in NamespaceManger.bind.

abhishekvickyvyas:
I have submitted my solution of this issue in pull request #1070 . Please check it and let me if any change need to be done in it.

edmondchuc:
Closing as it seems to be addressed by https://github.com/RDFLib/rdflib/pull/1070 already.

PR #1070, opened 2020-05-26
issue_771_add_key_error_if_spaces

abhishekvickyvyas:
Hello, 
Here is my contribution for issue #771 . I raise a key error if prefix have spaces in it. The changes are done is  /rdflib/NamespaceManger.bind.  Please let me know if error message need to be change or if we need to raise only warning instead of error message.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -708,8 +708,9 @@ class NamespaceManager(object):
         # When documenting explain that override only applies in what cases
         if prefix is None:
             prefix = ""
-        if (len(prefix.split(" ")) > 1):
-            raise KeyError("\'" + prefix + "\' does not look like a valid prefix as it has spaces, change it into correct format")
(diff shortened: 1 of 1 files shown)


nicholascar:
Hi @abhishekvickyvyas thanks for addressing this - it's a real error in RDFlib - but your solutions is odd! Perhaps you're using a JavaScript code style in Python??

Rather than:

```python
 if (len(prefix.split(" ")) > 1):
            raise KeyError("\'" + prefix + "\' does not look like a valid prefix as it has spaces, change it into correct format")
```

How about:

```python
 if " " in prefix:
            raise KeyError("'{}' does not look like a valid prefix as it has spaces".format(prefix))
```

We do also need a test for any PR (RDFlib policy) so you'll need to write a small test function for this.

If you make this change, and add a test, we'll accept the PR. Thanks (maintainers)

PR #1072, opened 2020-05-26
issue 928 fix

### LINKED ISSUE #928: Can we inhance using Json to JSONLD to RDF Converters?
I was going through the code, can there be an enhanced feature where we just load in the data in form of JSON (without semantic's and specify general types of key's, their data types and required or not required) which goes to rdflib, auto-generates the JSON-LD based on converting techniques which are extracted from jsonld.org and later referred to RDF mapping.

This is because most of the web data out there is less semantic, Even though for converting a JSON data to a Graph database or Sparql/ Triplet based its hard.
Since we need to rewrite the schema or maybe if someone can give me a hint. AS I am looking in to use AWS NEPTUNE and it accepts RDFXML  or triplets graph formats.

Any Thoughts or concepts?

ayush609:
created rdfjson2rdf.py file in tools folder
along with json-ld support for serialization fron rdflib-jsonld

Closes #928


nicholascar:
@rob-metalinkage can you please review?

rob-metalinkage:
Going to do a meta-review before I look at it in any depth..

a solution claims to have been implemented - and inside individual commits there are comments like:
"created an example in example folder"

IMHO the issue itself needs to be updated with details of the nature of the solution and links to such examples.   If it was a bug then a test case would be expected - and I think for a new feature at the very least documentation needs review in addition to the code, which should have tests.

nicholascar:
Hi @ayush609 sorry but we can't accept this PR! You've added lots and lots of code (from json-ld.org?) but provided no tests and therefore there's no way for use to see what's going on.  You've not broken any existing tests but that's because your code is isolated - no testing. Also, we (maintainers) think that the underlying Issue, #928, is invalid anyway, so we don't want to have it addressed.

PR #1075, opened 2020-05-27
issue 981 fix

ayush609:
extensiondecider file to decide the correct extension based on the rdf format and if incorrect file is given create new file with correct extension


ashleysommer:
@ayush609 
There is already two PRs for a fix for #981
See: https://github.com/RDFLib/rdflib/pull/1046
and https://github.com/RDFLib/rdflib/pull/1064

nicholascar:
I've chose PR #1046 for this. It uses the in-built `guess_format()` function so is much shorter, and *includes tests*! Please, please include tests for all PRs.

PR #1076, opened 2020-05-27
Issue fix #981

### LINKED ISSUE #981: Auto-detect RDF type from file extension in parse()

raghav17083:
According to the issue, Graph().parse() should automatically identify the extension of the location.
I have added the functionality in graph.py by modifying the parse() method when the format is given "None".

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1059,11 +1059,10 @@ class Graph(Node):
         >>> os.remove(file_name)
 
         """
-        # if source is not None:
-        #     if(isinstance(source,str)):
-        #         location=source
+        if source is not None:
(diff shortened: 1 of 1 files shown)


nicholascar:
I've chose PR #1046 for this. It does use the in-built `guess_format()` function and *includes tests*! Please, please, include tests for all PRs.

PR #1077, opened 2020-05-27
Fix for issue #1071

### LINKED ISSUE #1071: Regular expression in sh:pattern not processed correctly
Discovered because I am using pySHACL. I originally reported this as a pySHACL bug but it appears to be a rdflib bug. See https://github.com/RDFLib/pySHACL/issues/47

I have the following:

```
graph_data = """
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix sch:  <http://schema.org/> .
@prefix sh:   <http://www.w3.org/ns/shacl#> .
@prefix ex:  <http://example.org/> .
@prefix xsd:  <http://www.w3.org/2001/XMLSchema#> .

ex:JohnDoe a ex:XXXX .
ex:JohnDoe ex:name "hello.txt" .
"""

shape_data = """
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix sch:  <http://schema.org/> .
@prefix sh:   <http://www.w3.org/ns/shacl#> .
@prefix ex:   <http://example.org/> .
@prefix xsd:  <http://www.w3.org/2001/XMLSchema#> .

ex:PersonShape
  a sh:NodeShape ;
  sh:targetClass ex:XXXX ;
  sh:property ex:PersonShape-name .

ex:PersonShape-name
  a sh:PropertyShape ;
  sh:path ex:name ;
  sh:minCount 1 ;
  sh:pattern  ".*.txt" .
"""
        
data  = rdflib.Graph().parse( data = graph_data, format = 'ttl' )
shape = rdflib.Graph().parse( data = shape_data, format = 'ttl' )

print( f"{data.serialize( format = 'ttl' ).decode( 'utf8' )}" )

report = validate( data, shacl_graph=shape, abort_on_error = False, meta_shacl = False, debug = True, advanced = True )

print( report[2] )
```

The ```sh:pattern``` should be ```".*\.txt"```, but when I do that, the following errors are generated:

``` 
... notation3.py", line 1591, in strconst  "bad escape")
... notation3.py", line 1615, in BadSyntax  raise BadSyntax(self._thisDoc, self.lines, argstr, i, msg)

  File "<string>", line unknown
BadSyntax
```

At least according to http://www.datypic.com/books/xquery/chapter19.html, I am using the escape correctly.

As commented on in the pySHACL bug...

> This problem is not of pySHACL but rdfLib library as rdfLib processes the shape file into rdf graph and during that processing, it is decoding the regular expression which is essentially the problem you are referring to

> Can you try this?: `sh:pattern ".*\\.txt" .`

Adding the extra escape does work.

This is probably going to be expected behavior and not an actual bug, but incase something can be done because an additional escape is required to resolve the problem should not be needed.

debarghadas411:
Modified "strconst" function to solve the issue.

Closes #1071

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -1598,9 +1598,9 @@ class SinkParser:
                         i,
                         "unterminated string literal (2)",
                     )
-                k = "abfrtvn\\\"'".find(ch)
+                k = "abfrtvn.\\\"'".find(ch)
                 if k >= 0:
-                    uch = "\a\b\f\r\t\v\n\\\"'"[k]
+                    uch = "\a\b\f\r\t\v\n.\\\"'"[k]
                     ustr = ustr + uch
                     j = j + 1
                 elif ch == "u":


ashleysommer:
Thanks @debarghadas411 
I don't believe your solution actually fixes the problem.
The problem is we want the string "*\.txt" to be parsed literally as "*\.txt", with slash intact and not escaped, so that it can be fed into the python regex module.

Your code fix removes the "\." and replaces it with ".", thus the resulting text is "*.txt", this is not what we want.

Note, we need to be very careful when changing any subtle behavior like this in the file `notation3.py`. Most of this code was written more than 15 years ago, and a lot of it written by Tim Berners-Lee. It is fragile parsing code, and a change like this can break things in unintended and unpredictable ways.

debarghadas411:
Hi @ashleysommer,
Sorry I could'nt understand the problem properly. Can you give an example so that I can work on it properly.

ashleysommer:
> Sorry I could'nt understand the problem properly. Can you give an example so that I can work on it properly.

Oh, github reformatted my response, removed some escape characters, made it unintelligible.
I've edited it, the problem should be clearer now.

ashleysommer:
Note, another problem I've just encountered.
This fix attempts to resolve only the problem with an escaped dot like `\.`.
However in regex we can have lots of different escapes, like `\w`, `\s`, `\(`, etc. There needs to be a much more robust solution to this problem.

debarghadas411:
Thanks @ashleysommer 
Trying my best to come up with a more robust solution.

nicholascar:
@debarghadas411 sorry but we - maintainers - can't merge this PR! What you're doing here is not following the requirements of the Turtle to allow a microformat, here Regex, within it. We have to err on the side of respecting the W3C/RDF specs and then escape any microformat as needed. You're just going to have to use double backslash to escape things.

Issue #841, opened 2018-08-07
graph.items() should raise an exception if the node isn't a list


vporton:
`graph.items()` should raise an exception if the node isn't a list.

As a temporary measure you can add

        if not graph.value(self, RDF.first):
            raise ...

vporton:
Ugh, I was wrong: This would also raise an exception on an empty list.

Not sure what to do.

ashleysommer:
I've been encountering this same issue, and I too don't know what the ideal solution is. I'm not sure if there is any way of checking the node before traversing it, to know if it is a valid list or not.

The way I am working at the moment, I am just treating a non-a-list and a empty-list as the same thing in my application logic.

CaptSolo:
@ashleysommer how does one represent an empty-list as an RDF collection?

vporton:
@CaptSolo 
```
L rdf:rest rdf:nil
```

In Turtle empty list can also be made as `()`.

CaptSolo:
@vporton thanks for the hint.

Turns out an empty list is represented by `rdf:nil`: https://www.w3.org/TR/rdf-schema/#ch_nil

`.list()` could be modified to distinguish between a non-list (= throw an exception) and an empty list. Would it be safe to do so = will that not break existing implementations that rely on this behavior?

jiggylepcha:
Hey I want to solve this PR . As `list` is a keyword in python why is it acting like a variable in graph.items()?

vporton:
> Hey I want to solve this PR . As `list` is a keyword in python why is it acting like a variable in graph.items()?

First, it is an issue not a PR. Which PR do you refer to? I don't understand (or don't remember) where `list` is used like a variable.

Second, `list` is not a keyword. It can be used as a variable if one wishes (however, this is not recommended, because the user may expect that it is the list type, so you could rename the variable).

jiggylepcha:
Hey I raised the exception when node in `graph.items()` is not a list but the Travis CI build is failing because it is raising this exception.

nicholascar:
Hi all, we - maintainers - have reviewed this Issue and PR #1078. We've decided not to make any change here as we think that existing behaviour - the `.items()` function stops `yield`ing if no `first` value is found, is correct. So, currently, a Collection will `yeild` `item()` values and empty Collections and non-Collections won't `yield` anything. 

If we allowed this change, we would have expensive checking for all sorts of things going on.

PR #1078, opened 2020-05-27
Fixes issue #841: graph.items() should raise an exception if the node isn't a list

jiggylepcha:
1. changes the variable name `list` to `rdf_collection` to avoid any clashes with the data type `list`
2. raises an error in `graph.items()` when node is not a list


nicholascar:
@jiggylepcha, thanks for this addition. You've supplied a negative tests - checking for input that raises an error - but please can you supply a positive test - where the code can be seen to correctly handle collections content? Such a test will show us that the code's really working. Unless you can point to another test that show this, thanks!

nicholascar:
Sorry but we - maintainers - won't merge this in, as per discussion in the Issue #841, we prefer existing behaviour.

Issue #724, opened 2017-03-23
Please make the parser more robust


cskyan:
I encountered some ParseError exceptions when I tried to parse the ntriples files. Some of them are quite easy to be fixed during the runtime such as empty lines, codec issues, etc. I hope that the parser could pre-process the files and deal with these problems or ignore the invalid records. At least, we need to know which lines in our data file have problems. Because we cannot make sure that the downloaded files strictly follow the standard format. If the package just raises the exception without correcting it, it will take more time to parse the whole file. Maybe the impact of neglected data could be accepted when we are processing a large data set.
In my case, I directly modify the code bellow.
[https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/parsers/ntriples.py#L154](url)
I insert a `continue` code here to let the program proceed. Otherwise, I cannot get the remaining data when I encounter a ParseError. I know that it is not a good way to skip this exception but it is the fastest way to continue my project. Hope that this suggestion would be accepted.

joernhees:
hmm, i'm against changing the default behavior (it's correct to raise an error if the input format is broken)... but maybe providing an `ignore_errors` flag via `parse()` would be a nice feature here...

cskyan:
Sure, `ignore_errors` would be a helpful feature in parsing the raw data. And the program should also put the ignored records into the log. Another expected feature is that it could incrementally parse the files. Say if I have some invalid records in my raw data, after the first parsing I can correct them and parse these revised records. It would save lots of time if this feature could also be implemented.

gromgull:
This is a good feature for the #283 clean-up, then there will be an error-callback.

mehak16163:
Hi I have attempted to solve this issue. Please go through my PR #1080 
Please suggest any changes to be made.

karish-grover:
I have been working on this on this issue for quite some time now. I have made incorporated a few parsing errors, but I was wondering if you could explain in detail, the exact errors that you want us to handle so that I can keep those in mind before sending out a PR.

PR #1080, opened 2020-05-27
Fix for Issue #724

mehak16163:
For this fix of issue #724 I have

1. Added _ignore_error_ method for each of the parsers included in rdflib/plugins/parsers. 
2. Added a couple of test cases to verify whether the new keyword works as expected. 
3. In graph.py where the parse method is called, I have made the changes to include the keyword. 
4. Added the logging functionality when the ignore_errors = True, so that we can identify the lines which were not correctly parsed.


mehak16163:
Hi @nicholascar, 
Can you please go through this pull request and suggest any changes to be done?

nicholascar:
Hi @mehak16163, thanks for this PR but there are some problems with it that mean we, the maintainers, have to reject it.

It doesn't make sense to ignore parsing errors in any RDF format other than n-triples as the triples are inter-connected so if you ignore a parsing error, you will not get a valid graph. in n-triples, you could ignore a single line parsing fail as each line is a subject, predicate, object, triple, but this is the only format you can do this for.

Also, you've tried to present a passing and a failing test for your changes but your passing test will always pass and doesn't notice what the state of the parsed RDF is, since you've ignored errors, and your failing test will always fail, since you are parsing invalid RDF. So your given tests don't actually tell us the code is working (the code can't work, except for n-triples, as I said).

Sorry but I have to close this PR.

Issue #837, opened 2018-07-02
`Graph.subjects` and `Graph.predicates` should return only unique values


mertzjames:
When calling the `Graph` method `subjects` or `predicates`, the generator should only return unique values.  For example:

```python
from rdflib import Graph
graph = Graph()
graph.parse("https://www.w3.org/People/Berners-Lee/card")
list(graph.subjects())
```

returns the following:
```
[rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card'),
 rdflib.term.BNode('N170b6415b06f45c98cb7a09398bbe818'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('http://www.w3.org/data#W3C'),
 rdflib.term.URIRef('http://www.ecs.soton.ac.uk/~dt2/dlstuff/www2006_data#panel-panelk01'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('http://wiki.ontoworld.org/index.php/_IRW2006'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
 rdflib.term.URIRef('https://www.w3.org/People/Berners-Lee/card#i'),
[...]

PR #1068, opened 2020-05-25
Fixes #837

nilay77:
- Fixed Graph.subjects, Graph.predicates, and Graph.objects to return only unique values. 
- Added test case for the above 3 methods

PR #1083, opened 2020-05-27
fixedIssue837

Kshitiz-Jain:
Fixed for #837 , made necessary changes in `subjects`, `predicates` and `objects` API.
For the memory issue, I introduced a variable `uniqueLimit`, this can be customised by the user in accordance with how much memory he can deal with. By default it is -1 i.e. no memory will be used and all values will be returned.
The case where uniqueLimit <  uniqueValues is also handled. 
Added tests for all 3 APIs in ./rdflib/test_graph.py. Following are the details for the testcases :
1. Where the unqiueLimit=0 or uniqueLimit=-1(default)
2. Where  uniqueLimit > uniqueValues
3. Where  uniqueLimit <  uniqueValues

Please suggest changes if any.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -615,20 +615,45 @@ class Graph(Node):
         self.remove((subject, predicate, None))
         self.add((subject, predicate, object_))
 
-    def subjects(self, predicate=None, object=None):
+    def subjects(self, predicate=None, object=None, uniqueLimit =-1):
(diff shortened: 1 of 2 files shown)


nilay77:
I have already submitted a PR for the same issue: #1068 . 
In your implementation, you could optimize your code to not check `uniqueLimit=-1` in every iteration as its value will remain the same.

PR #1084, opened 2020-05-28
Fixes #833

### LINKED ISSUE #833: Allow to return the number of triples added from Graph.addN

nilay77:
- Added a class variable `size` in Memory Store to track the number of triples in the store. This allows returning the number of triples in the store in O(1) instead of O(n)


rdflib/plugins/memory.py
```diff
@@ -49,6 +52,15 @@ def add(self, triple, context, quoted=False):
         # exits.
         subject, predicate, object = triple
         spo = self.__spo
+
+        try:
+            # Triple already in store
+            if(spo[subject][predicate][object]==1):
```
ashleysommer, line 58: Like the fast key tests below, you can replace this without an if statement like so:
```python
try:
    # Triple already in store
    _ = spo[subject][predicate][object]
except:
    # New triple being added
    self.__size += 1
```

rdflib/plugins/memory.py
```diff
@@ -49,6 +52,15 @@ def add(self, triple, context, quoted=False):
...
+                pass
```
ashleysommer, line 59: I think you could actually `return` here, rather than `pass`.
After spending the cpu time to check if the triple already exists in the index, we should at least take advantage of that answer and exit the method early if we know it exists.
(This assumes that if it exists in the `spo` index, then it also exists in the `pos` and the `osp` indexes)

rdflib/plugins/memory.py
ashleysommer: Don't need to check if the value `== 1`, because when the triple is removed, the whole key+value is removed from the corresponding dict.

nilay77: Yes, that makes sense. I'll change it.

rdflib/plugins/memory.py
nilay77: Your assumption makes sense, I did not think of this. I'll update the code. Thanks!

nilay77:
- Added functionality to return 1 on the addition of a new triple in Memory and IOMemory store, and 0 otherwise.
- Modified addN method from store.py to return the number of new triples/quads added. 
- Modified add method from Graph to return 1 on the addition of a new triple, and 0 otherwise.
- Modified addN method from Graph to return the number of new triples/quads added.
- Added test cases for add and addN method of Graph.

nilay77:
- Fixes #833: Added functionality to return the count of the number of triples added from Graph.addN

test/test_issue833.py
```diff
@@ -0,0 +1,61 @@
...
+        cheese = self.cheese
+
+        self.quads.append((tarek, likes, pizza, self.graph))
+        self.quads.append((tarek, likes, cheese, self.graph))
+        self.quads.append((tarek, likes, bob, self.graph))
+        self.quads.append((bob, likes, cheese, self.graph))
+        self.quads.append((bob, hates, pizza, self.graph))
+
+    def testAddN(self):
+        asserte = self.assertEqual
```
mwatts15, line 39: you don't really need to define `asserte`. It's defined in other places because it's used so many times in the same method.

test/test_issue833.py
```diff
@@ -0,0 +1,61 @@
...
+        tarek = self.tarek
+        bob = self.bob
+        likes = self.likes
+        hates = self.hates
+        pizza = self.pizza
+        cheese = self.cheese
```
mwatts15, line 45: Recommend to just define these on the module level -- less code and no reason not to.

(Don't ask me why it's done that way in other tests. IDK. It's kinda dumb.)

nilay77: Thanks for all the suggested changes. I have added them, please have a look.

mwatts15: Looks OK to me.

(I was actually recommending to define the `URIRef`s _outside_ of the class, like on line 7.)

rdflib/plugins/memory.py
```diff
@@ -38,15 +38,29 @@ def __init__(self, configuration=None, identifier=None):
...
+            1 in case of a new addition, 0 otherwise
         """
         # add dictionary entries for spo[s][p][p] = 1 and pos[p][o][s]
         # = 1, creating the nested dictionaries where they do not yet
         # exits.
         subject, predicate, object = triple
         spo = self.__spo
+        try:
+            # Triple already in store
+            _ = spo[subject][predicate][object]
```
tgbugs, line 58: What is the performance impact here?

rdflib/plugins/memory.py
```diff
@@ -38,15 +38,29 @@ def __init__(self, configuration=None, identifier=None):
         self.__namespace = {}
         self.__prefix = {}
 
+        # number of triples in the store
+        self.__size = 0
```
tgbugs, line 42: Are there cases where this variable could get out of sync with the actual number in the store?

rdflib/store.py
```diff
@@ -226,14 +226,18 @@ def addN(self, quads):
...
-                s,
-                p,
-                o,
-            )
-            self.add((s, p, o), c)
-
+            assert c is not None, \
+                "Context associated with %s %s %s is None!" % (s, p, o)
+            added = self.add((s, p, o), c)
+            if (isinstance(added,int)):
```
tgbugs, line 234: Stores that do not support count triples should not be forced to deal with this branch. Should probably dispatch on whether or not the store supports returning counts.

rdflib/plugins/memory.py
```diff
@@ -249,6 +262,7 @@ def namespaces(self):
             yield prefix, namespace
 
     def add(self, triple, context, quoted=False):
+        count_before = self.__len__()
```
tgbugs, line 265: Why not use `len(self)`?

rdflib/plugins/memory.py
nilay77: As far as I [know](https://stackoverflow.com/questions/1963507/time-complexity-of-accessing-a-python-dict), on an average, it should be O(1).

rdflib/plugins/memory.py
nilay77: I do not think so. The number of triples only changes when the add or remove function is called, and in both the cases this variable is updated accordingly.

rdflib/plugins/memory.py
nilay77: That should work too. I thought `self.__len__()` would be a direct call and `len(self)` would be an indirect call, thus I preferred the former. Should I change it to `len(self)`?

rdflib/store.py
nilay77: Stores that do not support count triples would not return anything from their `add `method, in which case `None` will be returned from this (i.e. `addN`) method. I could not understand how this would be an issue?

mwatts15: Since there hasn't been any movement on this: I was advocating for an explicit tag on stores to know if a count is returned, but at this point, we weren't using the return value of add/addN for anything, so until another feature comes along, this should be fine.

As far as "dealing with the branch", I assume you're thinking of performance based on your general comment, but this isn't a tight loop since we're making a function call, so there's barely in additional cost here.

nicholascar:
Hi @nilay77 sorry but this PR is now outdated due to changes in the default Store and the names of Stores. We've introduced a new, more efficient, Store and renamed the old Memory Store to SimpleMemory.

Secondly, while I don't disagree that this PR will change `len()` from O(n) to O(1), why is this particularly valuable? Even for the largest graphs, an O(n) count isn't going to kill any computers and, with this code, you've now added the additional variable and counting operations for every add/addN/remove call.

Finally, this store can't really be used for much - it's not the default - since it's not capable of some Store functions needed for many RDF formats.

Won't merge and closing, sorry!

Issue #543, opened 2015-11-12
NamespaceManager.bind(prefix, namespace, replace=True) leaves existing prefix in place  


Jeroen537:
I want to change an existing and bound namespace prefix. From the documentation of NamespaceManager I understand that "bind" with parameter "replace=True" should achieve this. However, while this does associate the namespace with the new prefix, the old prefix remains bound to the same namespace.
This can make it impossible to obtain a serialization with the new prefix. Is this intended behaviour?

Code example:

``` python
from rdflib import Graph, Namespace

g = Graph()

g.parse(data="""<?xml version="1.0" encoding="UTF-8"?>
<rdf:RDF
   xmlns:ns1="http://somewhere/else#"
   xmlns:ns2="http://nowhere/else#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
>
  <rdf:Description rdf:about="http://somewhere/else#y">
    <ns2:Q rdf:resource="http://somewhere/else#z"/>
  </rdf:Description>
  <rdf:Description rdf:about="http://somewhere/else#root">
    <ns1:P rdf:resource="http://somewhere/else#y"/>
    <ns1:P rdf:resource="http://somewhere/else#x"/>
  </rdf:Description>
</rdf:RDF>
""")

for ns in g.namespaces():
    print('%s (%s)' % (ns))

# prints:
#
# xml (http://www.w3.org/XML/1998/namespace)
# ns2 (http://nowhere/else#)
# ns1 (http://somewhere/else#)
# rdfs (http://www.w3.org/2000/01/rdf-schema#)
# xsd (http://www.w3.org/2001/XMLSchema#)
# rdf (http://www.w3.org/1999/02/22-rdf-syntax-ns#)
#
# (as expected)

print(g.serialize().decode())

# prints:
#
# <?xml version="1.0" encoding="UTF-8"?>
# <rdf:RDF
#    xmlns:ns1="http://somewhere/else#"
#    xmlns:ns2="http://nowhere/else#"
#    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
# >
#   <rdf:Description rdf:about="http://somewhere/else#y">
#     <ns2:Q rdf:resource="http://somewhere/else#z"/>
#   </rdf:Description>
#   <rdf:Description rdf:about="http://somewhere/else#root">
#     <ns1:P rdf:resource="http://somewhere/else#y"/>
[...]

PR #1086, opened 2020-05-28
Fix for Issue #543 

abhishekvickyvyas:
We added a unbind function which can be called in same way in which we call bind function. We pass the prefix that need to be unbind. The example can be given as:- 

from rdflib import Graph, Namespace
g = Graph()
g.namespace_manager.bind('abc', Namespace('http://xmlns.com/foaf/0.1'))
print('--- namespaces before unbinding ----')
for ns in g.namespaces():
    print('%s:%s' % (ns))
print
g.namespace_manager.unbind('abc')
print("\n\n")
print('--- namespaces after unbinding ----')
for ns in g.namespaces():
    print('%s:%s' % (ns))
print

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -694,6 +694,9 @@ class NamespaceManager(object):
 
             return self.__cache_strict[uri]
 
+    def unbind(self,prefix):
+        self.store.unbind(prefix)
+
     def bind(self, prefix, namespace, override=True, replace=False):
         """bind a given namespace to the prefix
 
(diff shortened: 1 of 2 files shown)


tgbugs:
This won't solve the problem. The namespace and prefix have to be removed from the [trie](https://github.com/RDFLib/rdflib/blob/b9446fa3f5d2797f31defd03dcc3f6e5af2a817c/rdflib/namespace.py#L537) as well I suspect. This PR will need tests to be added and by doing so I suspect you will spot the issue.

Issue #948, opened 2019-12-19
Prefixed names does not allow escaping (turtle 1.1)


jbulow:
(feature request)
It is not possible to use escaping when using namespace prefixes. Creation of URIRefs warns about the problem and the serializer fails to create turtle 1.1 output.

```
import rdflib as rdf
foo = rdf.Namespace('http://www.example.com/foo/')
g = rdf.Graph()
nsm = rdf.namespace.NamespaceManager(g)
nsm.bind('foo', foo)
g.add((foo['sub'], foo['pred'], foo['obj']))
g.add((foo['sub'], foo['pred'], foo['obj\/bar']))
g.add((foo['sub'], foo['pred'], foo['obj#boo']))
g.serialize(format='turtle').decode('utf-8').splitlines()
```

nicholascar:
In PyCharm, I get a PEP* (Python style guide) warning "PEP 8: invalid escape sequence '\/'". 

It all works without escaping the '/'.

Can you describe what the purpose of with request is since it seems you can just "do" it without escaping?

jbulow:
Sorry, I misunderstood my own problem. :-)

The problem is when serializing the graph. The prefix foo is only applied to the first one of the three triples: (I removed the unnecessary escape in 'obj\/bar'):



``` 
['@prefix foo: <http://www.example.com/foo/> .',
 '@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .',
 '@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .',
 '@prefix xml: <http://www.w3.org/XML/1998/namespace> .',
 '@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .',
 '',
 'foo:sub foo:pred foo:obj,',
 '        <http://www.example.com/foo/obj#boo>,',
 '        <http://www.example.com/foo/obj/bar> .',
 '']
```

I think it should be:

```
['@prefix foo: <http://www.example.com/foo/> .',
 '@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .',
 '@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .',
 '@prefix xml: <http://www.w3.org/XML/1998/namespace> .',
 '@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .',
 '',
 'foo:sub foo:pred foo:obj,',
 '        foo:obj\#boo>,',
[...]

PR #1087, opened 2020-05-28
Fix Issue #948 prefixed names does not allow escaping in turtle 1.1

megha19104:
We have provided the solution for issue #948 . In this we have allowed objects to have special characters with forward slash to make it a valid one.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -833,25 +833,43 @@ def is_ncname(name):
     return 0
 
 
-def split_uri(uri, split_start=SPLIT_START_CATEGORIES):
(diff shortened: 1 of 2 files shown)


tgbugs:
As the test results show, this approach will not work. The place in the code to start on this is probably in https://github.com/RDFLib/rdflib/blob/037ea51e5f4863a7f98ff59972fcd34d39a7ed97/rdflib/namespace.py#L836, but the issue related to curies vs qnames will also have to be addressed to really address #948.

gjhiggins:
Fixed in 9c1b55cb870f273800ce4869fdebb54b0b63081a

Issue #629, opened 2016-05-11
support for xsd:duration


wrobell:
The script below gives output

```
[(rdflib.term.URIRef('file:///home/wrobell/projects/uddo/#c1'), None),
 (rdflib.term.URIRef('file:///home/wrobell/projects/uddo/#c2'), None)]
```

It would be nice to get time difference as xsd:duration instead of nulls.

The script

```
from pprint import pprint
import rdflib
import io

data = """
@prefix : <#>.
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.

:C a rdfs:Class.

:start a rdf:Property;
    rdfs:domain :C; 
    rdfs:range xsd:dateTime.

:end a rdf:Property;
    rdfs:domain :C; 
    rdfs:range xsd:dateTime.

:c1 a :C;
    :start "2016-01-01 20:00:00"^^xsd:dateTime;
    :end "2016-01-02 20:01:00"^^xsd:dateTime.

:c2 a :C;
    :start "2016-01-01 20:05:00"^^xsd:dateTime;
    :end "2016-01-01 20:30:00"^^xsd:dateTime.
"""

graph = rdflib.Graph()

f = io.StringIO(data)
graph.parse(f, format='n3')

result = graph.query("""
SELECT ?c ?duration
WHERE {
    BIND(?end - ?start AS ?duration)
    ?c :start ?start;
        :end ?end.
}
""")

pprint(list(result))
```

gromgull:
SPARQL only defines - + etc for numeric types: https://www.w3.org/TR/sparql11-query/#OperatorMapping

Note that raw literals in RDFLib also do not support arithmetic: 

```python
In [110]: l1=rdflib.Literal(datetime.now())

In [111]: l2=rdflib.Literal(datetime.now())

In [112]: l1-l2
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-112-fe8dcc4af6cc> in <module>()
----> 1 l1-l2

TypeError: unsupported operand type(s) for -: 'Literal' and 'Literal'
```

(Although we support +, but in lexical space: 
```python
In [113]: l1+l2
[...]

PR #1089, opened 2020-05-28
Added DateTime arithmetic operations for Literals - Extension of #629

Akash-Sharma-1:
This PR adds some additional set of operations - dateTime / date / time subtraction to the Literals of RDFLib. This is an extension to the #629 issue of adding these operations to Literals in addition to the SPARQL queries.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -699,7 +699,7 @@ class Literal(Identifier):
         >>> a= Literal('2006-07-01T20:52:00',datatype=XSD.dateTime)
         >>> b= Literal('2006-11-01T12:50:00',datatype=XSD.dateTime)
         >>> (b-a)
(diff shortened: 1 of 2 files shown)


JervenBolleman:
@Akash-Sharma-1 thank you! it looks correct to me.

nicholascar:
Hi @Akash-Sharma-1, this looks great *but*, what happens for other Literals that are not dates? Your function returns `None` fro all non dates but since you've declared a `__sub__` function in term, we (maintainers) are concerned that this might override normal Python subtraction. 

Can you please add a couple of tests to test_literal to show that this function doesn't bread integer, float and decimal subtraction? If so, we will merge in.

FlorianLudwig:
@nicholascar I would suggest to throw a `TypeError` if subtraction is not supported for the types in question. That's what python does for other types, e.g. subtracting two strings.

gjhiggins:
Rebased and committed in https://github.com/RDFLib/rdflib/pull/1870

PR #1090, opened 2020-05-28
Fix Issue #790. Adding QueryBuilder.

### LINKED ISSUE #790: Feature: Query builder.
Hey,

I'd be very interested to know if people have already considered creating a query builder API?

I think it would help adoption if there was a convenient way to programatically build queries from within python in a LINQ kinda way.

Having such a DSL would not only allow for quicker generation of non prepared queries, as the entire parsing step would be omitted, it could also provide better tool support, as IDE's like pycharm would be capable of autocompleting and type checking queries.

I'd imagine something in the style of.

``` python


graph = rdflib.Graph()

query = SELECT(
                  v.s.AS(v.middle_aged_friend)
              ).WHERE(
                  (v.person, FOAF.knows, v.s),
                  (v.person, FOAF.age, v.age),
                  FILTER(l(35) < v.age)
                  FILTER(v.age < l(65))
             )

tim = r("http://www.w3.org/People/Berners-Lee/card#i")

result = g.query(q , initBindings={'person': tim})
```

I'd be more than happy to start working on this 😄.

Thoughts, suggestions, comments?

anubhavj99:
Fixes #790.
Developing a query-builder taking guidance from the examples in the issue.
#### Functions added:
1. SELECT
2. WHERE
4. OPTIONAL
5. UNION
6. ORDER BY, GROUP BY
7. AGGREGATES - ([functions](https://github.com/anubhavj99/rdflib/blob/cfee37467468ddbaafa870d2eca8b5fa2b32fffc/rdflib/query_builder.py#L128))
8. FILTER - ([functions](https://github.com/anubhavj99/rdflib/blob/cfee37467468ddbaafa870d2eca8b5fa2b32fffc/rdflib/query_builder.py#L162)), ([operations](https://github.com/anubhavj99/rdflib/blob/cfee37467468ddbaafa870d2eca8b5fa2b32fffc/rdflib/query_builder.py#L31))
9. LIMIT, OFFSET
10. FOR_GRAPH - setting graph for a query using
11. MOVE, ADD
12. INSERT, DELETE
13. NESTED QUERY ([example](https://github.com/anubhavj99/rdflib/blob/cfee37467468ddbaafa870d2eca8b5fa2b32fffc/test/test_issue790.py#L588))


magdasalatka:
What's the status? Can we get it merged?

nicholascar:
Hi @anubhavj99 , @Arshdeep25 & @magdasalatka apoligies for missing this PR, I don't know why we haven't addressed it yet. I'm going to put it to multiple review now.

nicholascar:
@joernhees I think you might like to review this since you were part of the discussion on Issue #790 that it claims to fix.

FlorianLudwig:
As commented in the issue (https://github.com/RDFLib/rdflib/issues/790#issuecomment-687613468) I would prefer to have separate QueryBuilders for different query types, e.g.

```
query = query_builder.Select().where(...)
```

instead of:

```
query = query_builder.QueryBuilder.Select().where(...)
```

As it allows for simplifications in the code and also avoids the question how to handle api usage like:

```
query = query_builder.QueryBuilder().SELECT(...).INSERT(...)
```

FlorianLudwig:
I can see the appeal of the current naming `SELECT().WHERE()` instead of the `pep8` conform `.where()` but I would like to ask: is that appeal big enough to not follow pep8 as requested in the dev-guide?

With every exception we get further away from doing more linting inside the CI

PR #1091, opened 2020-05-29
fix for issue #981

debarghadas411:
modified graph.parse method to add support for extracting file format from filename.
Also added a testcase to check the modifications made


nicholascar:
I've chosen PR #1046 to solve this Issue since it had much cleaner commit code and tests. You need to supply tests for all PRs!

Issue #936, opened 2019-10-02
HAVING clause with variable comparison not correctly evaluated


white-gecko:
On the graph

    <urn:a> <urn:p> 1 .
    <urn:b> <urn:p> 3 .
    <urn:c> <urn:q> 1 .

I would expect the query

    SELECT ?p ?o
    WHERE { ?s ?p ?o }
    GROUP BY ?p HAVING (?p != <urn:foo> )

to result in

?p | ?o
--|--
<urn:p> | one of 1 or 3
<urn:q> | 1

while it results in an empty result set.

This issue is demonstrated in pull request #935 .

https://github.com/RDFLib/rdflib/blob/1503f156037a0034ea32133ad9dc82aa508126a3/test/test_having.py#L8-L12

https://github.com/RDFLib/rdflib/blob/1503f156037a0034ea32133ad9dc82aa508126a3/test/test_having.py#L33-L39

white-gecko:
I think the problem is in the parser the `RelationExpression`: `RelationalExpression_{'expr': rdflib.term.Variable('p'), 'op': '!=', 'other': rdflib.term.URIRef('urn:foo'), '_vars': set()}`

Is comparing with `rdflib.term.Variable('p')` while it has to be `rdflib.term.Variable('__agg_1__')` because of the `GROUP BY ?p` as can be seen in the `Aggregate_Sample_`.

`
[...]

PR #1093, opened 2020-05-29
Fix issue #936 HAVING clause with variable comparison not correctly evaluated

devrb:
Modified the algebra.py file under sparql folder

Fixes #935, #936.


nicholascar:
@devrb can you please provide positive and negative tests to show how this PR solves the issue and doesn't introduce error? Thanks.

white-gecko:
Merging master and #935 apparently does not fail for the `GraphTestCase` as introduced by #935 .

nicholascar:
> No other test seem to fail.

So do you recommend merging this or not? If your tests pass without the change, we shouldn't merge it, unless someone can show that it accomplishes something desirable

gjhiggins:
I rewrote the test as pytest-compatible:
```py
import pytest
from rdflib import Graph

g = Graph()

data = """
<urn:a> <urn:p> 1 .
<urn:b> <urn:p> 3 .
<urn:c> <urn:q> 1 .
"""
g.parse(data=data, format="turtle")


def test_group_by():
    query = "SELECT ?p" "WHERE { ?s ?p ?o } " "GROUP BY ?p"
    qres = g.query(query)

    assert len(qres) == 2


def test_having_aggregate_eq_literal():
    query = (
        "SELECT ?p (avg(?o) as ?a) "
        "WHERE { ?s ?p ?o } "
        "GROUP BY ?p HAVING (avg(?o) = 2 )"
    )
    qres = g.query(query)

    assert len(qres) == 1


def test_having_primary_expression_var_neq_iri():
    query = "SELECT ?p " "WHERE { ?s ?p ?o } " "GROUP BY ?p HAVING (?p != <urn:foo> )"
    qres = g.query(query)

    assert len(qres) == 2
```

I was able to confirm that `test_having_primary_expression_var_neq_iri` fails without the change, passes with the change and the change cause no fails in the test suite:

```6332 passed, 3 skipped, 229 xfailed, 2 xpassed, 52 warnings```

However, the W3C SPARQL1.1 suite doesn't actually test HAVING clauses to any great degree of complexity:

```shell
$ grep -rni HAVING test/data/suites/w3c/sparql11/
test/data/suites/w3c/sparql11/aggregates/agg-avg-02.rq:7:HAVING (AVG(?o) <= 2.0)
test/data/suites/w3c/sparql11/aggregates/agg07.rq:6:HAVING ( COUNT(*) > 2 )
test/data/suites/w3c/sparql11/aggregates/agg03.rq:6:HAVING (COUNT(?O) > 2 )
test/data/suites/w3c/sparql11/aggregates/agg06.rq:5:HAVING (COUNT(*) > 0 )
```

aucampia (approved):
This looks okay to me, I think though that `rdflib.plugins.sparql.algebra.traverse` is a bit iffy, I think other places where it is used also has problems, but this change seems okay.

gjhiggins (approved):

PR #1094, opened 2020-05-29
Fix for Issue #543 (Namespace unbind required) including trie  and other modification 

abhishekvickyvyas:
For Issue #543  ,We added a unbind function which can be called in same way in which we call bind function. We pass the prefix that need to be unbind. In this PR we also take care of Trie and other cases in both rdflib/namespace.py  and  rdflib/plugins/memory.py .

The example can be given as:-

from rdflib import Graph, Namespace
g = Graph()
g.namespace_manager.bind('abc', Namespace('http://abc.com/xyz/0.1'))
print('--- namespaces before unbinding ----')
for ns in g.namespaces():
print('%s:%s' % (ns))
print
g.namespace_manager.unbind('abc')
print("\n\n")
print('--- namespaces after unbinding ----')
for ns in g.namespaces():
print('%s:%s' % (ns))
print

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -694,7 +694,10 @@ class NamespaceManager(object):
 
             return self.__cache_strict[uri]
 
-    def unbind(self,prefix):
+    def unbind(self, prefix):
+        for ns in self.namespaces():
+            if (ns[0] == prefix):
+                del self.__trie[str(ns[1])]
         self.store.unbind(prefix)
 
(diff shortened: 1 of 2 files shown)


nicholascar:
@abhishekvickyvyas would you be willing to update this PR since the in-memory graph has changes? Also, can you please provide some tests? Can't really accept a PR without tests. Thanks.

gjhiggins:
Subequently rebased and extended, close when https://github.com/RDFLib/rdflib/pull/1985 is merged

Issue #813, opened 2018-03-01
Error parsing simple DESCRIBE query


johnstonskj:
A simple query of the form `DESCRIBE <http://example.org/>` results in the following stacktrace.

```
  File "/Users/simonjo/anaconda/lib/python3.5/site-packages/rdflib/plugins/sparql/processor.py", line 75, in query
    query = translateQuery(parsetree, base, initNs)
  File "/Users/simonjo/anaconda/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py", line 766, in translateQuery
    P, PV = translate(q[1])
  File "/Users/simonjo/anaconda/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py", line 549, in translate
    M = translateGroupGraphPattern(q.where)
  File "/Users/simonjo/anaconda/lib/python3.5/site-packages/rdflib/plugins/sparql/algebra.py", line 269, in translateGroupGraphPattern
    if graphPattern.name == 'SubSelect':
AttributeError: 'NoneType' object has no attribute 'name'
```

The call to `translateGroupGraphPattern` in `algebra.py` states incorrectly that queries always include a where clause. This is not the case for DESCRIBE, see SPARQL 1.1 section 16.4.1 (SPARQL 1.0 section 10.4.1) "Explicit IRIs".

https://github.com/RDFLib/rdflib/blob/300db62c13779e85890ec6c3f0db2871557cdcb1/rdflib/plugins/sparql/algebra.py#L550-L551

luckyagarwal:
@johnstonskj  the following issue has now been fixed.

aucampia:
This was implemented in <https://github.com/RDFLib/rdflib/pull/2221> and should now be working. Please confirm if you can.

I will close this issue in some weeks if there is no further feedback.

aucampia:
Closing this as DESCRIBE was implemented.

PR #968, opened 2020-03-13
Concise Bounded Description

nicholascar:
This PR implements a Graph() function cbd() for 'Concise Bounded Description'. It extracts a subgraph from a source graph as per the rules in the W3C member submission [1].

This PR includes documentation for the method, tests and I want it mostly to be an example of a well-presented PR, not so much a rocket science improvement to the codebase!

It is also a precursor to implementing SPARQL's `DESCRIBE` queries though.

Please note this PR adds `black` to requirements.txt as I'd like `black` to be run on all code being submitted to rdfllib and, for rdflib 6.0.0, all existing code too.

[1] https://www.w3.org/Submission/CBD/

Issue #831, opened 2018-06-08
Rename prefix


Davidswinkels:
Love the Python library to make RDF! Keep it going.

One issue I don't know how to fix and I didn't find explicitly in the RDFLib 5.0.0 documentation is renaming a prefix. For example, I have been given a graph from an external source and don't like the way they did their prefixes. Therefore I want to rename the prefix at the start and in the graph. 
This is what I parsed from a link:

``` RDF file in n3 format (INCORRECT)
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix n80: <http://www.opengis.net/ont/geosparql#>.
n80:Geometry rdf:type <http://www.w3.org/2002/07/owl#Class>;
    rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql#>
.
n80:hasGeometry rdf:type <http://www.w3.org/2002/07/owl#ObjectProperty>;
    rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql#>
.
```
This is where I want to go:
``` RDF file in n3 format (CORRECT!)
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix geo: <http://www.opengis.net/ont/geosparql#>.
geo:Geometry rdf:type <http://www.w3.org/2002/07/owl#Class>;
    rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql#>
.
geo:hasGeometry rdf:type <http://www.w3.org/2002/07/owl#ObjectProperty>;
    rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql#>
.
```
This is how I tried to solve it, but which didn't work.

``` Python Script
from rdflib import Namespace, Graph

buildings_ontology = 'https://bag.basisregistraties.overheid.nl/query/model?format=ttl'
g = Graph()
g.parse(buildings_ontology, format = 'n3')
geo = Namespace('http://www.opengis.net/ont/geosparql#')
g.namespace_manager.bind('n80', geo, override = True, replace=True)
```
How would you rename the prefix at the start and in the graph?

Since I think there is no method or function that does that, I made an issue on GitHub.
For now this issue is fixed by downloading as text and replacing 'n80:' for 'geo:'.

[...]

PR #1096, opened 2020-05-29
Update Algebra.py add DESCRIBE

luckyagarwal:
fixes  RDFLib/rdflib#813

### DIFF
--- a/rdflib/plugins/sparql/algebra.py
+++ b/rdflib/plugins/sparql/algebra.py
@@ -778,14 +778,17 @@ def translateQuery(q, base=None, initNs=None):
     q[1] = traverse(
         q[1], visitPost=functools.partial(translatePName, prologue=prologue)
     )
-
-    P, PV = translate(q[1])
+    if(q[1].name!= "DescribeQuery"):
+        P, PV = translate(q[1])
     datasetClause = q[1].datasetClause
     if q[1].name == "ConstructQuery":
 
         template = triples(q[1].template) if q[1].template else None
 
         res = CompValue(q[1].name, p=P, template=template, datasetClause=datasetClause)
+        
+    elif(q[1].name== "DescribeQuery"):
+        res = CompValue(q[1].name, p=None, datasetClause=datasetClause, PV=q[1].var)
     else:
         res = CompValue(q[1].name, p=P, datasetClause=datasetClause, PV=PV)
 


luckyagarwal:
> #831 has been closed. What's this change meant to correct?

Sorry about that,  831 was accidentally mentioned for the corresponding pull request.

nicholascar:
Hi @luckyagarwal, thank you so much for addressing the issue of `DESCRIBE` queries not parsing. I'm really keen to see support for this added to RDFlib. 

I see that this is passing all tests but can you please supply a test demonstrating that this PR handles `DESCRIBE` queries? Without a test, I can't instantly see that this improvement does solve #813. I *think* it does but general policy is that PRs need to supply tests... thanks!

bgyori:
I bumped into this same issue and checked out this branch to solve it, but then I found that ultimately, `DESCRIBE` queries are not supported at all so I just got a different error saying so later.

nicholascar:
@bgyori are you keen to get `DESCRIBE` queries handled? They are the simplest of SPARQL queries so this shouldn't be too much of a problem I suppose - it just requires a bit of time for someone to read through *algebra.py* and actually line all the parts up since the functions of `DESCRIBE` can be fully implemented with other, existing, queries that already work or fairly simple `Graph` looping (looking for subject & object matches for the element `DESCRIBE`ed)

white-gecko:
As I understand it, it should not just be all statements matching the resource-to-describe in subject and object possition, but the CBD of the resource with is implemented in #968 .

nicholascar:
Yes, that’s why I implemented `cbd()`! I’ve just not found the time to revisit this to finish off the `DESCRIBE` parsing part!

white-gecko:
Futher, I think it should not be implemented in `algebra.py` but in the `evaluate.py`.
https://github.com/RDFLib/rdflib/blob/e56df6e189fba0e592391d21e3b394882032f186/rdflib/plugins/sparql/evaluate.py#L290-L291

nicholascar:
@luckyagarwal are you interested in updating this PR in like with comments from @white-gecko?

luckyagarwal:
Hey, I am sorry to say but not now because I am swamped with other works.

nicholascar:
@luckyagarwal are you keen to readdress this PR?

bpelakh:
@nicholascar I was looking at resuming this effort - is work underway for DESCRIBE support in 6.x? I was also wondering why your `cbd()` implementation reverted in a subsequent checkin.

PR #1097, opened 2020-05-29
Fixes for #896

sunnyjangra98:


nicholascar:
Hi @sunnyjangra98 this code looks interesting - especially maintaining the length count in the in-memory store implementation - but you've bundled a number of issues together and you've not included any tests, so we can't merge anything in. Also, what is `_add()`? We can't see it being used anywhere!

We will close this PR. If you'd like to re-add parts of this, please raise separate PRs for each issue addressed, and please include tests!

Issue #903, opened 2019-03-18
Support for new specifications in OWL 2


DhanshreeA:
Does rdflib currently support adding data ranges as per OWL2 specifications? For Example, can you create a custom datatype like: 
`g.add((URIRef(ns+"Answer"), RDF.type, RDFS.Datatype))
g.add((URIRef(ns+"Answer"), OWL.unionOf, (XSD.string, XSD.boolean, XSD.positiveInteger)))
g.add((URIRef(ns+"answer_value"), RDF.type, RDFS.Property))
g.add((URIRef(ns+"answer_value"), RDFS.domain, URIRef(ns+"Question")))
g.add((URIRef(ns+"answer_value"), RDFS.range, URIRef(ns+"Answer")))`

MuraliKrishna26:
Hey DhansheeA, I have worked on the issue and fixed it. Now you can add more than one data-type(or any property) at once as you mentioned in the code.
The below line will work fine 
       g.add((URIRef(ns+"Answer"), OWL.unionOf, (XSD.string, XSD.boolean, XSD.positiveInteger))

I have observed there may be a typo in the line : 
       g.add((URIRef(ns+"answer_value"), RDF.type, RDFS.Property))
Property term is defined in RDF class but not in RDFS.

ashleysommer:
@MuraliKrishna26 
> Property term is defined in RDF class but not in RDFS.

RDF.type is correct. Even though it is specified in the RDFS spec, the `type` predicate _intentionally_ uses `rdf:type`.
See https://en.wikipedia.org/wiki/RDF_Schema

DhanshreeA:
@ashleysommer I believe @MuraliKrishna26 is talking about `RDF.Property` and is correct in pointing out that Property is in RDF Class and not RDFS. :) I am a bit rusty on the spec itself, but at least within this Python implementation `RDFS.Property` does not seem to be the case.

ashleysommer:
> @MuraliKrishna26 is talking about RDF.Property and is correct in pointing out that Property is in RDF Class and not RDFS

Yes, you're right. I misread what @MuraliKrishna26 had written.

PR #1098, opened 2020-05-29
updated_graph.py

MuraliKrishna26:
fixes RDFLib/rdflib#903
Added code so that it can handle more than one predicate (taken in a tuple).

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -381,13 +381,36 @@ class Graph(Node):
         """
         self.__store.close(commit_pending_transaction=commit_pending_transaction)
 
+#     def add(self, triple):
+#         """Add a triple with self as context"""
+#         s, p, o = triple
+#         assert isinstance(s, Node), "Subject %s must be an rdflib term" % (s,)
+#         assert isinstance(p, Node), "Predicate %s must be an rdflib term" % (p,)
+#         assert isinstance(o, Node), "Object %s must be an rdflib term" % (o,)
+#         self.__store.add((s, p, o), self, quoted=False)
+
+    # In manageTuple function, each datatype from the tuple is added one by one
+    def manageTuples(self, triple):
+        s, p, tuplelist = triple
+        for i in tuplelist:
+            self.add((s, p, i))
+
     def add(self, triple):
         """Add a triple with self as context"""
         s, p, o = triple
-        assert isinstance(s, Node), "Subject %s must be an rdflib term" % (s,)
-        assert isinstance(p, Node), "Predicate %s must be an rdflib term" % (p,)
-        assert isinstance(o, Node), "Object %s must be an rdflib term" % (o,)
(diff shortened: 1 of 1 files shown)


nicholascar:
Hi @MuraliKrishna26, thanks for picking up this OWL2 issue. I *think* that your addition will do as expected, and it passes all tests, but please can you supply a test of your own demonstrating that your code now resolves that? You can perhaps use the example from Issue #903 that motivated this PR. It's RDFlib policy to have tests supplied for all PRs. Thanks!

ashleysommer:
@MuraliKrishna26 I'm not super familiar with OWL2 patterns, but I believe the RDF identifiers in the supplied object tuple are supposed to be turned into a rdf List mechanism, then the list added as the object to the triple, rather than multiple triples added with different object.

nicholascar:
The example Python from #903 is:

```python
g.add((URIRef(ns+"Answer"), RDF.type, RDFS.Datatype))
g.add((URIRef(ns+"Answer"), OWL.unionOf, (XSD.string, XSD.boolean, XSD.positiveInteger)))
g.add((URIRef(ns+"answer_value"), RDF.type, RDFS.Property))
g.add((URIRef(ns+"answer_value"), RDFS.domain, URIRef(ns+"Question")))
g.add((URIRef(ns+"answer_value"), RDFS.range, URIRef(ns+"Answer")))
```

This translates to RDF (Turtle) of:

```turtle
ns:Answer
    a rdfs:Dataype ;
    owl:unionOf (
        xsd:string
        xsd:boolean
        xsd:positiveInteger 
    ) ;
.

ns:answer_value
    a rdfs:Property ;
    rdfs:domain ns:Question ;
    rdfs:range ns:Answer ;
.
```

Yes, that syntax in the Turtle of `( xsd:string  xsd:boolean  xsd:positiveInteger )` represents "represents an `rdf:first`/`rdf:rest` list structure with the sequence of objects of the `rdf:first` statements being the order of the terms enclosed by `()`" as per the Turtle Spec [1] .

Here is `owl:unionOf` in Turtle from one of my projects describing the domain of a property:

```turtle
:vernacularName 
    a owl:DatatypeProperty ;
    rdfs:domain [ 
        a owl:Class ;
        owl:unionOf ( 
            :Host
            :SampleAnimal
        )
    ] ;
    rdfs:comment "The vernacular (common) name for the host or pest" ;
    skos:prefLabel "vernacular name" ;
.
```

So, back to my comment above: we need a test in this PR that shows this additional code doing the job it is claiming to do, not just a test for the error.

[1] https://www.w3.org/TR/turtle/#collections

nicholascar:
@MuraliKrishna26, no tests supplied so I'll close this in a week if I don't hear back from you here.

nicholascar:
Closing PR due to no response from @MuraliKrishna26

PR #1101, opened 2020-05-29
Fixed Blanknode issue 980

AyushGupta22:
We take the new map through which we were assigning new ids to each new blank nodes of different graphs. If two blank nodes came from the same graph then we assign the same id.


white-gecko:
This is superseded by #1107 and #1108 .

PR #1102, opened 2020-05-29
fix #876 Updated term.py to add xsd:normalizedString and xsd:token support for Literals

### LINKED ISSUE #876: xsd:normalizedString support is not implemented
`xsd:normalizedString` support is not implemented.

It should be turned into a regular Python string.

I would qualify it a low priority bug, but it is a bug nonetheless.

bhadoo-aditya:
The lexical space of xsd:normalizedString is unconstrained (any valid XML character may be used). Its value space is the set of strings after whitespace replacement—i.e., after any occurrence of #x9 (tab), #xA (linefeed), and #xD (carriage return) have been replaced by an occurrence of #x20 (space) without any whitespace collapsing. 
https://github.com/RDFLib/rdflib/issues/876


edmondchuc:
Merging this closes https://github.com/RDFLib/rdflib/issues/876.

edmondchuc:
Spec reference: https://www.w3.org/TR/xmlschema11-2/#normalizedString

`xsd:token` is a derived datatype of `xsd:normalizedString`. Therefore, this PR should also apply to `xsd:token` as well. See https://www.w3.org/TR/xmlschema11-2/#token

edmondchuc:
Note that while `xsd:normalizedString` has `whiteSpace = replace`, `xsd:token` instead has `whiteSpace = collapse`.

rdflib/term.py
```diff
@@ -1629,6 +1636,19 @@ def _castLexicalToPython(lexical, datatype):
...
 
+def _normalise_XSD_STRING(st):
+    """
+    Replaces \t, \n, \r (#x9 (tab), #xA (linefeed), and #xD (carriage return)) with space without any whitespace collapsing
+    """
+    if isinstance(st,str):
+        try:
+            return st.replace('\t',' ').replace('\n',' ').replace('\r',' ')
+        except:
+            return None
```
edmondchuc, line 1647: Not clear why this needs to be in an try except.

edmondchuc:
This PR performs normalisation on `xsd:normalizedString` and `xsd:token`. For `xsd:token`, this PR also strips trailing whitespace (left and right) and replaces contiguous whitespace with a single whitespace. 

Dockerfile base image version tag bumped from `python:3.6` to `python:3.7`. 

Improved some tests for `xsd:normalizedString` and added tests for `xsd:token`.

nicholascar (approved):

Issue #816, opened 2018-03-16
Issue parsing Turtle with info URI


scossu:
I am trying to read a TTL stream and parse into a graph by replacing a public URI (`http://ex.org/res/`) with an internal URI (`info:ns/`). I want to do this using the tools provided by RDFLib rather than writing tedious and error-prone pattern matching code. 

The following is a minimal example: 

~~~
>>> from rdflib import URIRef, Graph
INFO:rdflib:RDFLib Version: 4.2.2
>>> g1 = Graph().parse(data='<http://ex.org/res/1> a <http://auth.edu/ns#Resource> .', format='turtle')
>>> ttl1 = g1.serialize(format='turtle', base='http://ex.org/res')
>>> print(ttl1.decode())
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

</1> a <http://auth.edu/ns#Resource> .

>>> g2 = Graph().parse(data=ttl1, format='turtle', publicID=URIRef('info:ns'))
~~~

The last line triggers the following error:

~~~
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/graph.py", line 1043, in parse
    parser.parse(source, self, **args)
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/plugins/parsers/notation3.py", line 1870, in parse
    p.loadStream(source.getByteStream())
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/plugins/parsers/notation3.py", line 434, in loadStream
    return self.loadBuf(stream.read())    # Not ideal
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/plugins/parsers/notation3.py", line 440, in loadBuf
    self.feed(buf)
  File "/home/scossu/code/lakesuperior/virtualenv/lib/python3.5/site-packages/rdflib/plugins/parsers/notation3.py", line 466, in feed
    i = self.directiveOrStatement(s, j)
[...]

PR #1104, opened 2020-05-29
Issue parsing turtle #816

AyushGupta22:
This issue is not valid perfectly. As the public id part is coded in such a way. That you can merge the absolute uri of public id with the relative uri of the base of ttl. But here both are absolute so public id uri is overwritten by ttl base uri. If you want to make public id as base then follow the comment given in notation3.py at line 623 regarding the base. Though we could not able to exactly replicate the error as given but we found that it is not getting updated.

#816


nicholascar:
Hi @AyushGupta22, reading your comments here, it seems that you're saying there is no issue in the RDFlib tools regarding the Turtle parsing and your PR changes here then make one change to n-triples parsing. This seems to be irrelevant.

Can you clarify why you;re making an n-triples change? Also, you'll need a test for that n-triples change if you want to keep it but I think you need to explain it as well as provide a test.

nicholascar:
Closing this PR as it's remained unattended by the proponent for over a year

PR #1106, opened 2020-05-29
Expanding nested triples to triple and inserting them into graph

Vipank:


mwatts15:
What issue is this PR meant to resolve? If there's an existing issue in the tracker, please link to it in the description of this PR. Otherwise, for a feature change or significant bug fix that you've identified, it's better to write an issue first explaining what the desired change is. Then you can link to that issue in your PR so that maintainers and community members can review properly.

nicholascar:
Even if we had an Issue here being resolved, there are no tests included in this PR so I wouldn't be able to merge it. Must have unit tests!

nicholascar:
@Vipank thanks for this PR but you've not supplied any description or tests and not replied for a couple of months so we're going to close this without merging.

Also, if you wanted to add N-depth triples, it should be your custom implementation code that implements loops within loops etc and then just calls `g.add()` whenever all the S, P, O values are available, rather than expecting RDFlib to unpack your nested object.

Issue #980, opened 2020-03-19
Blank-nodes collisions


nleguillarme:
Hi.

If I understand correctly the graphs merging process explained [here](https://rdflib.readthedocs.io/en/stable/merging.html), the following piece of code should create a graph with two distinct blank nodes :

```python
from rdflib import Graph

graph1 = """
_:0 <http://purl.obolibrary.org/obo/RO_0002350> <http://www.gbif.org/species/0000001> .
"""
graph2 = """
_:0 <http://purl.obolibrary.org/obo/RO_0002350> <http://www.gbif.org/species/0000002> .
"""

g = Graph()
g.parse(data=graph1, format="nt")
g.parse(data=graph2, format="nt")

for triple in g:
    print(triple)
```

However, when executing the code, I get the following output :

`(rdflib.term.BNode('Ne3fd8261b37741fca22d502483d88964'), rdflib.term.URIRef('http://purl.obolibrary.org/obo/RO_0002350'), rdflib.term.URIRef('http://www.gbif.org/species/0000002'))
(rdflib.term.BNode('Ne3fd8261b37741fca22d502483d88964'), rdflib.term.URIRef('http://purl.obolibrary.org/obo/RO_0002350'), rdflib.term.URIRef('http://www.gbif.org/species/0000001'))`

Am I missing something ? (versions : rdflib 4.2.2, python 3.7.5)

white-gecko:
I think you understand it correctly. I think this is related to issue #892 . The rdflib uses the blank identifiers as they are.

Changing this behavior now would break some things and as we are in the feature freeze for 5.x I moved it to the 6.0.0 milestone.

Actually I have a use case where I need to parse multiple files within the same context of blank identifiers. When executing SPARQL queries I need to have individual contexts per query. Maybe it would be a good idea to introduce some blank context object which can be handed over to the parse method and the query method. We have to put this on the roadmap for 6.0.0.

nleguillarme:
Thank you for your reply. However I don't really understand... does that mean that there is no graph merging mechanism currently implemented in rdflib ? This would be in contradiction with what is said in the doc :

> In RDFLib, blank nodes are given unique IDs when parsing, so graph merging can be done by simply reading several files into the same graph

https://rdflib.readthedocs.io/en/stable/merging.html

sanyam19106:
but both the graph have common subject and predicate and object is different.

vikash18086:
we solve this issue as follows,
[...]

PR #1107, opened 2020-05-31
Allow distinct blank node contexts from one NTriples parser to the next (#980)

mwatts15:
Partial solution to #980. Allows to re-use a blank nodes across calls to parse while still allowing to keep the same context when desired.

Didn't touch derived parsers (NT, NQuads). Can do if the proposed interface is acceptable.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -102,7 +102,7 @@ version = re.sub("[0-9]+\\.[0-9]\\..*", "\1", release)
 exclude_trees = ["_build", "draft"]
 
 # The reST default role (used for this markup: `text`) to use for all documents.
-# default_role = None
+default_role = 'py:obj'
 
 # If true, '()' will be appended to :func: etc. cross-reference text.
 add_function_parentheses = True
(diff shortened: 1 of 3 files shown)


nicholascar:
Thanks @mwatts15, looks good! So, you're slowly chipping away at the long-standing RDFlib issue of BNodes? I do like the idea of passing a context to make BNs the same, if that's what a user wants. Or just to leave them defined per-graph by leaving out a context. Gives the users options.

mwatts15:
> So, you're slowly chipping away at the long-standing RDFlib issue of BNodes? 
Eh. I don't really like blank nodes, but I think people will want to use them in what I'm building, so I'd like to make it less painful for myself. Also, have seen a few PRs on RDFLib that I could review, but it's easier for me to just give an example of what might be a better approach.

Going to make another PR since this one doesn't completely address #980 (doesn't handle the `format='nt'` option to parse). Will also cover NQuads. Implementation will be different to support that.

PR #1108, opened 2020-06-01
BNode context dicts for NT and N-Quads parsers

mwatts15:
Address remainder #980. Also add similar behavior for N-Quads.


nicholascar:
Thanks @mwatts15. Funny how the test coverage *decreases* when all functions altered are already under test and new whole tests are added...

PR #1112, opened 2020-06-04
Added unit test for issue #977.

### LINKED ISSUE #977: Serialization shows only partial substitution of namespace.
I noticed something strange when serializing a graph to turtle, after binding a namespace prefix in code, the output shows a mix of prefix notation and full IRIs. I am using 4.2.2. While developing this part of the app, I noticed that in the output, the substitution comes and goes in an unpredictable manner.

In code:
```python
graph.namespace_manager.bind('mbse', URIRef('http://example.com/onto/v1/mbse#'))
graph.namespace_manager.bind('requirements', URIRef('http://example.com/api/v1/requirements/'))
```
Now, look at the "requirements" prefix:
```
@prefix co: <http://purl.org/co/> .
@prefix mbse: <http://example.com/onto/v1/mbse#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix requirements: <http://example.com/api/v1/requirements/> .

<http://example.com/api/v1/requirements/92c6e875-c721-4b2c-82ab-076de020d97e> a mbse:Requirement,
        owl:NamedIndidual ;
    co:member <http://example.com/api/v1/requirements/131fe437-5b49-4ab0-bee5-5ad5b0aa7a2f>,
        <http://example.com/api/v1/requirements/238f0f5d-28d6-4349-861d-d1d3a0cb3ce6>,
        <http://example.com/api/v1/requirements/42d6cd38-6858-48a8-88a6-db886eaf3bea>,
        <http://example.com/api/v1/requirements/4e33d5ee-5494-48a6-8aaa-d86240eecfbe>,
        <http://example.com/api/v1/requirements/5990ebbb-7023-4a4e-9bb6-34aff2bb6599>,
        <http://example.com/api/v1/requirements/708ef762-0c7e-4a11-84a3-4f84ddd28d42>,
        <http://example.com/api/v1/requirements/78ea66bc-ff55-4ffc-8c3d-a931c52e6811>,
        <http://example.com/api/v1/requirements/7e24a77f-cf1b-4285-b53b-e86230927a74>,
        <http://example.com/api/v1/requirements/8e410082-b85b-4184-8728-75f404415469>,
        requirements:a130d8ff-4482-4ffa-89c1-93da3e9fdf75,
        requirements:f384f799-4eac-44f2-b131-c1d47672fb55,
        requirements:f5252fbf-fe18-4959-a4a5-1a9b7b18dd0b,
        requirements:ff701a5d-dad6-41e8-8601-6651826245d4 .
```

Only a few are substituted: I can't make any sense of this behaviour.

dannmartens:
@nicholascar writes: If the tests here pass, it shows that `master` now addresses #977.

Closes #977

### DIFF
--- a/test/test_issue977.py
+++ b/test/test_issue977.py
@@ -11,15 +11,15 @@ class TestIssue977(unittest.TestCase):
         self.graph.bind('isbn', 'urn:isbn:')
         self.graph.bind('webn', 'http://w3c.org/example/isbn/')
         # Populate graph.
-        self.graph.add((URIRef('urn:isbn:1503280780'), RDFS.label, Literal('Moby Dick') ))
-        self.graph.add((URIRef('http://w3c.org/example/isbn/1503280780'), RDFS.label, Literal('Moby Dick') ))        
+        self.graph.add((URIRef('urn:isbn:1503280780'), RDFS.label, Literal('Moby Dick')))
+        self.graph.add((URIRef('http://w3c.org/example/isbn/1503280780'), RDFS.label, Literal('Moby Dick')))        
 
     def test_namespace_manager(self):
         assert 'isbn', 'urn:isbn:' in tuple(self.graph.namespaces())
         assert 'webn', 'http://w3c.org/example/isbn/' in tuple(self.graph.namespaces())
 
     def test_turtle_serialization(self):
-        serialization = self.graph.serialize(None, format='turtle').decode('utf-8')
+        serialization = self.graph.serialize(None, format='turtle')
         print(f'Test Issue 977, serialization output:\n---\n{serialization}---')
         # Test serialization.
         assert '@prefix webn:' in serialization, "Prefix webn not found in serialization!"
--- a/test/test_issue977.py
+++ b/test/test_issue977.py
@@ -1,0 +1,30 @@
+import unittest
+
+from rdflib import Graph, XSD, RDF, RDFS, URIRef, Literal
+
+
+class TestIssue977(unittest.TestCase):
+
+    def setUp(self):
+        self.graph = Graph()
+        # Bind prefixes.
+        self.graph.bind('isbn', 'urn:isbn:')
+        self.graph.bind('webn', 'http://w3c.org/example/isbn/')
+        # Populate graph.
+        self.graph.add((URIRef('urn:isbn:1503280780'), RDFS.label, Literal('Moby Dick') ))
+        self.graph.add((URIRef('http://w3c.org/example/isbn/1503280780'), RDFS.label, Literal('Moby Dick') ))        
+
+    def test_namespace_manager(self):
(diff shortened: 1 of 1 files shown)


nemesamade:
Hello, I can confirm that URN-based prefixes are still not being serialized into the .ttl format. Is this issue being looked into? I see that this has been labeled wontfix. Are there any plans to resolve this?

nicholascar:
@nemesamade this PR points to the issue but provides no indication as to how to solve it. Since the resulting turtle is valid - the ISBN URN is ﻿`<urn:isbn:1503280780>` - I suppose it's only a stylistic issue.

I'm not sure who is motivated to solve this one but if you would like to, we (maintainers) would be happy to review any PRs!

nemesamade:
@nicholascar Thanks for the status update. I might try and dig in and solve this one if I can find the time then, as it's a rather annoying inconvenience for a project I'm working on.

nicholascar:
@nemesamade great, well if you can get close to a solution (i.e. find all the bits) but need a hand, let me know!

nemesamade:
I've opened a PR (#1274) that addresses and resolves this issue by modifying namespace.py, along with adding some tests for URN namespace handling by compute_qname and turtle serialization. The change didn't break any existing tests and passes the new tests so it should be an easy fix:)

aucampia:
These test should maybe be committed with [@unittest.expectedFailure](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure)

nicholascar:
> These test should maybe be committed with @unittest.expectedFailure

We do need a standard way of adding tests for un-solved issues but can't afford for them to fail so this does look like a solution! I'll try and add this to the PR.

@aucampia if we start to add lots of tests annotated like this, is there a way to find all those tests in the future? I don't know of a unittest tool to find such tests. Would we just have to full text search all files in `test/` to find them?

aucampia:
> @aucampia if we start to add lots of tests annotated like this, is there a way to find all those tests in the future? I don't know of a unittest tool to find such tests. Would we just have to full text search all files in `test/` to find them?

If I run the unit tests directly with `unittest` they get reported as

```
$ poetry run python test/test_literal.py 
.........../home/iwana/d/github.com/iafork/rdflib/rdflib/term.py:1410: UserWarning: Parsing weird boolean, 'abcd' does not map to True or False
  category=UserWarning,
/home/iwana/d/github.com/iafork/rdflib/rdflib/term.py:1410: UserWarning: Parsing weird boolean, '10' does not map to True or False
  category=UserWarning,
.......x
Ran 19 tests in 0.008s

OK (expected failures=1)
```

Though sadly when running with nose they don't get reported in a special way.

I think there may be more reporting options but I would have to look into it more.

nicholascar:
Rerunning tests as this should pass due to fix mentioned above.

nicholascar:
Open/Close retest

PR #1121, opened 2020-06-22
drop support for python 2

t0b3:
2to3 the whole codebase and drop support for py2 to simplify maintaining in the future


rdflib/compat.py
```diff
@@ -2,9 +2,9 @@
 Utility functions and objects to ease Python 2/3 compatibility,
 and different versions of support libraries.
 """
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+
+
+
```
white-gecko, line 7: Could you please also remove the empty lines here and in other similar cases?

rdflib/extras/infixowl.py
```diff
@@ -278,47 +278,50 @@ def castToQName(x):
...
                 if childList:
-                    return (
-                        str(prefix)
-                        + u" THAT "
-                        + u" AND ".join(
-                            [str(manchesterSyntax(x, store)) for x in childList]
-                        )
-                    )
+                    return str(prefix) + ' THAT ' + ' AND '.join(
+                        [str(manchesterSyntax(x, store)) for x in childList])
```
white-gecko, line 287: Would it still be possible to split 2to3 changes and code formatting in two separate commits? This would improve the clarity for code reviews and possible future code blame.

rdflib/compat.py
t0b3: Sure. I am not happy with that neither. Will commit that soon...

rdflib/extras/infixowl.py
t0b3: I'll fix the worst of them, ok

t0b3:
@white-gecko ready for review now

rdflib/plugins/sparql/evaluate.py
```diff
@@ -357,10 +357,10 @@ def _buildQueryStringForServiceCall(ctx, match):
             service_query = "BASE <" + base + "> " + service_query
     sol = ctx.solution()
     if len(sol) > 0:
-        variables = " ".join(map(lambda v: v.n3(), sol))
-        variables_bound = " ".join(map(lambda v: ctx.get(v).n3(), sol))
+        variables = ' '.join([v.n3() for v in sol])
+        variables_bound = ' '.join([ctx.get(v).n3() for v in sol])
```
white-gecko, line 361: Is this fully equivalent? As of https://docs.python.org/3.8/library/functions.html#map map operated in parallel, does `[v.n3() for v in sol]` also run in parallel?

rdflib/plugins/stores/sparqlstore.py
```diff
@@ -425,27 +425,27 @@ class SPARQLUpdateStore(SPARQLStore):
...
         STRING_LITERAL2,
-        STRING_LITERAL_LONG1,
-        STRING_LITERAL_LONG2,
+        STRING_LITERAL_LONG1, 
+        STRING_LITERAL_LONG2
     )
-    IRIREF = u'<([^<>"{}|^`\\]\\\\\[\\x00-\\x20])*>'
-    COMMENT = u"#[^\\x0D\\x0A]*([\\x0D\\x0A]|\\Z)"
+    IRIREF = '<([^<>"{}|^`\\]\\\\[\\x00-\\x20])*>'
+    COMMENT = '#[^\\x0D\\x0A]*([\\x0D\\x0A]|\\Z)'
```
white-gecko, line 439: This is always a headache to find the correct amount of backslashes. Is it correct to remove one? Is this due to some 2to3 change? do we have some way to express this regex in a more clear representation?

rdflib/tools/csv2rdf.py
```diff
@@ -102,7 +100,7 @@ def toProperty(label):
     firstNm => firstNm
 
     """
-    label = re.sub("[^\w]", " ", label)
+    label = re.sub("[^\\w]", " ", label)
```
white-gecko, line 103: why do we need to escape this backslash now?

rdflib/plugins/sparql/evaluate.py
```diff
@@ -357,10 +357,10 @@ def _buildQueryStringForServiceCall(ctx, match):
...
         service_query = (
-            service_query + "VALUES (" + variables + ") {(" + variables_bound + ")}"
+            service_query + 'VALUES (' + variables + ') {(' + variables_bound + ')}'
```
white-gecko, line 363: in this case the single vs. double quote question applies

test/test_hex_binary.py
```diff
@@ -19,29 +19,29 @@ def _test_integer(self, i):
 
         l = Literal(hex_i, datatype=XSD.hexBinary)
         bin_i = l.toPython()
-        self.assertEquals(int(binascii.hexlify(bin_i), 16), i)
+        self.assertEqual(int(binascii.hexlify(bin_i), 16), i)
```
white-gecko, line 22: Just for other reviewers. Some background is here: https://stackoverflow.com/a/931011/414075 and here: https://docs.python.org/3/library/unittest.html#deprecated-aliases

test/test_n3.py
```diff
@@ -132,13 +132,13 @@ def testBaseSerialize(self):
         self.assertEqual(list(g), list(g2))
 
     def testIssue23(self):
-        input = """<http://example.com/article1> <http://example.com/title> "this word is in \u201Cquotes\u201D"."""
+        input = """<http://example.com/article1> <http://example.com/title> "this word is in \\u201Cquotes\\u201D"."""
```
white-gecko, line 135: Again some backslash confusion.

test/test_trig.py
```diff
@@ -31,7 +31,7 @@ def testRepeatTriples(self):
         self.assertEqual(len(g.get_context("urn:b")), 1)
 
         s = g.serialize(format="trig")
-        self.assertTrue("{}".encode("latin-1") not in s)  # no empty graphs!
+        self.assertTrue(b"{}" not in s)  # no empty graphs!
```
white-gecko, line 34: I understand, we don't want latin-1 but why do we use `b"{}"` in this case?

test/test_turtle_serialize.py
```diff
@@ -41,7 +41,7 @@ def testTurtleBoolList():
 
 
 def testUnicodeEscaping():
-    turtle_string = " <http://example.com/A> <http://example.com/B> <http://example.com/aaa\u00F3bbbb> . <http://example.com/A> <http://example.com/C> <http://example.com/zzz\U00100000zzz> . <http://example.com/A> <http://example.com/D> <http://example.com/aaa\u00f3bbb> ."
+    turtle_string = " <http://example.com/A> <http://example.com/B> <http://example.com/aaa\\u00F3bbbb> . <http://example.com/A> <http://example.com/C> <http://example.com/zzz\\U00100000zzz> . <http://example.com/A> <http://example.com/D> <http://example.com/aaa\\u00f3bbb> ."
```
white-gecko, line 44: backslash?

test/test_util.py
```diff
@@ -243,17 +243,17 @@ def test_util_from_n3_expectliteralwithtrailingbackslash(self):
 
     def test_util_from_n3_expectpartialidempotencewithn3(self):
         for n3 in (
-            "<http://ex.com/foo>",
+            '<http://ex.com/foo>',
             '"foo"@de',
-            u"<http://ex.com/漢字>",
-            u"<http://ex.com/a#あ>",
+            '<http://ex.com/漢字>',
+            '<http://ex.com/a#あ>',
```
white-gecko, line 249: double vs. single quote

test/test_util.py
```diff
@@ -243,17 +243,17 @@ def test_util_from_n3_expectliteralwithtrailingbackslash(self):
...
             # '"\\""', # exception as '\\"' --> '"' by orig parser as well
             '"""multi\n"line"\nstring"""@en',
         ):
             self.assertEqual(
-                util.from_n3(n3).n3(),
+                util.from_n3(n3).n3(),
```
white-gecko, line 254: trailing white space introduced

examples/film.py
```diff
@@ -130,17 +130,17 @@ def main(argv=None):
             rating = None
             while not rating or (rating > 5 or rating <= 0):
                 try:
-                    rating = int(input("Rating (on five): "))
+                    rating = int(eval(input("Rating (on five): ")))
```
white-gecko, line 133: shouldn't `int(input())` be enough? Why do we need `eval()?` Also in the other cases, isn't a string fine?

nicholascar:
Hi all, I'll get to this PR on the weekend. A little unsure about how to proceed with merging it until just before the 6.0.0 release since there are so many other PRs that need to go in too but I'll git it a go!

nicholascar (approved):
After Ashley Edmond & I have reviewed, we think this PR is correct in all cases, including those with comments from Natanael so we will merge.

Issue #1035, opened 2020-05-03
pip install rdflib doesn't install all dependencies


nicholascar:
When I install rdflib with pip:

```
pip install rdflib
```

I don't get `requests` installed, even though it's in the *requirements.txt* file. 

Is this correct behaviour?

tgbugs:
Yes it is correct. You have to install with `pip install rdflib[sparql]` because requests is an extra requirement. There is a bug though which is that requests probably needs to be added to tests_require.

nicholascar:
I think I just never noticed before because I’ve always installed `requests` with `rdflib` because I tend to be doing HTTP actions in my APIs and always have needed it anyway.

We can re-examine all dependencies as we remove the Py 2/3 requirements like `six` now that Py 2 is gone.

tgbugs:
Related: https://github.com/RDFLib/rdflib/pull/969

white-gecko:
Yes I had added it for sparql in #969 already, but it was also missing for the tests which should be fixed in #1036.

PR #969, opened 2020-03-14
Add requirement for requests to setup.py

white-gecko:
The requests package is required by the sparql plugin at several places.

The current issue is reproducible with:

In a new virtualenv

```
$ pip install git+https://github.com/RDFLib/rdflib@master
$ python3
>>> import rdflib
>>> graph = rdflib.Graph()
>>> graph.query("select * {?s ?p ?o}")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/natanael/.virtualenvs/rdflib-try/lib/python3.7/site-packages/rdflib/graph.py", line 1073, in query
    result = plugin.get(result, query.Result)
  File "/home/natanael/.virtualenvs/rdflib-try/lib/python3.7/site-packages/rdflib/plugin.py", line 107, in get
    return p.getClass()
  File "/home/me/.virtualenvs/rdflib-try/lib/python3.7/site-packages/rdflib/plugin.py", line 69, in getClass
    module = __import__(self.module_path, globals(), locals(), [""])
  File "/home/me/.virtualenvs/rdflib-try/lib/python3.7/site-packages/rdflib/plugins/sparql/__init__.py", line 37, in <module>
    from .processor import prepareQuery, processUpdate
  File "/home/me/.virtualenvs/rdflib-try/lib/python3.7/site-packages/rdflib/plugins/sparql/processor.py", line 18, in <module>
    from rdflib.plugins.sparql.evaluate import evalQuery
  File "/home/me/.virtualenvs/rdflib-try/lib/python3.7/site-packages/rdflib/plugins/sparql/evaluate.py", line 20, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
>>>
```

PR #1123, opened 2020-06-23
update setup.py to include requests needed by Graph.parse()

### LINKED ISSUE #1122: requests not included in setup.py
although requests is defined in  requirements.txt it is not in setup.py and hence not picked up by pip install

rob-metalinkage:
Fixes #1122

## Proposed Changes
updates setup.py

suggest reviewing requirements.txt and setup.py and syncing.
  -
  -
  -

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import re
 from setuptools import setup, find_packages
 
 kwargs = {}
-kwargs["install_requires"] = ["isodate", "pyparsing"]
+kwargs["install_requires"] = ["isodate", "pyparsing", "requests"]
 kwargs["tests_require"] = [
     "html5lib",
     "networkx",


tgbugs:
This should not be merged per https://github.com/RDFLib/rdflib/pull/969 and https://github.com/RDFLib/rdflib/issues/1035. The correct solution is to install via `pip install rdflib[sparql]`. requests is not a core dependency.

rob-metalinkage:
i piut this  comment on #1122 

See the direction (pip install rdflib[sparql]) ,  fair enough...

but a few things seem off -

there seems to be inconsistency between requirements.txt and setup.py (and not sure if this is a good thing to have similar info in both places anyway?)
Parsing a graph from a HTTP request has nothing to do with SPARQL - so this isnt very intuitive a solution
probably the parse module should report a sensible error with the correct import suggestion if you ask it for a HTTP source and it doesnt have the install available
the documentation for parse is quite explicit HTTP is supported without any mention of the requirement.
I think all these things need to be addressed for the proposed solution to be acceptible.

tgbugs:
Unless something changed very recently in the internals (which my review of the commit logs it did not), rdflib doesn't depend on requests for fetching http resources.

ashleysommer:
the `requests` library is only used in two locations in `rdflib`.

Here: https://github.com/RDFLib/rdflib/blob/e4fe0fdbd4de7e1183418f302315b51a14602e03/rdflib/plugins/sparql/evaluate.py#L316-L318

And here: https://github.com/RDFLib/rdflib/blob/e4fe0fdbd4de7e1183418f302315b51a14602e03/rdflib/plugins/stores/sparqlconnector.py#L60

The only reason you would need `requests` when doing `Graph.parse()` is if you have configured your Graph to use the sparqlconnector backend.

The correct solution here would be to take the requirement out of `requirements.txt` I think it was put there accidentally.

I've been experimenting with a way of eliminating the need for `requests` in plugins/sparql/evaluate.py (by using the built-in python3 urllib.requests library), but it looks like it cannot be avoided in plugins/stores/sparqlconnector.py due to the use of a persistent Session per thread.

rob-metalinkage:
So the HTTP request from parse works fine without requests? If so good and the doco is fine.

white-gecko:
Ok, so we can close this I guess.

PR #1125, opened 2020-06-26
quick fix for Link header handling for json-ld

datadavev:
Fixes https://github.com/RDFLib/rdflib-jsonld/issues/85

## Proposed Changes

  - Add support for handling Link header responses

PR is a minimal implementation to get rdflib-jsonld to work with schema.org remote context reference.

TBD:
- Can multiple Link headers returned (e.g. to reference different content types)?
- Do other RDF formats leverage the Link header in HTTP response is specific to JSON-LD?

### DIFF
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -114,6 +114,20 @@ class URLInputSource(InputSource):
 
         req = Request(system_id, None, myheaders)
         file = urlopen(req)
+        #Fix for https://github.com/RDFLib/rdflib-jsonld/issues/85
+        #TODO: Check if Link header is supported for other formats
+        if format == "json-ld":
+            # Check for a Link header in response, rfc8288
+            _link = file.getheader("Link", None)
+            if not _link is None:
+                #TODO: Scan for content type in potentially multiple Link headers
+                # Grab the possibly relative URL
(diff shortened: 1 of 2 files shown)


ashleysommer:
Hi @datadavev 
Is the cgi import and empty <B> fix related to the json-ld link header fix? Looks like you might've pushed that commit into the wrong branch.

datadavev:
> Hi @datadavev
> Is the cgi import and empty <B> fix related to the json-ld link header fix? Looks like you might've pushed that commit into the wrong branch.

Ah, indeed - I'll adjust. Thanks

nicholascar:
Hi @datadavev, there are two issues with this PR:

1. JSON-LD isn't yet in RDFlib core, although we hope it will be soon, so that this PR won't apply for many cases. If we were to pass it, it would need to do a better job of knowing that JSON-LD was enable, not just looking for the format "json-ld".
2. `Link` headers in HTTP can be used for all kinds of things, e.g. description of related resources, not just redirects, so your Link header parsing function needs a lot more information to only get redirects.

Sorry but I'm going to have to close this. Please feel free to re-issue it after we've included JSON-LD in RDFlib core (Coming soon!) and after having a bit more of a go at parsing Link headers a bit more carefully.

PR #1134, opened 2020-07-27
Add a  __dir__ attribute to a closed namespace

tonyfast:
## Proposed Changes

  - This change is introduced for improved affordance when working with rdflib in interactive settings like IPython. 
  - The __dir__ attribute gives attribute completion.
  - Use the mangled attribute from the closed namespace object to give better interactive completion.

Closes #918

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -214,7 +214,9 @@ class ClosedNamespace(object):
     def __repr__(self):
         return "rdf.namespace.ClosedNamespace(%r)" % str(self.uri)
 
-
+    def __dir__(self):
+        return list(self._ClosedNamespace__uris)
+    
 class _RDFNamespace(ClosedNamespace):
     """
     Closed namespace for RDF terms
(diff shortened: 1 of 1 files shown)


nicholascar:
Hi @tonyfast, thanks for this PR. I don;t know much about iPython myself but fully support people using it.

Can you please just put a comment in about how this approach is better than this approach: PR #918? Code is very similar. If we merge your code, we will close #918 (or vice versa!).

tonyfast:
Oh wow. yea that pr is basically the same thing. In #918, I personally can't get at the `__uris` property the author is using there. Since the `getattr` and `getitem` work the same on the closed namespace the key completions (which returns getitem completion) should be equivalent to the getattr/`__dir__` completion.

I made this PR on 5.0 so maybe it is update to date with the api.

nicholascar:
@tonyfast we'll merge this one in and close the older #918. We'll leave it to Jupyter users to fag any issues (I've only just started making RDFlib Jupyter training material myself!).

PR #1140, opened 2020-08-13
improved Graph().parse()

nicholascar:
## Proposed Changes

  - raises a `ParseError`, not a mysterious XML SAX error, when failing to parse in `graph().parse()`
  - default parse format is now Turtle, not RDF/XML
  - understands format "xml", i.e. `file.xml`, is RDF/XML
  - includes tests for passing & failing to parse
  - updates all tests that relied on `parse()` default being XML to use `format="xml"` - they all now work
  - removes pointless `Graph().load()` method since it just called `parse()` with bad defaults


ashleysommer (approved):

ashleysommer:
Will add back in the `graph.load()` function but it will create a deprecation warning.

PR #1151, opened 2020-08-27
Deprecate some more Graph API surface

ashleysommer:
Deprecate some more Graph API surface that will be removed in RDFLib 6.0.0

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -130,7 +130,7 @@ via triple pattern:
     >>> g.add((statementId, RDF.type, RDF.Statement))
     >>> g.add((statementId, RDF.subject,
     ...     URIRef("http://rdflib.net/store/ConjunctiveGraph")))
-    >>> g.add((statementId, RDF.predicate, RDFS.label))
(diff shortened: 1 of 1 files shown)


nicholascar (approved):

ashleysommer:
Can't merge this yet. A lot of tests are failing.
Turns out some parts of rdflib (like a lot of tests) do `from rdflib.graph import RDFS`
With RDFS removed as an import here, those locations now fail.
I'll update this soon.

nicholascar:
> Turns out some parts of rdflib (like a lot of tests) do from r`dflib.graph import RDFS` 

Yes, I've seen this and occasionally have done that myself, by mistake. It would be good to update everything so that `RDFS`, along with all namespaces are *always* imported from the same place, `rdflib.namespace`, not `rdflib.graph`. Good to finally address this small niggle

ashleysommer:
Ok, turns out I was wrong. It wasn't used in the tests, but in the inline doctests in graph.py, I didn't know Travis tests those too.
I've updated those so should pass tests now.

ashleysommer:
All tests pass, good to merge 👍

PR #1156, opened 2020-08-31
remove import side-effects

### LINKED ISSUE #1135: Importing rdflib outputs the version
When importing rdflib, an identifier and the version number are output. There is no need to output on the successful loading of an import. It should remain silent when loaded. For background, see the "[Rule of Silence](http://www.catb.org/~esr/writings/taoup/html/ch01s06.html#id2878450)."

```
Python 3.6.5 (default, Oct  8 2018, 17:39:55)
[GCC 7.3.0 20180125 (Cray Inc.)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rdflib
RDFLib Version: 5.0.0

```

FlorianLudwig:
importing rdflib has side-effects, if you do so in an interactive terminal the logging module gets configured (or reconfigured if you already have done so...). This is quite unexpected.

Fixes #1135

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -83,39 +83,6 @@ __all__ = [
     "util",
 ]
 
-import sys
-
-import logging
-
-logger = logging.getLogger(__name__)
-_interactive_mode = False
-try:
-    import __main__
-
-    if (
-        not hasattr(__main__, "__file__")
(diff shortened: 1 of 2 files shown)


ashleysommer:
Looks good, thanks.

PR #1162, opened 2020-09-11
Small improvement to serialize docs

FlorianLudwig:
Docs state that serialize returns a `string` but returns `bytes`.


ashleysommer:
Thanks. This is a hangover from the python2 days when it did return `str`, back when `str` was byte-encoded.
Ever since RDFLib 4.0, serialize() returns `bytes`, this was never updated in the documentation.

Note, we are planning a change to `serialize()` in RDFLib 6.0, so it _does_ return a unicode string.
But I will merge this for now anyway.

PR #1167, opened 2020-09-17
Move to using graph.parse() rather than deprecated graph.load()

ashleysommer:
Replaced all usages of graph.load() in rdflib, it is deprected, move to use graph.parse().


nicholascar (approved):
Good, as discussed!

PR #1172, opened 2020-09-28
Fix serialize with multiple disks on windows

### LINKED ISSUE #1170: serialize fails on windows machine with multiple drives
In the serialize method of graph.py, you urlparse the given  destination:https://github.com/RDFLib/rdflib/blob/e56df6e189fba0e592391d21e3b394882032f186/rdflib/graph.py#L985

On windows machines this strips the leading drive letter (e.g. `C:\`).
On some machines the shutil.move call seems to work nevertheless, but on some it doesn't:
https://github.com/RDFLib/rdflib/blob/e56df6e189fba0e592391d21e3b394882032f186/rdflib/graph.py#L996

```
  File "...", line 133, in store
    self._graph.serialize(destination=path_graph, format="xml")
  File "...\lib\site-packages\rdflib\graph.py", line 982, in serialize
    shutil.move(name, path)
  File "...\lib\shutil.py", line 802, in move
    copy_function(src, real_dst)
  File "...\lib\shutil.py", line 432, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "...\Python38\lib\shutil.py", line 261, in copyfile
    with open(src, "rb") as fsrc, open(dst, "wb") as fdst:
FileNotFoundError: [Errno 2] No such file or directory: "\\Users\\urba\\.osp_ontologies\\graph.xml"
```

Why do you strip the drive and why does it work sometimes and sometimes not?

urbanmatthias:
Fixes #1170 

## Proposed Changes
I replaced the `path` variable in the `shutil.move` call with `location`.
The former is the output of an `urlencode` call, which strips the drive letter on windows machines.
If the user has multiple drives, `shutil.move` will fail if the drive letter is not there.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -992,10 +992,11 @@ class Graph(Node):
             stream = os.fdopen(fd, "wb")
             serializer.serialize(stream, base=base, encoding=encoding, **args)
             stream.close()
+            dest = path if scheme == "file" else location
             if hasattr(shutil, "move"):
-                shutil.move(name, location)
+                shutil.move(name, dest)
             else:
-                shutil.copy(name, location)
+                shutil.copy(name, dest)
                 os.remove(name)
 
     def parse(
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -993,9 +993,9 @@ class Graph(Node):
             serializer.serialize(stream, base=base, encoding=encoding, **args)
             stream.close()
             if hasattr(shutil, "move"):
-                shutil.move(name, path)
+                shutil.move(name, location)
             else:
-                shutil.copy(name, path)
+                shutil.copy(name, location)
                 os.remove(name)
 
     def parse(


ashleysommer:
Hi @urbanmatthias 
I don't think this fix is quite complete, because it will now fail for UNC file locations, which used to work, like:
`file:///tmp/outfile.ttl`
In that kind of string, `scheme` is "file" and `path` is `/tmp/outfile.ttl` so `path` would be the correct variable to use.

Also, just a note, this works with Windows paths too:
`file://c:\\tmp\\outfile.ttl`
In this example, `scheme` is "file" and `path` is `c:\\tmp\\outfile.ttl` so again `path` is the correct variable to use.

A good middleground for this fix might be to test `if scheme == "file"` then use `path` otherwise, use `location`.

rdflib/graph.py
```diff
@@ -993,9 +993,9 @@ def serialize(
             serializer.serialize(stream, base=base, encoding=encoding, **args)
             stream.close()
             if hasattr(shutil, "move"):
```
ashleysommer, line 995:
```suggestion
            dest = path if scheme == "file" else location
            if hasattr(shutil, "move"):
```

rdflib/graph.py
```diff
@@ -993,9 +993,9 @@ def serialize(
...
-                shutil.move(name, path)
+                shutil.move(name, location)
```
ashleysommer, line 996:
```suggestion
                shutil.move(name, dest)
```

rdflib/graph.py
```diff
@@ -993,9 +993,9 @@ def serialize(
...
             else:
-                shutil.copy(name, path)
+                shutil.copy(name, location)
```
ashleysommer, line 998:
```suggestion
                shutil.copy(name, dest)
```

ashleysommer (changes requested):

urbanmatthias:
Yes you are right, thanks. I applied your suggested changes.

ashleysommer:
Need review from one more Maintainer before we can merge.

nicholascar (approved):
Looks OK to me. I folowed back through graph.py to check all the URL parsing etc. and I think this will work fine.

ashleysommer:
@nicholascar 
As far as I can tell, this is only used in the `graph.serialize()` method and only when you've set a "destination" location. So unlikely to affect anything else in graph.py or URL parsing.
I'll merge.

ashleysommer (approved):

PR #1175, opened 2020-10-03
Remove requests

### LINKED ISSUE #1122: requests not included in setup.py

nicholascar:
Removes requests from RDFlib requirements entirely to 

Tidies requirements.txt & requirements-dev.txt

Closes #1122

### DIFF
--- a/rdflib/plugins/stores/sparqlconnector.py
+++ b/rdflib/plugins/stores/sparqlconnector.py
@@ -1,15 +1,13 @@
 import logging
-import threading
 from urllib.request import urlopen, Request
 from urllib.parse import urlencode
 from urllib.error import HTTPError, URLError
 import base64
 
-import os
-
 from io import BytesIO
 
 from rdflib.query import Result
+from rdflib import BNode
 
 log = logging.getLogger(__name__)
 
@@ -72,12 +70,12 @@ class SPARQLConnector(object):
         self._method = method
 
     def query(self, query, default_graph: str = None, named_graph: str = None):
-
         if not self.query_endpoint:
             raise SPARQLConnectorException("Query endpoint not set!")
 
         params = {"query": query}
-        if default_graph:
+        # this test ensures we don't have a useless (BNode) default graph URI, which calls to Graph().query() will add
+        if default_graph is not None and type(default_graph) != BNode:
             params["default-graph-uri"] = default_graph
 
         headers = {"Accept": _response_mime_types[self.returnFormat]}
@@ -93,7 +91,10 @@ class SPARQLConnector(object):
         if self.method == "GET":
             args["params"].update(params)
             qsa = "?" + urlencode(args["params"])
-            res = urlopen(Request(self.query_endpoint + qsa, headers=args["headers"]))
+            try:
+                res = urlopen(Request(self.query_endpoint + qsa, headers=args["headers"]))
+            except Exception as e:
+                raise ValueError("You did something wrong formulating either the URI or your SPARQL query")
         elif self.method == "POST":
             args["headers"].update({"Content-Type": "application/sparql-query"})
             try:
@@ -102,7 +103,6 @@ class SPARQLConnector(object):
                 return e.code, str(e), None
         else:
             raise SPARQLConnectorException("Unknown method %s" % self.method)
-
         return Result.parse(
             BytesIO(res.read()), content_type=res.headers["Content-Type"].split(";")[0]
         )
--- a/rdflib/plugins/stores/sparqlconnector.py
+++ b/rdflib/plugins/stores/sparqlconnector.py
@@ -1,6 +1,9 @@
 import logging
 import threading
-import requests
+from urllib.request import urlopen, Request
+from urllib.parse import urlencode
+from urllib.error import HTTPError, URLError
+import base64
 
 import os
 
@@ -26,7 +29,6 @@ _response_mime_types = {
 
 
 class SPARQLConnector(object):
-
     """
     this class deals with nitty gritty details of talking to a SPARQL server
     """
@@ -37,10 +39,13 @@ class SPARQLConnector(object):
         update_endpoint=None,
         returnFormat="xml",
         method="GET",
+        auth=None,
         **kwargs
     ):
         """
-        Any additional keyword arguments will be passed to requests, and can be used to setup timesouts, basic auth, etc.
+        auth, if present, must be a tuple of (username, password) used for Basic Authentication
+
+        Any additional keyword arguments will be passed to to the request, and can be used to setup timesouts etc.
         """
 
         self.returnFormat = returnFormat
@@ -48,18 +53,12 @@ class SPARQLConnector(object):
         self.update_endpoint = update_endpoint
         self.kwargs = kwargs
         self.method = method
-
-        # it is recommended to have one session object per thread/process. This assures that is the case.
-        # https://github.com/kennethreitz/requests/issues/1871
-
-        self._session = threading.local()
-
-    @property
-    def session(self):
-        k = "session_%d" % os.getpid()
-        self._session.__dict__.setdefault(k, requests.Session())
-        log.debug("Session %s %s", os.getpid(), id(self._session.__dict__[k]))
-        return self._session.__dict__[k]
+        if auth is not None:
+            assert type(auth) == tuple, "auth must be a tuple"
+            assert len(auth) == 2, "auth must be a tuple (user, password)"
+            base64string = base64.b64encode(bytes('%s:%s' % auth, 'ascii'))
+            self.kwargs.setdefault("headers", {})
+            self.kwargs["headers"].update({"Authorization": "Basic %s" % base64string.decode('utf-8')})
 
     @property
     def method(self):
@@ -72,7 +71,7 @@ class SPARQLConnector(object):
 
         self._method = method
 
-    def query(self, query, default_graph=None):
+    def query(self, query, default_graph: str = None, named_graph: str = None):
 
         if not self.query_endpoint:
             raise SPARQLConnectorException("Query endpoint not set!")
@@ -84,7 +83,6 @@ class SPARQLConnector(object):
         headers = {"Accept": _response_mime_types[self.returnFormat]}
 
         args = dict(self.kwargs)
-        args.update(url=self.query_endpoint)
 
         # merge params/headers dicts
         args.setdefault("params", {})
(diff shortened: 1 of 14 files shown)


rdflib/plugins/stores/sparqlconnector.py
```diff
@@ -94,47 +92,44 @@ def query(self, query, default_graph=None):
 
         if self.method == "GET":
             args["params"].update(params)
+            qsa = "?" + urlencode(args["params"])
+            res = urlopen(Request(self.query_endpoint + qsa, headers=args["headers"]))
         elif self.method == "POST":
             args["headers"].update({"Content-Type": "application/sparql-query"})
-            args["data"] = params
+            try:
+                res = urlopen(Request(self.query_endpoint, data=query.encode(), headers=args["headers"]))
```
ashleysommer, line 100: Don't you need `method='POST'` here?

rdflib/plugins/stores/sparqlconnector.py
```diff
@@ -94,47 +92,44 @@ def query(self, query, default_graph=None):
 
         if self.method == "GET":
             args["params"].update(params)
+            qsa = "?" + urlencode(args["params"])
+            res = urlopen(Request(self.query_endpoint + qsa, headers=args["headers"]))
         elif self.method == "POST":
             args["headers"].update({"Content-Type": "application/sparql-query"})
-            args["data"] = params
```
ashleysommer, line 99: Why don't we need `params` in `args["data"]` anymore?

rdflib/plugins/stores/sparqlconnector.py
```diff
@@ -94,47 +92,44 @@ def query(self, query, default_graph=None):
...
         args.setdefault("params", {})
         args["params"].update(params)
         args.setdefault("headers", {})
         args["headers"].update(headers)
 
-        res = self.session.post(**args)
-
-        res.raise_for_status()
-
-    def close(self):
```
ashleysommer, line 139: Why do we remove `close()` method?

ashleysommer:
@nicholascar 
Why did you put the Python3.6+ changes into this `requests` removal PR?

rdflib/plugins/stores/sparqlconnector.py
nicholascar: No, if you specify `data=...` then urllib.Request becomes a POST automatically

rdflib/plugins/stores/sparqlconnector.py
nicholascar: The only `params` allowed for a `POST` SPARQL query is the query itself, [as per the spec](https://www.w3.org/TR/sparql11-protocol/#query-operation), so we don't allow any others. Auth params for secure endpoints I've catered for with a separate `auth` param.

rdflib/plugins/stores/sparqlconnector.py
nicholascar: Because the Store is actually stateless with each operation and there's nothing to close. I originally dropped `open()` too but it's needed if the Store is called by `Graph("SPARQLStore")` as opposed to `SPARQLStore()`

nicholascar:
Me being lazy. I saw Python 3.5 failures and wanted to jump over those! I thought we agreed to drop 3.5 so I was really just testing that out.

Also, since this PR is all about dropping dependencies, I wanted to reduce the encumbrances as much as possible.

Since there is a remaining Python 3.6 error, I'm going to have to address that in further edits here so could reintroduce Python 3.5 if you like.

We did agree to drop Python 3.5 right?

nicholascar:
Weird that it fails tests here as it passed everything locally on Python 3.6: I ran on that version in particular to test. I think it's just a naming collision with `json` packages so perhaps easily solvable in the week.

ashleysommer:
Yes, we did agree to drop python 3.5 for RDFLib v6.0.0, and have a minimum 3.6. Just wasn't expecting that change to come as part of the "Remove requests" PR. But it should be fine.

nicholascar:
Regarding the single failure here, the message is `OSError: [Errno 99] Cannot assign requested address`. This is due to something in the setting of the testing machine, not the content of RDFlib code I think as the error doesn't appear running tests locally. It's something about setting ports or colliding with ports in use. Any thoughts on this one?

rdflib/plugins/stores/sparqlconnector.py
white-gecko: https://docs.python.org/3/library/urllib.request.html#urllib.request.Request

rdflib/plugins/stores/sparqlconnector.py
white-gecko: We still need to submit `default-graph-uri` and `named-graph-uri` parameters. So the `params` is still required.

Also I think we should support both `application/x-www-form-urlencoded` and `application/sparql-query` content-types. But I'm unsure how we would determine which content-type is supported by the store we are speaking with.

rdflib/plugins/stores/sparqlstore.py
```diff
@@ -381,8 +386,35 @@ def _is_contextual(self, graph):
...
+
+    def subject_objects(self, predicate=None):
+        """A generator of (subject, object) tuples for the given predicate"""
+        for t, c in self.triples((None, predicate, None)):
+            yield t[0], t[2]
+
+    def predicate_objects(self, subject=None):
+        """A generator of (predicate, object) tuples for the given subject"""
+        for t, c in self.triples((subject, None, None)):
+            yield t[1], t[2]
```
white-gecko, line 417: Why do we need these methods?

white-gecko (commented):
It is hard for me to understand the complete pull request, as it mixes a lot of different things. Would it be possible for you to split this into several unrelated pull-requests?

white-gecko:
> Regarding the single failure here, the message is `OSError: [Errno 99] Cannot assign requested address`. This is due to something in the setting of the testing machine, not the content of RDFlib code I think as the error doesn't appear running tests locally. It's something about setting ports or colliding with ports in use. Any thoughts on this one?

This might show that there are to many open connections in to short time: https://stackoverflow.com/questions/11981933/python-urllib2-cannot-assign-requested-address?noredirect=1

This could be a limitation on the travis machine. Maybe we can also perform the tests with some local endpoint. Which might not be related to that issue.

rdflib/plugins/stores/sparqlstore.py
nicholascar: These are all just convenience methods that are present in the normal `Graph()` interface and they are commonly used, so I thought they could just be present in the SPARQLStore too. This store returns quads, not triples, so I had to copy these methods into the store's own list of methods to re-implement them, else they would just break when someone called them.

nicholascar:
> > Regarding the single failure here, the message is `OSError: [Errno 99] Cannot assign requested address`. This is due to something in the setting of the testing machine, not the content of RDFlib code I think as the error doesn't appear running tests locally. It's something about setting ports or colliding with ports in use. Any thoughts on this one?
> 
> This might show that there are to many open connections in to short time: https://stackoverflow.com/questions/11981933/python-urllib2-cannot-assign-requested-address?noredirect=1
> 
> This could be a limitation on the travis machine. Maybe we can also perform the tests with some local endpoint. Which might not be related to that issue.

When I perform the test locally, they all pass so I do think it is a Travis limitation.

nicholascar:
> We still need to submit `default-graph-uri` and `named-graph-uri` parameters. 

OK, let me think about that! I might have missed something there.

nicholascar:
> Would it be possible for you to split this into several unrelated pull-requests?

Not really: the focus here was just to get rid of requests and requests is only use in 2 places in the entire codebase: a couple of tests and by SPARQLStore. This PR's really only just removing those references and then patching SPARQLStore to ensure that everything works. 

The only optional things not about that are the convenience methods such as `Graph().subjects()` that you asked about in another comment.

nicholascar:
Open/Close to trigger Travis to see if we can overcome the port hogging test failure

nicholascar:
Here's a demo fo `default-graph-uri` working , compared with the NG in the query:

```python
from rdflib.plugins.stores.sparqlstore import SPARQLStore
ENDPOINT = "http://cgi.surroundaustralia.com:7200/repositories/cgi-vocs"
NAMED_GRAPH = "http://resource.geosciml.org/classifierscheme/cgi/2016.01/classification-method-used"
# NG in query, Graph then SPARQLStore
q = """
    PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
    SELECT (COUNT(?s) AS ?c)
    WHERE {{
        GRAPH <{}> {{
            ?s a skos:Concept .
        }}
    }}
    """.format(NAMED_GRAPH)

g = Graph("SPARQLStore")
g.open(ENDPOINT)
for r in g.query(q):
    assert int(r['c']) == 14

st = SPARQLStore(query_endpoint=ENDPOINT)
for r in st.query(q):
    assert int(r['c']) == 14

# NG separate from query, Graph then SPARQLStore
q2 = """
    PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
    SELECT (COUNT(?s) AS ?c)
    WHERE {
        ?s a skos:Concept .
    }
    """

g = Graph("SPARQLStore", identifier=NAMED_GRAPH)
g.open(ENDPOINT)
for r in g.query(q2):
    assert int(r['c']) == 14

st = SPARQLStore(query_endpoint=ENDPOINT)
for r in st.query(q2, queryGraph=NAMED_GRAPH):
    assert int(r['c']) == 14
```

white-gecko:
Are you sure it is executed via POST?

nicholascar:
> Are you sure it is executed via POST?

No, there seems to be no logic that allows SPARQLStore to ever use POST in SparqlConnector. SPARQLUpdateStore has a `postAsEncoded` variable but this also seems never to be used. In testing I manually set SparqlConnector's `self.method` value to POST to try it out - it works fine for most queries - but I haven't done as much testing as for GET.

I will write some POST tests next.

I think there were quite a few combinations of things not tested in the original PSARQLStore code but fixing them all wasn't the purpose of this PR, just to remove requests! I've had to fix lots of things to get it all working without requests though.

nicholascar:
@ashleysommer @white-gecko please can you approve this PR? The single error in Travis above is, again `urlopen error [Errno 99] Cannot assign requested address` which is a Travis machine capacity issue, not a "real" test failure. 

There are an endless number of things that I could keep fixing in SPARQLStore but that's not the goal of this PR - just the removal of requests.

Actually I'm using SPARQLStore in a production system now so will keep working on it. If this PR can be merged, I'll use `master` in the system (as opposed to 5.0.0) and that will let me keep testing SPARQLStore.

ashleysommer:
@nicholascar 
I'm not concerned about the one failing Travis issue. That is something I'll have to look into going forward, as I'm pretty sure it is bad practice to stand up a real web server in a testing environment for running tests against, and its bad practice to make real web requests in a testing environment tool.

If you're confident the POST issue raised by @white-gecko above isn't related to these changes (and it was like that already when using `requests`) then we can accept and merge this.
We'll have to create a new issue for the POST-related problem.

nicholascar:
> its bad practice to make real web requests in a testing environment tool.

Yes, I think we are going to have to better mock up things like the SPARQL endpoints that are needed for some of the tests so things like Fuseki aren't needed to be bundled into the tests.

> If you're confident the POST issue raised by @white-gecko above isn't related to these changes (and it was like that already when using requests) then we can accept and merge this.

Yes I am confident: that's a different thing and part of the incomplete or now slightly broken SPARQLStore implementation that needs love and attention.

I'd like to raise a bunch of Issues for SPARQLStore and then fix them with some PRs straight after this one which I'll discuss this arvo. Having this PR through will set a base layer for me to test behaviour with. I can think of a whole bunch more tests like the ones for GET I implemented but for POST and also things like transactions testing etc. I'm concerned that transaction rollback etc isn't actually working but we don't normally notice since SPARQL Services that I use are highly available and thus don't often disappear mid way through a session.

ashleysommer (approved):

ashleysommer:
@nicholascar 
Did you run `black` on these files after these changes?
I just ran `black` on the codebase for my PR, and I'm seeing lots of `black` errors in plugins/stores/sparqlstore and in sparqlwrapper and other sparql-related files.

rdflib/plugins/stores/sparqlconnector.py
white-gecko: The `params` attribute is still required or at least the information about the default-graph.

If you consider the following:

Load some context aware quad store (e.g. the QuitStore) with the following data:
```turtle
graph <http://example.org/> {
    <http://example.org/ExampleInstance> a <http://example.org/Example>
}
graph <http://othergraph.org/> {
    <http://example.org/OtherInstance> a <http://example.org/Example>
}
```

The query endpoint is assumed to be available at `http://localhost:5000/sparql`

```python
class SPARQLStoreQuitStoreTestCase(unittest.TestCase):
    store_name = "SPARQLStore"
    path = "http://localhost:5000/sparql"
    create = False

    def setUp(self):
        store = SPARQLStore(query_endpoint=self.path, method="POST")
        self.conjunctivegraph = ConjunctiveGraph(store=store)

    def tearDown(self):
        self.conjunctivegraph.close()

    def test_Query(self):
        query = "select distinct ?inst where {?inst a <http://example.org/Example>}"
        graph = self.conjunctivegraph.get_context(URIRef("http://example.org/"))
        res = graph.query(query, initNs={})
        assert len(res) == 1, len(res)
        for i in res:
            assert type(i[0]) == URIRef, i[0].n3()
            assert i[0] == URIRef("http://example.org/ExampleInstance"), i[0].n3()
```

The query is executed as POST request but does not convey the information about the default graph.

This issue is a combination of removing this line and sending the wrong Content-Type.
Actually, what the request was doing before this change and the merge of #1022 was sending a proper POST request with a `application/x-www-form-urlencoded` content type according to the SPARQL 1.1 Protocol (query via URL-encoded POST; https://www.w3.org/TR/2013/REC-sparql11-protocol-20130321/#query-operation).

white-gecko:
Are you sure the `urlopen error [Errno 99] Cannot assign requested address` is not introduced by removing the session? As the `requests.Session()` is implementing connection pooling (https://2.python-requests.org/en/master/user/advanced/). It might be that by removing the session there are to many open connections which bring travis to the limit of pots to be used. On your local machine you might have more open ports available. I'm unsure if this could also happen in some use cases, where a lot of queries are sent.

PR #1176, opened 2020-10-03
URI Validation Performance Improvements

jbmchuck:
I've found _is_valid_uri to be a hotspot in serialization and
deserialization.

Benchmarking this version of the function against the original
with 1e7 urls results in runtime improvement of nearly 3x.

Tests against real RDF graphs with 2e6+ triples results in runtime
improvement of 10-12% in serialization and deserialization.

PR #1177, opened 2020-10-03
URI Validation Performance Improvements

jbmchuck:
I've found _is_valid_uri to be a hotspot in serialization and
deserialization.

Benchmarking this version of the function against the original with 1e7
urls results in runtime improvement of ~5x

Tests against real RDF graphs with 2e6+ triples results in runtime
improvement of ~10-12% in serialization and deserialization.

Functionally there is one diff - I don't believe the `ord(c)` check is
necessary.  In the original code effectively the check for each char
is that:

    ord(c) > 256
            OR
    c not in _invalid_uri_chars

Since for all c in _invalid_uri_chars ord(c) <= 256 there is no case where
the `ord(c) > 256` condition holds true but `c not in _invalid_uri_chars`
does not also.

I tried a few other approaches including performing this check in a list
comprehension, e.g.

    return bool([c for c in _invalid_uri_chars if c in uri])

but this was significantly slower for large numbers of calls, though faster than
the original code by ~2x.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -74,7 +74,7 @@ _invalid_uri_chars = '<>" {}|\\^`'
 
 
 def _is_valid_uri(uri):
-    return all([ord(c) > 256 or not c in _invalid_uri_chars for c in uri])
+    return not bool([c for c in uri if c in _invalid_uri_chars and ord(c) <= 256])
 
 
 _lang_tag_regex = compile("^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$")
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -74,8 +74,8 @@ _invalid_uri_chars = '<>" {}|\\^`'
 
 
 def _is_valid_uri(uri):
-    for c in uri:
-        if c in _invalid_uri_chars:
+    for c in _invalid_uri_chars:
+        if c in uri:
             return False
     return True
 
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -74,7 +74,10 @@ _invalid_uri_chars = '<>" {}|\\^`'
 
 
 def _is_valid_uri(uri):
-    return not bool([c for c in uri if c in _invalid_uri_chars and ord(c) <= 256])
+    for c in uri:
+        if c in _invalid_uri_chars:
+            return False
+    return True
 
 
 _lang_tag_regex = compile("^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$")


jbmchuck:
Benchmark script:

    import time

    _invalid_uri_chars = '<>" {}|\\^`'

    def _is_valid_uri_orig(uri):
        return all([ord(c) > 256 or not c in _invalid_uri_chars for c in uri])

    def _is_valid_uri_new(uri):
        return not bool([c for c in uri if c in _invalid_uri_chars and ord(c) <= 256])

    sample_uris = [
        "http://foo.{x}.com".format(x=x) for x in range(1, int(1e7))
    ]


    while True:
        start_t = time.time()
        for sample_uri in sample_uris:
            _is_valid_uri_orig(sample_uri)
        print("orig: ", time.time()-start_t)
        start_t = time.time()

        start_t = time.time()
        for sample_uri in sample_uris:
            _is_valid_uri_new(sample_uri)
        print("new: ", time.time()-start_t)
        start_t = time.time()

ashleysommer:
Hi @jbmchuck 
I see this approach is a tiny bit slower than your proposal in https://github.com/RDFLib/rdflib/pull/1176
Why did you close that one and go this route?

It looks like you're getting yourself confused with the logical requirements of the `_is_valid_uri` function, amongst these optimisations.

The function must return True if:
1) for each character `c` in URI `ord(c)` is `<= 256`
2) for each character `c` in URI `c` is _not_ in `_invalid_uri_chars`

Another way of looking at this is, the function will return False if:
1) any character `c` in URI `ord(c)` is `> 256`
2) any character `c` in URI is in `_invalid_uri_chars`

Both of your commits in this PR:
https://github.com/RDFLib/rdflib/pull/1177/commits/fec4ed05a7bbbc20236ef7610bb465a70e709ca2 and https://github.com/RDFLib/rdflib/pull/1177/commits/c95965080089b43f752beeb33f9fb1968c8c0fdd do not capture the above logic.

jbmchuck:
Hi @ashleysommer ,

The approach in 1176 is incorrect, it incorrectly flags anything with non-ascii chars as invalid.  I think it's actually slightly slower than the current (https://github.com/RDFLib/rdflib/pull/1177/commits/6484fcd2ac771c111577133e23be4a250c0e20de) which is about 5x the original func.

FlorianLudwig:
When profiling RDFlib I came across the same issue but actually looked into how to get rdflib to do less validations in the first place.

@jbmchuck  approach looks correct, readable and fast to me. Looking forward to having this merged.

ashleysommer:
We'll discuss this in today's RDFLib maintainers meeting and will probably merge it.

PR #1180, opened 2020-10-07
Support parsing paths specified with pathlib

AnjoMan:
pathlib was added to the standard librarly as of Python 3.4.
This adds support for calling Graph.parse on a file specified
using a pathlib object.

## Proposed Changes

  -  if the source is specified as a Path object or subtype, call str on it
  - `pathlib` objects implement __str__  precisely for this purpose: https://docs.python.org/3/library/pathlib.html

### DIFF
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -12,6 +12,7 @@ want to do so through the Graph class parse method.
 
 import codecs
 import os
+import pathlib
 import sys
 
 from io import BytesIO, TextIOBase, TextIOWrapper, StringIO, BufferedIOBase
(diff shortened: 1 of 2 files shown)


AnjoMan:
please add the label `hacktoberfest-accepted`

ashleysommer:
Hi @AnjoMan 
This looks like a good change, especially now that python 2 (and python <=3.5) are no longer supported in future versions of RDFLib, and Pathlib is becoming more popular in the Python community.

We will discuss this more formally tomorrow at our RDFLib maintainers meeting.

I'm sorry, we are not participating in DigitalOcean's Hacktoberfest campaign.

PR #1181, opened 2020-10-07
Issue 738: support empty publicID

### LINKED ISSUE #738: RDFLib doesn't accept an empty public ID
I want to set the public id to be an empty string, so that the ID values on my xml elements will not be amended.

If I run

``` python

g = rdflib.Graph()
g.load("foo.xml")

for subj, obj, pred in g:
    print(f"El\tSubject:\t{s};\n\tPredicate:\t{p};\n\tObject:\t\t{o};")
```
where the contents of `foo.xml` are:
``` xml
<?xml version="1.0" encoding="UTF-8"?>
<!--Generated by CYME 8.0r02, CYME International T&D on 2017-02-13 7:51:55 PM-->
<rdf:RDF xmlns:cim="http://iec.ch/TC57/2013/CIM-schema-cim16#"
         xmlns:cyme="http://www.cyme.com/CIM/1.0.2#"
         xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
    <cim:SwitchInfo rdf:ID="_AB16765A-B19E-4454-A58F-868D23C6CD26">
        <cim:IdentifiedObject.name>DEFAULT</cim:IdentifiedObject.name>
        <cim:SwitchInfo.ratedVoltage>25000.000000</cim:SwitchInfo.ratedVoltage>
        <cim:SwitchInfo.ratedCurrent>100.000000</cim:SwitchInfo.ratedCurrent>
    </cim:SwitchInfo>
</rdf:RDF>

```

I get the following output: all my
```
 scratch.py:24
El	Subject:	file:///home/anton/Development/profile_validator/test_cim.xml#_AB16765A-B19E-4454-A58F-868D23C6CD26;
	Predicate:	http://iec.ch/TC57/2013/CIM-schema-cim16#SwitchInfo.ratedVoltage;
	Object:		25000.000000;
El	Subject:	file:///home/anton/Development/profile_validator/test_cim.xml#_AB16765A-B19E-4454-A58F-868D23C6CD26;
	Predicate:	http://iec.ch/TC57/2013/CIM-schema-cim16#IdentifiedObject.name;
	Object:		DEFAULT;
El	Subject:	file:///home/anton/Development/profile_validator/test_cim.xml#_AB16765A-B19E-4454-A58F-868D23C6CD26;
	Predicate:	http://www.w3.org/1999/02/22-rdf-syntax-ns#type;
	Object:		http://iec.ch/TC57/2013/CIM-schema-cim16#SwitchInfo;
El	Subject:	file:///home/anton/Development/profile_validator/test_cim.xml#_AB16765A-B19E-4454-A58F-868D23C6CD26;
	Predicate:	http://iec.ch/TC57/2013/CIM-schema-cim16#SwitchInfo.ratedCurrent;
	Object:		100.000000;
```

As I expect from [what the docs say](http://rdflib.readthedocs.io/en/stable/apidocs/rdflib.html) about how to use `Graph.parse`, the base uri of each xml element includes the document file path - since I did not give a value for `publicID` it uses the default behaviour for `publicID=None`

Where I'm finding an issue is that

``` python

g = rdflib.Graph()
g.load("foo.xml", publicID="")

for subj, obj, pred in g:
    print(f"El\tSubject:\t{s};\n\tPredicate:\t{p};\n\tObject:\t\t{o};")
```
I get the same output, which to me is unexpected because the value of `publicID` is not `None` and therefore the base uri should be whatever I gave it.

I'm finding this behaviour in Python 3.6 with RDFLib 4.2.2

AnjoMan:
This adds support for specifying a publicID as '';
instead of evaluating the falsy-ness of the input,
we want to use the provided string, even if it is
''.

Fixes #738

## Proposed Changes

  - when the user specifies `publicID` as '', we should not trigger the `None` default behaviour


AnjoMan:
If you like this contribution, please add the #hacktoberfest-accepted label

ashleysommer:
Hi @AnjoMan
This looks like a good change, we will discuss this more formally tomorrow at our RDFLib maintainers meeting. We will have to run some more tests internally to ensure everything in the RDF/XML serializer operates correctly after this change.

I'm sorry, we are not participating in DigitalOcean's Hacktoberfest campaign.

ashleysommer:
Hi @AnjoMan 
After discussing this, we have decided we will need to test this solution further. You have implemented the change just for the RDF/XML parser, but we need to test what happens if you use `publicID=""` for other parsers, like TRIG, or N3, or JSON-LD. While your change does fix the problem you are seeing, it could also potentially introduce breakages in unexpected places, so we need to be cautious about how to proceed with this.

nicholascar (approved):

white-gecko (commented):
As can be seen for the turtle example, with your fix we would introduce some inconsistency across the serializers. The fix does also not work in the same way for the current Turtle parser/notation3 parser as we call `graph.absolutize()` in this case.

What is the usecase for an emptyID?

In m eyes this relative URI thing without explicitly defining a base in the file is a tough problem.

white-gecko (changes requested):
This issue needs to be solved for all parsers at once.

PR #1186, opened 2020-10-08
Remove the usage of assert in the SPARQLConnector

white-gecko:
Fixes the usage of `assert` as introduced in #1175 in the SPARQLConnector.

## Proposed Changes

In my eyes it is no good practice to have assert in production code, so it should not be in a library.
The Python reference states:

> The current code generator emits no code for an assert statement when optimization is requested at compile time.
> https://docs.python.org/3/reference/simple_stmts.html#assert

So I think it should be replaced by if and raise an exception.

There are many more usages of `assert` which should be replaced at some point as well.


nicholascar:
Merging since the only test fails are that cannot assign address issue!

PR #1192, opened 2020-10-23
optimize sparql.Bindings

FlorianLudwig:
This improves performance 13% on the rdflib-benchmark with the default (memory) store and a store size of 4k.

Moving straight to `collections.ChainMap` was considered but 
1) `Bindings` behaves differently when one key is in multiple dicts, as they are iterated over multiple times and also counted on `__len__` multiple times
2) The API is a bit different, increasing the changes size

I do think that the differences for `__len__` and `__iter__` are unintentional / bugs. If so, I would rework this MR to fix those.

### DIFF
--- a/rdflib/plugins/sparql/sparql.py
+++ b/rdflib/plugins/sparql/sparql.py
@@ -52,12 +52,12 @@ class Bindings(MutableMapping):
         self.outer = outer
 
     def __getitem__(self, key):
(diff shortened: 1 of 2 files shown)


nicholascar (approved):

nicholascar:
Test passing and this has been here for a while so I'll merge

PR #1193, opened 2020-10-23
do not use current time in sparql TIMEZONE

FlorianLudwig:
Calculating the timezone should use date date and time from the
datetime object the timezone is requested from instead from the
current time. This makes a difference if the timezone for the given
location has changed in the past. In this particular case it is probably
not possible as iso 8601 seems not have a method to specify the
time zone via a location.
Therefore this cleans up unnecessary code and removing the
call to now in every sparql context creation gives a nice
performance increase.

Speedup varies depending on benchmark between 3% and 12%

(To be honest I am surprised myself with those 12% but could reproduce the result)


FlorianLudwig:
@nicholascar could you also have a look at this merge request? It is rather small :) It is just about making `now` in queries lazy.

PR #1203, opened 2020-11-21
tests: retry on network error (CI)

FlorianLudwig:
The CI does fail at random. A flaky CI is an ignored CI. An ignored CI is pretty much a useless CI. I would like to improve the "ci experience" over all - with the CI providing useful feedback to merge requests this could reduce the burden on maintainers.

This merge request adds a retry function for network based tests.

As the issue is not reproducible I made several small random clean-up changes to trigger travis. I removed the commits from the branch afterwards.

Each travis run, does run the test suite 4 times with different interpreters.

- https://travis-ci.org/github/RDFLib/rdflib/builds/745056525 ✅
- https://travis-ci.org/github/RDFLib/rdflib/builds/745193124 ✅
- https://travis-ci.org/github/RDFLib/rdflib/builds/745193628 ✅
- https://travis-ci.org/github/RDFLib/rdflib/builds/745194332 ✅
- https://travis-ci.org/github/RDFLib/rdflib/builds/745195934 ✅
- https://travis-ci.org/github/RDFLib/rdflib/builds/745197044 ✅

From this test series I like to argue that the change does improve the situation.

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -1,0 +1,23 @@
+import time
+import urllib.error
+
+import rdflib
+import rdflib.query
+
+
+MAX_RETRY = 3
+def query_with_retry(graph: rdflib.Graph, query: str, **kwargs) -> rdflib.query.Result:
+    backoff = 0
+    for i in range(MAX_RETRY):
+        try:
+            result = graph.query(query, **kwargs)
+            result.bindings # access bindings to ensure no lazy loading
+            return result
+        except urllib.error.URLError as e:
+            if i == MAX_RETRY -1:
+              raise e
+
+            backoff_s = 1.2 ** backoff
+            print(f"Network eroror {e} during query, waiting for {backoff_s}s and retrying")
+            time.sleep(1)
+            backoff += 1
--- a/test/helper.py
+++ b/test/helper.py
@@ -5,8 +5,36 @@ import rdflib
 import rdflib.query
 
 
-MAX_RETRY = 3
+MAX_RETRY = 10
+BACKOFF_FACTOR = 1.5
 def query_with_retry(graph: rdflib.Graph, query: str, **kwargs) -> rdflib.query.Result:
+    """Query graph an retry on failure, returns preloaded result
+
+    The tests run against outside network targets which results
+    in flaky tests. Therefor retries are needed to increase stability.
+
+    There are two main categories why these might fail:
+
+     * Resource shortage on the server running the tests (e.g. out of ports)
+     * Issues outside the server (network, target server, etc)
+
+    As fast feedback is important the retry should be done quickly.
+    Therefor the first retry is done after 100ms. But if the issue is
+    outside the server running the tests it we need to be good
+    citizenship of the internet and not hit servers of others at 
+    a constant rate. (Also it might get us banned)
+
+    Therefor this function implements a backoff mechanism.
+
+    When adjusting the parameters please keep in mind that several
+    tests might run on the same machine at the same time
+    on our CI, and we really don't want to hit any rate limiting.
+
+    The maximum time the function waits is:
+
+    >>> sum((BACKOFF_FACTOR ** backoff) / 10 for backoff in range(MAX_RETRY))
+    11.3330078125
+    """
     backoff = 0
     for i in range(MAX_RETRY):
         try:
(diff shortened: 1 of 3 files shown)


FlorianLudwig:
@nicholascar please have a look, thanks :)

white-gecko:
The idea and the changes look good. I tried to find such retry situations in the builds you have linked above, but I could not find the string "retrying" in them. Could you point me to a specific build log, that shows the retry?

white-gecko:
In general I like the idea but I also think if we should maybe just replace those external queries with queries to a simple local endpoint (e.g. fuseki) that we are running in a docker container within the CI.

FlorianLudwig:
> The idea and the changes look good. I tried to find such retry situations in the builds you have linked above, but I could not find the string "retrying" in them. Could you point me to a specific build log, that shows the retry?

i am no expert on nose but it is my understanding that prints are not outputted to the console during tests.

white-gecko:
ah ok

FlorianLudwig:
> In general I like the idea but I also think if we should maybe just replace those external queries with queries to a simple local endpoint (e.g. fuseki) that we are running in a docker container within the CI.

I agree with that but that might be more work. We still might need a retry mechanism than, as it looks like there is a limit on open sockets.

In any case, I think this is an improvement and can (and should be) merged even if the tests will be refactored in the future to not require internet connection at all

white-gecko (approved):

test/helper.py
```diff
@@ -0,0 +1,23 @@
...
+            result = graph.query(query, **kwargs)
+            result.bindings # access bindings to ensure no lazy loading
+            return result
+        except urllib.error.URLError as e:
+            if i == MAX_RETRY -1:
+              raise e
+
+            backoff_s = 1.2 ** backoff
+            print(f"Network eroror {e} during query, waiting for {backoff_s}s and retrying")
+            time.sleep(1)
```
kernc, line 22: Comment and action are in discord. `backoff` is a needless complication, imho.

test/test_sparqlstore.py
```diff
@@ -74,19 +77,21 @@ def test_counting_graph_and_store_queries(self):
             """
         g = Graph("SPARQLStore")
         g.open(self.path)
-        c = 0
-        for r in g.query(q):
-            c += 1
+        count = 0
+        result = helper.query_with_retry(g, query)
+        for _ in result:
+            count += 1
```
kernc, line 83: If by renaming you meant to leave code in better shape than you found it, why not simply:
```py
result = ...
count = len(result)
```

nicholascar:
> In general I like the idea but I also think if we should maybe just replace those external queries with queries to a simple local endpoint (e.g. fuseki) that we are running in a docker container within the CI.

> In any case, I think this is an improvement and can (and should be) merged even if the tests will be refactored in the future to not require internet connection at all

I agree to all of this and thanks @FlorianLudwig for another great PR - tests failing due to connection issues is a real pain. Can you please just address the two comments from @kernc above?

test/helper.py
FlorianLudwig: Hi @kernc,

good catch! I fixed the code but kept the backoff as I believe it is best practise to do this and not a needless complication.

I added my reasoning as comment into the code.

test/test_sparqlstore.py
FlorianLudwig: actually `len` only works if the result is already retrieved. The only reason it works in this context is this line:

https://github.com/RDFLib/rdflib/pull/1203/files#diff-4dd682d5c858fe7d3f39400e3da7a0ea9e77232bab93c6d51e5cb523844665b2R42

As that is somewhat counter intuitive, I will fix `len` in a separate MR. Afterwards I am fine with removing those for loops (there are more of those in the tests).

FlorianLudwig:
@nicholascar thanks! I addressed the two comments. Looking forward to having this merged :)

test/helper.py
kernc: Imho it's way overenginnered and simply `time.sleep(1)` would do.

test/test_sparqlstore.py
kernc: Didn't check, just though it should work based on these `query.Result` lines:
https://github.com/RDFLib/rdflib/blob/56dc4207ce6e7b11ed7b45fb4fd4020ba548e718/rdflib/query.py#L245-L251

Alternatively, even something like `count = len(list(result))` would be better.

PR #1211, opened 2020-12-08
Add documentation and type hints for rdflib.query.Result and rdflib.graph.Graph

edmondchuc:
## Add documentation and type hints for `rdflib.query.Result`. Update documentation and type hints for `rdflib.graph.Graph`.

I am keen to work on adding documentation and type hints to rdflib. `rdflib.query.Result`'s `serialize()` is an example of this work. 
Please leave any comments on styling and conventions. 

## Proposed Changes

 - Progressively update documentation and add type hints everywhere
 - Remove sphinx-style documentation such as parameter and return types and update with newer Python type hints.

For the second point, an example would be to remove https://github.com/RDFLib/rdflib/blob/c5ff12717c02cf21abc8dd00ed8f0ec27071ba5d/rdflib/graph.py#L1150 and replace it with using python type hints. 

E.g. 
```python3
def query(
        self,
        ...
        **kwargs
    ) -> query.Result:
```
Sphinx also understands this and adds hyperlinks automatically as well.

If this is acceptable, we should draft a guideline on adding/updating documentation and type hints to the rdflib codebase.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1130,13 +1130,13 @@ class Graph(Node):
     def query(
         self,
         query_object,
-        processor="sparql",
-        result="sparql",
+        processor: str = "sparql",
+        result: str = "sparql",
         initNs=None,
         initBindings=None,
-        use_store_provided=True,
+        use_store_provided: bool = True,
         **kwargs
-    ):
+    ) -> query.Result:
         """
         Query this graph.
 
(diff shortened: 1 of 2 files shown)


FlorianLudwig:
Thanks for improving the docs, that is really welcome :+1: 

My opinion: the `:returntype:` in the doc string can be removed if there is a type declaration but parameter descriptions, like `:param destination: Path of file output.` cannot.

PR #1213, opened 2020-12-14
ENH: Make ClosedNamespace extend Namespace

kernc:
Useful for testing URIRef is from a particular namespace by means of:
```py
is_from_rdf_namespace = URIRef(RDF.givenName).startswith(RDF)
                                              ^^^^^^^^^^
```

where `URIRef.defrag()` currently wouldn't cut it.

Fixes https://github.com/RDFLib/rdflib/issues/542
Closes https://github.com/RDFLib/rdflib/pull/597
Closes https://github.com/RDFLib/rdflib/issues/596

## Proposed Changes

  - Have `rdflib.namespace.ClosedNamespace` be a `str` like other Namespaces are.

-----

Excellent library, btw. Almost works. :laughing: With the move to Py3-only, the code could use a more thorough refresher.


nicholascar:
@hsolbrig you've previously worked in this space! Please review with reference to your previous work which isn't merged. Thanks.

nicholascar:
Thanks for the PR @kernc. I'm just putting this to @hsolbrig who has worked around this area in RDFlib before. Once he's taken a look, I will formally (looks fine though).

kernc:
With the prospect of maybe getting this in, I took liberty of minor additional refactoring of the surrounding code. :grin: Everyone benefits.
I'll be equally happy if you just reapply https://github.com/RDFLib/rdflib/pull/551. It was re-scheduled already for 5.0.0, but was somehow missed. :roll_eyes:

rdflib/namespace.py
```diff
@@ -123,28 +118,27 @@ def __new__(cls, value):
...
         return self.term(key)
 
     def __getattr__(self, name):
         if name.startswith("__"):  # ignore any special Python names!
             raise AttributeError
-        else:
-            return self.term(name)
+        return self.term(name)
 
     def __repr__(self):
```
aucampia, line 136: I may be wrong here, but AFAIK `f"Namespace({self!r})"` will call `self.__repr__`,  if this is the case it will result in infinite recursion.

aucampia: I guess just doing `return f"Namespace({self})"` should be fine and do what was happening before.

rdflib/namespace.py
```diff
@@ -163,55 +157,54 @@ def __new__(cls, value):
...
+    Trying to create terms not listed is an error
     """
 
-    def __init__(self, uri, terms):
-        self.uri = uri
-        self.__uris = {}
-        for t in terms:
-            self.__uris[t] = URIRef(self.uri + t)
+    def __new__(cls, uri, terms):
+        try:
```
aucampia, line 177: If the expectation is that this calls `Namespace.__new__` (which it should I think), then the exception handling should not be there, as it is already in `Namespace.__new__`

rdflib/namespace.py
```diff
@@ -163,55 +157,54 @@ def __new__(cls, value):
...
-    def __str__(self):
-        return str(self.uri)
+        try:
+            return self.term(name)
+        except KeyError as e:
+            raise AttributeError(e)
 
     def __repr__(self):
-        return "rdf.namespace.ClosedNamespace(%r)" % str(self.uri)
+        return f"{self.__module__}.{self.__class__.__name__}({self!r})"
```
aucampia, line 204: Here again I think `{self!r}` should just be `{self}`

rdflib/namespace.py
```diff
@@ -163,55 +157,54 @@ def __new__(cls, value):
...
 
     def __dir__(self):
-        return list(self._ClosedNamespace__uris)
```
aucampia, line 214: I guess this never worked, I vaguely recall hitting my head against this and being confused with the resulting error.

test/test_aggregate_graphs.py
```diff
@@ -122,7 +122,7 @@ def setUp(self):
             graph.parse(StringIO(n3Str), format="n3")
 
         self.graph4 = Graph(memStore, RDFS)
-        self.graph4.parse(RDFS.uri)
+        self.graph4.parse(RDFS)
```
aucampia, line 125: Won't this work fine as it was? And does it make sense to change it? Maybe can add both with and without .uri

rdflib/namespace.py
edmondchuc: This `f"Namespace({self!r})"` is fine as long as `__str__` is implemented. Since we are subclassing `Namespace` from `str`, this is fine and does not result in an infinite recursion.

edmondchuc: I can't see any issues with `return f"Namespace({self})"` though either. Maybe @kernc can explain further?

rdflib/plugins/sparql/operators.py
```diff
@@ -35,7 +35,7 @@
 
 
 # closed namespace, langString isn't in it
-RDF_langString = URIRef(RDF.uri + "langString")
+RDF_langString = URIRef(RDF + "langString")
```
edmondchuc, line 38: Replace 
```python
RDF_langString = URIRef(RDF + "langString")
```
with 
```python
RDF_langString = RDF.langString
```

edmondchuc:
@nicholascar @ashleysommer This is ready to be merged.

edmondchuc:
Merged some docstring tests and test cases from this pull request https://github.com/RDFLib/rdflib/pull/597. Merging this PR should close https://github.com/RDFLib/rdflib/pull/597.

nicholascar:
Thanks very much @edmondchuc 

@aucampia @kernc I'm going to leave this PR un-merged just for a couple of days for you to review/comment but will merge by Friday at the latest as we have a sprint on this week to produce RDFlib 6.0.0 and I want this done so we can move on. 

Since tests pass and we have new tests here, we will be fine if it's merged now without further changes, even if there are more things that could be added. Can always do follow-up PRs!

rdflib/namespace.py
aucampia: Just checked:

```
20210707T222944 iwana@teekai.zoic.eu.org:~/d/github.com/iafork/rdflib
$ git diff
diff --git a/test/test_namespace.py b/test/test_namespace.py
index 59ff0ece..92b9e495 100644
--- a/test/test_namespace.py
+++ b/test/test_namespace.py
@@ -155,3 +155,8 @@ class NamespacePrefixTest(unittest.TestCase):
 
         ref = URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')
         self.assertTrue(ref in RDF, "_RDFNamespace does not include rdf:type")
+
+    def test_repr(self):
+        ns_str = "http://example.com/name/space"
+        ns = Namespace(ns_str)
+        self.assertEqual(ns_str, f"{ns!r}")
20210707T222949 iwana@teekai.zoic.eu.org:~/d/github.com/iafork/rdflib
$ ^C
20210707T223015 iwana@teekai.zoic.eu.org:~/d/github.com/iafork/rdflib
$ poetry run python run_tests.py --tests test/test_namespace.py 
Running nose with: --tests test/test_namespace.py --attr= --where=./ --with-doctest --doctest-extension=.doctest --doctest-tests
.....E....
ERROR: test_repr (test.test_namespace.NamespacePrefixTest)
  File "/home/iwana/d/github.com/iafork/rdflib/test/test_namespace.py", line 162, in test_repr
    self.assertEqual(ns_str, f"{ns!r}")
  File "/home/iwana/d/github.com/iafork/rdflib/rdflib/namespace.py", line 144, in __repr__
    return f"Namespace({self!r})"
  File "/home/iwana/d/github.com/iafork/rdflib/rdflib/namespace.py", line 144, in __repr__
    return f"Namespace({self!r})"
  File "/home/iwana/d/github.com/iafork/rdflib/rdflib/namespace.py", line 144, in __repr__
    return f"Namespace({self!r})"
  [Previous line repeated 235 more times]
RecursionError: maximum recursion depth exceeded while calling a Python object

Ran 10 tests in 0.005s

FAILED (errors=1)
```

aucampia: The `!r` at the end there means it will call __repr__ and not str, which is the function itself.

Will make a PR to the branch.

aucampia: Please merge: https://github.com/kernc/rdflib/pull/1

rdflib/namespace.py
aucampia: Please merge: https://github.com/kernc/rdflib/pull/1/files

aucampia (changes requested):

aucampia:
To make things easier, I can also make a PR to master to merge fixes for `__repr__` after this is merged to master, think it is better to get this in for 6.0.0 than not and other than for __repr__ I see nothing wrong.

rdflib/namespace.py
kernc: @aucampia, @edmondchuc 
[`__repr__` output](https://docs.python.org/3/reference/datamodel.html#object.__repr__):
> should look like a valid Python expression that could be **used to recreate an object with the same value**

So one of these would be better, I think:
```py
return f"Namespace({super().__repr__()})"
return f"Namespace('{self}')"  # assuming no quotes in namespace URI
```

rdflib/namespace.py
```diff
@@ -169,35 +181,38 @@ def __new__(cls, value):
...
-        return URIRef(str(self).__mod__(*args, **kwargs))
+        return URIRef(super().__mod__(*args, **kwargs))
 
     def format(self, *args, **kwargs):
-        return URIRef(str.format(self, *args, **kwargs))
+        return URIRef(super().format(*args, **kwargs))
 
     def __repr__(self):
-        return "URIPattern(%r)" % str(self)
+        return f"URIPattern({self!r})"
```
kernc, line 190: @aucampia Is it [the same issue](https://github.com/RDFLib/rdflib/pull/1213/files#r633624221) here?

rdflib/namespace.py
edmondchuc: @aucampia Nice, thanks for the check.

edmondchuc:
@aucampia thanks for the additional checks and tests. @kernc if you could merge in those new PRs on your fork, then we can review it once more before we merge it in on Friday.

rdflib/namespace.py
aucampia: yes, it also causes infinite recursion.

rdflib/namespace.py
aucampia: @kernc I updated PR with your suggestion and added tests to validate.

aucampia (approved):

nicholascar (approved):

PR #1220, opened 2020-12-28
fix foaf ClosedNamespace

### LINKED ISSUE #1210: Missing property for built-in FOAF namespace
The property 'gender' specified in http://xmlns.com/foaf/spec/ is missing from the built-in namespace. Error is thrown when referencing FOAF.gender.

nicholascar:
ensuring FOAF ClosedNamespace contains all FOAF classes & properties

Fixes #1210

## Proposed Changes

  - the existing FOAF ClosedNamespace contains only the primary classes & properties listed at http://xmlns.com/foaf/spec/ but not all of them, erroneously taken from the "FOAF Core" & "Social Web" coloured box listing, not the "A-Z of FOAF terms" as it should have
  - this fix includes all of them


white-gecko (approved):
Looks good to me

PR #1234, opened 2021-01-23
Add sentence on names not valid as Python IDs.

### LINKED ISSUE #1233: Leading digits in local part of prefixed names and `rdflib.Namespace`
According to [RDF 1.1 Turtle](https://www.w3.org/TR/turtle/#sec-iri), the local part of prefixed names may include leading digits. For example, ` http://lod.nal.usda.gov/nalt/286017 ` can be prefixed as `nalt:286017`.  However, the rdflib Namespace mechanism does not handle this URI as expected:
```
>>> from rdflib import Namespace, Graph, RDFS
>>> NALT=Namespace("http://lod.nal.usda.gov/nalt/")
>>> g = Graph()
>>> g.parse('nalt.nt', format='nt')
<Graph identifier=Nc26a8583c0c14cb9b323bf7dc55375fe (<class 'rdflib.graph.Graph'>)>
>>> list(g.subject_objects(RDFS.label))
[(rdflib.term.URIRef('http://lod.nal.usda.gov/nalt/'), rdflib.term.Literal('NAL Thesaurus'))]
>>> list(g.predicates(NALT.28601))
  File "<ipython-input-24-460dbe57146e>", line 1
    list(g.predicates(NALT.28601))
                          ^
SyntaxError: invalid syntax
```
Is this a known issue (I couldn't find one here) and, if so, could it be fixed?

tombaker:
Fixes #1233

## Proposed Changes

  - Add sentence pointing out that name strings may be valid for use in an RDF namespace but not valid as Python identifiers (e.g., "1234" because it starts with a digit) and must therefore be addressed with the "item" syntax (as illustrated in the example above).
 
This point is currently made as a brief in-line comment to a line in the code example, but the fact that name strings may be valid for RDF but not as Python identifiers deserves to be made more explicitly.

### DIFF
--- a/docs/namespaces_and_bindings.rst
+++ b/docs/namespaces_and_bindings.rst
@@ -17,6 +17,8 @@ The :mod:`rdflib.namespace` defines the :class:`rdflib.namespace.Namespace` clas
 	n['first%20name'] # as item - for things that are not valid python identifiers
 	# = rdflib.term.URIRef(u'http://example.org/first%20name')
 
+Note that if a name string is valid for use in an RDF namespace but not valid as a Python identifier, such as '1234', it must be addressed with the "item" syntax (using the "attribute" syntax will raise a Syntax Error).
+
 The ``namespace`` module also defines many common namespaces such as RDF, RDFS, OWL, FOAF, SKOS, etc. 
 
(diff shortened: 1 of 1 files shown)


white-gecko (approved):

white-gecko:
Sounds like a good idea

nicholascar (approved):

nicholascar:
Thanks @tombaker! Always a pleasure to fold in more documentation.

The PR is approved but it seems we may have a technical Travis issue so we will merge it in as soon as that's resolved. Online documentation will update only with the next RDFlib release which is due in a month or two.

white-gecko:
As this is a documentation issue, we don't need the CI here, right?

tombaker:
On 2021-02-05 04:30, Nicholas Car wrote:
> Thanks @tombaker! Always a pleasure to fold in more documentation.

Thank you @nicholascar! I will look for other places where a sentence here or there could make rdflib easier to use.

nicholascar:
> As this is a documentation issue, we don't need the CI here, right?

In this case yes but of course some documentation affects the doctests, so we can't always make this call..

PR #1237, opened 2021-01-27
Adding Namespace.__contains__()

JaimieMurdock:
No issue for this, just a quick addition for usability/literacy while working with the library.

## Proposed Changes

  - Adds a `Namespace.__contains__()` method to enable syntax sugar of `ref in ns`. This is a natural way to discover if an arbitrary `URIRef` is in the `Namespace`.
  - Current method would be manually writing `ref.startswith(ns)`, which doesn't preserve semantics

This has possible extension to `NamespaceManager`, testing if a `URIRef` is in any managed namespace.

```
class NamespaceManager():
    def __contians__(self, ref):
        return any(ref.startswith(ns) for x, ns in self.namespaces())
```

Note that in `NamespaceManager`, namespaces are stored as `URIRef`s, not `Namespace` objects :/

Hoping this starts a discussion about some usability concerns.

### DIFF
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -112,7 +112,11 @@ class Namespace(str):
     rdflib.term.URIRef(u'http://example.org/Person')
     >>> n['first-name'] # as item - for things that are not valid python identifiers
     rdflib.term.URIRef(u'http://example.org/first-name')
-
+    >>> n.Person in n
+    True
+    >>> n2 = Namespace("http://example2.org/")
+    >>> n.Person in n2
+    False
     """
 
     def __new__(cls, value):
@@ -141,6 +145,9 @@ class Namespace(str):
 
     def __repr__(self):
         return "Namespace(%r)" % str(self)
+    
+    def __contains__(self, other):
+        return other.startswith(self) # test namespace membership with "ref in ns" syntax
 
 
 class URIPattern(str):
--- a/rdflib/namespace.py
+++ b/rdflib/namespace.py
@@ -146,8 +146,8 @@ class Namespace(str):
     def __repr__(self):
         return "Namespace(%r)" % str(self)
     
-    def __contains__(self, other):
-        return other.startswith(self) # test namespace membership with "ref in ns" syntax
+    def __contains__(self, ref):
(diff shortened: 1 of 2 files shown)


white-gecko:
Hey @JaimieMurdock, thank you for submitting a pull-request. I understand your motivation but I'm unsure if this extra code is worth the convenience as `other.startswith(self)` is a single expression as well. I understand the argument for the NamespaceManagemer, but as I see it would work without the proposed change but sure would make more sense if both classes get the `__contains__()` method.
So I'm 50/50 for this change, maybe @nicholascar or @ashleysommer have a stringer opinion ;-)

JaimieMurdock:
I'll add the `NamespaceManager.__contains__()` method to a commit shortly.

The major motivator is "Readability counts." from [PEP 20](https://www.python.org/dev/peps/pep-0020/). Also, was the "natural" way of writing it for a relatively-new-to-Python student working on the project.

JaimieMurdock:
Starting to peel back the layers on Namespace management and there's quite a bit of counter-intuitive typing - had to add a method to `ClosedNamespace` and `_RDFNamespace` as well. I'll write up a proper Issue today.

JaimieMurdock:
Worth noting that #597 proposed the `__contains__` functionality implemented here for 5.0.0.

nicholascar:
I'm happy for any considered work to make any Pythonic ways of querying RDFlib objects work. I suspect that work to implement this function, even if the use of the output is essentially the same as the startwith way of working, will address inconsistencies or old coding styles in the code base so it is worth it.

So, if you're prepared to dive into Namespace to implement stuff and the PR passes all tests, great!

ashleysommer:
I'll close this PR and re-open it to hopefully trigger Drone to see the PR and run its tests.

ashleysommer:
Nope, looks like Drone only kicks in for _new_ pull requests.
However the close/open did trigger the New travis config on travisci**.com** which does work a little bit better than the old one from travisci**.org**, see here: https://github.com/RDFLib/rdflib/pull/1237/checks?check_run_id=1861893743

PR #1246, opened 2021-02-05
Add Drone Continuous-Integration, Fix some tests with deprecations

ashleysommer:
Added a drone backend on our Linode server, and added a drone config file to our repository.
Tweaked some tests to get these passing. We may eventually be able to say goodbye to Travis.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,60 @@
+---
+kind: pipeline
+name: python-3-6
+type: docker
+platform:
+  os: linux
+  arch: amd64
+
+steps:
+- name: test
+  image: python:3.6
+  commands:
+  - pip install -U setuptools pip
+  - bash .travis.fuseki_install_optional.sh
+  - pip install --default-timeout 60 -r requirements.txt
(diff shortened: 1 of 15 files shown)


white-gecko (changes requested):
I like the idea of having a new CI!
I have a question about drone: is it necessary to replicate all of the steps' commands for each python version we are testing?

Please do not introduce any other code changes in a pull-request, that is meant to introduce a new CI. Also I think these changes might interfere with #1222 .

ashleysommer:
@white-gecko 
No, unfortunately the code duplication is required in this case. Drone doesn't have quite as sophisticated matrix runner functionality that Travis has. See here: https://docs.drone.io/pipeline/docker/examples/languages/python/
We could parametrically create the runners using a jsonnet file (as shown at the bottom of that example page) but I wanted to keep it all in the .yml file if possible.

I intended this PR to be just a test of the new Drone backend. I can close it now.
I needed to add those little test fixes in order to get the test passing so I could validate how Drone works when everything passes.
I will go ahead and create a new PR with just the .drone.yml file.

PR #1247, opened 2021-02-05
Add a Drone config file.

ashleysommer:
Replaces #1246
Unfortunately the code duplication is required in this case. Drone doesn't have quite as sophisticated matrix runner functionality that Travis has. See here: https://docs.drone.io/pipeline/docker/examples/languages/python/
We could parametrically create the runners using a jsonnet file (as shown at the bottom of that example page) but I wanted to keep it all in the .yml file if possible.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,60 @@
+---
+kind: pipeline
+name: python-3-6
+type: docker
+platform:
+  os: linux
+  arch: amd64
+
+steps:
+- name: test
+  image: python:3.6
+  commands:
+  - pip install -U setuptools pip
+  - bash .travis.fuseki_install_optional.sh
+  - pip install --default-timeout 60 -r requirements.txt
+  - pip install --default-timeout 60 -r requirements.dev.txt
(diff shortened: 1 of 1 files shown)


white-gecko:
Could we then just remove the travis integration as well? Or do we want to keep it for some days?

white-gecko (approved):
Very nice!

white-gecko:
We should still keep the following on our todo list:

From .travis.yml
```
…
arch:
 - amd64
 - ppc64le
…
branches:
  only:
    # only build master and release branches (merge request are built anyhow)
    - master
    - /^\d+\.\d+\.\d+(-.*)?$/
…
after_success:
    - if [[ $HAS_COVERALLS ]] ; then coveralls ; fi

notifications:
    irc:
        channels: "chat.freenode.net#rdflib"
```

- [ ] Architectures: Do we want to test on ppc64le as well? Maybe we just drop it. Would it be possible to test other archs like armhf and arm64 in an emulator with drone?
- [ ] Branches: Do we nee to configure something for the release branches and prs or is this default already?
- [ ] Coveralls: Report coverage somwhere
- [ ] IRC: Report builds to irc or maybe matrix. Maybe we only need this if something is failing on master.

PR #1253, opened 2021-02-11
Changed to support passing fully qualified queries through the graph …

### LINKED ISSUE #1252: Damned if you do, damned if you don't
So passing a fully qualified (with all needed prefixes) SPARQL query through an rdflib graph to a SPARQL endpoint is pretty much impossible. 

This line means that if you pass a null initNs in, it gets updated to include all namespace prefixes the graph knows about:

https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1171

You can bypass this by passing in an empty dict, but then this line in sparqlstore.py:

https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/stores/sparqlstore.py#L170

does the *opposite*, and any dicts get updated (overridden) to include the default RDF and RDFS namespaces. Which is fine, except that some SPARQL endpoints don't allow for prefix redefinition.

This makes it difficult to pass user supplied SPARQL queries (that follow best practice of being fully qualified) through the RDFlib interface.

jimmccusker:
…and store by passing empty initNs dict.

Fixes #1252

## Proposed Changes

  - Only inject store prefixes if the initNs dict is non-empty.

### DIFF
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -167,7 +167,7 @@ class SPARQLStore(SPARQLConnector, Store):
         self.debug = DEBUG
         assert isinstance(query, str)
 
-        if initNs is not None:
+        if initNs is not None and len(initNs) > 0:
             query = self._inject_prefixes(query, initNs)
 
         if initBindings:
--- a/test/test_sparqlstore.py
+++ b/test/test_sparqlstore.py
@@ -67,6 +67,19 @@ class SPARQLStoreDBPediaTestCase(unittest.TestCase):
         for i in res:
             assert type(i[0]) == Literal, i[0].n3()
 
+    def test_query_with_added_rdf_prolog(self):
+        prologue = """\
+        PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+        PREFIX xyzzy: <http://www.w3.org/2004/02/skos/core#>
+        """
+        query = """\
+        SELECT ?label WHERE
+            { ?s a xyzzy:Concept ; xyzzy:prefLabel ?label . } LIMIT 10
+        """
+        res = helper.query_with_retry(self.graph, prologue + query)
+        for i in res:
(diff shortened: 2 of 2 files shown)


white-gecko:
I understand your issue and was surprised, that it is so hard to control how the namespaces are set for the query. I also see, how your fix solves your problem. But this also raises the question. How is it possible to produce a query without any prefix definitions?

white-gecko:
I do not see the relation between the added taste case and your fix.

nicholascar:
> How is it possible to produce a query without any prefix definitions?

If you send:

```sparql
SELECT * WHERE { ?x a abc:X }
```
...and you've not included a prefix for `abc` but the Store has such a prefix, then it will use that without error. But perhaps you meant: `abc` for `<http://abc.com/>` but the store meant `<http://abc.org/>`? So you should set an `initNs` to make `abc` mean `<http://abc.com/>`...

> I do not see the relation between the added taste case and your fix.

Like @white-gecko, I don't think the test case is actually testing the issue the fix is meant to fix. `rdf` is defined in all `Graph` objects but it's not defined differently or used differently in the query in the test case so I don't see what the impact is.

@jimmccusker: are you saying that your new test case fails without your fix and passes with it? I can't get my head around why that would be. Can you explain?

jimmccusker:
The new test adds an explicit prefix for `rdf:`, which ends up being duplicated. Many SPARQL endpoints prohibit this duplication, so any fully qualified query will fail. In fact, we have had issues in the past where you can pass the same prefixed query to the same Graph object twice, and on the first pass, it succeeds, but the Graph caches the prefixes that were used, which get added on the second pass.

nicholascar:
OK, well I understand the issue now - thanks for the explanation.

Passing all tests so looks good to me. We really do have to sort out this behaviour!

Does this fix solve both the issues you identified in Issue #1252?

nicholascar (approved):

nicholascar:
By the way: I'm gearing up now for the 6.0.0 release so small updates like this actually become available for pip users.

PR #1258, opened 2021-02-19
handle encoding of base64Binary Literals

### LINKED ISSUE #1257: Literals with xsd:base64Binary datatype serialize incorrectly
Looking at `rdflib.term` source this appears to be because the lexical is taken as the decoded value, because no base64 encoder is defined. This is in contrast to `xsd:hexBinary` where the lexical is the encoded value.

Example:

```
>>> byts = b'foo'
>>> enc = hexlify(byts)
>>> lit = Literal(enc, datatype=XSD.hexBinary)
>>> print(lit.toPython(), lit)
b'foo', 666f6f   # expected
```
vs
```
>>> enc = b64encode(byts)
>>> lit = Literal(enc, datatype=XSD.base64Binary)
>>> print(lit.toPython(), lit)
b'foo', foo  # unexpected
```

This means for instance that Graph containing this literal in the o position actually serializes like:
```
@prefix ns1: <http://example.com/> .

ns1:foo ns1:bar "foo"^^xsd:base64Binary,
```

when I expect:
```
@prefix ns1: <http://example.com/> .

ns1:foo ns1:bar "Zm9v"^^xsd:base64Binary,
```

I can work around by using `rdflib.term.bind` to handle this datatype in a custom fashion, but I believe the fix is easy enough

delocalizer:
Fixes #1257

## Proposed Changes

  - add base64 encoding for `base64Binary` datatype, by analogy to hexlify/unhexlify for `hexBinary`
  - (cosmetic) remove use of redundant lambda when handling decode
 
tests pass

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -38,7 +38,6 @@ import logging
 import warnings
 import math
 
-import base64
 import xml.dom.minidom
 
 from datetime import date, time, datetime, timedelta
@@ -53,6 +52,7 @@ from isodate import (
     parse_duration,
     duration_isoformat,
 )
+from base64 import b64decode, b64encode
 from binascii import hexlify, unhexlify
 
 import rdflib
@@ -1435,6 +1435,7 @@ _XSD_DAYTIMEDURATION = URIRef(_XSD_PFX + "dayTimeDuration")
 _XSD_YEARMONTHDURATION = URIRef(_XSD_PFX + "yearMonthDuration")
 
 _OWL_RATIONAL = URIRef("http://www.w3.org/2002/07/owl#rational")
+_XSD_B64BINARY = URIRef(_XSD_PFX + "base64Binary")
 _XSD_HEXBINARY = URIRef(_XSD_PFX + "hexBinary")
 # TODO: gYearMonth, gYear, gMonthDay, gDay, gMonth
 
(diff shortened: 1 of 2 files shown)


white-gecko:
This is a case I have not encountered so far. Could you please provide a test that shows what was failing before and that it is fixed now?

white-gecko:
Will there be any interference with https://github.com/RDFLib/rdflib/pull/1222 ?

delocalizer:
The test is by analogy with the existing [test_hex_binary](https://github.com/RDFLib/rdflib/blob/master/test/test_hex_binary.py)  (minus the integer stuff that relies on hex-specific features of `format` builtin and `int` constructor)
it fails on master but passes with this PR

delocalizer:
> Will there be any interference with #1222 ?

I don't believe so, the issue is not with the serialization _per se_, rather that encoding was not defined for `xsd:base64Binary` type so the decoded value was being used as the lexical. Now I think about it, a more descriptive summary of this PR might be "make handling of `xsd:base64Binary` type consistent with the existing handling of `xsd:hexBinary`."

nicholascar (approved):

aucampia:
@white-gecko @ashleysommer this PR is still pending on your reviews

aucampia (approved):

nicholascar:
This PR has had a couple of reviews, passes all tests and introduces more tests, so I'm merging. It's pretty low-risk PR too (just an updated Literal type).

PR #1263, opened 2021-02-26
Speedup with Cython and Turtle parsing.

rchateauneu:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -1358,16 +1358,16 @@ class SinkParser:
 
             i += 1
             lastslash = False
-            # start = i # TODO first char .
+            start = i
             ln = ""
             while i < len_argstr:
                 c = argstr[i]
-                if not lastslash and c == "\\":
+                if c == "\\" and not lastslash:  # Very rare.
                     lastslash = True
-                    i += 1
-
-                elif lastslash or c not in allowedChars:
-
+                    if start < i:
+                        ln += argstr[start:i]
+                    start = i + 1
+                elif c not in allowedChars or lastslash: # Most common case is "a-zA-Z"
                     if lastslash:
                         if c not in escapeChars:
                             raise BadSyntax(
@@ -1377,7 +1377,7 @@ class SinkParser:
                                 i,
                                 "illegal escape " + c,
                             )
-                    elif c == "%":
+                    elif c == "%": # Very rare.
                         if (
                             argstr[i + 1] not in hexChars
                             or argstr[i + 2] not in hexChars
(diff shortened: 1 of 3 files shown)


rdflib/plugins/parsers/notation3.py
```diff
@@ -1570,8 +1585,8 @@ def strconst(self, argstr, i, delim):
             if ch == delim1:
                 j = i
                 continue
-            elif ch in ('"', "'") and ch != delim1:
-                ustr = ustr + ch
+            elif ch in "\"'" and ch != delim1:
```
FlorianLudwig, line 1588: To me the original looks more readable and - at least in cpython - it is actually faster. Didn't test cython though.

FlorianLudwig:
Overall this looks like a good MR to me. A lot of cleanup and cython :+1: 

I am curious why travis fails on powerpc - I cannot see any errors in the logs. Is it just me being blind?

FlorianLudwig:
Maybe it might be a good idea to split this up into two merge requests, one for the cleanup of the python code and one for setting up cython.

rchateauneu:
> 
> 
> It might be a good idea to split this up into two merge requests, one for the cleanup of the python code and one for setting up cython.

Yes, why not. I am discovering rdflib code, so sorry for these errors.

Please do not hesitate to separately take any change of think is worth.

rchateauneu:
> 
> 
> Overall this looks like a good MR to me. A lot of cleanup and cython +1
> 
> I am curious why travis fails on powerpc - I cannot see any errors in the logs. Is it just me being blind?

And, I restored setup.py to what it was, and I get the same error message on ppc only:
/plugins/serializers/nt.py:28: UserWarning: NTSerializer does not use custom encoding.

  warnings.warn("NTSerializer does not use custom encoding.")

/home/travis/build/rchateauneu/rdflib/rdflib/plugins/serializers/nquads.py:24: UserWarning: NQuadsSerializer does not use custom encoding.

  warnings.warn("NQuadsSerializer does not use custom encoding.")

.....

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

rdflib/plugins/parsers/notation3.py
rchateauneu: Indeed the effect of replacing "x = x + y" by "x += y" is very small. However, the other changes have a significant effect on my platform.

FlorianLudwig:
@ddeka2910 can you help here?

rchateauneu:
Thanks Ludwig. I have recreated a new clone from scratch, but the problem still happens.

white-gecko:
I have converted this one to a draft as the turtle parsing part is covered in #1266 and this one should be changes to only cover the cython improvements.

white-gecko:
That would be great

Issue #1261, opened 2021-02-25
why the loading of files is so slow? 


m1ci:
Hi,
I am loading a 165 MB file and the loading process takes ages. I would expect to be loaded in a matter of seconds.
Am I doing smth wrong or this is a normal?
Here is my code which is very simple:

```
import rdflib
g = rdflib.Graph()
g.load('data.ttl', format="turtle");
```

white-gecko:
In my setup it took 4 minutes

```
time python import.py
python import.py  235,80s user 3,08s system 99% cpu 4:00,02 total
```

I think our turtle parser is not very good.

To make some testing easier I've changed the code to:

```
import sys
import rdflib
g = rdflib.Graph()
g.parse(sys.argv[1], format=sys.argv[2])
```

Also I converted the file with rapper (raptor2-utils) to n-triples

```
$ time rapper -i turtle wordnet.ttl > wordnet.nt
rapper: Parsing URI file:///tmp/milan/wordnet.ttl with parser turtle
rapper: Serializing with serializer ntriples
rapper: Parsing returned 3103323 triples
rapper -i turtle wordnet.ttl > wordnet.nt  10,66s user 0,98s system 99% cpu 11,653 total
$ time rapper -i turtle -o turtle wordnet.ttl > wordnet_rapper.ttl
rapper: Parsing URI file:///tmp/milan/wordnet.ttl with parser turtle
rapper: Serializing with serializer turtle
rapper: Parsing returned 3103323 triples
rapper -i turtle -o turtle wordnet.ttl > wordnet_rapper.ttl  32,84s user 0,84s system 99% cpu 33,712 total
```

And compared parsing if the n-triples file with the turlte parser as well as with the n-triples parser

```
$ time python import.py wordnet.nt nt
python import.py wordnet.nt nt  140,03s user 4,35s system 97% cpu 2:28,76 total
$ time python import.py wordnet.nt ttl
python import.py wordnet.nt ttl  320,25s user 3,22s system 99% cpu 5:24,02 total
$ … (to be continued)
```



Maybe somebody wants to do this comparison with Cython #1250 .

white-gecko:
After doing some profiling, I suspect the `uri_ref2` and `skipSpace` methods of the `notation3.py`

https://github.com/RDFLib/rdflib/blob/9f0f296a3b36cbca9771cd0f0c25492a06a33acd/rdflib/plugins/parsers/notation3.py#L1155

[...]

PR #1266, opened 2021-02-28
Speedup of Turtle parse. Mostly micro-optimisations.

rchateauneu:
Contributes to fixing #1261 

## Proposed Changes

  - Minor changes to the code to optimize execution time

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -509,14 +509,13 @@ class SinkParser:
             if tok not in self.keywords:
                 return -1  # No, this has neither keywords declaration nor "@"
 
-        len_tok = len(tok)
+        i_plus_len_tok = i + len(tok)
         if (
-            argstr[i : i + len_tok] == tok
-            and (argstr[i + len_tok] in _notKeywordsChars)
-            or (colon and argstr[i + len_tok] == ":")
+            argstr[i : i_plus_len_tok] == tok
+            and (argstr[i_plus_len_tok] in _notKeywordsChars)
+            or (colon and argstr[i_plus_len_tok] == ":")
         ):
-            i += len_tok
-            return i
+            return i_plus_len_tok
         else:
             return -1
 
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -137,7 +137,7 @@ def join(here, there):
         return here + frag
 
     # join('mid:foo@example', '../foo') bzzt
-    if here[bcolonl + 1 : bcolonl + 2] != "/":
+    if here[bcolonl + 1] != "/":
         raise ValueError(
             "Base <%s> has no slash after "
             "colon - with relative '%s'." % (here, there)
@@ -503,18 +503,19 @@ class SinkParser:
         """
 
         assert tok[0] not in _notNameChars  # not for punctuation
-        if argstr[i : i + 1] == "@":
-            i = i + 1
+        if argstr[i] == "@":
+            i += 1
         else:
             if tok not in self.keywords:
                 return -1  # No, this has neither keywords declaration nor "@"
 
+        len_tok = len(tok)
         if (
-            argstr[i : i + len(tok)] == tok
-            and (argstr[i + len(tok)] in _notKeywordsChars)
-            or (colon and argstr[i + len(tok)] == ":")
+            argstr[i : i + len_tok] == tok
+            and (argstr[i + len_tok] in _notKeywordsChars)
+            or (colon and argstr[i + len_tok] == ":")
         ):
-            i = i + len(tok)
+            i += len_tok
             return i
         else:
             return -1
@@ -527,10 +528,11 @@ class SinkParser:
 
         assert tok[0] not in _notNameChars  # not for punctuation
 
-        if argstr[i : i + len(tok)].lower() == tok.lower() and (
-            argstr[i + len(tok)] in _notQNameChars
+        len_tok = len(tok)
+        if argstr[i : i + len_tok].lower() == tok.lower() and (
+            argstr[i + len_tok] in _notQNameChars
         ):
-            i = i + len(tok)
+            i += len_tok
             return i
         else:
             return -1
@@ -802,10 +804,10 @@ class SinkParser:
             res.append(("<-", self._store.newSymbol(Logic_NS + "implies")))
             return i + 2
 
-        if argstr[i : i + 1] == "=":
+        if argstr[i] == "=":
             if self.turtle:
                 self.BadSyntax(argstr, i, "Found '=' in Turtle mode")
-            if argstr[i + 1 : i + 2] == ">":
+            if argstr[i + 1] == ">":
                 res.append(("->", self._store.newSymbol(Logic_NS + "implies")))
                 return i + 2
             res.append(("->", DAML_sameAs))
@@ -845,8 +847,8 @@ class SinkParser:
         if j < 0:
             return j  # nope
 
-        while argstr[j : j + 1] in "!^":  # no spaces, must follow exactly (?)
-            ch = argstr[j : j + 1]
+        while argstr[j] in "!^":  # no spaces, must follow exactly (?)
+            ch = argstr[j]
             subj = res.pop()
             obj = self.blankNode(uri=self.here(j))
             j = self.node(argstr, j + 1, res)
@@ -880,7 +882,7 @@ class SinkParser:
         if j < 0:
             return j  # eof
         i = j
-        ch = argstr[i : i + 1]  # Quick 1-character checks first:
+        ch = argstr[i]  # Quick 1-character checks first:
 
         if ch == "[":
             bnodeID = self.here(i)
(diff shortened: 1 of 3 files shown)


rdflib/plugins/parsers/notation3.py
```diff
@@ -1570,8 +1584,8 @@ def strconst(self, argstr, i, delim):
             if ch == delim1:
                 j = i
                 continue
-            elif ch in ('"', "'") and ch != delim1:
-                ustr = ustr + ch
+            elif ch in "\"'" and ch != delim1:
```
FlorianLudwig, line 1587: The old version (check if char in tuple) vs new version (check if char in str) is faster and more readable imho

FlorianLudwig:
Thank you for splitting up the merge request!

This one looks good to me

FlorianLudwig (approved):

rdflib/plugins/parsers/notation3.py
white-gecko: `in ['"', "'"]`

rchateauneu: Hi
I pushed ['"', "'"] to my repository in "master" branch. 
Do you want a PR ?

FlorianLudwig: @white-gecko 

```
>>> timeit.timeit("'a' in {'a', 'b', 'c', 'd'}")
0.031824612990021706
>>> timeit.timeit("'a' in ['a', 'b', 'c', 'd']")
0.03463121401728131
>>> timeit.timeit("'a' in 'abcd'")
0.0507955830253195
```

The difference is not big, so using a set or a list does not make a big difference.

rchateauneu: Yes, I am surprised that a string search is not faster as I imagined that it would be implemented with strchr(). Here is what I have on Python 3.6, Windows 7, the trend is confirmed:

>>> timeit.timeit("'a' in {'a', 'b', 'c', 'd'}")
0.03046370870097803
>>> timeit.timeit("'a' in ['a', 'b', 'c', 'd']")
0.028151531781134054
>>> timeit.timeit("'a' in 'abcd'")
0.031605845198640736
>>> timeit.timeit("'a' in {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
'w', 'x', 'y', 'z'}")
0.03088098463007327
>>> timeit.timeit("'a' in ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
'w', 'x', 'y', 'z']")
0.027731048504392675
>>> timeit.timeit("'a' in 'abcdefghijklmnopqrstuvwxyz'")
0.033609475274715805

white-gecko:
on master
```
time python import.py wordnet.ttl ttl
python import.py wordnet.ttl ttl  208,93s user 2,40s system 99% cpu 3:31,53 total
```

with these fixes
```
$ time python import.py wordnet.ttl ttl
python import.py wordnet.ttl ttl  166,64s user 2,99s system 99% cpu 2:50,03 total
```

So around 20% faster in this case :+1: This is a very good result.

What concerns me, is that although the tests are passing some newly introduced lined are not covered.
Maybe somebody is willing to contribute test cases so we can be sure, that there is no degradation of the correctness.

https://coveralls.io/builds/37530950/source?filename=rdflib%2Fplugins%2Fparsers%2Fnotation3.py

rdflib/plugins/parsers/notation3.py
white-gecko: So for @rchateauneu the lists are faster, what do we do now? (Actually I did not know of this set syntax, I just new the curly brace for dicts.)

rchateauneu: @white-gecko

I did the change in my fork of this single-line:
https://github.com/rchateauneu/rdflib/compare/9652eca3e955...5c6e279820bb

It is ready for a PR if you wish so. It is a very simple change so if sets are preferred, I can do it too.

rdflib/plugins/parsers/notation3.py
```diff
@@ -1239,20 +1234,31 @@ def uri_ref2(self, argstr, i, res):
     def skipSpace(self, argstr, i):
         """Skip white space, newlines and comments.
         return -1 if EOF, else position of first non-ws character"""
+
+        # Most common case is a non-commented line starting with few spaces and tabs.
+        try:
+            while True:
+                ch = argstr[i]
+                if ch in " \t":
```
white-gecko, line 1242: We should use lists or sets (what ever we decide) also here.

rdflib/plugins/parsers/notation3.py
```diff
@@ -1239,20 +1234,31 @@ def uri_ref2(self, argstr, i, res):
...
+                    i += 1
+                    continue
+                elif ch not in "#\r\n":
```
white-gecko, line 1245: … list or set …

rdflib/plugins/parsers/notation3.py
```diff
@@ -1262,14 +1268,15 @@ def variable(self, argstr, i, res):
         if j < 0:
             return -1
 
-        if argstr[j : j + 1] != "?":
+        if argstr[j] != "?":
             return -1
-        j = j + 1
+        j += 1
         i = j
         if argstr[j] in "0123456789-":
```
white-gecko, line 1275: … list or set …

rdflib/plugins/parsers/notation3.py
```diff
@@ -1300,8 +1307,9 @@ def bareWord(self, argstr, i, res):
         if argstr[j] in "0123456789-" or argstr[j] in _notKeywordsChars:
```
white-gecko, line 1307: … list or set …

rdflib/plugins/parsers/notation3.py
```diff
@@ -1319,46 +1327,46 @@ def qname(self, argstr, i, res):
         c = argstr[i]
         if c in "0123456789-+.":
```
white-gecko, line 1328: … l or s … ;-)

rdflib/plugins/parsers/notation3.py
white-gecko: I did consequently change such constructs to lists and then to sets. (lists very much slower for me, sets slightly faster then strings)
I did this comparison, as just one timing to get the first element is not representative. If this is not:
https://gist.github.com/white-gecko/1a787e1911c479b3a13abc3e7103e7f7

FlorianLudwig: One reason for the differences we see might be do to the way timeit is used. While the `in` check might be faster with `set`, the construction of the `set` is slower than for a string.

As far as i remember (!): In the real code, python does optimize the creation when executed as part of a module - but not when executing a string with timeit since it's not a module.

So I added another example go @white-gecko gist - maybe @rchateauneu can test this one as well :)

Btw with those micro optimizations it's best to have nothing else running to get proper results...

rchateauneu: Also, one aspect to consider - which is very theoretical indeed, but still - is that Cython, when working on bytes, and with specific hints, might sometimes be able to handle Python chars and strings, just like strings.
To be completely honest, I do not have a clear answer in the general case, especially here where these strings/lists/sets are short.

I did a little test with dis, which gives a hint about the bytecode (which is not necessarily the actual execution time):
```
import dis

def func1(a):
    return x in "abc"
    
def func2(a):
    return x in ["a", "b", "c"]
    
def func3(a):
    return x in {"a", "b", "c"}
    
for func in [func1, func2, func3]:
    bytecode = dis.Bytecode(func)
    print(func.__name__)
    for instr in bytecode:
        print("    ", instr.opname)

```
The result is:
```
func1
     LOAD_GLOBAL
     LOAD_CONST
     COMPARE_OP
     RETURN_VALUE
func2
     LOAD_GLOBAL
     LOAD_CONST
     COMPARE_OP
     RETURN_VALUE
func3
     LOAD_GLOBAL
     LOAD_CONST
     COMPARE_OP
     RETURN_VALUE

```

FlorianLudwig: overall - this is a really small wins we are talking about here. The big wins come from optimizations of the algorithm. And @rchateauneu found awesome wins there (e.g. replacing a big loop with `.find` :))

While I started this discussion and it is quite interesting - I would merge any version. It doesn't actually matter in the big picture.

Readability is something to consider though.

white-gecko:
@rchateauneu why did you merge the cython setup into this pull-request? It would be better, if we can deal with these micro-optimizations and the cython setup in #1269 separately. I even think we can merge this pull-request faster, if you can reset this pull-request to eefaa374a01dbfef45da3984699b83f75962a0cb and maybe cherry-pick 3597d555c780404180b4295a1a53b742a3ca1765 to it, if it brings significant improvement. The optimizations to the store should be part of an additional pull-request.

rchateauneu:
My mistake for adding Cython. I do that and tell you.

white-gecko:
See: https://github.com/RDFLib/rdflib/pull/1272

PR #1270, opened 2021-03-03
Speedup notation3 only

rchateauneu:
cherry picked from commit 9652eca3e9557e728879c6339c752b8451ecd47f
cherry picked from commit 3597d555c780404180b4295a1a53b742a3ca1765

PR #1272, opened 2021-03-04
speedup notation3/turtle parser

### LINKED ISSUE #1262: Possible optimisation of notation3.py ?
The file rdflib/plugins/parsers/notation3.py contains very often this pattern to access a single character from a string:

ch = argstr[i : i + 1]  # Quick 1-character checks first
if argstr[j : j + 1] == "=":
ch2 = argstr[i + 1 : i + 2]

It seems to be equivalent to these lines which would be simpler and faster:

ch = argstr[i]  # Quick 1-character checks first
if argstr[j] == "=":
ch2 = argstr[i + 1]

So, is there a specific reason to select a substring instead of a single character ?

white-gecko:
Fixes #1266, #1270 (replaces).
Fixes #1262.
Related to #1261

@rchateauneu, @FlorianLudwig 

## Proposed Changes

  - some micro optimizations
  - Replace access to substring of length 1: `[i : i + 1]` with access to chars `[i]`
  - replace `while` loop with with find:
https://github.com/RDFLib/rdflib/compare/master...white-gecko:feature/speedupN3?expand=1#diff-702447d86f727e47363ffbc864d3cef01dbceee6b0652f0e992c71947324dc95L1202
https://github.com/RDFLib/rdflib/compare/master...white-gecko:feature/speedupN3?expand=1#diff-702447d86f727e47363ffbc864d3cef01dbceee6b0652f0e992c71947324dc95R1201
  - move most common case of non-commented line starting with a view spaces to the beginning of `skipSpace()`
  - replace `char in "abc"` with `char in {'a', 'b', 'c'}` constructs (set instead of string)

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -137,7 +137,7 @@ def join(here, there):
         return here + frag
 
     # join('mid:foo@example', '../foo') bzzt
-    if here[bcolonl + 1 : bcolonl + 2] != "/":
+    if here[bcolonl + 1] != "/":
         raise ValueError(
             "Base <%s> has no slash after "
             "colon - with relative '%s'." % (here, there)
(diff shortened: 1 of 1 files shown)


white-gecko:
The last thing we were looking for are new tests for the changed lines of the parser, but maybe passing the current tests is ok already, and we make a new pr for additional tests. @nicholascar @ashleysommer  what do you think?

white-gecko:
@nicholascar and @ashleysommer: any objections to merge?

nicholascar (approved):

nicholascar:
I'm happy for you to merge.

PR #1274, opened 2021-03-08
Correct behaviour of compute_qname for URNs

nemesamade:
Fixes: Allow URN qnames to be computed correcly by compute_qname
Removed colon from ALLOWED_NAME_CHARS, as having it included results in URN identifiers never being split by split_uri and thus no qname being computed by compute_qname for URNs.
I've also added some tests for correct behaviour of compute_qname for URNs and turtle serialization (where this issue surfaced initally, see the end of https://github.com/RDFLib/rdflib/issues/977)


nemesamade:
Just a note: removing the colon from ALLOWED_NAME_CHARS doesn't break any existing tests, and passes the 2 new tests so it should be an easy fix:) The colon was added originally to modify the behavior of is_cname, but was subsequently filtered out in the is_cname function (https://github.com/RDFLib/rdflib/pull/663#issuecomment-295404539) as it caused issues there too. Seems like it causes issues for split_uri as well, it just wasn't noticed due to no testing for URNs.

nicholascar (approved):
Great PR - thanks for diving in and fixing this.

white-gecko (approved):

PR #1280, opened 2021-03-14
Allow parentheses in uri

fredrik-aschehoug:
Fixes: Allow URI's containing parentheses to be split correctly.

I could not find any documentation stating that parentheses are not allowed in the name part of the URI. I noticed that in my serialized graphs, URI's containing parentheses did not serialize properly. Yago3 for example uses parentheses extensively. 

Please excuse me if my assumption that parentheses are allowed is wrong.


nicholascar:
Yes, "(" & ")" are indeed allowed, they are `sub-delims`, that is used for delimitation in some places within the URI but not everywhere. See the URI syntax https://tools.ietf.org/html/rfc3986#section-3.3:

```
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

Since the theory checks out, the PR passes all tests and you've supplied new tests, I'm happy to approve this.

nicholascar (approved):

PR #1281, opened 2021-03-18
Issue rdflib jsonld 85 - link header support in URLInputSource

datadavev:
Fixes specifically rdflib-jsonld-85 by adding general support for following link headers in URLInputSource.

## Proposed Changes

  - Added method parse_link_header
  - Follow link header when media type and optionally a provided profile match
  - Added test cases


nicholascar:
@datadavev thanks for this PR. I recall the issue you are addressing now. I'll just need to think about this for a day or two and chat with another person (see below) about this before approving. Just want to ensure the handling doesn't get in the way of other Content Negotiation stuff.

@rob-metalinkage can you take a look at this? This may solve an issue for JSON-LD but create issues for other RDF formats. We need to discuss before accepting.

rob-metalinkage:
A quick explanation of when jsonld should be following a link header - on parsing inputs? 

My only concern is that the parser should work properly if the JSON-LD uses contexts - and should support serialisation respecting use of contexts rather than brute-force turning everything into full URIs - JSON-LD should support serialisation to specific schemas with contexts.

PR #1288, opened 2021-03-30
pathname2url removal

tgbugs:
Use pathlib's as_uri function instead of pathname2url.


tgbugs:
I also switched collections to have only lf line endings instead of crlf.

nicholascar:
@tgbugs there's a line number difference of 1 in your changed `container.py`. Can you identify the change?

PR #1289, opened 2021-04-10
dataset examples file

nicholascar:
A short examples file showing how to use Dataset. We just don't have one!

### DIFF
--- a/examples/datasets.py
+++ b/examples/datasets.py
@@ -1,0 +1,146 @@
+"""
+An RDFLib Dataset is a slight extension to ConjunctiveGraph: it uses simpler terminology
+and has a few additional convenience method extensions, for example add() can be used to
+add quads directly to a specific Graph within the Dataset.
+
+This example file shows how to decalre a Dataset, add content to it, serialise it, query it
+and remove things from it.
+"""
+
+from rdflib import Dataset, URIRef, Literal, Namespace
+
+#
+#   Create & Add
+#
+
+# Create an empty Dataset
+d = Dataset()
+# Add a namespace prefix to it, just like for Graph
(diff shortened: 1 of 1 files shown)


white-gecko (changes requested):
It currently fails with

```
  File "datasets.py", line 67, in <module>
    print(d.serialize(format="trig").decode().strip())
AttributeError: 'str' object has no attribute 'decode'
```

as the serialize methods return strings in our current master branch. So I think the example should work on the current master branch.

white-gecko:
So far I have only used `ConjunctiveGraph` not `Dataset` why do we have both? This might confuse people to decide which one to choose.

nicholascar:
Yes, `Dataset` is a very small enhancement to `ConjunctiveGraph` that is, I think, confusing. I wasn't involved in RDFlib years ago when it was made but I propose de-duplicating CJ & Dataset. I'm not sure which class people would prefer to keep. I think this might be subject to a poll.

white-gecko:
If Dataset is an enhancement of ConunctiveGraph, then we should make both classes do the same, and make it just a naming decision. If we cam to a conclusion which name to take, we throw a warning, when the other one is used, telling that it is the same as the one we chose and telling that the other one is deprecated and will be removed some day.

nicholascar:
@white-gecko @ashleysommer: You can see in PR #1296 that I'm proceeding through all the docs. If you can approve this example, I'll then tackle the comment above about "we should make both classes do the same". I think there are only a couple of differences between the two classes but I'll study this properly in the other PR.

nicholascar:
Merging as the test failure is the false "assign address" one

PR #1296, opened 2021-04-24
Complete clean up of docs for 6.0.0

nicholascar:
I've walked through all .rst files in `docs` and ensured all examples are working . I've updated some incidental elements in the rest of the code base discovered via docco updates, e.g. `Dataset` & `ConjunctiveGraph` `parse()` format keyword default to match `Graph` (now `None`, not "xml").


nicholascar:
@rchateauneu @aucampia  could either of you assist with the test that is failing here? I can't reproduce the error running the test locally and can't understand why the error is occurring at all

nicholascar:
Solved by forcing 0-padding for weird gYear!

PR #1305, opened 2021-05-10
Fix transitive_objects/subjects docstrings and signatures

### LINKED ISSUE #1303: transitive methods typo
While trying to understand some puzzling output from `graph.transitive_subjects()` I noted that the docstrings for graph.transitive_subjects and graph.transitive_objects (starting at graph.py line 882) are identical. I think perhaps this is not what was intended?

The puzzling output in question is as follows:
```
>>> pprint(list(cg.transitive_subjects(RDF.type, pome.Person)))
[rdflib.term.URIRef('http://prosopOnto.medieval.england/2006/04/pome#Person'),
 rdflib.term.URIRef('http://example.com/thisgraph#Hugh_Despenser'),
 rdflib.term.URIRef('http://example.com/thisgraph#Audley_Henry_de'),
 rdflib.term.URIRef('http://example.com/thisgraph#Thomas_earl_of_Warwick_d_1242'),
.
.
. etc.
]
```

The URI.ref for `pome.Person` is listed first even though it does not appear as a subject. Is this the intended behavior?

aucampia:
Fixes #1303

## Proposed Changes

  - Fix signatures of both methods to use predicate.
  - Fix incorrect docstring for transitive_subjects

This is a breaking change, because signature of a user facing method was changed. (breaks in case of using named attributes.)

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -879,11 +879,11 @@ class Graph(Node):
             for rt_2 in self.transitiveClosure(func, rt, seen):
                 yield rt_2
 
-    def transitive_objects(self, subject, property, remember=None):
-        """Transitively generate objects for the ``property`` relationship
+    def transitive_objects(self, subject, predicate, remember=None):
+        """Transitively generate objects for the ``predicate`` relationship
 
         Generated objects belong to the depth first transitive closure of the
-        ``property`` relationship starting at ``subject``.
+        ``predicate`` relationship starting at ``subject``.
         """
         if remember is None:
             remember = {}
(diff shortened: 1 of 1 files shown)


aucampia:
@white-gecko hope this is what you had in mind [here](https://github.com/RDFLib/rdflib/issues/1303#issuecomment-836612019)

white-gecko (approved):
Thank you.

nicholascar (approved):

nicholascar:
Given that test pass there must be none that actually test the `transitive_objects` or `transitive_subjects` functions. Pity...

aucampia:
> Given that test pass there must be none that actually test the `transitive_objects` or `transitive_subjects` functions. Pity...

Yeah, I looked and I could not find any, will try add some when I have time.

PR #1310, opened 2021-05-13
Don't sort variables in TXTResultSerializer

### LINKED ISSUE #1249: TXTResultSerializer should not sort the columns alphabetically
The order of the fields is defined in the SPARQL query provided by the user. The library should not change this order. The user has defined the order of the data in the query, and the library has no need to change that order.

https://github.com/RDFLib/rdflib/blob/d78f0a40d73903b1ca364af6fd782edb61a29ce0/rdflib/plugins/sparql/results/txtresults.py#L46

aucampia:
Fixes #1249

## Proposed Changes

  - Eliminate sorting of columns in TXTResultSerializer

### DIFF
--- a/rdflib/plugins/sparql/results/txtresults.py
+++ b/rdflib/plugins/sparql/results/txtresults.py
@@ -43,7 +43,7 @@ class TXTResultSerializer(ResultSerializer):
             return "(no results)\n"
         else:
 
-            keys = sorted(self.result.vars)
+            keys = self.result.vars
             maxlen = [0] * len(keys)
             b = [
                 [_termString(r[k], namespace_manager) for k in keys]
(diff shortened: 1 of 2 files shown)


white-gecko:
Does this introduce any speedup or something like that?

aucampia:
> Does this introduce any speedup or something like that?

I did not test it, I presume it would, but I don't think it will be significant. This sort happens once per query, not per row.

There is not really a production or high volume use case for txtresults as it is not standardized and using csv, tsv, json, xml would be better for such cases. 

It is more an issue of correctness, as when I run a select and ouput it as txt it would be best if the column order is the same as the order I specified in my select (same as it would be for csv and tsv as far as I know).

white-gecko (approved):

nicholascar (approved):

PR #1315, opened 2021-05-15
Add special handling for gYear and gYearMonth

### LINKED ISSUE #747: xsd:gYear and xsd:gYearMonth error
although support for xsd:gYear was added in #636 it appears to interpret things improperly:
(schema details here: https://www.w3.org/TR/xmlschema-2/#gYear)

```python
Literal('2002', datatype='http://www.w3.org/2001/XMLSchema#gYear')
```
returns:
```python
rdflib.term.Literal('2002-01-01', datatype=rdflib.term.URIRef('http://www.w3.org/2001/XMLSchema#gYear'))
```
and
```python
Literal('2002-01', datatype='http://www.w3.org/2001/XMLSchema#gYearMonth')
```
returns:
```python
rdflib.term.Literal('2002-01-01', datatype=rdflib.term.URIRef('http://www.w3.org/2001/XMLSchema#gYearMonth'))
```

### LINKED ISSUE #806: problem with parsing "<http://www.w3.org/2001/XMLSchema#gYear>"
Given input triple `<http://dbpedia.org/resource/Australian_Labor_Party> <http://dbpedia.org/ontology/formationYear> "1891"^^<http://www.w3.org/2001/XMLSchema#gYear> .`
After using the parser and print the object
- g = Graph()
- g.parse('./sample.nt',format='nt')
- for s,p,o in g:
- print o.n3()

it became `"1891-01-01"^^<http://www.w3.org/2001/XMLSchema#gYear> `, and this will cause trouble for some sparql sever.
Please help to fix this.

### LINKED ISSUE #1228: Turtle de/serialize incorrectly interprets gYear
In 5.0.0, when a `gYear` is parsed and then re-serialised, from any format to any other format, RDFlib adds "-01-01" to the literal so, for example, `"1982"^^xsd:gYear` goes to `"1982-01-01"^^xsd:gYear` which is invalid.

Testing code:

```python
from rdflib import Graph

data =  """
        PREFIX sdo: <https://schema.org/>
        PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

        <x:> sdo:startDate "1982"^^xsd:gYear .
        """

g = Graph().parse(data=data, format="ttl")
assert "1982-01-01" not in g.serialize(format="ttl").decode()


data2 = """<?xml version="1.0" encoding="UTF-8"?>
<rdf:RDF
   xmlns:sdo="https://schema.org/"
>
  <rdf:Description rdf:about="x:">
    <sdo:startDate rdf:datatype="http://www.w3.org/2001/XMLSchema#gYear">1982</sdo:startDate>
  </rdf:Description>
</rdf:RDF>"""

g2 = Graph().parse(data=data2, format="xml")
assert "1982-01-01" not in g2.serialize(format="xml").decode()
```

aucampia:
This should solve various issues that occur with hanlding of xsd:gYear
and xsd:gYearMonth values.

Fixes #1228
Fixes #806
Fixes #747

## Proposed Changes

  - Add specific handling for `xsd:gYear` and `xsd:gYearMonth` in `_SpecificPythonToXSDRules`
  - Add tests for some known issues that remain after this fix

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -1437,6 +1437,8 @@ _XSD_YEARMONTHDURATION = URIRef(_XSD_PFX + "yearMonthDuration")
 _OWL_RATIONAL = URIRef("http://www.w3.org/2002/07/owl#rational")
 _XSD_B64BINARY = URIRef(_XSD_PFX + "base64Binary")
 _XSD_HEXBINARY = URIRef(_XSD_PFX + "hexBinary")
+_XSD_GYEAR = URIRef(_XSD_PFX + "gYear")
+_XSD_GYEARMONTH = URIRef(_XSD_PFX + "gYearMonth")
 # TODO: gYearMonth, gYear, gMonthDay, gDay, gMonth
 
 _NUMERIC_LITERAL_TYPES = (
@@ -1558,6 +1560,8 @@ _GenericPythonToXSDRules = [
 ]
 
 _SpecificPythonToXSDRules = [
+    ((date, _XSD_GYEAR), lambda val: val.strftime("%04Y")),
+    ((date, _XSD_GYEARMONTH), lambda val: val.strftime("%04Y-%02m")),
     ((str, _XSD_HEXBINARY), hexlify),
     ((bytes, _XSD_HEXBINARY), hexlify),
     ((str, _XSD_B64BINARY), b64encode),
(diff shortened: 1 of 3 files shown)


nicholascar (approved):
Looks good and addresses some annoying issues. Thanks again @aucampia!

PR #1319, opened 2021-05-17
Speedup __ctx_to_str.

rchateauneu:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/rdflib/plugins/stores/memory.py
+++ b/rdflib/plugins/stores/memory.py
@@ -509,15 +509,13 @@ class Memory(Store):
             return None
         try:
             # ctx could be a graph. In that case, use its identifier
-            ctx_str = "{}:{}".format(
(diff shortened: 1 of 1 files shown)


aucampia (approved):
Looks good to me. Would be interesting to know what the speedup is here, but good change even if it provided no speedup.

rchateauneu:
The speedup is between 0.5% and 1% when loading "orkg.nt"
    g = Graph()
    g.parse("orkg.nt", format="nt")
    
    
mem_org
mem

The profiles are obtained with:
```
py -3.6 -m cProfile -o mem.pstats import_orgk.py
py -3.6 gprof2dot.py -f pstats mem.pstats -o mem.dot
dot -Tpng -o mem.png mem.dot
```

FYI, using "%s:%s" is a bit faster.

FlorianLudwig (approved):
even without any speedup, code is cleaner, LGTM :)

PR #1321, opened 2021-05-18
Speedup Literal.__hash__ and Literal.__eq__ by accessing directly _da…

rchateauneu:
Literal.__hash__ and Literal.__eq__ use self.datatype and self. language properties. It is faster to directly use the private members Literal._dataype and Literal.__language.

Test importing orkg.nt

Before:
image

After:
  -
image

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -969,6 +969,7 @@ class Literal(Identifier):
         """
         # don't use super()... for efficiency reasons, see Identifier.__hash__
         res = str.__hash__(self)
+        # Directly accessing the member is faster than the property.
         if self._language:
             res ^= hash(self._language.lower())
         if self._datatype:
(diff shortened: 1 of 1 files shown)


rchateauneu:
It failed only Python 3.7, but 3.6 and 3.8 worked OK. This is strange because the changes are very simple.

The failed test is:
```
ERROR: test_query_with_added_rdf_prolog (test.test_sparqlstore.SPARQLStoreDBPediaTestCase)
...
  File "/usr/local/lib/python3.7/socket.py", line 716, in create_connection
    sock.connect(sa)
OSError: [Errno 99] Cannot assign requested address
```

I tend to think that it is a transient error related to a temporary network problem.

nicholascar (approved):
Don't worry about the "cannot assign address" test failure you're noticing: that's spurious and won't prevent this PR going through.

PR #1331, opened 2021-05-30
Replace incomplete example in intro_to_sparql.rst

tombaker:
Fixes incomplete example in intro_to_sparql.rst

TimBL's FOAF page no longer has any `foaf:knows` triples, so a beginner will not be able to reproduce the results obtained in the example, as is. The query string is also missing a PREFIX binding for foaf:.

## Proposed Changes

  - As TimBL's FOAF page no longer has any `foaf:knows` triples, uses Danbri's FOAF page, which does.
  - Minor rewordings in third paragraph, for clarity.
  - Assigns query string to variable so that it can be used in a second example (below).
  - Minor rewording in paragraph starting "The results are tuples...".
  - Provides second, short example, illustrating the use of `initNs`
  - Uses f-string illustrating different ways to access result values.

### DIFF
--- a/docs/intro_to_sparql.rst
+++ b/docs/intro_to_sparql.rst
@@ -17,56 +17,62 @@ Queries can be evaluated against a graph with the
 :meth:`rdflib.graph.Graph.update`.
 
 The query method returns a :class:`rdflib.query.Result` instance. For
-SELECT queries, iterating over this return
+SELECT queries, iterating over this returns
 :class:`rdflib.query.ResultRow` instances, each containing a set of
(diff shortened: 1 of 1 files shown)


nicholascar:
Hi @tombaker thnks very much for the updated docco! I am preparing a rather large total doc update in readiness for a 6.0.0 release of RDFlib soon, but I hand't picked up on this particular issue so this is very much appreciated.

tombaker:
@nicholascar Thank you for letting me know that my modest contributions are appreciated :-) If I come across other places where small improvements might be made, will submit them for consideration.

PR #1337, opened 2021-06-09
attempt at adding coveralls support to drone runs

ashleysommer:
added coveralls step to drone
removed coverage collection from python3.6 and python3.8, run it only in python3.7


nicholascar:
Trusting you on this one @ashleysommer!

nicholascar (approved):

PR #1339, opened 2021-06-17
Require setuptools, rdflib/plugins/sparql/__init__.py and rdflib/plugin.py import pkg_resources

hrnciar:
## Proposed Changes

  - list setuptools as a runtime dependency

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import re
 from setuptools import setup, find_packages
 
 kwargs = {}
-kwargs["install_requires"] = ["isodate", "pyparsing"]
+kwargs["install_requires"] = ["isodate", "pyparsing", "setuptools"]
 kwargs["tests_require"] = [
     "html5lib",
     "networkx",


aucampia:
Can you clarify what problem this is supposed to fix?

hrnciar:
In Fedora, we package RPM package for rdflib. I am not the maintainer of that package, but the maintainer of Python in Fedora.

We used to automatically generate a runtime dependency on setuptools for all RPM packages with Python projects with console_scripts entrypoints. Reason for this was that it was necessary for the way how was entrypoint script generated. Nowadays the entrypoint script looks differently. Particularly, they do not require setuptools to be installed on runtime in order to execute (at least on Python 3.8+ -- because it has importlib.metadata).

Hence we are about to stop generating the automatic runtime dependency on setuptools for packages with console_scripts entrypoints. We went through all the affected packages in Fedora and identified some packages that will be negatively impacted in Fedora because when we stop adding the requirement automatically, they will have missing dependencies, because they use setuptools (pkg_resources) on runtime, but don't specify a dependency.

As long as you import from pkg_resources in the actual code, the project requires setuptools on runtime and the requirement needs to be specified, otherwise it might be missing.

PR #1346, opened 2021-06-27
Add @expectedFailure unit tests for #1294 and type annotations for compare.py

aucampia:
This is related to #1294

## Proposed Changes

  - Add type annotations for compare.py
  - Add @expectedFailure unit tests for #1294

### DIFF
--- a/rdflib/compare.py
+++ b/rdflib/compare.py
@@ -93,11 +93,25 @@ __all__ = [
 ]
 
 from rdflib.graph import Graph, ConjunctiveGraph, ReadOnlyGraphAggregate
-from rdflib.term import BNode, Node
+from rdflib.term import BNode, Node, URIRef
 from hashlib import sha256
 
 from datetime import datetime
 from collections import defaultdict
+from typing import (
+    Set,
+    Dict,
+    TYPE_CHECKING,
+    Union,
+    List,
+    Tuple,
+    Callable,
+    Optional,
+    Iterator,
+)
+
+if TYPE_CHECKING:
+    from _hashlib import HASH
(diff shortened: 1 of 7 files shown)


nicholascar:
@rob-metalinkage you'd better look this over

rob-metalinkage (commented):
Good to have some tests - I think it would be better to characterise the tests around the expected behaviour in a systematic way that supports regression testing - rather than tests per issue, but thats a much bigger piece of work.  I'm not familiar with the @expectedFailure and how it fits into planning to _fix_ the problem - from what I read its useful to catch unexpected fixes when the code is changing.

aucampia:
@rob-metalinkage I will update the PR to make the tests more descriptive. The idea of adding `@expectedFailure` test to the code base is to establish what behaviour should be there, even if the behaviour is not there currently, it should then make it easier for someone who is changing this or fixing this to understand the expectations of the code.

nicholascar:
@aucampia can you finish of this PR shortly? If so, we can merge it into the 6.0.0 candidate

aucampia:
@rob-metalinkage I added some better explanations for the tests, and tests names are also more descriptive and not issue specific.

Again this does not fix the problem, but it sets constraints on what a fix for this issue should do. The hope is that it makes it easier for whoever looks at this later.

I may look at this again in the future myself, but this at least provides a very minimal test that should pass which makes the issue easier to pinpoint, and together with type annotations makes it easier for me to see what the intent was of the original author.

nicholascar (approved):
Thanks for the swift work @aucampia !

PR #1347, opened 2021-06-27
Replace Sleepycat with BerkeleyDB

nicholascar:
Replaces `bsddb3` module with `berkeleydb`, as recommended for Python 3.6+ and replaces all references to 'Sleepycat' with 'BerkeleyDB'. Updates store name, documentation and examples. Also provides tests.


nicholascar:
I'll wait until the other PRs improving the address assigning issue in tests are merged before re-submitting this to test so hopefully it passes everything then.

white-gecko:
Sorry, for coming in late. But I'm unsure, if it is a good idea to change old entries in the changelog (e.g. 2011/03/17). Why do you think it is good to replace sleepycat with berkeleydb in the old entries? I think an entry, that states that sleepycat was renamed to berkeleydb in the latest changelog entry would do the trick.

nicholascar:
I think you’re right: I probably accidentally renamed Sleepycat here when I did a find and replace. I’ll name it back

PR #1352, opened 2021-07-07
Improve running tests locally

edmondchuc:
## Proposed Changes

  - Add running tests instructions to README
  - Simpler on-boarding experience for contributors who want to contribute and run tests locally with the options to run the tests on the host system or in a Docker test running environment.
  - Fixes `OSError: [Errno 49] Can't assign requested address` on some OSs such as macOS.
  - Removes the requirement to have BerkeleyDB installed on the host system (problematic for macOS and Windows users).


nicholascar (approved):
Good additions @edmondchuc. There has been a lot of work for the "cannot assign address" using Mock SPARQL endpoints, instead of DBPedia and in auto-allocating free local IP addresses, but I'm not sure if they are all merged yet. I don't think this container work will get in the way of that so I think this is good to go.

ashleysommer (approved):

aucampia:
> Good additions @edmondchuc. There has been a lot of work for the "cannot assign address" using Mock SPARQL endpoints, instead of DBPedia and in auto-allocating free local IP addresses, but I'm not sure if they are all merged yet. I don't think this container work will get in the way of that so I think this is good to go.

Please do ping me if you get any more spurious failures from tests so I can have a look, as far as I know they should be gone in master now and in forks that are up to date with master.

PR #1366, opened 2021-07-19
Fixes 1190 - pin major version of pyparsing

### LINKED ISSUE #1190: SPARQL engine breaks on import with next major pyparsing library
Hi all,

I found a looming incompatibility with a dependency that is currently in alpha status for its next major release.  If this compatibility issue is not addressed, the SPARQL engine in `rdflib` will not import.

`rdflib` currently imports `pyparsing` at version `2.4.7`.  In a few inconsistently-reproducing cases, I'd found that adding this line to a Python script would cause an import failure: `import rdflib.plugins.sparql`.  Stack trace:

```bash
source venv_broken/bin/activate \
  && python3 -c 'import pyparsing ; print(pyparsing.__version__)'
3.0.0a2
source venv_broken/bin/activate \
  && python3 -c 'import rdflib.plugins.sparql'
RDFLib Version: 5.0.0
  File "<string>", line 1, in <module>
  File "/home/ajn/local/src/test/venv_broken/lib/python3.6/site-packages/rdflib/plugins/sparql/__init__.py", line 33, in <module>
    from . import parser
  File "/home/ajn/local/src/test/venv_broken/lib/python3.6/site-packages/rdflib/plugins/sparql/parser.py", line 184, in <module>
    Param('prefix', PN_PREFIX)) + Suppress(':').leaveWhitespace()
  File "/home/ajn/local/src/test/venv_broken/lib/python3.6/site-packages/rdflib/plugins/sparql/parserutils.py", line 114, in __init__
    self.name = name
AttributeError: can't set attribute
```
[Here](https://github.com/RDFLib/rdflib/blob/dde9db804d30fe0ab0c3291c105093ed691b0ef4/rdflib/plugins/sparql/parserutils.py#L114) is that line in today's `master` branch for context.

*Discovery background:*

I'm still not sure how I've been triggering `pip` to download an alpha release.  My best guess at the moment comes from this only happening when I run `python setup.py install` in a project's source tree to install the project into a virtual environment for some tests.  A bit maddeningly, this has happened in my Mac environment running Python 3.8 from MacPorts; an Ubuntu 18.04 environment running Python 3.6; but *not* an Ubuntu 20.04 environment running Python 3.8.

*Minimal steps to reproduce:*
1. Create virtual environment.
2. Install packages with this requirements file:
```
pyparsing == 3.0.0a2
rdflib
requests
```
3. Activate environment, and in some script import `rdflib.plugins.sparql`.

EDITED TO ADD: I'd accidentally truncated the last line of the stack trace reporting the `AttributeError`.

cmungall:
Fixes #1190 -- currently rdflib sparql engine breaks with pyparsing 3b

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 html5lib
 isodate
-pyparsing
+pyparsing==2.*


ajnelson-nist:
I oppose this patch, with rationale noted in [this comment](https://github.com/RDFLib/rdflib/issues/1190#issuecomment-882764171).

PR #1375, opened 2021-07-23
fixes #1374 adding RDF.li property to its correspoding namespace

### LINKED ISSUE #1374: RDF.li not in namespace (since 6.0)
After updating to the newest version of rdflib I started to have the following error whenever I use RDF.li

`AttributeError: term 'li' not in namespace 'http://www.w3.org/1999/02/22-rdf-syntax-ns#'`

I see that is not defined in 22-rdf-syntax-ns, but is part of RDF.

> There is an rdf:li special property element that is equivalent to rdf:_1, rdf:_2 in order, explained in detail in section 7.4. The mapping to the container membership properties is always done in the order that the rdf:li special property elements appear in XML — the document order is significant. The equivalent RDF/XML to Example 17 written in this form is shown in Example 18.

https://www.w3.org/TR/rdf-syntax-grammar/#section-Syntax-list-elements

> Note that RDF/XML uses li as a convenience element to avoid having to explicitly number each member. The li element assigns the properties _1, _2, and so on as necessary. The element name li was chosen to be mnemonic with the term "list item" from HTML.

https://www.w3.org/TR/1999/REC-rdf-syntax-19990222/

carlosvega:
Fixes #1374 

## Proposed Changes

  - Add li property to RDF namespace

## To be done

 - Is not on me to decide how to handle `_{positive_integer}` forms.

### DIFF
--- a/rdflib/namespace/_RDF.py
+++ b/rdflib/namespace/_RDF.py
@@ -31,6 +31,10 @@ class RDF(DefinedNamespace):
     type: URIRef  # The subject is an instance of a class.
     value: URIRef  # Idiomatic property used for structured values.
 
+    # There is an rdf:li special property element that is equivalent to rdf:_1, rdf:_2 
+    # See https://www.w3.org/TR/rdf-syntax-grammar/#section-Syntax-list-elements
+    li: URIRef
+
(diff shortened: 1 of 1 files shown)


white-gecko:
Following from #1374, I think, this can be closed.

PR #1380, opened 2021-07-28
fix query example

AlexMetsai:
Fixes #

## Proposed Changes

  -
  -
  -


nicholascar:
Thanks @AlexMetsai, I just merged an equivalent PR to this, so the issue is addressed, so I'll close this PR without merging.

PR #1382, opened 2021-07-29
Iterate over dataset return quads

### LINKED ISSUE #1351: Add __iter__ method to Dataset
#### Motivation

It is convenient to be able to iterate over all triples in a `Graph` instance without invoking `graph.triples((None, None, None))`. When working with a small instance of `Dataset`, there are occasions where one will want to observe all of the quads in a similar manner.

#### Notes

This is trivial to implement by simply having an introduced `__iter__` method invoke `self.quads((None, None, None, None))` in exactly [the same way that it is done in `Graph`](https://github.com/RDFLib/rdflib/blob/538446e08f992b7fc333151367dd216611dcd8e3/rdflib/graph.py#L500).

juanjosegzl:
Fixes #1351

## Proposed Changes

  - Add iter to Dataset to iterate over it

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union, Type, cast, overload
+from typing import Optional, Union, Type, cast, overload, Generator, Tuple
 import logging
 from warnings import warn
 import random
@@ -29,6 +29,9 @@ assert Namespace  # avoid warning
 
 logger = logging.getLogger(__name__)
 
+# Type aliases to make unpacking what's going on a little more human friendly
+ContextNode = Union[BNode, URIRef]
+DatasetQuad = Tuple[Node, URIRef, Node, Optional[ContextNode]]
 
 __doc__ = """\
 
@@ -1913,7 +1916,7 @@ class Dataset(ConjunctiveGraph):
             else:
                 yield s, p, o, c.identifier
 
-    def __iter__(self):
+    def __iter__(self) -> Generator[DatasetQuad, None, None]:
         """Iterates over all quads in the store"""
         return self.quads((None, None, None, None))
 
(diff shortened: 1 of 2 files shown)


rdflib/graph.py
```diff
@@ -1896,6 +1912,11 @@ def quads(self, quad):
             else:
                 yield s, p, o, c.identifier
 
+    def __iter__(self):
```
kvjrhall, line 1915: Note: I'm just some guy commenting and should not be considered to be someone with any authority here.

I'd recommend putting a type hint here so that the codebase gradually shifts towards full type hint / mypy support. I believe that the following would work:

```
from typing import Optional, Union

# For python < 3.9, else built-in tuple and collections.abc.Generator
from typing import Generator, Tuple

# Type aliases to make unpacking what's going on a little more human friendly
ContextNode = Union[BNode, URIRef]
DatasetQuad = Tuple[Node, URIRef, Node, Optional[ContextNode]]
...
def __iter__(self) -> Generator[DatasetQuad, None, None]:
...
```
I think that the alias that I suggested above for `DatasetQuad` may even be able to be made more specific. The usage of `URIRef` reflects the RDF specification that the properties of triples are all URI references. Analogous restrictions exist on the subjects (I believe that literals cannot be subjects, for example). It may be helpful to look into what the context field can be. I believe the context here can only be `None`, a `BNode`, or a `URIRef`, but it's been a little bit since I looked around.

nicholascar: > I'd recommend putting a type hint here

Yes please! Our (the maintainers') unofficial policy is that type hinting should be present for all new RDFlib additions.

nicholascar: > # Type aliases to make unpacking what's going on a little more human friendly

Cool!

nicholascar:
Thanks @kvjrhall: I have added your type hinting suggestions to this PR as well as a test. If it passes all tests still - it passes all tests with Py3.7 & Py3.9 on my machine - I'll merge.

nicholascar:
OK, everything passes, as expected, so merging. Thanks @juanjosegzl  and @kvjrhall

PR #1384, opened 2021-08-03
Update intro_to_sparql.rst

sinaahmadi:
One line missing in the SPARQL example where the query should be run as in `qres = g.query(knows_query)`

Fixes #

## Proposed Changes

  -
  -
  -


nicholascar:
@sinaahmadi thanks!

PR #1385, opened 2021-08-04
Fix/1369 custom resolver

### LINKED ISSUE #1369: URLInputSource can be abused to retrieve arbitrary documents if used naïvely
This is mostly related to rdflib-jsonld, but the dereferencing implementation is in rdflib, hence raising it here.

# Scenario

If a web service takes POSTed JSON-LD data, e.g. as part of a [Linked Data Notifications](https://www.w3.org/TR/ldn/) implementation, rdflib will attempt to resolve any URL in the `@context`. This can lead to:

* attackers being able to probe internal networks, by having rdflib request potential non-public URLs
* reflection attacks, if the same or slightly-different URLs are repeated multiple times in the `@context`
* resource exhaustion, as the entire remote file is loaded into memory before JSON parsing is attempted (admittedly an rdflib-jsonld issue)
* denial of service, if web or task workers are tied up waiting for extended periods for HTTP requests to complete
* attackers being able to probe the local filesystem using `file://` URLs

# Problem

rdflib provides no way to control how external references are resolved, nor a way to implement caching of external resources.

An implementor should be able to:

* add URLs to a safelist, if e.g. they only expect certain JSON-LD contexts to be used
* provide local copies of remote resources, to obviate needing to make HTTP requests
* hook in a caching mechanism

These things should either be possible directly, or there should be an obvious way to hook them in.

# Resolution

A new `Resolver` base class should be added that takes responsibility for resolving external references and returning `InputSource` instances, probably encapsulating the `create_input_source()` behaviour in a `resolve()` method. There should be a default implementation that resolves everything called e.g. `DefaultResolver`. Maybe this resolver has an instantiation parameter like `resolve_schemes=('file', 'http', 'https')` so it's easy to turn off dereferencing.

An optional `resolver` argument should be added to `Graph.parse()`, so that implementors can override the default behaviour. This is then passed down to the `Parser.parse()` plugin implementation, defaulting to an instance of `DefaultResolver` if not specified.

Finally, rdflib-jsonld can be updated to use the `resolver` instead of `create_input_source` directly.

Maybe there should also be a way to install a global default resolver.

### LINKED ISSUE #1844: URLInputSource can be abused to retrieve arbitrary documents if used naïvely
### Discussed in https://github.com/RDFLib/rdflib/discussions/1543

<div type='discussions-op-text'>

<sup>Originally posted by **alexdutton** July 20, 2021</sup>
This is mostly related to rdflib-jsonld, but the dereferencing implementation is in rdflib, hence raising it here.

# Scenario

If a web service takes POSTed JSON-LD data, e.g. as part of a [Linked Data Notifications](https://www.w3.org/TR/ldn/) implementation, rdflib will attempt to resolve any URL in the `@context`. This can lead to:

* attackers being able to probe internal networks, by having rdflib request potential non-public URLs
* reflection attacks, if the same or slightly-different URLs are repeated multiple times in the `@context`
* resource exhaustion, as the entire remote file is loaded into memory before JSON parsing is attempted (admittedly an rdflib-jsonld issue)
* denial of service, if web or task workers are tied up waiting for extended periods for HTTP requests to complete
* attackers being able to probe the local filesystem using `file://` URLs

# Problem

rdflib provides no way to control how external references are resolved, nor a way to implement caching of external resources.

An implementor should be able to:

* add URLs to a safelist, if e.g. they only expect certain JSON-LD contexts to be used
* provide local copies of remote resources, to obviate needing to make HTTP requests
* hook in a caching mechanism

These things should either be possible directly, or there should be an obvious way to hook them in.

# Resolution

A new `Resolver` base class should be added that takes responsibility for resolving external references and returning `InputSource` instances, probably encapsulating the `create_input_source()` behaviour in a `resolve()` method. There should be a default implementation that resolves everything called e.g. `DefaultResolver`. Maybe this resolver has an instantiation parameter like `resolve_schemes=('file', 'http', 'https')` so it's easy to turn off dereferencing.

An optional `resolver` argument should be added to `Graph.parse()`, so that implementors can override the default behaviour. This is then passed down to the `Parser.parse()` plugin implementation, defaulting to an instance of `DefaultResolver` if not specified.

Finally, rdflib-jsonld can be updated to use the `resolver` instead of `create_input_source` directly.

Maybe there should also be a way to install a global default resolver to easily implement these protections without having to track down every `Graph.parse()` call.

Happy to put together a PR if/when an approach is agreed.</div>

alexdutton:
Fixes #1369

## Proposed Changes

This adds a new Resolver framework, and updates the JSON-LD parser to use it.

As a result, implementors have much more control over which URLs referenced in JSON-LD documents will be resolved.

This PR implements a default-deny approach to URL resolution, which will need explaining in release notes and documentation. Implementors can revert to the old behaviour via environment variable if they are happy with the risks.

This also provides a base for future extensibility, e.g. caching or using locally-held copies of external resources.

No tests or documentation yet, but all existing tests pass when using the new `PermissiveResolver`.

### DIFF
--- a/rdflib/exceptions.py
+++ b/rdflib/exceptions.py
@@ -10,6 +10,8 @@ __all__ = [
     "ObjectTypeError",
     "ContextTypeError",
     "ParserError",
+    "ResolutionError",
+    "ResolutionForbiddenError",
 ]
 
 
@@ -104,12 +106,12 @@ class ResolutionError(Error):
 class ResolutionForbiddenError(Error):
     """Resolution of the given URL is not allowed."""
 
-    def __init__(self, location):
+    def __init__(self, url):
         Error.__init__(self, self.__doc__)
-        self.location = location
+        self.url = url
 
     def __str__(self):
-        return f"Resolution of '{self.location}' is not allowed."
+        return f"Resolution of '{self.url}' is not allowed."
 
 
 class UniquenessError(Error):
--- a/rdflib/exceptions.py
+++ b/rdflib/exceptions.py
@@ -87,6 +87,31 @@ class ParserError(Error):
         return self.msg
 
 
+class ResolutionError(Error):
+    """Resolution error.
+
+    This includes the case when the URI scheme is not supported.
+    """
+
+    def __init__(self, msg):
+        Error.__init__(self, msg)
+        self.msg = msg
+
+    def __str__(self):
+        return self.msg
+
+
+class ResolutionForbiddenError(Error):
+    """Resolution of the given URL is not allowed."""
+
+    def __init__(self, location):
+        Error.__init__(self, self.__doc__)
+        self.location = location
+
+    def __str__(self):
+        return f"Resolution of '{self.location}' is not allowed."
+
+
 class UniquenessError(Error):
     """A uniqueness assumption was made in the context, and that is not true"""
 
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -18,8 +18,6 @@ import sys
 from io import BytesIO, TextIOBase, TextIOWrapper, StringIO, BufferedIOBase
 
 from urllib.request import Request
-from urllib.request import url2pathname
-from urllib.parse import urljoin
 from urllib.request import urlopen
 from urllib.error import HTTPError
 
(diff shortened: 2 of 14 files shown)


nicholascar:
Hi @alexdutton I am watching this PR with interest - I think it's a great idea to have "everything" available to RDFlib users, and, if scary, off by default - so look forward to you bringing it out of draft.

Just a couple of questions:

1. Has this work of yours been motivated by the JSON-LD parser & serializer now being in RDFlib core? 
2. Do you have your own code like this in action already, just not in RDFlib?
3. Can you please add some notes to the top of the new file `rdflib/resolver.py` to explains what it's doing, perhaps something like `rdflib/compare.py` (though doesn't have to be as long!) This just helps with people and file browsing

alexdutton:
Hi @nicholascar

1. I didn't realise the JSON-LD support had been moved into core when I first realised this was an issue, but still saw the fix as being in core as that's where the resolution behaviour was.                 
2. No, I don't have similar code elsewhere. This was precipitated by needing to implement a Linked Data Notifications inbox for the COAR Notify initiative, which involves receiving JSON-LD documents to a publicly accessible HTTP endpoint.
3. Yep, I'll do that now :-)

nicholascar:
@alexdutton this PR is still in draft. Are you planning on taking it out of draft soon so we can review it formally?

ramcaat:
This is a critical vulnerability. Will this be fixed in 6.3 release? Do we have a date for it?

aucampia:
> This is a critical vulnerability. Will this be fixed in 6.3 release? Do we have a date for it?

The plan is to fix in in 6.3, but sadly I have very little time to work on this. I will be having some days off work later this month and I hope to finish it then.

PR #1386, opened 2021-08-05
Allowing Container Membership Properties in RDF namespace (#873)

### LINKED ISSUE #873: RDF namespace does not allow rdf:Seq valid terms
Hi,

According to the [RDFS recommendation](https://www.w3.org/TR/rdf-schema/#ch_collectionvocab), elements of an rdf:Seq can be ordered by using predicates of the form:

`C rdf:_nnn O`

"where `nnn` is the decimal representation of an integer greater than 0 with no leading zeros". However in rdflibish notation these do not work:

```
>>> RDF["_1"]
Exception: term '_1' not in namespace 'http://www.w3.org/1999/02/22-rdf-syntax-ns#'
>>> RDF._1
Exception: term '_1' not in namespace 'http://www.w3.org/1999/02/22-rdf-syntax-ns#'
```

My current workaround is to mint new URIRefs for these, but I guess a more elegant solution could be provided?

Thanks,
Albert

mwatts15:
- Also, deleting _RDFNamespace since it is not used in rdflib since #1074

Fixes #873


nicholascar (approved):

nicholascar:
With our new DeclaredNamespaces showing up issues, we still have a few other namespaces to fix too, [like XSD](https://github.com/RDFLib/rdflib/issues/1391).

PR #1394, opened 2021-08-18
Make graph and other methods chainable

wssbck:
## Proposed Changes

Makes certain graph methods, like `remove` or `add` chainable. This will make it possible to write more concise code involving multiple graph modifications, along the lines of:

```python
g = Graph().add(...).addN(...).add(...).remove(...)
```

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -402,8 +402,7 @@ class Graph(Node):
         Might be necessary for stores that require closing a connection to a
         database or releasing some resource.
         """
-        self.__store.close(commit_pending_transaction=commit_pending_transaction)
-        return self
+        return self.__store.close(commit_pending_transaction=commit_pending_transaction)
 
     def add(self, triple):
         """Add a triple with self as context"""
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -354,15 +354,18 @@ class Graph(Node):
     def destroy(self, configuration):
         """Destroy the store identified by `configuration` if supported"""
         self.__store.destroy(configuration)
+        return self
 
     # Transactional interfaces (optional)
     def commit(self):
         """Commits active transactions"""
         self.__store.commit()
+        return self
 
     def rollback(self):
         """Rollback active transactions"""
         self.__store.rollback()
+        return self
 
     def open(self, configuration, create=False):
         """Open the graph store
@@ -379,6 +382,7 @@ class Graph(Node):
         database or releasing some resource.
         """
         self.__store.close(commit_pending_transaction=commit_pending_transaction)
+        return self
 
     def add(self, triple):
         """Add a triple with self as context"""
@@ -387,6 +391,7 @@ class Graph(Node):
         assert isinstance(p, Node), "Predicate %s must be an rdflib term" % (p,)
         assert isinstance(o, Node), "Object %s must be an rdflib term" % (o,)
         self.__store.add((s, p, o), self, quoted=False)
+        return self
 
     def addN(self, quads):
         """Add a sequence of triple with context"""
@@ -398,6 +403,7 @@ class Graph(Node):
             and c.identifier is self.identifier
             and _assertnode(s, p, o)
         )
+        return self
 
     def remove(self, triple):
         """Remove a triple from the graph
@@ -406,6 +412,7 @@ class Graph(Node):
         from all contexts.
         """
         self.__store.remove(triple, context=self)
+        return self
 
     def triples(self, triple):
         """Generator over the triple store
@@ -622,6 +629,7 @@ class Graph(Node):
         ), "p can't be None in .set([s,p,o]), as it would remove (s, *, *)"
         self.remove((subject, predicate, None))
         self.add((subject, predicate, object_))
+        return self
 
     def subjects(self, predicate=None, object=None):
         """A generator of subjects with the given predicate and object"""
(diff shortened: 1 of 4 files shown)


nicholascar:
@wssbck this change might only need a few test changes for it to pass. I guess that, since most of the methods like `add()` have always returned `None`, this is what a bunch of tests look for but I don't suppose there's any reason why they _must_ look for such an outcome. So returning `self` will just require those tests to be changed. Looks like there's only about 8 of those test anyway, so it won't be a big deal to alter.

Are you still keen to see this PR through?

wssbck:
@nicholascar I would like to continue, yes. I got distracted by, well, life.

wssbck:
@nicholascar I added returns to more methods and adjusted all the tests. I hope I have not missed anything. The build is passing now.

nicholascar:
@wssbck thanks for updating this PR. I'm happy that it's now doing what you want it to do and that it's passing tests _however_ I want to check one thing with my co-maintainers still:

@ashleysommer @white-gecko: this PR makes `add()` and `.remove()` return `self` and I have two questions for you:

1. can you think of any down-stream projects that would be negatively affected by this?
2. is there any performance penalty incurred by this passing back? I don't mean data in `self` - it's just a reference - but does passing it back, even if it's not used, cost any extra memory or processing? I assume no but need to check

ashleysommer:
Good question regarding performance. I don't think it will change anything, but I'll do a couple of benchmarks and get a number to compare.

ashleysommer:
Ok, I created this simple benchmark (its a simple python method benchmark, not an rdflib benchmark):
https://gist.github.com/ashleysommer/86a30e4b3b356198db058a7c6b0393e1

Conclusion: zero performance difference when returning `self` rather that not returning anything.

wssbck:
@nicholascar this PR changes return values of several methods from implicit `None` to explicit object references. From a practical standpoint it shouldn't cause anyone issues unless their code, for some reason, relies on the implicit return of `None`. I really cannot come up with a reasonable scenario where it could happen, though. It would be like relying on the return value of `print` (which also returns `None`).

The practical effect is that these methods can now be chained, potentially making some downstream code more compact.
 
Since the return values are now explicit, they are easier to reason about. It is, to my experience, always better to try to explicitly return something related to (or convenient because of) the purpose of the function/method rather than rely on the default behaviour of the language.

nicholascar:
Thanks @ashleysommer, I'll add you to formally review then and we are good to merge

nicholascar (approved):

rdflib/graph.py
```diff
@@ -379,6 +403,7 @@ def close(self, commit_pending_transaction=False):
         database or releasing some resource.
         """
         self.__store.close(commit_pending_transaction=commit_pending_transaction)
+        return self
```
ashleysommer, line 406: My only comment is that graph.close() probably shouldn't return self, because it doesn't make sense to chain any other methods after close().

ashleysommer (changes requested):
Looks good

rdflib/graph.py
wssbck: I changed it to an explicit return of the last command. Thanks for spotting :-)

nicholascar:
Since @ashleysommer's comment has been addressed, I'm merging

nicholascar:
@wssbck you might like to update the rdflib documentation files with a new PR to show chained use in action

wssbck:
I will try to find time shortly, thanks for reviewing and merging!

aucampia:
I was looking at fixing type checking, since I had some issues using 6.0.0 in typed python, and noticed that serialize will now sometimes return `self`, but it will be a bit inconsistent, as sometimes it will still return `None`. I think for serialize there is actually a good case to make that it should not return self, as the return value somewhat depends on the arguments.

Just a note, I will fix the typing to correspond with what things are now, and then hopefully with type checking enabled in CI issues like this will be easier to spot.

nicholascar:
> serialize will now sometimes return self

Doesn't serialize return a string/bytes, e.g. https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1013 & https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1033?

> ...as sometimes it will still return None

As above, when would serlialize() return None? Do you perhaps mean something like add()? That might return self/None and it was the focus of this PR more than serialize(). Just trying to make sure I understand the point!

> ...I will fix the typing to correspond with what things are now...

Great, please do and I'll happily assist

aucampia:
@nicholascar `Graph.serialize()` used to return `None` when a destination is supplied (see [here](https://github.com/RDFLib/rdflib/blob/6.0.0/rdflib/graph.py#L1014-L1024) and [here](https://github.com/RDFLib/rdflib/blob/6.0.0/rdflib/graph.py#L1077-L1098))

This change makes `Graph.serialize()` return self if a destination is supplied.

However if a destination is supplied and it is a network location `serialize()` warns and returns `None` (see [here](https://github.com/RDFLib/rdflib/blob/506244c74459c264b4577e463013427cfdc50913/rdflib/graph.py#L1125-L1129)).

I am actually unsure if this make sense, maybe this should rather raise a `ValueError` or `RuntimeException` than return `None` or self. But I guess if it does not throw it makes most sense to return `self`, which is what I changed it to in #1407.

nicholascar:
@aucampia ok, thanks for that explanation. 

I think you're right: always return the same thing (`self`) and return an actual error, rather than print a warning, when actually encountering an error condition.

Would you be able to PR to update this?

aucampia:
@nicholascar I only thought that throwing maybe better than logging a warning after I wrote the code for #1407. I updated the PR now to make `Graph.serialize` throw a ValueError instead of raise an exception if the `destination` paramter is not a local file.

PR #1400, opened 2021-09-05
Revert black

nicholascar:
revert black to version agreed for the 6.0.0 release


ashleysommer:
Is this supposed to include the "add dateTimeStamp, fundamental & constraining facets, 7-prop data model" commit?

PR #1403, opened 2021-09-07
Add JSON-LD to guess_format()

ajnelson-nist:
Now that the `rdflib` package provides the functionality formerly in the
package `rdflib-jsonld`, it would be helpful for the utility function
`guess_format` to recognize file extensions indicative of JSON-LD
content.

This patch adds recognition of two file extensions, preserving syntax of
the `SUFFIX_FORMAT_MAP` variable:
* `.json`
* `.jsonld`

The latter is in recognition of the IANA media type
`application/ld+json` reporting the `.jsonld` file extension.
References:
https://www.iana.org/assignments/media-types/application/ld+json
via https://www.iana.org/assignments/media-types/media-types.xhtml
via https://w3c.github.io/json-ld-syntax/#iana-considerations

Signed-off-by: Alex Nelson <alexander.nelson@nist.gov>

## Proposed Changes

  - Add "json" and "json-ld" to recognized extensions
  - Add "json" example to docstring for `guess_format`

### DIFF
--- a/rdflib/util.py
+++ b/rdflib/util.py
@@ -364,6 +364,8 @@ SUFFIX_FORMAT_MAP = {
     "svg": "rdfa",
     "nq": "nquads",
     "trig": "trig",
+    "json": "json-ld",
+    "jsonld": "json-ld",
 }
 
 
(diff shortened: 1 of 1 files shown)


nicholascar:
Good update. Happy to merge.

Though this is simple, please next time add a test too, even if trivial? Thanks!

ajnelson-nist:
Ah, I apologize, I had delegated the test to the docstring, thinking a docstrings test would pick it up.

I think I see where to add the test now.  I'll add a follow-on PR.

PR #1407, opened 2021-09-10
Fix mypy type errors and add mypy to .drone.yml

### LINKED ISSUE #1311: Policy on type annotations
There are not many changes needed to make to get mypy to pass, and this opens the door to adding more type annotations which makes it easier for users of the library and helps prevents errors.

Are you open to a PR that adds support for mypy to the CI pipelines and makes changes needed in the codebase for it to pass? Any preferences for config and styling (i.e. `import typing as t; import typing_extensions as te;`)?

Would really make the experience of using rdflib easier IMO.

aucampia:
Fixes #1311

Add `mypy` to .drone.yml and fix type errors that come up.

Not type checking examples or tests.

Other changes:

- fix return value for `Graph.serialize` (refs #1394)
- remove default value for
  `rdflib.plugins.sparql.algebra.translateAlgebra` (refs #1322)
- add .dockerignore to reduce context size and make docker quicker
  to run.
- add .flake8 config to ignore line length as black is managing
  formatting.
- add mypy to docker-compose, makefile and tox.ini
- fix the way csv2rdf is invoked to ensure that the right code gets
  executed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.tox
+.venv
+.mypy_cache
+.git
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,8 +26,9 @@ steps:
   - pip install --default-timeout 60 -r requirements.dev.txt
   - pip install --default-timeout 60 coveralls && export HAS_COVERALLS=1
   - python setup.py install
-  - black --config black.toml --check ./rdflib | true
+  - black --config black.toml --check ./rdflib || true
   - flake8 --exit-zero rdflib
+  - mypy --show-error-context --show-error-codes rdflib
   - PYTHONWARNINGS=default nosetests --with-timer --timer-top-n 42 --with-coverage --cover-tests --cover-package=rdflib
   - coverage report --skip-covered
   - coveralls
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+# https://flake8.pycqa.org/en/latest/user/configuration.html
+[flake8]
+extend-ignore =
+    # E501: line too long
+    # Disabled so that black can control line length.
+    E501,
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ tests:
 	docker-compose -f docker-compose.tests.yml up test-runner
 	docker-compose -f docker-compose.tests.yml down
 
+.PHONY: build
 build:
 	docker-compose -f docker-compose.tests.yml build
 
@@ -14,3 +15,6 @@ reformat:
 
 check-format:
 	black --config ./black.toml --check .
+
+check-types:
+	docker-compose -f docker-compose.tests.yml up check-types
--- a/docker-compose.tests.yml
+++ b/docker-compose.tests.yml
@@ -14,3 +14,11 @@ services:
     volumes:
       - .:/rdflib
     command: ["/rdflib/run_tests_with_coverage_report.sh"]
+
+  check-types:
+    build:
+      context: .
+      dockerfile: test/Dockerfile
+    volumes:
+      - .:/rdflib
+    command: ["python", "-m", "mypy", "--show-error-context", "--show-error-codes" ,"/rdflib/rdflib"]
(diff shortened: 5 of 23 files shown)


aucampia:
@nicholascar This change will make drone fail if `mypy` fails, if you want I can change it so that errors are discarded (i.e. add `|| true`).

nicholascar (approved):
Looks sensible to me so over to @ashleysommer / @white-gecko to just check the mypy / drone implications.

aucampia:
I included the following as per discussion in #1394

> Change Graph.serialize to throw for invalid destinations
> 
> `Graph.serialize` was just logging a warning if the supplied `destination`
> was not valid, but it would be better to fail if the method does not
> perform the requested operation.
> 
> This change makes `Graph.serialize` throw if the supplied `destination`
> can't be written to.

nicholascar:
All good with the inclusions of the error raise in this PR. @ashleysommer can you please take a look?

ajnelson-nist:
I hope you'll pardon my butting in.  I've been trying `mypy` in some other projects.  I've found some cases require the addition of a `py.typed` file, with some extra setup references needed to export the `py.typed` file.  One place I did this is [here](https://github.com/dfxml-working-group/dfxml_python/pull/20/commits/9e781474434ee4635f95201c99350ca413f37e01).  I did this from reviewing [PEP 561](https://www.python.org/dev/peps/pep-0561/) and from finding it necessary from some tinkering.

Is it more appropriate for `rdflib` to have, or not to have, a `py.typed` file?

aucampia:
> I hope you'll pardon my butting in. I've been trying `mypy` in some other projects. I've found some cases require the addition of a `py.typed` file, with some extra setup references needed to export the `py.typed` file. One place I did this is [here](https://github.com/dfxml-working-group/dfxml_python/pull/20/commits/9e781474434ee4635f95201c99350ca413f37e01). I did this from reviewing [PEP 561](https://www.python.org/dev/peps/pep-0561/) and from finding it necessary from some tinkering.
> 
> Is it more appropriate for `rdflib` to have, or not to have, a `py.typed` file?

I was not aware of this, thanks for the info. I think it should have py.typed in this case then. I will try test this and see what is up, maybe make another PR to add it.

nicholascar:
From the PEP: “if a top-level package includes it, all its sub-packages MUST support type checking as well.”

So please do go ahead and add a py.typed file but just consider where it should be added. I assume to the rdflib/ folder, not the repo parent folder!

Unless test/ also needs one…

Need to get rid of nose too… Ugh, lots to do…

aucampia:
> So please do go ahead and add a py.typed file but just consider where it should be added. I assume to the rdflib/ folder, not the repo parent folder!

I added it in rdflib/ - and put it on this PR. I think it only really matters for what is distributed in wheels anyway, so even if we enforce typing in test/ it is not needed there as others should not be importing tests.


This will prevent the following error from occuring:
```
$ python3.9 -m mypy --show-error-context --show-error-codes  test_rdflib.py
test_rdflib.py:1: error: Skipping analyzing "rdflib": found module but no type hints or library stubs  [import]
test_rdflib.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

`test_rdflib.py`:
```
import rdflib

# python3.9 -m mypy test_rdflib.py

# mypy: warn-unused-configs, disallow-any-generics, disallow-subclassing-any
# mypy: no-disallow-untyped-calls, disallow-untyped-defs, disallow-incomplete-defs
# mypy: check-untyped-defs, disallow-untyped-decorators, no-implicit-optional,
# mypy: warn-redundant-casts, warn-unused-ignores, warn-return-any,
# mypy: no-implicit-reexport, strict-equality,

g = rdflib.Graph()
```

ajnelson-nist:
I'm glad my feedback was helpful!

I look forward to seeing a release with this PR.  Adding type checking to my RDF projects still requires `# type: ignore` on all of my `rdflib` imports, though I feel confident that won't be needed much longer.

PR #1418, opened 2021-09-21
Make Result.serialize work more like Graph.serialize

### LINKED ISSUE #1338: Make serialize() on a CONSTRUCT result act like normal g.serialize()
Currently (RDFlib 6.0.0a0 AKA `master`), `g.serialize()` defaults to turtle format and is decoded, however calling `serialize()` on a SPARQL `CONSTRUCT` result (i.e. on a `Result` object) returns the RDFlib 5.0.0 form of an RDF/XML, UTF-8-encoded result.

The old style `serialize()` in `Result` should be updated to match the main `Graph()` `serialize()`.

aucampia:
**DO NOT REVIEW**: I'm going to split everything off from this that I can to reduce it to just the breaking changes so we can merge it before 7.0.0.

~~**Reviewers beware**: This patch is pretty big and contains a lot of things that could be merged separately. I may at some point break off some of the changes to separate PRs.~~

This patch makes the following changes to `Result.serialize`.

* Return str by default instead of bytes.
* Use "txt" as the default tabular serialization format.
* Use "turtle" as the default graph serialization format.
* Support both typing.IO[bytes] and typing.TextIO destinations.

Corresponding changes are made to the specific serializers also.

This patch also changes how text is written to typing.IO[bytes] in
serializers to ensure that the buffer is flushed and
detatched from the TextIOWrapper once the serialization function
completes so it can be used normally afterwards.

This patch further includes a bunch of additional type hints.

Fixes #1338

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -351,15 +351,15 @@ class Graph(Node):
         self.default_union = False
 
     @property
-    def store(self):
+    def store(self) -> Store:  # read-only attr
         return self.__store
 
     @property
-    def identifier(self):
+    def identifier(self) -> Node:  # read-only attr
         return self.__identifier
 
     @property
-    def namespace_manager(self):
+    def namespace_manager(self) -> NamespaceManager:
         """
         this graph's namespace-manager
         """
@@ -368,8 +368,9 @@ class Graph(Node):
         return self.__namespace_manager
 
     @namespace_manager.setter
-    def namespace_manager(self, nm):
-        self.__namespace_manager = nm
+    def namespace_manager(self, value: NamespaceManager):
+        """this graph's namespace-manager"""
+        self.__namespace_manager = value
 
     def __repr__(self):
         return "<Graph identifier=%s (%s)>" % (self.identifier, type(self))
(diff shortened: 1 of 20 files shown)


rdflib/graph.py
```diff
@@ -1053,50 +1054,69 @@ def serialize(
...
+           The destination to serialize the graph to. This can be a path as a
+           :class:`string` or :class:`~pathlib.PurePath` object, or it can be a
+           :class:`~io.BufferedIOBase` like object. If this parameter is not
+           supplied the serialized graph will be returned.
+        :type destination: Optional[Union[str, io.BufferedIOBase, pathlib.PurePath]]
+        :param format:
+           The format that the output should be written in. This value
+           references a :class:`~rdflib.serializer.Serializer` plugin. Format
+           support can be extended with plugins, but `"xml"`, `"n3"`,
+           `"turtle"`, `"nt"`, `"pretty-xml"`, `"trix"`, `"trig"` and `"nquads"`
```
nicholascar, line 1102: Can you please add JSON-LD in here too, now that it's included?

rdflib/query.py
```diff
@@ -206,28 +207,110 @@ def parse(source=None, format=None, content_type=None, **kwargs):
...
 
-        :param destination: Path of file output or BufferedIOBase object to write the output to.
-        :param encoding: Encoding of output.
-        :param format: One of ['csv', 'json', 'txt', xml']
-        :param args:
-        :return: bytes
+            - `"csv"`: :class:`~rdflib.plugins.sparql.results.csvresults.CSVResultSerializer`
+            - `"json"`: :class:`~rdflib.plugins.sparql.results.jsonresults.JSONResultSerializer`
+            - `"txt"`: :class:`~rdflib.plugins.sparql.results.txtresults.TXTResultSerializer`
+            - `"xml"`: :class:`~rdflib.plugins.sparql.results.xmlresults.XMLResultSerializer`
```
nicholascar, line 306: Also JSON-LD too please!

test/test_serialize.py
```diff
@@ -1,43 +1,159 @@
...
+            if format == "ttl":
+                check(self.graph.serialize(fileref, encoding=encoding))
+            if encoding == sys.getdefaultencoding():
+                check(self.graph.serialize(fileref, format))
+                check(self.graph.serialize(fileref, format, None))
+                check(self.graph.serialize(fileref, format, None, None))
+
+            check(self.graph.serialize(fileref, format, encoding=encoding))
+            check(self.graph.serialize(fileref, format, None, encoding))
+
```
nicholascar, line 156: ...and maybe a JOSN-LD tests in here somewhere too

nicholascar:
@aucampia since you are doing work elsewhere, do you want to put this PR on hold or to close it perhaps?

Issue #1415, opened 2021-09-18
Submit your DefinedNamespaces


nicholascar:
As of rdflib 6.0.0, we have DefinedNamespaces which contain the elements of a namespace (ontology/vocabulary/profile etc) and make them available for type hinting and warning-based validation.

If people would like to create DefinedNamespaces for interesting/important/their favourite namespaces, we can include them in [rdflib's `rdflib.namespace` package](https://github.com/RDFLib/rdflib/tree/master/rdflib/namespace). This will make your namesapces more visible and less prone to error in use.

If you are interested, have a look at those already there and put a PR in to add yours.

dwinston:
I would be interested in a DefinedNamespace for [QUDT](https://qudt.org/). @nicholascar you mentioned on the rdflib-dev list that you are on the Tech Advisory Board of QUDT, so I presume you would be interested in this inclusion as well. Are you planning to do this soon-ish anyway? I don't have much free time to start a PR at the moment, but I can see myself attempting one at some point. It is quite a large namespace / set of namespaces, but I would find its inclusion highly useful!

dwinston:
Also, uhh...I'm seeing a default "If you're seeing this, you've successfully installed Tomcat. Congratulations!" page at qudt.org right now, so to be clear, I mean this: <https://lov.linkeddata.es/dataset/lov/vocabs/qudt>, or equivalently, the family of `qudt*.ttl` files I have stashed [here](https://github.com/polyneme/csv-ld/tree/03cb426913fe85615ff4d2469bc9a087f75d66bc/libttl).

![image](https://user-images.githubusercontent.com/1497854/133943770-0610eb4a-e3c1-47cc-9b13-24fe29ec1fa2.png)

nicholascar:
Looks like QUDT is back now! I suspect they updated their server to the latest version of TopQuadrant's EDG and there was just a momentary boo boo while they did.

> I would be interested in a DefinedNamespace for QUDT. @nicholascar you mentioned on the rdflib-dev list that you are on the Tech Advisory Board of QUDT, so I presume you would be interested in this inclusion as well.

Great, yes I and and yes I am! Silly I didn't think of that namespace before...

We have used a script to auto-process ontologies into DefinedNamespace but in this case there are multiple QUDT files so it might be easier just to SPARQL query them. I'll have a go.

westurner:
I share the same concerns about inlining vocabs into rdflib core; even for autocompletion in an IDE.

[...]

PR #1419, opened 2021-09-22
Add Brick DefinedNamespace

gtfierro:
An answer to the call in https://github.com/RDFLib/rdflib/issues/1415

## Proposed Changes

- Adds Brick classes, properties and EntityProperties into an RDFlib DefinedNamespace
- a generation script will be committed to the [Brick](https://github.com/BrickSchema/brick/) repository to facilitate updates to RDFlib when Brick releases are done (every ~6 months for bigger releases)

Let me know if the namespace is too large: I can impose some principles for reducing the number of individuals in the file, such as only including the first 2 levels of the Brick class hierachy


nicholascar:
Thanks @gtfierro! I don't think it's too big - ~1,000 lines of Python is nothing really, compared to some PyPI packages.

I'm going to try and make available, probably by adding to rdflib utils, a DefinedNamespace script that people can use, so we can compare that with your BRICK script eventually. 

OK, well this passes tests and looks fine to me so I'll merge in, without requiring further reviews. We may eventually split many namespaces off into a separate repo but, for now, I'll merge them into the core.

PR #1425, opened 2021-10-02
A new Turtle serializer

nicholascar:
This is a slightly improved, or at least changed, Turtle serializer that does more indenting and new lining.

In particular it:

* uses PREFIX instead of @prefix
* uses BASE instead of @base
* adds a new line at RDF.type, or 'a'
* adds a newline and an indent for all triples with more than one object (object list)
* adds a new line and ';' for the last triple in a set with '.'
    on the start of the next line
* uses default encoding (encode()) is used instead of "latin-1"

This serializer is called with `g.serialize(format="turtle2")`

Here is a comparison of turtle v. turtle2 outputs:

```
@prefix ex: <https://example.org/> .
@prefix ex2: <https://example2.org/> .

ex:b a ex:Thing ;
    ex:name "B" ;
    ex2:name "B" .

ex:c a ex:Thing ;
    ex:bnObj [ ex:lst ( ex:one ex:two ex:three ),
                ( ex:four ex:fize ex:six ) ] ;
    ex:lst ( ex:one ex:two ex:three ),
        ( ex:four ex:fize ex:six ) ;
    ex:lst2 ( ex:one ex:two ex:three ) ;
    ex:name "C" .

```

turtle2:

```
PREFIX ex: <https://example.org/>
PREFIX ex2: <https://example2.org/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

ex:b
    a ex:Thing ;
    ex:name "B" ;
    ex2:name "B" ;
.

ex:c
    a ex:Thing ;
    ex:bnObj [
            ex:lst
                (
                    ex:one
                    ex:two
                    ex:three
                ) ,
                (
                    ex:four
                    ex:fize
                    ex:six
                )
        ] ;
    ex:lst
        (
            ex:one
            ex:two
            ex:three
        ) ,
        (
            ex:four
            ex:fize
            ex:six
        ) ;
    ex:lst2 (
            ex:one
            ex:two
            ex:three
        ) ;
    ex:name "C" ;
.

```

### DIFF
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -161,6 +161,7 @@ register(
     "text/turtle", Serializer, "rdflib.plugins.serializers.turtle", "TurtleSerializer"
 )
 register("turtle", Serializer, "rdflib.plugins.serializers.turtle", "TurtleSerializer")
+register("turtle2", Serializer, "rdflib.plugins.serializers.turtle2", "TurtleSerializer2")
 register("ttl", Serializer, "rdflib.plugins.serializers.turtle", "TurtleSerializer")
 register(
     "application/n-triples", Serializer, "rdflib.plugins.serializers.nt", "NTSerializer"
--- a/rdflib/plugins/serializers/turtle2.py
+++ b/rdflib/plugins/serializers/turtle2.py
@@ -1,0 +1,331 @@
+"""
+Turtle2 RDF graph serializer for RDFLib.
+See <http://www.w3.org/TeamSubmission/turtle/> for syntax specification.
+
+This variant, turtle2 as opposed to just turtle, makes some small format changes
+to turtle - the original turtle serializer - which are:
+
+* using PREFIX instead of @prefix
+* using BASE instead of @base
+* adding a new line at RDF.type, or 'a'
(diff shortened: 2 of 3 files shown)


white-gecko:
How is the performance?
And why `PREFIX` instead of `@prefix`?

nicholascar:
> How is the performance?

I haven’t checked performance but since it’s only a few newline characters and spaces different from the current turtle serialiser, the performance will be the same.

> And why PREFIX instead of @prefix?

Because this syntax matches SPARQL better and thus it’s easier to convert between this turtle and SPARQL while testing queries and data.

> The many newlines might be matter of taste.

Yes, this is mostly a subjective thing however I will just point out that the motivation for this particular change is for better Git-style diffing. It also makes it easier to see multiple classes and list objects.

> I like, that it uses the default encoding.

There are just a few places in which the current serialiser implements latin-1, probably just a hang over from Python 2.

white-gecko:
Yes the encoding should also be part of my, never finished, #1222 ... Unifying all serializers and parsers to unicode would be a good overall improvement.

PR #1428, opened 2021-10-07
Add owl:rational and owl:real to match standard.

### LINKED ISSUE #1378: `rational` not in `rdflib.OWL.rational` after upgrade to rdflib 6.0.0
After upgrading to rdflib 6.0.0, I noticed that `rational` is no longer available as a term in the `OWL` namespace.

```python
In [1]: import rdflib
   ...: rdflib.OWL.rational
RDFLib Version: 6.0.0
AttributeError                            Traceback (most recent call last)
<ipython-input-1-df2c677dd20b> in <module>
      1 import rdflib
----> 2 rdflib.OWL.rational

~/.local/lib/python3.9/site-packages/rdflib/namespace/__init__.py in __getattr__(cls, name)
    203 
    204     def __getattr__(cls, name):
--> 205         return cls.__getitem__(name)
    206 
    207     def __repr__(cls):

~/.local/lib/python3.9/site-packages/rdflib/namespace/__init__.py in __getitem__(cls, name, default)
    194         if (cls._warn or cls._fail) and not name in cls:
    195             if cls._fail:
--> 196                 raise AttributeError(f"term '{name}' not in namespace '{cls._NS}'")
    197             else:
    198                 warnings.warn(

AttributeError: term 'rational' not in namespace 'http://www.w3.org/2002/07/owl#'
```

I assume that this has something to do with [this comment](https://github.com/RDFLib/rdflib/issues/953#issuecomment-575072954) on issue #953. Nevertheless, as said there, the `rational` is mentioned in [the OWL 2 standard](https://www.w3.org/TR/owl2-syntax/#Datatype_Maps), so it might make sense to bring it back.

sa-bpelakh:
Fixes #1378 

## Proposed Changes

  - Add `real` and `rational` to the `OWL` namespace, as per [standard](https://www.w3.org/2007/OWL/wiki/Syntax#a_rational)

### DIFF
--- a/rdflib/namespace/_OWL.py
+++ b/rdflib/namespace/_OWL.py
@@ -116,6 +116,8 @@ class OWL(DefinedNamespace):
     versionIRI: URIRef  # The property that identifies the version IRI of an ontology.
 
     # http://www.w3.org/2000/01/rdf-schema#Datatype
+    # NOTE: the following two elements don't appear in the OWL RDF documents but are defined in the OWL2 Recommentation
+    # at https://www.w3.org/TR/owl2-syntax/#Datatype_Maps
     rational: URIRef  # The value space is the set of all rational numbers. The lexical form is numerator '/' denominator, where both are integers.
     real: URIRef  # The value space is the set of all real numbers. Does not directly provide any lexical forms.
 


nicholascar:
I've just added a note to indicate where these two elements are defined - in the OWL2 spec but not in the RDF file.

It's a pain that all elements of OWL, PROV etc aren't in the RDF docs...

PR #1433, opened 2021-10-10
Export DCMITYPE

aucampia:
Not sure what went wrong here but for some reason master branch tries to
export DCMITYPE without also importing it.


nicholascar:
Oops, my mistake yesterday. There were discrepancies between the namespace __initi__.py and the rdflib __init__.py that contains all the namespaces. I fixed 3 but this was missed! merging right away. Thanks @aucampia

PR #1355, opened 2021-07-08
Dev stack update

ashleysommer:
Initial python 3.7 requirement update, add new Drone tests, add black.toml file, add .editorconfig file

~Draft PR, do not merge yet,~ testing Drone compatibility with Black

PR #1434, opened 2021-10-11
Bump Black version for real

ashleysommer:
Related to #1355, #1397, #1400, #1401

I know I have stated we're sticking with Black v21.6b0 as _the_ black version to use (and closing/reverting attempts to update it from dependbot). However this update is deliberate. There was a bug I encountered in black v21.6b0 in PySHACL, so I had to update it to a newer version. Thought I'd do the same in RDFLib, so the required version is known to be good, before we run into the same issue.

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -167,6 +167,7 @@ from rdflib.namespace import (
     CSVW,
     DC,
     DCAT,
+    DCMITYPE,
     DCTERMS,
     DOAP,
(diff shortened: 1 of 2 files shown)


nicholascar:
@ashleysommer no problems: I think the plan being we peg to a single version of black for as long as we can, but that circumstances may, and here have, forced us to update.

Funny lot of errors coming through the build here. Are these a result of other PRs?

PR #1439, opened 2021-10-12
Drone: Add Python 3.10 to the testing

cclauss:
Fixes #

## Proposed Changes

  - Add Python 3.10 to the testing

---

Nose is incompatible with Python 3.10 which has finally removed the alias between `collections` and `collections.abc`.
* Perhaps upgrade to https://docs.nose2.io or https://docs.pytest.org
```
  File "/usr/local/lib/python3.10/site-packages/nose/suite.py", line 106, in _set_tests
    if isinstance(tests, collections.Callable) and not is_suite:
AttributeError: module 'collections' has no attribute 'Callable'
```
https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -34,31 +34,61 @@ steps:
   - coveralls
 
 ---
-local Pipeline(name, image) = {
-  kind: pipeline
-  name: name
-  type: docker
-  platform:
-    os: linux
-    arch: amd64
-  steps: [
-    {
-      name: test
-      image: image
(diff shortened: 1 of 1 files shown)


nicholascar:
We are going to have to do work to get this passed. I do want to support 3.10 ASAP and I also want to move away from note, to pytest.

We might just have to leave this unmerged until the Callable fix is merged first.

PR #1441, opened 2021-10-12
Fix Graph.parse URL handling on windows

### LINKED ISSUE #1430: new version does not parse URL in windows
The old version(5) does parse remote URL in windows, however, the new version shows error
"[WinError 123] The syntax for the file name, directory name or disk label is incorrect:" when parse. Is it possible to change it?Thanks

aucampia:
Using `pathlib.Path("http://example.com/").exists()` on windows causes an exception as a URL
is not a valid path, while `os.path.exists("http://example.com/")` just
returns false.

This patch reverts _create_input_source_from_location to using
`os.path.exists()` instead of pathlib.Path to make it possible to parse
graphs from http URLs on windows.

Fixes #1430

## Proposed Changes

  - use `os.path.exists()` instead of `pathlib.Path().exists()` as the former returns false for URLs on windows while the latter rasies an exception.

### DIFF
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -307,10 +307,13 @@ def create_input_source(
 
 
 def _create_input_source_from_location(file, format, input_source, location):
-    # Fix for Windows problem https://github.com/RDFLib/rdflib/issues/145
-    path = pathlib.Path(location)
-    if path.exists():
-        location = path.absolute().as_uri()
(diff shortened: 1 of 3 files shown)


nicholascar:
I believe the PR authors and other contributors to #1430 that this fixes their issues and there are great tests, so I'm merging straight away.

PR #1442, opened 2021-10-12
Run two different flake8 configs

cclauss:
Related to https://github.com/RDFLib/rdflib/pull/1438#issuecomment-941206241

This will allow us to remove ignored tests gradually over multiple PRs that are easier to review and merge.

## Proposed Changes
Run two flake8 commands: 1) Just /rdflib, 2) The whole repo
```
  - flake8 --ignore=E203,E266,E402,E713,E722,E741,F401,F402,F403,F811,F821,F841,W503
           --max-complexity=41 --max-line-length=1656 --show-source --statistics rdflib
  - flake8 --ignore=E101,E126,E203,E265,E266,E303,E402,E711,E712,E713,E722,E731,E741,F401,F402,F403,F405,F523,F541,F811,F821,F841,W191,W291,W293,W503
           --max-complexity=41 --max-line-length=1653 --show-source --statistics .
```

PR #1444, opened 2021-10-14
drone.yml: Make black check a mandatory test

cclauss:
Move all conditionals into black.toml

Related to the discussion in #1442

## Proposed Changes

  - Make black check a mandatory test
  - Remove the excludes tests more files without slowing down the process.
```
+ black --check --config black.toml --diff .
All done! ✨ 🍰 ✨
281 files would be left unchanged.
```

### DIFF
--- a/black.toml
+++ b/black.toml
@@ -1,23 +1,4 @@
 [tool.black]
+include = '\.pyi?$'
 required-version = "21.9b0"
 target-version = ['py37']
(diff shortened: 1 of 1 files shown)


nicholascar:
@cclauss have you c,osed this PR as you no logger recommend these changes or because they are added to other PRs?

PR #1445, opened 2021-10-14
replace pkg_resources with importlib.metadata

### LINKED ISSUE #952: plugin discovery causes very bad startup time
https://github.com/RDFLib/rdflib/blob/e00480271a6de85ea3e24b63c177b34340a5c2dd/rdflib/plugin.py#L111 This line adds at least 0.3 seconds to startup time because entry point discovery is extremely slow. This number can very depending on the system. https://github.com/pypa/setuptools/issues/510 has more details.

A temporary fix that breaks plugin discovery is included below. However, a more robust solution is needed.

```patch
diff --git a/rdflib/plugin.py b/rdflib/plugin.py
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -108,6 +108,7 @@ def get(name, kind):
 
 
 try:
+    raise ImportError('iter_entry_points is unacceptably slow so skip it')
     from pkg_resources import iter_entry_points
 except ImportError:
     pass  # TODO: log a message
```

Panaetius:
Fixes #952

## Proposed Changes
- replaces pkg_resources.iter_entry_points with importlib.metdata.entry_points
- Conditionally installs importlib_metadata backport depending on python version

One thing to note, the `if hasattr(all_entry_points, "select"):` check is slightly awkward. In python 3.8 and 3.9, `entry_points()` returns a plain dict, but in python 3.10 and the backport (on 3.7) it returns a `SelectableGroups` objects that inherits from dict. Using `select()` on that object is slightly faster (~55ms) than using it through the dict interface. Since it inherits from dict, `isinstance(x, dict)` wouldn't work, and since `SelectableGroups` is not available in the 3.8/3.9 version, `isinstance(x, SelectableGroups)` wouldn't work either. hence the check via `hasattr`. This makes it perform equally fast across python versions.

### DIFF
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -37,6 +37,7 @@ from rdflib.query import (
 )
 from rdflib.exceptions import Error
 from typing import Type, TypeVar
+import sys
 
 __all__ = ["register", "get", "plugins", "PluginException", "Plugin", "PKGPlugin"]
 
@@ -110,10 +111,10 @@ def get(name: str, kind: Type[PluginT]) -> Type[PluginT]:
     return p.getClass()
 
 
-try:
-    from importlib.metadata import entry_points
-except ImportError:
+if sys.version_info < (3, 8):
     from importlib_metadata import entry_points
+else:
+    from importlib.metadata import entry_points
 
 all_entry_points = entry_points()
 if hasattr(all_entry_points, "select"):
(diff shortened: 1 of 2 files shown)


nicholascar:
Hmmm... I'll try and run locally tomorrow

aucampia:
For more info on what is going wrong here: https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

Panaetius:
I merged your PR, thanks for looking into this!

PR #1456, opened 2021-10-28
Fix JSON-LD data import adds trailing slashes to IRIs (#1443)

### LINKED ISSUE #1443: json-ld parser adds trailing slash to URLs that have no path
This is best described with an example. Here is a json-ld file that contains a URL that has no path:
```json
[
  {
    "@id": "https://sbolstandard.org/examples/model1",
    "http://sbols.org/v3#source": [
      {
        "@id": "http://virtualparts.org"
      }
    ]
  }
]
```

Using rdflib 6.0.2 (and 6.0.1) when the above file is loaded, the value of property `http://sbols.org/v3#source` is `http://virtualparts.org/` instead of `http://virtualparts.org`. Note the trailing `/`. This does not happen with other formats like 'nt11', 'xml', or 'ttl'. The issue appears to be with the json-ld parser, not the json-ld serializer.

I have written a self-contained program that demonstrates this issue: 

```python
import rdflib

print(f'Using rdflib {rdflib.__version__}')
g = rdflib.Graph()
g.add((rdflib.URIRef('https://sbolstandard.org/examples/model1'),
       rdflib.URIRef('http://sbols.org/v3#source'),
       rdflib.URIRef('http://virtualparts.org')))


def round_trip(rdf_format):
    print(f'-------------------- Testing {rdf_format} --------------------')
    data = g.serialize(format=rdf_format)
    # print()
    # print(f'Data:\n{data}')
    g2 = rdflib.Graph()
    g2.parse(data=data, format=rdf_format)

    g_objects = list(g.objects())
    g2_objects = list(g2.objects())
    if g_objects == g2_objects:
        print(f'Round trip {rdf_format} looks good')
    else:
        print(f'Oh no! Things are different with {rdf_format}!')
        print(f'g has {g_objects}')
        print(f'g2 has {g2_objects}')


round_trip('xml')
round_trip('ttl')
round_trip('nt11')
round_trip('json-ld')
```

Here is the output of the the program:
```
$ python3 examples/jsonldbug.py
Using rdflib 6.0.2
-------------------- Testing xml --------------------
Round trip xml looks good
-------------------- Testing ttl --------------------
Round trip ttl looks good
-------------------- Testing nt11 --------------------
Round trip nt11 looks good
-------------------- Testing json-ld --------------------
Oh no! Things are different with json-ld!
g has [rdflib.term.URIRef('http://virtualparts.org')]
g2 has [rdflib.term.URIRef('http://virtualparts.org/')]
```

newinnovations:
In norm_url leave url alone if it already contains a scheme/protocol.

Fixes #1443

### DIFF
--- a/rdflib/plugins/shared/jsonld/util.py
+++ b/rdflib/plugins/shared/jsonld/util.py
@@ -59,6 +59,8 @@ def norm_url(base, url):
     >>> norm_url('http://example.org/', 'http://example.org//one')
     'http://example.org//one'
     """
+    if "://" in url:
+        return url
     parts = urlsplit(urljoin(base, url))
     path = normpath(parts[2])
     if sep != "/":
--- a/test/jsonld/test_util.py
+++ b/test/jsonld/test_util.py
@@ -1,0 +1,79 @@
+import unittest
+from typing import NamedTuple
+
+from rdflib.plugins.shared.jsonld.util import norm_url
+
+
+class URLTests(unittest.TestCase):
+    @unittest.expectedFailure
+    def test_norm_url_xfail(self):
+        class TestSpec(NamedTuple):
+            base: str
+            url: str
+            result: str
+
+        tests = [
+            TestSpec(
+                "git+ssh://example.com:1231/some/thing/",
+                "a",
+                "git+ssh://example.com:1231/some/thing/a",
+            ),
+        ]
+
+        for test in tests:
+            (base, url, result) = test
+            with self.subTest(base=base, url=url):
+                self.assertEqual(norm_url(base, url), result)
+
+    def test_norm_url(self):
+        class TestSpec(NamedTuple):
+            base: str
+            url: str
+            result: str
+
+        tests = [
+            TestSpec("http://example.org/", "/one", "http://example.org/one"),
+            TestSpec("http://example.org/", "/one#", "http://example.org/one#"),
+            TestSpec("http://example.org/one", "two", "http://example.org/two"),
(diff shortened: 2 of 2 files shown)


nicholascar (approved):
Happy to approve but I think we are getting new tests that are old style - UnitTest, not pytest - coming though... so I suppose we might have to update these tests to pytest before the next release

aucampia:
@nicholascar pytest integrates very well with `unittest.UnitTest` so these do run and will report correctly when pytest runs.

nicholascar:
> pytest integrates very well with unittest.UnitTest

Good to know. I thought this was the case but hadn't checked up recently. I suppose then our focus just needs to be on the skipped tests and the reasons for skipping. I suppose there will be some, on-going, good reasons for skipping some but we should, in general, try and see if all currently skipped tests can be not skipped, whether written in modern pytests or older compatible forms!

I'll follow up with the Py 3.10 isodate issues this weekend if they are not solved by gweis before then so we can see all tests run on that too.

aucampia:
> I suppose then our focus just needs to be on the skipped tests and the reasons for skipping. I suppose there will be some, on-going, good reasons for skipping some but we should, in general, try and see if all currently skipped tests can be not skipped, whether written in modern pytests or older compatible forms!

A lot of the skipped tests should be changed to expected failures I think, at least most of the ones for the test suites, I will do a general review of test suites when I have time to make sure we run every test suite that is applicable and report tests that we don't run correctly (i.e. expected failure as failure and skipped only for tests we cannot run), possibly in conjuction with #1479 so that we run a EARL report on every test run.

PR #1461, opened 2021-11-14
Add GitHub actions workflow for validation

### LINKED ISSUE #1471: CI pipelines for Windows and MacOS
Currently the CI only runs for linux, but the result is that things have fallen into disrepair for windows and to a much lesser extent MacOS. It would be nice if we could run CI on these platforms also before merging patches.

Currently GitHub actions does provide windows and macos runners and I have done some work to set it up here: https://github.com/iafork/rdflib/compare/iwana-20211018T2122-pytest...iwana-20211113T1312-github-actions

I have not set it up for Windows yet, but this is mainly because windows is still having some weird issues which I'm independently debugging.

We can maybe start by running this in parallel to drone, I think drone can maybe test windows also but I don't think it can do macos.

Any input on this would be welcome.

aucampia:
Add GitHub actions workflow for validation

The workflow will run on windows, linux and macos.

Also:
- Limit berkeleydb in requirements.dev.txt to not install on windows as
  currently this fails to install in the windows workflow.
- Fix .editorconfig for yaml

Fixes #1471

This depends on fixes in #1476 which should probably be merged first.

Note that this fails on windows and will continue to fail until #1478 is merged.


gjhiggins:
>     * Remove berkeleydb from requirements.dev.txt as this won't work on
>       windows.
Maybe use conditional checking, [according to stackoverflow](https://stackoverflow.com/questions/29222269/is-there-a-way-to-have-a-conditional-requirements-txt-file-for-my-python-applica):
```berkeleydb; platform_system != "Windows"```

aucampia:
> Maybe use conditional checking, [according to stackoverflow](https://stackoverflow.com/questions/29222269/is-there-a-way-to-have-a-conditional-requirements-txt-file-for-my-python-applica):
> `berkeleydb; platform_system != "Windows"`
> 
> 
> Works as expected with [rdflib-leveldb](https://github.com/gjhiggins/rdflib-leveldb/blob/py3-migration/requirements.txt) on my Windows VM, switching to plyvel-wheels as instructed.

@gjhiggins will do thanks

aucampia:
> > Maybe use conditional checking, [according to stackoverflow](https://stackoverflow.com/questions/29222269/is-there-a-way-to-have-a-conditional-requirements-txt-file-for-my-python-applica):
> > `berkeleydb; platform_system != "Windows"`
> > Works as expected with [rdflib-leveldb](https://github.com/gjhiggins/rdflib-leveldb/blob/py3-migration/requirements.txt) on my Windows VM, switching to plyvel-wheels as instructed.
> 
> @gjhiggins will do thanks

@gjhiggins  done

PR #1464, opened 2021-11-15
Add import of os

gjhiggins:
1.
```
File "../test_magic_set.py", line 64, in test_AdornLiteral
    res = translateAlgebra(translateQuery(parsed_query))
File ".../rdflib/plugins/sparql/algebra.py", line 1427, in translateAlgebra
    os.remove("query.txt")
name 'os' is not defined
```
2. infixowl throws an Exception when it tries to split the OWL Namespace object. (Would have preferred to raise a separate PR, was unaware that GH simply concatenates commits into a single PR).
## Proposed Changes

  - Add `import os`
  - Change `str(p.split(OWL)[-1])` to `str(p.split(str(OWL))[-1])` to avoid Exception.
  -

### DIFF
--- a/rdflib/extras/infixowl.py
+++ b/rdflib/extras/infixowl.py
@@ -39,7 +39,7 @@ We can then access the rdfs:subClassOf relationships
 
 This can also be used against already populated graphs:
 
->>> owlGraph = Graph().parse(OWL) #doctest: +SKIP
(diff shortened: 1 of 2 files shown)


white-gecko (approved):
Looks good to me.

gjhiggins:
@white-gecko, given the straightforward nature of the change, I'll go ahead with the merge.

PR #1474, opened 2021-11-22
Fix shield for CI status

aucampia:
The old shield was still pointing to travis, this patch changes
it to use the Drone CI shield shield.

The link is to the status for all branches, I can't find a link
specifically to the master branch status, the links from the branch
status page goes to individual build numbers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 RDFLib
 ======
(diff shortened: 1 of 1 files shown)


nicholascar:
@ashleysommer can you help out with the master branch link?

nicholascar (approved):

PR #1485, opened 2021-12-01
SDO HTTPS and DN creator script

nicholascar:
Fixes downstream issues with other RDFlib packages that expect schema.org NS to be HTTPS (e.g. https://github.com/RDFLib/pyLODE/issues/155).

Adds a  tool script to create DefinedNamespaces too.

SDO seems to have reverted to HTTP from HTTPS accidentally.

## Proposed Changes

  - update SDO to HTTPS
  - add script to generate DefinedNamespaces


rdflib/tools/defined_namespace_creator.py
```diff
@@ -0,0 +1,182 @@
...
+
+sys.path.append(str(Path(__file__).parent.absolute().parent.parent))
+
+from rdflib import Graph, util
+from rdflib.namespace import DCTERMS, OWL, RDFS, SKOS
+
+
+def get_input_format(file_path):
+    input_format = util.guess_format(str(file_path))
+    if input_format is None:
```
aucampia, line 26: I think the latest util.guess_format does support this now after changes you made to it:

https://github.com/RDFLib/rdflib/blob/9379a69d6ec6e18819aa0c5a0d54849e7abb8223/rdflib/util.py#L354-L370

rdflib/tools/defined_namespace_creator.py
```diff
@@ -0,0 +1,182 @@
...
+        raise ValueError("The supplied namespace must end with '/' or '#'")
+
+
+def validate_object_id(object_id):
+    for c in object_id:
+        if not c.isupper():
+            raise ValueError("The supplied object_id must be an all-capitals string")
+
+
+def get_classes(g, target_namespace):
```
aucampia, line 50: Is this being used? Can't see where, but I may be overlooking it.

aucampia (approved):
Looks good to me, there are two comments but they are not wrong, just noting redundant code (I think).

rdflib/tools/defined_namespace_creator.py
nicholascar: Actually no! I started thinking about getting Classes & Properties - the main things in ontologies - but then reverted to a simpler collection of things in the namespace. I've left this function there because I hate deleting code but I will mark it unused.

rdflib/tools/defined_namespace_creator.py
nicholascar: Yes, I think so, so I will remove this and test.

PR #1487, opened 2021-12-01
Fix typing of create_input_source

aucampia:
Change the type of `data` parameter to also allow values of `Dict` type
as these are used with `PythonInputSource`.

NOTE: This fixes build failure on master.


nicholascar (approved):
You beat me to fixing this by a day!

Issue #1437, opened 2021-10-12
Support HexTuples RDF format


nicholascar:
The [HexTuples](https://github.com/ontola/hextuples) RDF format looks simple yet powerful. Support for this in RDFlib could be made easily using perhaps the [jsonlines](https://jsonlines.readthedocs.io/) JSON parser which supports NDJSON.

The reason for this request isn’t just the touted performance of the format in a browser but also because it’s extremely simple and explicit so I think it would assist new RDF users to differentiate BlankNodes, Named Nodes, datatypes and languages.

nicholascar:
@joepio, author of HexTiples spec, you might be interested in this

nicholascar:
Looks simple, yay! I’ll make sure correct attributions to the spec and implementation authors are added to any code in RDFlib.

gjhiggins:
Finalized in https://github.com/RDFLib/rdflib/commit/dabcacf3bfda83715e3bc4afb85d10e7ae2630ce

Issue #1332, opened 2021-05-30
A `URIRef` subclass breaks SPARQL querying


anatoly-scherbakov:
**Context.** I tried to subclass `rdflib.URIRef` and use it my code because I wanted to add additional functionality to it.

**Expectations.** I expected a subclass of `URIRef` to work exactly like `URIRef` itself works.

**Reality.** However, it appeared that SPARQL querying will not properly work in that case. Here is a full fledged example.

```python
import rdflib
from rdflib import Graph, FOAF


class IRI(rdflib.URIRef):
    """Internationalized Resource Identifier."""

    def __repr__(self):
        """Format IRI."""
        return f'<{self.toPython()}>'


def test_uriref_subclass():
    graph = Graph()
    graph.add((
        IRI('https://example.com/Alice'),
        IRI(FOAF.knows),
        IRI('https://example.com/Bob'),
    ))

    bindings = graph.query(
        'SELECT * WHERE { ?s <http://xmlns.com/foaf/0.1/knows> ?o . }',
    ).bindings

    # Here we see that bindings are empty: nothing was found.
    assert not bindings
```

_(This code should be runnable as-is.)_

Since we added a triple to graph, I expected to find that triple via a SPARQL query but to no avail because `FOAF.knows` is saved as a `IRI`, not as a `URIRef`.

This is a non-standard pattern of usage, but would be great to see if anyone had circumvented such a problem.

juanjosegzl:
I've been doing some experiments to deal with this and I was able to solve it by casting the IRI object to URIRef at the moment is saved in the graph [changes here](https://github.com/RDFLib/rdflib/compare/master...juanjosegzl:poc/query-subclass-uriref?expand=1#diff-a8bea910ef28d48bce0a390d9ab24c55ac32f1c9a39d9cd215224e5447b70c83).

query works well, problem is that now the node is returned as URIRef instead of IRI.

I will keep experimenting with the exact moment the cast could be done, I am thinking right now when the query is being evaluated. **What do you think about casting to solve this?**

I'll be posting my progress

anatoly-scherbakov:
> I will keep experimenting with the exact moment the cast could be done, I am thinking right now when the query is being evaluated. **What do you think about casting to solve this?**

I would like to point out a few questions here.

1. The developer who had subclassed `URIRef` might change its constructor. It is not guaranteed that `URIRef(s)`, where `s` is an instance of a custom `IRI` class, even works.
[...]

PR #1489, opened 2021-12-03
Hextuples Serializer

nicholascar:
This HexTuples Serializer partly addresses Issue #1437. No Parser is yet made


nicholascar:
@joepio we are some of the way there!

aucampia (approved):

nicholascar:
@joepio: now we have a parser as well! It's as simple as I could make it so it just loops through strings or files. No remote URI fetching of content yet.

@joepio: please can you at least verify that the following test data is valid Hextuples:

```json
        ["http://example.com/s01", "http://example.com/a", "http://example.com/Type1", "", "", ""]
        ["http://example.com/s01", "http://example.com/label", "This is a Label", "http://www.w3.org/2001/XMLSchema#string", "en", ""]
        ["http://example.com/s01", "http://example.com/comment", "This is a comment", "http://www.w3.org/2001/XMLSchema#string", "", ""]
        ["http://example.com/s01", "http://example.com/creationDate", "2021-12-01", "http://www.w3.org/2001/XMLSchema#date", "", ""]
        ["http://example.com/s01", "http://example.com/creationTime", "2021-12-01T12:13", "http://www.w3.org/2001/XMLSchema#dateTime", "", ""]
        ["http://example.com/s01", "http://example.com/age", 42, "http://www.w3.org/2001/XMLSchema#integer", "", ""]
        ["http://example.com/s01", "http://example.com/trueFalse", false, "http://www.w3.org/2001/XMLSchema#boolean", "", ""]
        ["http://example.com/s01", "http://example.com/op1", "http://example.com/o1", "", "", ""]
        ["http://example.com/s01", "http://example.com/op1", "http://example.com/o2", "", "", ""]
        ["http://example.com/s01", "http://example.com/op2", "http://example.com/o3", "", "", ""]
```

Since comparisons are against this, if this is correct, we should be fine!

My only query was whether a datatype of XSD string _must_ be indicated if a language is present. For completeness, I think it should be, hence the "label" line in the data above.

nicholascar:
OK, so roundtripping is failing and I haven't yet worked out why. Likely due to the Hextuples parser putting in bad default graph values

rescribet:
@nicholascar No that's not valid.

- All elements are always a string, not json literal `true` or `42`.
- The individuals encoding the meaning of a named or blank node are missing in the datatype column
- rdf:langString is the correct datatype if a language tag is present, as per the rdf spec
- Empty value in the graph position means these statements should end up in the default graph

"No remote URI fetching of content yet." Not sure what you mean by this, it is a serialization format.

nicholascar:
I have worked out why most of the round-tripping tests that are failing aren't working: it is due to my current serialization implementation adding in `xsd:string` for literals of unknown type. I think this is an rdflib implementation issue for, in RDF in general, a literal of undeclared type is interpreted as a string (https://www.w3.org/TR/turtle/#turtle-literals) so whether there's an `xsd:string` declaration there or not, it should be isomorphic. This will be partly solved by addressing the comments below.

Some other tests are failing due to something relating to HexT not being formula-aware but I can't work this out yet. Basically, HexT doesn't have the expressive power for Quoted Graphs in the Subject position (I think). So I will just get the HexT round-tripping to pass for all n-triples examples that are not formula-aware for now and ignore the N3 tests which are formular-aware.

@rescribet I think I can make most of those changes:

* All elements are always a string, not json literal true or 42.
    * easy
* The individuals encoding the meaning of a named or blank node are missing in the datatype column
    * easy
* rdf:langString is the correct datatype if a language tag is present, as per the rdf spec
    * easy
* Empty value in the graph position means these statements should end up in the default graph
    * yes, this is what my parser is doing, but the serializer is, so far, putting Blank Node IDs in the graph position for default graphs. I'll replace this with ""

* "No remote URI fetching of content yet." Not sure what you mean by this, it is a serialization format.
    * the rdflib `parse()` method can parse strings or files or content at the end of a URI (online). I just haven't implemented that "end of a URI" version yet.

rescribet:
Cool. I've just extracted our parser code into a modified version of our n-quads-parser package. It also has an rdflib-compatible class, but that was rdflib compatible at the time of writing the n-quads parser, no idea if that is still useful today.

At least it should be useful if people outside of rdflib would want to add this capability to their project https://github.com/ontola/hextuples-parser

rescribet:
@nicholascar A small but significant change, the individuals which were on the rdf namespace have been replaced with just `globalId` or `localId`. So:

```
["http://example.com/s01", "http://example.com/a", "http://example.com/Type1", "globalId", "", ""]
["http://example.com/s01", "http://example.com/b", "b432", "localId", "", ""]
["http://example.com/s01", "http://example.com/age", "42", "http://www.w3.org/2001/XMLSchema#integer", "", ""]
```

rdflib/plugins/serializers/hext.py
```diff
@@ -34,70 +35,85 @@ def __init__(self, store: Union[Graph, ConjunctiveGraph]):
...
+            return '["%s", "%s", "%s", "%s", "%s", "%s"]\n' % (
+                self._iri_or_bn(triple[0]),
+                triple[1],
+                value,
+                datatype,
+                language,
+                self._context(context)
+            )
+        else:  # do not return anything for non-IRIs or BNs, e.g. QuotedGraph, Subjects
+            return None
```
edmondchuc: Can the type comparisons be done using `isinstance()`? Or is there a requirement to use `type()`?

nicholascar: Is there an advantage to using `isinstance()`? No real reason not to: that's just what I started doing! Happy to change if `isinstance()` is _better_ in some way.

nicholascar: I presume you're concerned with type inheritance, in case people use subclasses of, say, `Literal`? I'll convert (and solve the reamining failing test).

edmondchuc: `isinstance()` caters for subclasses while `type()` does not. 

```python
class SpecialURIRef(URIRef):
    pass


special_uri = SpecialURIRef("123")

type(special_uri) == URIRef
>>> False

isinstance(special_uri, URIRef)
>>> True
```

Using `type()` could lead to similar issues like https://github.com/RDFLib/rdflib/issues/1332

nicholascar: Done in Commit https://github.com/RDFLib/rdflib/pull/1489/commits/75b797e0aacdd098d5ad7c863fc628d468f4dfe9

PR #1490, opened 2021-12-03
fix for issue 1484 raised and solved by Graham Klyne:

gjhiggins:
Error reading JSON-LD from text IO stream

Fix and test provided by Graham Klyne, transcribed into a separate branch.

Confirmatory results from the “validate.yaml” workflow are [here](https://github.com/DOACC/rdflib/actions/runs/1536643449/workflow)

### DIFF
--- a/rdflib/plugins/shared/jsonld/util.py
+++ b/rdflib/plugins/shared/jsonld/util.py
@@ -37,14 +37,12 @@ def source_to_json(source):
     try:
         # Use character stream as-is, or interpret byte stream as UTF-8
         if isinstance(stream, TextIOBase):
-            # use_stream = stream
-            return json.load(stream)
+            use_stream = stream
         else:
-            use_stream = TextIOWrapper(stream, encoding="utf-8")
+            use_stream = TextIOWrapper(stream, encoding='utf-8')
         return json.load(use_stream)
     finally:
         stream.close()
-
 
 VOCAB_DELIMS = ("#", "/", ":")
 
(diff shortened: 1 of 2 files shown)


nicholascar (approved):
Thanks Graham Klyne & Graham Higgins!

rdflib/plugins/shared/jsonld/util.py
```diff
@@ -17,22 +17,31 @@
...
+        return json.load(source.getCharacterStream())
+
     # TODO: conneg for JSON (fix support in rdflib's URLInputSource!)
     source = create_input_source(source, format="json-ld")
 
     stream = source.getByteStream()
     try:
-        return json.load(StringIO(stream.read().decode("utf-8")))
+        # Use character stream as-is, or interpret byte stream as UTF-8
+        if isinstance(stream, TextIOBase):
```
aucampia, line 39: I'm actually unsure if this condition will be true given this code from `rdflib.parser.FileInputSource`:

https://github.com/RDFLib/rdflib/blob/9379a69d6ec6e18819aa0c5a0d54849e7abb8223/rdflib/parser.py#L239-L244

gklyne: FWIW, I was trying to home in on the error that was triggered by trying to parse the output from `json.dump`, which specifically caused the stream result from `source.getByteStream()` to be a character stream.  I agree that seems not quite right, and the problem may lie deeper in the logic that sorts out the data stream to be parsed.  

But I was also trying to, identify a minimal fix that could solve my immediate problem.  Hopefully if someone chooses to delve deeper into the `create_input_source` logic, the test cases provided will still be helpful.

Issue #1470, opened 2021-11-17
NamespaceManager only binds subset of available namespaces


RinkeHoekstra:
A new `NamespaceManager` (e.g. as initialized with a new `Graph` instance) will only bind the XML, XSD, RDF and RDFS namespaces, but none for the other available namespaces.

https://github.com/RDFLib/rdflib/blob/d9e86c932695d780bcdcdfadda150e4c1bcf72ef/rdflib/namespace/__init__.py#L336-L339

It would be much more convenient if the default available namespaces were to be bound to their default prefixes upon initialization of a new `Graph`.

PR #1491, opened 2021-12-07
Add bindings for rdflib namespaces. Import DCAM.

edmondchuc:
Fixes https://github.com/RDFLib/rdflib/issues/1470

## Proposed Changes

  - Add bindings for rdflib namespaces.
  - Import `DCAM` to be available in the `namespace` package.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -348,7 +348,7 @@ class NamespaceManager(object):
         self.bind("owl", OWL)
         self.bind("prof", PROF)
         self.bind("prov", PROV)
-        self.bind("QB", QB)
+        self.bind("qb", QB)
         self.bind("rdf", RDF)
         self.bind("rdfs", RDFS)
         self.bind("schema", SDO)
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -333,10 +333,36 @@ class NamespaceManager(object):
         self.__trie = {}
         for p, n in self.namespaces():  # self.bind is not always called
             insert_trie(self.__trie, str(n))
-        self.bind("xml", XMLNS)
+
+        self.bind("brick", BRICK)
+        self.bind("csvw", CSVW)
+        self.bind("dc", DC)
+        self.bind("dcat", DCAT)
+        self.bind("dcmitype", DCMITYPE)
+        self.bind("dcterms", DCTERMS)
+        self.bind("dcam", DCAM)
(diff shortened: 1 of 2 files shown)


nicholascar:
Oh dear, this seems to have broken a couple of SPARQL tests!

rdflib/namespace/__init__.py
```diff
@@ -333,11 +333,37 @@ def __init__(self, graph):
...
+        self.bind("sh", SH)
+        self.bind("skos", SKOS)
+        self.bind("sosa", SOSA)
+        self.bind("ssn", SSN)
+        self.bind("time", TIME)
+        self.bind("vann", VANN)
+        self.bind("void", VOID)
         self.bind("xsd", XSD)
 
+        # TODO: XMLNS is not a DefinedNamespace.
```
edmondchuc, line 364: Is there unique properties in the XMLNS? If not, then we can remove the todo comment.

edmondchuc:
I have no idea how adding these additions would break the SPARQL tests. I'm keen to look into it if you have any ideas @nicholascar.

rdflib/namespace/__init__.py
nicholascar: No, there are no properties for it. It's not an RDF namespace as such, just a legacy XML one. I _think_ people need it for RDF/XML but no other RDF formats care about it.

nicholascar:
Perhaps the breaking is due to testing that uses a fixed set of namespace PREFIX statements in the SPARQL?

rdflib/namespace/__init__.py
edmondchuc: Cool thanks, I removed the todo comment in https://github.com/RDFLib/rdflib/pull/1491/commits/ce9e2cf4003b195f15911f45f28e99c2bc59a302.

edmondchuc:
@nicholascar I fixed the tests. 

Essentially what I did was change the SPARQL service queries for DBPedia's URL from `http` to `https` and now it works. 

```sparql
PREFIX ex:<http://example.org/>
select ?s ?p ?o
where {
     SERVICE <https://DBpedia.org/sparql> {
        VALUES (?s ?p ?o) {(ex:a ex:b 1) (<http://example.org/a> <http://example.org/b> 2)}
    }
} 
LIMIT 2
```

No idea why the `http` version works fine with the basic prefix bindings (`rdf`, `rdfs`, etc.) but did not work for the others 🤷‍♂️.

I think it is fine to make this change to the service query URL since it's a functional test to an external service.

Issue #1404, opened 2021-09-08
Unable to round-trip skolemize() and de_skolemize()


edwardanderson:
I would like to be able to perform a round-trip skolemize and deskolemize on a graph. While `graph.skolemize()` works expectedly, I'm not able to convert the URIs back into blank nodes.

Convert a blank node to URI:

```python
from rdflib import Graph


ttl = '''
@prefix wd: <http://www.wikidata.org/entity/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

wd:Q1203 foaf:knows [ a foaf:Person;
    foaf:name "Ringo" ].
'''

graph = Graph()
graph.parse(data=ttl, format='turtle')
skolemized = graph.skolemize()
skolemized.namespace_manager.bind('foaf', 'http://xmlns.com/foaf/0.1/')
print(skolemized.serialize(format='turtle'))
```

```turtle
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

<http://www.wikidata.org/entity/Q1203> foaf:knows <http://rdlib.net/.well-known/genid/rdflib/n1c65f973c8e54cb2a6d199e36ec26a22b1> .

<http://rdlib.net/.well-known/genid/rdflib/n1c65f973c8e54cb2a6d199e36ec26a22b1> a foaf:Person ;
    foaf:name "Ringo" .
```

All good. However `graph.de_skolemize()` returns the URI unchanged:

```python
deskolemized = skolemized.de_skolemize()
deskolemized.namespace_manager.bind('foaf', 'http://xmlns.com/foaf/0.1/')
print(deskolemized.serialize(format='turtle'))
```

```turtle
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

<http://www.wikidata.org/entity/Q1203> foaf:knows <http://rdlib.net/.well-known/genid/rdflib/n1c65f973c8e54cb2a6d199e36ec26a22b1> .

<http://rdlib.net/.well-known/genid/rdflib/n1c65f973c8e54cb2a6d199e36ec26a22b1> a foaf:Person ;
    foaf:name "Ringo" .
```

I wonder if this is to do with the type checking inside [`do_de_skolemize2()`](https://rdflib.readthedocs.io/en/stable/_modules/rdflib/graph.html#Graph.de_skolemize)? As none of the subjects or objects in the graph are `isinstance(_, Genid)`, no deskolemize operation appears to be performed.

[...]

PR #1493, opened 2021-12-09
Fix BNode.skolemize() returning a URIRef instead of an RDFLibGenid.

edmondchuc:
Fixes https://github.com/RDFLib/rdflib/issues/1404

## Proposed Changes

- Correctly return `RDFLibGenid` instead of `URIRef` as required by `Graph.de_skolemize()`.
- This fixes skolemize/de-skolemize round-trip which was previously completely broken.
- Add tests.

Why `RDFLibGenid` and not `Genid`? See `URIRef.de_skolemize()`.

https://github.com/RDFLib/rdflib/blob/327de4ecc9e7a9a7c73b55d1b59b6e1e600f64ed/rdflib/term.py#L308-L328

Essentially, `RDFLibGenid` will correctly grab the original blank node ID before skolemization.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -446,7 +446,7 @@ class BNode(Identifier):
         if basepath is None:
             basepath = rdflib_skolem_genid
         skolem = "%s%s" % (basepath, str(self))
-        return URIRef(urljoin(authority, skolem))
+        return Genid(urljoin(authority, skolem))
 
 
 class Literal(Identifier):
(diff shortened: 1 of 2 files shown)


edmondchuc:
Let me add some tests for the skolemization/de-skolemization round-trip.

gjhiggins (approved):
Looks good to me, fwiw.

nicholascar (approved):
Excellent testing there @edmondchuc!

Issue #1469, opened 2021-11-16
How can I check for values to be valid URIRefs?


sabinem:
When I try to parse a Graph such as this here below, I get an error about a not well formed URIRef: 

```
<?xml version="1.0" encoding="utf-8" ?>
<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
         xmlns:dcat="http://www.w3.org/ns/dcat#"
         xmlns:dc="http://purl.org/dc/terms/">

  <dcat:Catalog rdf:about="https://swisstopo/catalog-endpoint.rdf">
    <dcat:dataset>
      <dcat:Dataset rdf:about="https://swisstopo/data/123">
        <dc:title xml:lang="en">some title</dc:title>
        <dc:description xml:lang="en">some dataset</dc:description>
        <dcat:accrualPeriodicity about="Hello World"/>
      </dcat:Dataset>
    </dcat:dataset>

    <dcat:dataset>
      <dcat:Dataset rdf:about="https://swisstopo/data/345">
        <dc:title xml:lang="en">some title</dc:title>
        <dc:description xml:lang="en">some dataset</dc:description>
      </dcat:Dataset>
    </dcat:dataset>

  </dcat:Catalog>

</rdf:RDF>
```

```
>g.parse('catalog,rdf')
ParserError: file:///home/sabine/Work/ckan-docker/catalog.rdf:11:8: Invalid property attribute URI: http://www.w3.org/1999/02/22-rdf-syntax-ns#about
```
So far so good. 
But when I build that catalog: how can I check that "Hello World" is not a valid URIRef?

```
>from rdflib import URIRef
>In [19]: x = URIRef("Hello world")                                                                                   
Hello world does not look like a valid URI, trying to serialize this will break.
```

I get this warning, but I what I would like to have would be an error, so that I can react on it. 
Would you have any advice for me on that? What does rdflib offer in that regard on validating URIRefs?

edmondchuc:
Unfortunately, looks like when you create a `URIRef`, it only logs a warning and does not raise an exception until you try to serialize the graph.

You can do something similar with the code below.

```python
from urllib.parse import quote_plus

from rdflib import URIRef, Graph, RDF, SDO
from rdflib.term import _is_valid_uri


class InvalidURIRef(Exception):
    """Raise when URIs are invalid."""
    pass


def create_uriref(uri):
    """Create a URIRef with the same validation func used by URIRef

[...]

PR #1494, opened 2021-12-09
Forbid truthy values for lang when initializing Literal

veyndan:
With the proposed change, an exception will be raised when `lang` is a truthy value. I found this quite confusing when messing around with the library that `Literal("Hello", lang={})` was somehow considered a valid language tag.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -547,7 +547,7 @@ class Literal(Identifier):
                 "per http://www.w3.org/TR/rdf-concepts/#section-Graph-Literal"
             )
 
-        if lang and not _is_valid_langtag(lang):
+        if lang is not None and not _is_valid_langtag(lang):
             raise ValueError("'%s' is not a valid language tag!" % lang)
 
         if datatype:
--- a/test/test_literal.py
+++ b/test/test_literal.py
@@ -1,10 +1,13 @@
 from decimal import Decimal
+from typing import Any, Optional, Type
 import unittest
 import datetime
 
 import rdflib  # needed for eval(repr(...)) below
 from rdflib.term import Literal, URIRef, _XSD_DOUBLE, bind, _XSD_BOOLEAN
 from rdflib.namespace import XSD
+
+import pytest
 
 
 class TestLiteral(unittest.TestCase):
@@ -42,15 +45,39 @@ class TestLiteral(unittest.TestCase):
         self.assertEqual(l.datatype, rdflib.XSD["boolean"])
 
 
+class TestNewPT:
+    # NOTE: TestNewPT is written for pytest so that pytest features like
+    # parametrize can be used.
+    # New tests should be added here instead of in TestNew.
+    @pytest.mark.parametrize(
+        "lang, exception_type",
+        [
+            ({}, TypeError),
+            ([], TypeError),
+            (1, TypeError),
+            (b"en", TypeError),
+            ("999", ValueError),
+            ("-", ValueError),
+        ],
+    )
+    def test_cant_pass_invalid_lang(
+        self,
+        lang: Any,
+        exception_type: Type[Exception],
+    ):
+        """
+        Construction of Literal fails if the language tag is invalid.
+        """
+        with pytest.raises(exception_type):
+            Literal("foo", lang=lang)
+
+
 class TestNew(unittest.TestCase):
+    # NOTE: Please use TestNewPT for new tests instead of this which is written
+    # for unittest.
     def testCantPassLangAndDatatype(self):
         self.assertRaises(
             TypeError, Literal, "foo", lang="en", datatype=URIRef("http://example.com/")
-        )
-
-    def testCantPassInvalidLang(self):
-        self.assertRaises(
-            ValueError, Literal, "foo", lang="999"
         )
 
     def testFromOtherLiteral(self):


gjhiggins:
The change that you propose doesn't have quite the desired effect. An Exception _is_ raised but it's raised in an unexpected context and the message is misleading, suggesting that a bytes-like object might be acceptable when it isn't:
```python
>>> Literal("foo", lang={})
  File ".../rdflib/rdflib/term.py", line 551, in __new__
    if lang is not None and not _is_valid_langtag(lang):
  File ".../rdflib/rdflib/term.py", line 92, in _is_valid_langtag
    return bool(_lang_tag_regex.match(tag))
TypeError: expected string or bytes-like object
```

But, you raise a valid concern, the checks performed by RDFLib are limited to whether the language tag is "" and if so rebinds it to `None` or whether the tag matches the regexp `"^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$"`.

The main issue here is what a user might reasonably expect of RDFLib as a library. I feel we could do better - something like this pytest example:

```python
def test_literal_language_tag():

    with pytest.raises(Exception) as e_info:
        Literal("xxxx", lang="")
    assert (
        repr(e_info.value) == """ValueError("language tag cannot be 'None' in the absence of a datatype!")""")

    with pytest.raises(Exception) as e_info:
        Literal("Hello", lang={})
    assert (
        repr(e_info.value) == """TypeError("language tag '{}' is <class 'dict'>, must be a string!")""")

    with pytest.raises(Exception) as e_info:
        Literal("xxxx", lang=b"")
    assert (
        repr(e_info.value) == """TypeError("language tag 'b''' is <class 'bytes'>, must be a string!")""")

    with pytest.raises(Exception) as e_info:
        Literal("Hello", lang="999")
    assert repr(e_info.value) == """ValueError("'999' is not a valid language tag!")"""

    # 'Eastern Armenian language written in Latin script, as used in Italy'
    Literal("դաշտերից", lang="hy-Latn-IT-arevela")

    # Japanese (family)
    Literal("日本語の表記体系", lang="jpx")

    # Don't be deceived by the above, no checking is performed.
    # This really shouldn't pass muster but it does
    Literal("xxxx", lang="is-utter-nonsense")

    # However, if conformance-checking were implemented, say using ozbigcat's
    # (https://github.com/ozbigcat/RFC5646-ISO639-ISO15924-python)
    # RFC5646 checking contribution with an added 'check_rfc5646_conformance'
    # flag (allowing for backwards-compatiblity) we'd see ...
    with pytest.raises(Exception) as e_info:
        Literal("xxxx", lang="is-utter-nonsense", check_rfc5646_conformance=True)
    assert (
        repr(e_info.value) == """ValueError("'is-utter-nonsense' is not a valid RFC5646 language tag!")""")

    # And, of course, “tlhIngan Hol”, aka Klingon
    Literal("pIn'a' qan paQDI'norgh", "tlh", check_rfc5646_conformance=True)
```

Or, are expectations more along the lines of RDFLib remaining relatively lightweight and the onus being on the user to provide sanitised, validated input rather than looking to RDFLIb for support?

veyndan:
> The change that you propose doesn't have quite the desired effect. An Exception _is_ raised but it's raised in an unexpected context and the message is misleading, suggesting that a bytes-like object might be acceptable when it isn't[.]

Ah, that's weird, perhaps there's some issue in the way that I'm running it as I'm getting the following:

```python
>>> Literal("foo", lang={})
rdflib.term.Literal('foo', lang={})
>>> Literal("foo", lang={"hello": "world"})
  File "/Users/veyndan/IdeaProjects/rdflib/rdflib/term.py", line 551, in __new__
    if lang and not _is_valid_langtag(lang):
  File "/Users/veyndan/IdeaProjects/rdflib/rdflib/term.py", line 92, in _is_valid_langtag
    return bool(_lang_tag_regex.match(tag))
TypeError: expected string or bytes-like object
```

> But, you raise a valid concern, the checks performed by RDFLib are limited to whether the language tag is "" and if so rebinds it to `None` and whether the tag matches the regexp `"^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$"`.
> 
> The main issue here is what a user might reasonably expect of RDFLib as a library. I feel we could do better[.]

Yeah, I agree that we could do better but I just didn't want to create a large PR with multiple API changes. Instead, I was hoping to create some other PRs with smaller changes so the discussion can be focused on a single change.

> ```python
> with pytest.raises(Exception) as e_info:
>     Literal("xxxx", lang="")
> assert (
>     repr(e_info.value) == """ValueError("language tag cannot be 'None' in the absence of a datatype!")""")
> ```

I like this a lot. Failing as close as possible to the source of the issue is always good! I can create a separate PR for this if interested.

> ```python
> with pytest.raises(Exception) as e_info:
>     Literal("Hello", lang={})
> assert (
>     repr(e_info.value) == """TypeError("language tag '{}' is <class 'dict'>, must be a string!")""")
> 
> with pytest.raises(Exception) as e_info:
>     Literal("xxxx", lang=b"")
> assert (
>     repr(e_info.value) == """TypeError("language tag 'b''' is <class 'bytes'>, must be a string!")""")
> ```

Could we just declare `lang` with the type `Optional[str]`? From what I understand about mypy (it's been a while since I've used Python tbh), just changing the annotation to `Optional[str]` won't change how the program is run, but perhaps we can assume people will run mypy over their code and we gain the benefit of succinctness by not having to do these type checks? I can create a separate PR for this if interested.

> ```python
> with pytest.raises(Exception) as e_info:
>     Literal("Hello", lang="999")
> assert repr(e_info.value) == """ValueError("'999' is not a valid language tag!")"""
> 
> # 'Eastern Armenian language written in Latin script, as used in Italy'
> Literal("դաշտերից", lang="hy-Latn-IT-arevela")
> 
> # Japanese (family)
> Literal("日本語の表記体系", lang="jpx")
> 
> # Don't be deceived by the above, no checking is performed.
> # This really shouldn't pass muster but it does
> Literal("xxxx", lang="is-utter-nonsense")
>
> # However, if conformance-checking were implemented, say using ozbigcat's
> # (https://github.com/ozbigcat/RFC5646-ISO639-ISO15924-python)
> # RFC5646 checking contribution with an added 'check_rfc5646_conformance'
> # flag (allowing for backwards-compatiblity) we'd see ...
> with pytest.raises(Exception) as e_info:
>     Literal("xxxx", lang="is-utter-nonsense", check_rfc5646_conformance=True)
> assert (
>     repr(e_info.value) == """ValueError("'is-utter-nonsense' is not a valid RFC5646 language tag!")""")
> 
> # And, of course, “tlhIngan Hol”, aka Klingon
> Literal("pIn'a' qan paQDI'norgh", "tlh", check_rfc5646_conformance=True)
> ```
>
> Or, are expectations more along the lines of RDFLib remaining relatively lightweight and the onus being on the user to provide sanitised, validated input rather than looking to RDFLIb for support?

My 2 cents are that I'd opt for putting the onus on the user. We could point them to https://github.com/ozbigcat/RFC5646-ISO639-ISO15924-python in the documentation, and then the user can decide whether they need these checks in their code by including or not including (what could be) a library.

```python
assert rfc5646_conformance(foo)  # Completely optional and the function can be provided by a separate library

Literal("xxxx", lang=foo)
```

gjhiggins:
I saw your edits, sorry for any confusion. I appreciate your observations about keeping RDFLib light and in order to preseve the potential utility of the rfc5646 solution for those occasions where it might come in handy, I've distilled out a patch which I'll include as a recipe (along with a reference to osbigcat's rfc5646 repos) in an RDFLIb "Cookbook" which I've (just) decided to create as an addition to the RDFLib documentation.

w.r.t. to the PR, ultimately, I boiled it down to just this:
```diff
diff --git a/rdflib/term.py b/rdflib/term.py
index 796a76b3..36d14db6 100644
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -548,8 +548,13 @@ class Literal(Identifier):
                 "per http://www.w3.org/TR/rdf-concepts/#section-Graph-Literal"
             )
 
-        if lang and not _is_valid_langtag(lang):
-            raise Exception("'%s' is not a valid language tag!" % lang)
+        if lang is not None:
+            if not isinstance(lang, str):
+                raise TypeError(
+                    f"language tag '{lang}' is {type(lang)}, must be a string!"
+                )
+            elif not _is_valid_langtag(lang):
+                raise ValueError(f"'{lang}' is not a valid language tag!")
 
         if datatype:
             datatype = URIRef(datatype)
```
and a minimalist test:
```py
import pytest
from rdflib import Literal

def test_literal_language_tag():

    with pytest.raises(Exception) as e_info:
        Literal("Hello", lang={})
    assert (
        repr(e_info.value)
        == """TypeError("language tag '{}' is <class 'dict'>, must be a string!")"""
    )

    with pytest.raises(Exception) as e_info:
        Literal("Hello", lang="999")
    assert repr(e_info.value) == """ValueError("'999' is not a valid language tag!")"""

```
Cheers,

Graham

nicholascar:
@veyndan, are you happy to adopt @gjhiggins's suggested updates and the test? 

@gjhiggins where are you making up the CookBook? That seems like a great idea and perhaps we can put a bit of effort into a CookBook and release it with rdflib 7.0.0 perhaps March next year? I've wanted to come up with rdflib patterns or a CookBooks style thing for some time, so if there is momentum here...

nicholascar:
@veyndan @gjhiggins since IANA maintains an authoritative list of the language tags [RFC 5646 Registry Format and Maintenance](https://datatracker.ietf.org/doc/html/rfc5646#section-3): 

> The IANA Language Subtag Registry ("the registry") contains a comprehensive list of all of the subtags valid in language tags.

with the Registry actually at https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

Why can't we test not for string, not for non binary etc but for actual values in that list? Storing a copy of the list in rdflib with function to regenerate it per release and another function for users to locally append to the rdflib list (for local use, if our master list is out of date) wouldn't be too hard.

gjhiggins:
> @veyndan @gjhiggins since IANA maintains an authoritative list of the language tags [RFC 5646 Registry Format and Maintenance](https://datatracker.ietf.org/doc/html/rfc5646#section-3):

> with the Registry actually at https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

> Why can't we test not for string, not for non binary etc but for actual values in that list? Storing a copy of the list in rdflib with function to regenerate it per release and another function for users to locally append to the rdflib list (for local use, if our master list is out of date) wouldn't be too hard.

That's pretty much what I had, following ozbigcat's suggestion - the repos includes an updater for when the IANA list changes and [simple instructions](https://github.com/ozbigcat/RFC5646-ISO639-ISO15924-python#to-update-the-record).

Given that the IANA source _does_ arbitrarily change, my concern was not to add yet another chore to the RDFLib maintenance burden for what is arguably a minor nuance of support. However, it's a reasonably clean change which a proficient user could easily make, so I thought it would make a good starter for a Cookbook.

There is also the danger that if the library starts checking lang tags against the spec, some well-meaning soul will inevitably point out that Literal values aren't checked for validity w.r.t. a declared datatype and for consistency's sake, perhaps they should be :)

gjhiggins:
> @gjhiggins where are you making up the CookBook? That seems like a great idea and perhaps we can put a bit of effort into a CookBook and release it with rdflib 7.0.0 perhaps March next year? I've wanted to come up with rdflib patterns or a CookBooks style thing for some time, so if there is momentum here...

Just locally atm, while I collect some "recipes" as a side effect of working through two-dozen-odd issues related to the identifier-as-context isssue, checking if they're still extant w.r.t. the rebased PR. As soon as I've got a few putative entries I can begin to hallucinate a structure and will create a repos for the content. (e,g, [Edmond Chuc's response](https://github.com/RDFLib/rdflib/issues/1469#issuecomment-990415448) to #1469 is another excellent candidate for a Cookbook recipe).

FWIW, I was the original perpetrator of the RDFLib docs, back in the day (mostly created from pulling in stuff the other devs had written) and although I haven't communicated as much, I was looking to assist you with refreshing/reworking the RDFLib docs (happily, I've managed to regain access my my readthedocs a/c).

Again, this is also a side-effect of the id-as-ctx work, f'rinstance the confusion around the param bindings for `Dataset.graph()` is quite understandable given that the API is completely undocumented.

nicholascar:
> not to add yet another chore to the RDFLib maintenance burden for what is arguably a minor nuance of support

OK, since the benefits are low, let's not add this burden!

nicholascar:
Last thing: policy is to see a test added as well that indicates how that the change is working as expected, so please add a short test (to an existing test file if one look appropriate or add a new file) and then, i all good, I'll approve.

aucampia:
@veyndan not sure if you are working on tests here, if you are let me know, if not I'm going to make a PR to include some tests from https://github.com/veyndan/rdflib/pull/2

veyndan:
I'm not sure what tests to add here tbh, as I'm still not convinced that we should add tests for non-strings (per our discussion in #1498). So the truthy value for string is `""` which (debatably) should (and does) pass, so apart from that I'm not sure what tests make sense assuming that we only want to test those which abide by the type definition of `Optional[str` (which is in #1498, but we can assume that both the PRs will be merged at the same-ish time).

aucampia:
> I'm not sure what tests to add here tbh, as I'm still not convinced that we should add tests for non-strings (per our discussion in #1498). So the truthy value for string is `""` which (debatably) should (and does) pass, so apart from that I'm not sure what tests make sense assuming that we only want to test those which abide by the type definition of `Optional[str`] (which is in #1498, but we can assume that both the PRs will be merged at the same-ish time).

Quoting response on https://github.com/RDFLib/rdflib/pull/1498#issuecomment-996254758

> > I understand that `lang` being anything other than a string is invalid. My point is that we have "restricted" the type of `lang` to a string with the type annotations. Is it really necessary to do a runtime check when we already have the type annotations. This is a general question — it's not specific to `lang`.
> 
> The answer to this question is not really very straight forward. In brief, I think having something like this would be wrong:
> 
> ```python
> if not isinstance(lang, str):
>     raise TypeError(f"Type of lang {lang!r} is invalid")
> ```
> 
> And actually if you did this, mypy would error out, because `warn_unreachable` is enabled for rdflib, and the above would be unreachable as far as mypy is concerned, and within this is probably the best answer I can give you to this question, if mypy determines that the type checking is unreachable, then it probably should not be there.
> 
> However in this case (state of [veyndan#2](https://github.com/veyndan/rdflib/pull/2)) `mypy` will not (and did not) determine that anything is unreachable, because nothing is unreachable.
> 
> Now as to the tests, there is a well defined value space for the language tag: it must be a string that conforms to the RDF spec for lang tag, and this value space is something that has to be checked outside of the type system as it is not expressible in the type system, and if this check does not pass an error should be raised, and we know this check should never pass for any value in this array, because they do not lie within the value space: `[ "", {}, [], 1, False, True ]` - so no check that adequately verifies that a value lies within the valid value space of language tag should succeed for these, and the way in which it should fail should follow the guidelines python sets for errors, which is that `ValueError` should occur for invalid values, and `TypeError` for invalid types.
> 
> So to me, it is reasonable to test that a value of `{}` for lang raises `TypeError` - because I know that there must be runtime validation of the value space, and I know that this should follow python's guidance on error checks, which is that Values outside of the value range but with right type should raise `ValueError`, and values outside of the value space with wrong type should raise `TypeError`. And again, I emphasize, what the tests I added in [veyndan#2](https://github.com/veyndan/rdflib/pull/2) check for is that the value space validation for lang is sufficiently strict.

aucampia:
I  made a pull request against your branch to add tests: https://github.com/veyndan/rdflib/pull/4

This follows the reasoning explained here: https://github.com/RDFLib/rdflib/pull/1498#issuecomment-996254758

aucampia (changes requested):
As indicated, if a PR fixes a runtime problem (as this test will) then it should have tests that reproduce the runtime problem that is being fixed.

The PR i made against your branch include tests which I think are sufficient but other tests which cover similar cases can also be accepted.

veyndan:
@nicholascar This PR should be good to go now (thanks @aucampia for adding the tests!).

aucampia (approved):
Looks good to me, will wait for one more approval from another maintainer.

nicholascar (approved):
Thanks for the detailed considerations for such a small PR!

PR #1496, opened 2021-12-10
Remove deprecated Statement class

### LINKED ISSUE #1158: Resolve .term.Statement deprecation
`.term.Statement` has been deprecated since 2012. Is there any plan to actually remove it or un-deprecate it?

https://github.com/RDFLib/rdflib/blob/master/rdflib/term.py#L1699

veyndan:
Fixes #1158 

Per the issue, it was mentioned that this was supposed to be removed in v5.1.0 or v6.0.0, but it is still present in the codebase.

I'm not sure whether we need to wait for a major version to release this as it is _technically_ a breaking change — but as the issue mentioned, it's been deprecated for about 10 years at this point, so perhaps it's fine to remove it in a minor version. I have no strong opinions on this, just thought I'd bubble this up as I think it's well overdue for removal.

### DIFF
--- a/rdflib/store.py
+++ b/rdflib/store.py
@@ -158,7 +158,6 @@ class Store(object):
             from rdflib.term import Literal
             from rdflib.graph import Graph, QuotedGraph
             from rdflib.term import Variable
-            from rdflib.term import Statement
 
             self.__node_pickler = np = NodePickler()
             np.register(self, "S")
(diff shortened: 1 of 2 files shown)


nicholascar:
Since this removal has triggered no test fails, I'm going to merge it without further reviews.

nicholascar (approved):

PR #1497, opened 2021-12-11
Raise a more specific Exception when lang isn't valid

veyndan:
@gjhiggins I pulled part of https://github.com/RDFLib/rdflib/pull/1494#issuecomment-991124245 into a separate PR, as I think it's good but outside the scope of the other PR.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -548,7 +548,7 @@ class Literal(Identifier):
             )
 
         if lang and not _is_valid_langtag(lang):
-            raise Exception("'%s' is not a valid language tag!" % lang)
+            raise ValueError("'%s' is not a valid language tag!" % lang)
 
         if datatype:
             datatype = URIRef(datatype)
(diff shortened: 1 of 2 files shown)


aucampia (approved):

aucampia:
As mentioned in #1498 - I think this change is very good - but this is so intertwined with #1498 that it may make more sense to just incorporate this into #1498 - especially so that it is easier to add the additional tests for #1498

veyndan:
Perhaps we could just merge this PR? It doesn't seem like there are any objections to this change so far, and then I can just update the base branch of the other PR.

aucampia:
> Perhaps we could just merge this PR? It doesn't seem like there are any objections to this change so far, and then I can just update the base branch of the other PR.

I'm also fine with that, but ultimately it has to get to the state in this PR: https://github.com/veyndan/rdflib/pull/2

veyndan:
@nicholascar Would it be possible to merge this if there are no objections?

aucampia:
@veyndan I may be wrong here, but I think @nicholascar is focused on getting 6.0.3 out at the moment so this will have to wait.

nicholascar:
We can merge this simple one!

PR #1500, opened 2021-12-12
Quench nt test userwarn

gjhiggins:
Fed up with seeing Serializer Userwarnings in the test output. Is the `latin-1` format testing anything specific? Can it be simply swapped out for `utf-8` or am I missing something? (I'm not strong on the parse/serializer side of things).

### DIFF
--- a/rdflib/plugins/serializers/jsonld.py
+++ b/rdflib/plugins/serializers/jsonld.py
@@ -68,8 +68,7 @@ class JsonLDSerializer(Serializer):
         encoding = encoding or "utf-8"
         if encoding not in ("utf-8", "utf-16"):
             warnings.warn(
-                "JSON should be encoded as unicode. "
-                + "Given encoding was: %s" % encoding
+                "JSON should be encoded as unicode. " f"Given encoding was: {encoding}"
             )
 
         context_data = kwargs.get("context")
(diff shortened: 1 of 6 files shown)


aucampia (approved):

aucampia:
> Can it be simply swapped out for `utf-8` or am I missing something? (I'm not strong on the parse/serializer side of things).

If we are not specifically testing character sets support and handing of passing invalid char sets then I think valid charsets (e.g. `utf-8` in this case ) is more appropriate than charsets that cause warnings (e.g. `ascii`/1latin-1` in this case). So I think this change makes a lot of sense.

There should be tests to check the handling of passing of invalid charsets but they should be in a separate place and not mixed in with tests for something different, and they could (should?) even go as far as expecting the warnings. And I also think we should rather error out in some of these cases, especially if the internal handling is to just use utf-8 - I guess there are different approaches here but I think erroring out when an unsupported charset is supplied will result in less surprises for users than warning would. Though this is something we can consider as a seperate issue and discuss potentially for future major versions as this will break compatibility and warrant a major version bump.

I made a bunch of round trip tests that I still have to break off from https://github.com/RDFLib/rdflib/pull/1418 that cover character sets, and tests for handling of invalid charsets could maybe go into something like that.

nicholascar:
In all the 1.1 versions of the W3C Specifications we see UTF-8 mandated, e.g.:

* N-Triples: https://www.w3.org/TR/n-triples/#h2_n-triples-changes
* Turtle: https://www.w3.org/TR/turtle/#h3_sec-mime

So I think we can always use UTF-8. I wondered if there were reasons this wasn't already the case but presumably it's just for the historical reasons of different character support in Python < 3 and perhaps in W3C Specs < 1.1.

nicholascar (approved):

Issue #893, opened 2019-01-20
Dataset is not unpickled correctly


iddan:
Dataset is unpickeld as ConjuctiveGraph because it extends ConjuctiveGraph and inherits the `__reduce__()` method. It should be unpickled as Dataset instead.

iddan:
I've managed to fix it:
```python
from rdflib.graph import Dataset

class FixedDataset(Dataset):
    def __reduce__(self):
        return (type(self), (self.store, self.default_union))

    def __getstate__(self):
        return self.store, self.identifier, self.default_context, self.default_union

    def __setstate__(self, state):
        self.store, self.identifier, self.default_context, self.default_union = state

# TEST

import pickle
from rdflib.namespace import Namespace

example = Namespace("http://example.com#")

fixed = FixedDataset()
one = fixed.graph(example.one)

one.add((example.dan, example.knows, example.lisa))
one.add((example.lisa, example.knows, example.dan))

two = fixed.graph(example.two)

two.add((example.ben, example.knows, example.lisa))
two.add((example.lisa, example.knows, example.ben))

assert set(fixed.quads((None, None, None, None))) == set(fixed.quads((None, None, None, None)))

picklestring = pickle.dumps(fixed)
new_fixed = pickle.loads(picklestring)

for new_graph, graph in zip(new_fixed.graphs(), fixed.graphs()):
    assert new_graph.identifier == graph.identifier
    for new_triple, triple in zip(new_graph.triples((None, None, None)), graph.triples((None, None, None))):
        assert new_triple == triple
```

gjhiggins:
Fixed in #1501

PR #1501, opened 2021-12-12
Fix for issue 893

gjhiggins:
As iddan reports in issue #893 “`Dataset` is unpickeld as `ConjunctiveGraph` because it extends `ConjunctiveGraph` and inherits the `__reduce__()` method. It should be unpickled as `Dataset` instead.”

A fix was provided but unfortunately overlooked. Rectifying this oversight.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1948,6 +1948,15 @@ class Dataset(ConjunctiveGraph):
         )
         return pattern % self.store.__class__.__name__
 
+    def __reduce__(self):
+        return (type(self), (self.store, self.default_union))
+
+    def __getstate__(self):
+        return self.store, self.identifier, self.default_context, self.default_union
+
+    def __setstate__(self, state):
+        self.store, self.identifier, self.default_context, self.default_union = state
+
     def graph(self, identifier=None, base=None):
         if identifier is None:
             from rdflib.term import rdflib_skolem_genid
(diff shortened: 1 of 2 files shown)


gjhiggins:
Um. Needs some investigation. Looks like the fix works but the test needs some refinement.

gjhiggins:
Probably don't need to keep the test as part of the standard suite, can be removed later, it's not as if it's going to be subject to regression.

nicholascar (approved):

aucampia:
Seems this is failing sporadically:

- https://github.com/RDFLib/rdflib/runs/4501225806?check_suite_focus=true
- https://github.com/RDFLib/rdflib/runs/4505918939?check_suite_focus=true

gjhiggins:
> Seems this is failing sporadically:
> 
>     * https://github.com/RDFLib/rdflib/runs/4501225806?check_suite_focus=true
> 
>     * https://github.com/RDFLib/rdflib/runs/4505918939?check_suite_focus=true

Hmm. Investigating.

nicholascar:
Yes, I thought it was fine to merge as tests passed but then I too notice the intermittent failing.

If we can sort this out shortly, I'll release 6.0.3 with what we have. There seems to be no end of new, very good, PRs but I want a 6.0.3 out before the work on Graph IDs that @gjhiggins is leading as that's a large item and it would be safe to make a stable release just before it.

aucampia:
I think this does fix it: https://github.com/RDFLib/rdflib/pull/1504 - and makes sense as a cause for the sporadic failure.

Issue #1390, opened 2021-08-12
Bringing back RDFa support to cRDFLib


fathoni:
RDFa support was [dropped ](https://github.com/RDFLib/pyrdfa3) from RDFLib since RDFLib version 5.0. There have been changes to the RDFLib 5.0 which breaks its compatibility with pyRDFa3, and unfortunately, there is currently no plan on maintaining its compatibility according to the [discussion](https://github.com/RDFLib/pyrdfa3/issues/33) on pyRDFa3's GitHub page.  I think It would be great to bring RDFa compatibility back to the current version.  Thank you!

PR #1503, opened 2021-12-13
Restore pyrdfa

gjhiggins:
Fix for #1390, “Bringing back RDFa support to RDFLib”.

Tactical decision required:

Follow the original approach and copy the pyRdfa and pyMicrodata sources into the RDFLib source tree
or
Add `pyrdfa3==3.5.3` and `git+https://github.com/RDFLib/pymicrodata.git#egg=pyMicrodata` to requirements.

Also, it opens a bit of a can of worms in that using a wikipedia URL results in:
```
.../site-packages/urllib3/connectionpool.py:1013:
InsecureRequestWarning: Unverified HTTPS request is being made to host 'en.wikipedia.org'.
Adding certificate verification is strongly advised.
See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
```


gjhiggins:
Dunno what's happening with drone, maybe I need to re-do the fork into my repos.

gjhiggins:
> Tactical decision required:
> 
> Follow the original approach and copy the pyRdfa and pyMicrodata sources into the RDFLib source tree or Add `pyrdfa3==3.5.3` and `git+https://github.com/RDFLib/pymicrodata.git#egg=pyMicrodata` to requirements.

Alternatively ... add pyrdfa3 to requirements to avoid mypy errors preventing drone build from succeeding.

Long-term, if RDFa capability is to be preserved, pyrdfa3 will need to be brought into mypy scope at some point.

PR #1504, opened 2021-12-13
Fix for issue893

gjhiggins:
Serialization changes with run, sort the list of graphs before iterative comparison.


aucampia (approved):

nicholascar:
Makes sense!

PR #958, opened 2020-02-11
Move Store API to work with identifiers, not graphs

oohlaf:
This is pull request #409 rebased against master to fix #167

Issue #939, opened 2019-10-14
rdflib 4.2.2: ConjunctiveGraph.parse() return a Graph object


flacout:
The parse function of ConjunctiveGraph modify in place the ConjunctiveGraph but return a Graph object. This is a minor issue but it can create issues down the road if the parse methode is called like so ` g = g.parse(source="test.ttl", format='turtle')`

Demonstration:

 ```python
from rdflib import Graph, ConjunctiveGraph

g = ConjunctiveGraph()
g.parse(source="test.ttl", format='turtle')
print(type(g)) # <class 'rdflib.graph.ConjunctiveGraph'>

g = ConjunctiveGraph()
g = g.parse(source="test.ttl", format='turtle')
print(type(g)) # <class 'rdflib.graph.Graph'>
```

Issue #698, opened 2017-01-24
Dataset.graph should not allow adding random graphs to the store


gromgull:
A combination of this: 

https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1356

and this: 

https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1628-L1630

means you can pass an existing graph into the dataset. This will then be added directly. But there is no guarantee this new graph has the same store, and the triples will not be copied over. 

This is chaos, but wasn't flagged earlier since it's chaos all the way down :) #167

gromgull:
I realised due to #679

mgberg:
To clarify, if I were to run
```
import rdflib
from rdflib.namespace import FOAF
foaf = rdflib.Graph(identifier = FOAF)
foaf.parse("http://xmlns.com/foaf/spec/index.rdf", format="xml")
ds = rdflib.Dataset()
ds.add_graph(foaf)
```
I assume the expected behavior should be that `len(foaf)` should equal `len(ds)`, when in fact `len(ds)` is 0.

edmondchuc:
Here's a reproducible test.

```python
from rdflib import RDF, RDFS, Dataset, Graph, URIRef


def test_working():
    """
    This test passes because the graph is created through the Dataset.graph method.
    """
    ds = Dataset(default_union=True)
    graph_name = URIRef("urn:graph")
    graph = ds.graph(graph_name)

    graph.add((URIRef("urn:class"), RDF.type, RDFS.Class))
    assert len(graph) == 1
    assert len(ds) == 1

    ds.add_graph(graph)
    assert any(g.identifier == graph_name for g in ds.graphs())
    assert len(list(ds.objects(None, None))) == 1

    retrieved_graph = ds.get_graph(graph_name)
    assert retrieved_graph.identifier == graph_name
    assert isinstance(retrieved_graph, Graph)

    assert len(retrieved_graph) == 1
    assert len(ds) == 1

    ds.remove_graph(retrieved_graph)
    assert len(retrieved_graph) == 0
    assert len(ds) == 0



def test_broken():
    """
    This test fails because the graph is created first and then added to the Dataset.
    """
    ds = Dataset(default_union=True)
    graph_name = URIRef("urn:graph")
    graph = Graph(identifier=graph_name)

    graph.add((URIRef("urn:class"), RDF.type, RDFS.Class))
    assert len(graph) == 1
    assert len(ds) == 0

[...]

Issue #307, opened 2013-06-20
Dataset / ConjunctiveGraph


gromgull:
This issue is to collect to discussion of dataset vs. conjunctive graph from: #301, #302, #303  and friends. 

Similarities between ConjunctiveGraph and DataSet: 
- Both allow you to group triples into separate graphs. You can either look at it as "quads" (with a fourth _context_ element), or as a collection of graphs, each containing triples, the two views are interchangeable.
- Both should have _quad methods_, like `addN`, `add_quad`, `remove_quad`, `quads`, etc. 
- Both should have _graph methods_, like `get_context` (or `get_graph`), listing, deleting, etc. 

The difference between ConjuctiveGraph and DataSet: 
- ConjunctiveGraph allows (as the named implies) querying the conjunction (union) of all graphs 
- ConjunctiveGraph has a named default graph, implicitly named with a bnode if nothing else is given.
- DataSet has the concept of a default graph, but the exact definition is open. 
- DataSet tracks the existence of empty graphs, a ConjuctiveGraph only contains the graphs that contain triples

Some more concerns: 
- Our Store API support (somewhat implicitly) that idea of union of graphs if no context is specified. I would rather not change the store API at this point, i.e. I would not like to have a solution that requires stores to change to accemodate yet-another-graph class
- For SPARQL, it is often interesting to on-the-fly define a new DataSet, i.e. selecting some subset of graphs from another DataSet, or even from several datasets. It would be nice to be able to do this without copying all the triples. 

I suggest: 
- Add a new class `ContextAwareGraph`, that has flags for `union_default_graph` and  `track_empty_graphs`. This class also provides quads/graph methods. 
- Let ConjunctiveGraph and DataSet extend this class, each setting the flags the way they want. 

Not clear to me just now:
- How are the empty graphs of a `DataSet` persisted? Looking at the current code I would say they are not. I.e. if I open a `DataSet` on a Sleepycat store, create some graphs, exit my program. Next time I open, only graphs that contained triples will exists. 
- What about python operators on Context-aware graphs? `__sub__`, `__contains__`, etc. This is #225.

PR #1505, opened 2021-12-14
Identifier as context

gjhiggins:
Fixes # A few issues

## Proposed Changes
A rebase of the rebased PR #958 “Move Store API to work with identifiers, not graphs”

This is a Draft PR because the rebased #958 is basically just the start, a number of issues remain unresolved, even with the changes contained in these commits.

This change is the latest in series of changes initiated in 2012 with #167 and reignited in 2013 by #301 (and friends). The subtle differences between the way Dataset and ConjunctiveGraph models work has been discussed at length in various issue threads, in an attempt to get an overview, I collated all the early discussions in a [gist](https://gist.github.com/gjhiggins/41c6e1cbc56a41f72609fd8f585ace41)

If there are any misperceptions about the relative complexity of the exercise, gromgull summarised the, ahem, context in #307 including the assertion:

“For SPARQL, it is often interesting to on-the-fly define a new DataSet, i.e. selecting some subset of graphs from another DataSet, or even from several datasets. It would be nice to be able to do this without copying all the triples.”

but was moved to observe four years later in #698:

“A combination of https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L1698 and https://github.com/RDFLib/rdflib/blob/master/rdflib/graph.py#L2028 means you can pass an existing graph into the dataset. This will then be added directly. But there is no guarantee this new graph has the same store, and the triples will not be copied over. This is chaos, but wasn't flagged earlier since it's chaos all the way down :)”

The issues list contains around a couple of dozen issues related to the whole "identifier as context" business. I'm gradually working my way through them, discovering whether they remain extant or whether this PR fixes them.

My modus operandi is to create a test to check (in [`test_dataset_anomaly_wip.py`](https://github.com/gjhiggins/rdflib/blob/identifier-as-context/test/test_dataset_anomaly_wip.py)) and when I've characterised the issue, migrate it to [`test_dataset_anomalies.py`](https://github.com/gjhiggins/rdflib/blob/identifier-as-context/test/test_dataset_anomalies.py) with annotations. I've also created a separate test which implements all the current Dataset docstrings [`test_identifier_as_context.py`](https://github.com/gjhiggins/rdflib/blob/identifier-as-context/test/test_identifier_as_context.py)

Example: Issue  #939, “rdflib 4.2.2: ConjunctiveGraph.parse() return a Graph object” is fixed by the changes in this PR, as attested by [`test_issue939_parse_return_inconsistent_type`](https://github.com/gjhiggins/rdflib/blob/838f9c724305654eb0b871d87f18cbfeed9ca971/test/test_dataset_anomalies.py#L255)

A note to reviewers - in order to get the tests to pass without extensive skipping, I've been obliged to make a couple of changes in addition to the rebased PR and for those I have provided a rationale, prefixed with `DEVNOTE`. I had to add three failing named-graph-related DAWG tests to the skippedtests list - SPARQL update CLEAR remains an issue atm.


gjhiggins:
The basics look quite good - serialization of default graph and contexts looks sane and as expected.
Attached is results from [wip](https://github.com/gjhiggins/rdflib/blob/838f9c724305654eb0b871d87f18cbfeed9ca971/test/test_dataset_anomaly_wip.py#L145), methodically iterating the serializers over the dataset object and then its contexts.

[anomalies-wip-results.txt](https://github.com/RDFLib/rdflib/files/7719567/anomalies-wip-results.txt)

gjhiggins:
Got too scrappy to be useful in this state, close and reopen when more coherent.

PR #1508, opened 2021-12-15
Issue 1484 fix

### LINKED ISSUE #1429: Graph parse module does not handle URI / IRI with non-ascii characters
To reproduce behavior:

```python
from rdflib import Graph

g = Graph()
g.parse("https://dbpedia.org/page/Almería")
```

I was able to bypass it locally by editing the function in `parser.py`
```python
    path = pathlib.Path(location)
    if path.exists():
        location = path.absolute().as_uri()

    base = pathlib.Path.cwd().as_uri()

    concept = location.split('/')[-1]                # I added this line and the line below
    location = location.replace(concept, quote(concept))
    absolute_location = URIRef(location, base=base)

        file = open(filename, "rb")
    else:
        input_source = URLInputSource(absolute_location, format)

    auto_close = True
    # publicID = publicID or absolute_location  # Further to fix
    # for issue 130

    return absolute_location, auto_close, file, input_source
```
according to the help [here](https://stackoverflow.com/questions/4389572/how-to-fetch-a-non-ascii-url-with-urlopen)

Is there a better way to deal with this scenario or should this be a PR?

gjhiggins:
Fixes #1429

## Proposed Changes

  - add an internal iri-to-uri conversion utility

> why park this in the "restore-pyrdfa" branch? This fix will work for non-HTML sources with non-ASCII chars in the IRI, perhaps something like `http://example.com/thing/Almería.ttl` for instance. So this is a generic fix for `parse()`

I was both unclear and pedantic - I _meant_ that the specific dbpedia example URL provided (which actually pointed to a text/html resource) is parsed without issue with the implementation coded therein and pyRfda enabled.

Providing a generic fix for `parse()` proved (as ever, with RDFLib) more tricky ....

Tactics:

1. The encoding of query param content "=" as "%3D" causes the failure of
`GraphTestCase::testGuessFormatForParse` in that
 `https://linked.data.gov.au/def/agrif?_format=text/turtle`
is converted to:
`https://linked.data.gov.au/def/agrif?_format%3dtext/turtle`
which returns a resource of text/html mimetype, triggering a "no plugin found" Exception.
So, either decode the encoded "%3D" and friends or skip the conversion of query params altogether - I chose the former.

2. urllib omits the empty fragment identifier and the reassembled uri is stripped of any trailing "#" that the iri might have had, which causes the failure of
`NamedGraphWithFragmentTest::test_named_graph_with_fragment`
so it has to be explicitly restored to the end of the uri.

3. It proved desirable to restrict the conversion just "http" and "https" schemes to avoid issues with filesystem-located resources.


gjhiggins:
Well, _that_ didn't work.

PR #1510, opened 2021-12-16
Improve error messages from with-fuseki.sh

aucampia:
The error message will now clarify that logs are dumped afterwards.

Also dump only 15 lines of the fuseki logs by default when an error.


nicholascar (approved):
Great improvement

PR #1511, opened 2021-12-16
Fix pipeline triggers

aucampia:
The pipeline was triggering twice for pull requests made from branches
in the main repo, once for the pull request and once for a push to the
branch.

This changes the trigger for push to only happen on master branch.


nicholascar (approved):
Changed as requested

PR #1514, opened 2021-12-16
Change python version used for mypy to 3.7

### LINKED ISSUE #1513: setup.cfg incorrectly specifies mypy should run on Python 3.6
I was getting strange behavior from `mypy` failing on this one-liner program, but just when I was a `mypy` test in the `rdflib` source directory:

```python
from __future__ import annotations
```

Someone else having this problem [found it was user error](https://github.com/python/mypy/issues/9009#issuecomment-645459102), where they accidentally configured their environment to run in Python 3.6 where that one-liner wouldn't work.

Running `mypy --verbose` showed me `setup.cfg` was picked up for configuration, and that's where we see [this line](https://github.com/RDFLib/rdflib/blame/eb151d3b753ce9d705412228ee3b679852a22ef6/setup.cfg#L25).

From `tox.ini`, it seems at least this line should be revised to read Python 3.7.  Is this a one-liner patch to fix in `setup.cfg`, or is there something else that needs to happen between `setup.cfg` and/or `tox.ini` and/or `.drone.yml`?

aucampia:
Support for python version 3.6 is dropped, and using 3.6 prevents
certain features from python 3.7 from being useable without type
failures.

Fixes #1513

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ exclude_lines =
     if __name__==.__main__.:
 
 [mypy]
-python_version = 3.6
+python_version = 3.7
 warn_unused_configs = True
 ignore_missing_imports = True
 disallow_subclassing_any = False


veyndan (approved):

nicholascar (approved):

nicholascar:
We can merge this with just one maintainer approval since it's straightforward!

Issue #1447, opened 2021-10-14
Type annotations? 


pudo:
I've just noticed that rdflib 6.0 marks itself as `py.typed` but that basic types, like `URIRef`, `Literal` and `Namespace` have no type annotations. This makes it impossible to write strictly typed mypy code against the library. Is there some thinking behind this - i.e. are you planning to roll out actual type annotations?

aucampia:
The plan is more or less to do gradual typing, and maybe eventually getting to the point of being able to use this library with strictly typed code, however this will be a lengthy process. `py.typed` is there mainly to inform type checkers that typing information from rdflib comes from rdflib and not somewhere else. Without it type checkers will complain that rdflib is untyped.

Is the presence of `py.typed` with only partial type hints causing specific problems, and if so could you share more detail about it?

eggplants:
Ref: https://github.com/eggplants/rdflib/tree/type_more_strict

Now I'm trying to make mypy more strict.

---

pyparsing will be published as a typed package in the next release soon. (Ref: https://github.com/pyparsing/pyparsing/pull/364)

isodate also will be done. (Ref: https://github.com/gweis/isodate/pull/72)

eggplants:
html5lib has stubs in typeshed as types-html5lib.

berkeleydb is not exist any stubs. So I just created as [berkeleydb-stubs](https://github.com/eggplants/berkeleydb-stubs).

networkx is still on a long way to go. (Ref: https://github.com/networkx/networkx/issues/3988)

aucampia:
For anyone interested, I have a bunch of typing in https://github.com/aucampia/rdflib/tree/iwana-20220126T2212-typing - I will start integrating it with master branch quite soon, but only bit by bit. Don't be discourgaged to add any types though, I rebase regularly as I use this to make it easier to fix bugs.

CC: @ajnelson-nist

aucampia:
@pudo we added a bunch of type hints in 6.2.0 and and will add more in the next release, furthermore there are various options to mitigate the missing type hints in parts of RDFLib, including the option to treat it as untyped using:

https://github.com/python/typing/discussions/911#discussioncomment-1536037

```
[mypy]
[mypy-rdflib.*]
follow_imports = skip
```

We also describe the type hints and the gradual typing process in detail here <https://rdflib.readthedocs.io/en/latest/type_hints.html>.

Unless there is something more you want in this PR I will close it.

[...]

PR #1515, opened 2021-12-17
Start support for mypy --strict

ajnelson-nist:
This patch series attempts to begin a "80/20" strategy suggested on other Issue threads (cited in References) around finding "most commonly used APIs" to address issues introduced with the addition of the `py.typed` file.

The first patches are the minimal necessary to get `mypy --strict` to pass on the test program `test/test_issue1447.py`, which exercises (IIRC, all of) the `rdflib` functions I've used across several projects.  Overall, the patch series adds annotations to methods under `rdflib/`, annotations to a few existing tests under `test/`, a new test under `test/`, and a practice of running the new and just-revised tests to `mypy --strict` in the various testing infrastructures.  Feedback is welcome on some of the type-supporting decisions (especially what I call out as worth squinting eyes) and the sustainability of the testing practice.

I have a few outstanding questions that would be good to answer before merging this PR, and/or give their own Issues if you'd rather merge sooner:

1. I guessed by name pattern on part of integrating the new test.  I assume Issue 1447 is the right reference, though we can start different Issues if it'd be better to focus more tests on specific things.  On the other hand, the 1447 test script could also be a "Feature request" zone where others could contribute (even trivial) function calls that meet their own needs for strict type review.
2. How, and especially where, should `mypy --strict` be called?  Would `docker-compose.tests.yml` get another `command:` line to run `mypy --strict` on just that file?  [Issue 1513](https://github.com/RDFLib/rdflib/issues/1513) came up as I was testing things, and my initial two guesses of which `mypy` call to revise weren't what ended up being revised, whereas my draft of this PR noted the docker script that has its own `mypy` step, and I missed that Github Actions seems to be what runs the CI now.  I'm proceeding in this PR with just modifying Github Actions.
3. Inlined in the 1447 test program - is there a better `rdflib` class name or type designation for `Node`s that are not `Literal`s, analogous to SHACL's [`sh:BlankNodeOrIRI`](https://www.w3.org/TR/shacl/#NodeKindConstraintComponent)?  I frequently find the need to define a `Union` type like I did in the test program.
4. Inlined in the test program - it's not clear to me why `prepareQuery` is not exposed to `mypy` via `rdflib/plugins/sparql/__init__py`, while the test program itself would run just fine calling `rdflib.plugins.sparql.prepareQuery()`.  Any ideas?
5. Squinting review request - The hardest part of this patch series to draft, and now probably the one worth the sternest review, was getting `Literal.toPython()` to work.  The first patch makes that the only `.toPython` method that is typed returning `Any`, while the rest return `str`.  It's possible to make it return a `Union` of types from the various mapping dictionaries, but that causes extra work (either `assert` or `typing.cast`) for any caller trying to pass `mypy --strict`.  As written in the first patch, the typing and value-testing requires minimal amount of annotation work.  Does this practice pass your review, though?

Test status:

`pytest`, `mypy`, and `black` continue to pass as well as they did, as parameterized in `.github/workflows/validate.yml` and `Makefile`, before the addition of this patch.  For review notes:

* `black` currently reports 4 formatting errors in files not touched by this patch series, which I see is `|| true`'d out in the Github Actions CI.
* `test/test_xmlliterals.py::testHTML` currently fails in the `master` branch on my laptop, though, I'm not sure why it's passing on the Github Actions CI.

I'll do catch-up `merge`s if someone lets me know that either is fixed, even a `rebase` if you prefer seeing PRs with all patches bearing a green checkmark from CI.

References:

* https://github.com/RDFLib/rdflib/issues/1447
* https://github.com/python/typing/discussions/911

Disclaimer:

Participation by NIST in the creation of the documentation of mentioned software is not intended to imply a recommendation or endorsement by the National Institute of Standards and Technology, nor is it intended to imply that any specific software is necessarily the best available for the purpose.


aucampia:
Thanks for this, I will have a look at it tomorrow night.

ajnelson-nist:
Well, that's an unfortunate series of X's.

The call to `mypy` that I added was only supposed to review four of the scripts under `test/`.  On my local machine, those are the only four reviewed.  In the Github Action, everything's reviewed.

I determined the difference is in how the package *install* is run:

1. `python setup.py install` triggers this behavior.
2. `pip install .` triggers this behavior.
3. `pip install --editable .` *does not* trigger this behavior.
4. `pip install --editable ${some_git_submodule_dir}/rdflib` *does not* trigger this behavior.
4. `pip install ${some_git_submodule_dir}/rdflib` *does not* trigger this behavior.

Of course, all of my testing directly with this repository was with the last three modes.  (Modes 4 and 5 were how I tested these modifications in other `rdflib`-using repositories I'd like to eventually test with `mypy --strict`.)

So, I guess I need to highlight my original question 2 - how and where would it be appropriate---and heck, even functional---for `mypy --strict` to be called to review test applications?

ajnelson-nist:
Evidence of the `pip install -e` flag getting this patch series to work is in [this CI-debugging PR](https://github.com/ajnelson-nist/rdflib/pull/7).

ajnelson-nist:
Also, I just caught that [PR 1498](https://github.com/RDFLib/rdflib/pull/1498) touches on some of the same content within `Literal`.  Thanks to @veyndan for the `_Any` member definition, it didn't occur to me to try that.

@aucampia , I'd be fine revising this patch series if 1498 is merged first.

veyndan:
> Thanks to @veyndan for the `_Any` member definition, it didn't occur to me to try that.

It was actually @aucampia who added that bit (https://github.com/RDFLib/rdflib/pull/1498/commits/a65e710214c9b9d8a389052372b16e7a14980b77) 👏

aucampia:
> I guessed by name pattern on part of integrating the new test.  I assume Issue 1447 is the right reference, though we can start different Issues if it'd be better to focus more tests on specific things.  On the other hand, the 1447 test script could also be a "Feature request" zone where others could contribute (even trivial) function calls that meet their own needs for strict type review.

I'm not entirely sure if this pattern is considered ideal - I know it has been used but there has also been some resistance to it, and I would say I somewhat agree that it may be better to try and decouple tests from the issue they were in. I think a name like `test_strict_typing.py` may be more appropriate.

I made a pull request to update the developers guide on where tests should go (https://github.com/RDFLib/rdflib/pull/1517) and will run it past others to see what the consensus is on this. I think for now it is fine, but the file name may need to change later.

> How, and especially where, should mypy --strict be called? Would docker-compose.tests.yml get another command: line to run mypy --strict on just that file? Issue 1513 came up as I was testing things, and my initial two guesses of which mypy call to revise weren't what ended up being revised, whereas my draft of this PR noted the docker script that has its own mypy step, and I missed that Github Actions seems to be what runs the CI now. I'm proceeding in this PR with just modifying Github Actions.

I think ideally this should be setup in `setup.cfg` using module specific mypy config: https://mypy.readthedocs.io/en/stable/config_file.html#per-module-and-global-options

So that will entail something like this in setup.cfg:

```ini
[mypy]
python_version = 3.7
warn_unused_configs = True
ignore_missing_imports = True
disallow_subclassing_any = False
warn_unreachable = True

[mypy-test.test_issue1447]
strict = True
```
I have tried it thought and it is not quite doing what I'm expecting, there are some other fallbacks but I will try a couple of things to get it working.

Will respond to other points when I can, hopefully tomorrow.

aucampia:
> How, and especially where, should mypy --strict be called? Would docker-compose.tests.yml get another command: line to run mypy --strict on just that file? Issue 1513 came up as I was testing things, and my initial two guesses of which mypy call to revise weren't what ended up being revised, whereas my draft of this PR noted the docker script that has its own mypy step, and I missed that Github Actions seems to be what runs the CI now. I'm proceeding in this PR with just modifying Github Actions.



@ajnelson-nist I made a PR against your branch to make the validation pass with the selective files running in strict mode: https://github.com/ajnelson-nist/rdflib/pull/8

Would recommend a rebase onto master though as there has been a couple of changes related to mypy and typing since your branch.

rdflib/namespace/__init__.py
```diff
@@ -101,6 +101,9 @@ class Namespace(str):
     False
     """
 
+    def __init__(self, *args, **kwargs) -> None:
```
aucampia, line 104: This is not needed if you type __new__

aucampia, line 104: You should rather just type `__new__` - I think this will have other runtime consequences and typing new works. I will make a PR against your branch to type `__new__`.

aucampia: There are a couple of other places you also do this, I made a PR https://github.com/ajnelson-nist/rdflib/pull/9 that types `__new__` in all cases instead of adding `__init__`.

aucampia:
Please merge https://github.com/ajnelson-nist/rdflib/pull/9 so that you are adding typing to `__new__` instead of adding `__init__`

rdflib/term.py
```diff
@@ -21,6 +21,7 @@
 
 """
 import re
+import typing
```
aucampia, line 24: If the file already uses `from typing import X` I would say it is best to do the same for `Any`

test/test_issue1447.py
```diff
@@ -0,0 +1,140 @@
...
+    python_iri: str = kb_https_uriref.toPython()
+    assert python_iri == "https://example.org/kb/y"
+
+
+def main() -> None:
+    test_rdflib_query_exercise()
+
+
+if __name__ == "__main__":
+    main()
```
aucampia, line 140: I think it is best to not add this, if pytest features are used this likely may not work, better to just run:

```bash
python -m pytest test/test_issue1447.py
```

aucampia:
> 4\. Inlined in the test program - it's not clear to me why `prepareQuery` is not exposed to `mypy` via `rdflib/plugins/sparql/__init__py`, while the test program itself would run just fine calling `rdflib.plugins.sparql.prepareQuery()`.  Any ideas?

mypy has an option:

```
  --no-implicit-reexport    Treat imports as private unless aliased (inverse:
                            --implicit-reexport)
```

This is enabled in --strict mode:

```
  --strict                  Strict mode; enables the following flags: --warn-
                            unused-configs, --disallow-any-generics,
                            --disallow-subclassing-any, --disallow-untyped-
                            calls, --disallow-untyped-defs, --disallow-
                            incomplete-defs, --check-untyped-defs, --disallow-
                            untyped-decorators, --no-implicit-optional,
                            --warn-redundant-casts, --warn-unused-ignores,
                            --warn-return-any, --no-implicit-reexport,
                            --strict-equality
```

For more info see https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport

aucampia:
> * Inlined in the 1447 test program - is there a better `rdflib` class name or type designation for `Node`s that are not `Literal`s, analogous to SHACL's [`sh:BlankNodeOrIRI`](https://www.w3.org/TR/shacl/#NodeKindConstraintComponent)?  I frequently find the need to define a `Union` type like I did in the test program.

Maybe:

```python
Subject = typing.Union[rdflib.BNode, rdflib.URIRef]
```

After all, from https://www.w3.org/TR/rdf11-concepts/#section-triples

> An RDF triple consists of three components:
> - the subject, which is an IRI or a blank node
> - the predicate, which is an IRI
> - the object, which is an IRI, a literal or a blank node

aucampia:
> 5\. Squinting review request - The hardest part of this patch series to draft, and now probably the one worth the sternest review, was getting `Literal.toPython()` to work.  The first patch makes that the only `.toPython` method that is typed returning `Any`, while the rest return `str`.  It's possible to make it return a `Union` of types from the various mapping dictionaries, but that causes extra work (either `assert` or `typing.cast`) for any caller trying to pass `mypy --strict`.  As written in the first patch, the typing and value-testing requires minimal amount of annotation work.  Does this practice pass your review, though?

For now typing `Literal.toPython()` as returning Any is the only option, and maybe the right option also as typing it as more specific types seem to fail. It is anyway not wrong, just over broad at worst, but at best it is maybe correct.

We need to add a lot more typing to help figure out the tightest constraints that make sense, but your changes are a step in the right direction and best to keep them minimal and merge them. I have a bunch more typing changes that are not in PRs yet, and there are a lot more to add, but it is easier if we make them in small batches - it will take some time getting to the point where everything works in strict mode.

aucampia:
[If you give me access to your branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) then I can do some rebases and merge the PRs I made, up to you of course - don't mind either way - but I want to keep this mergeable so we can get it in as soon as we are ready and so it does not hold up other typing changes that will be made.

aucampia:
I think in the long run we should just get mypy passing on test directory without strict, and then generally start adding additional errrors/warnings - not something to add here, but just a general view on what to do to get things moving along.

rdflib/namespace/__init__.py
ajnelson-nist: Agreed, I'm fine not adding the `__init__` methods.  I'll make sure the `Node.__init__()` gets reverted too.

rdflib/term.py
ajnelson-nist: Agreed, I'd missed the symbol import below.  Reverted.

test/test_issue1447.py
ajnelson-nist: Thanks, I wasn't clear on what the transition status was from `unittest` to `pytest`.  The recent doc updates help clarify the matter too.  Removing.

aucampia:
I'm happy with this change, something which is maybe debateable is on which test files strict mode makes sense. I'm fine with it on the ones you enabled it for but would just want the other maintainers to give some feedback there. (CC: @ashleysommer @nicholascar @white-gecko ).

As I said in a comment before, I want to make `mypy` run on all test in normal non strict mode, basically all that will be needed for this is removal of `ignore_errors = True` ([this](https://github.com/RDFLib/rdflib/blob/7bf70e4dab159291a291cb84fccc406d2bbfe429/setup.cfg#L38)) after fixing the issues raised - but it would be good to have it run on strict mode on at least some tests specifically designed to verifying typing, and for that I think `test/test_typing.py` is ideal. I will look at that after this PR is merged as I think the setup here provides a good basis for it.

test/test_issue1447.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
ajnelson-nist: I did see your earlier remark about naming this to reflect its purpose of exercising strict typing.  I'm fine with going away from the number-style name.  I'll rename and add a remark on the purpose being to exercise strict typing, to cover `test_typing.py` being too broad.

ajnelson-nist:
I've adjusted a few more style matters, but otherwise think this patch series is ready for review and merge.

@aucampia , I agree with the general direction of eventually removing `ignore_errors`.

ajnelson-nist:
Also, it seems I'm getting some CI chaos from networking-stack tests.  Hopefully that doesn't confuse review.

aucampia (approved):

nicholascar (approved):
Definitely steps in the right direction! Yes the literal typing is still not great. Last year I added a whole bunch more tests for literal types but not mypy or type hinting. It gets really hard to cater for all XSD datatypes AND Python datatypes AND custom datatypes since there's no guarentee that they all line up.

PR #1516, opened 2021-12-17
Increase fuseki start timeout from 15 to 30 seconds

aucampia:
Fuseki quite regularly takes longer than 15 seconds to start on MacOS.

We could maybe later consider only running it on Linux as it is not
likely to result in platform specific errors.

### DIFF
--- a/with-fuseki.sh
+++ b/with-fuseki.sh
@@ -7,7 +7,7 @@ script_dirname="$( dirname -- "${0}" )"
 set -eo pipefail
 
 wait_http_okay() {
-    local timeout=15
+    local timeout=30
     for var in "${@}"
     do
         eval "local ${var}"


nicholascar:
> We could maybe later consider only running it on Linux as it is not
likely to result in platform specific errors.

I think we should consider this very soon. While we have reduced the duplicate testing, we are still testing lots of Python versions and platforms and it seems a bit excessive. 

Let’s try and get all the tests running on the lowest version only Py37 and all the optional tests - BerkeleyDB, SPARQL ones using Fuseki etc. - just on the easiest platform (Linux).

I don’t think we can predict exactly which tests might have changes per Python version or platform but there are probably a few chunks, like these SPARQL ones that are pretty safe to not always test.

aucampia:
> Let’s try and get all the tests running on the lowest version only Py37 and all the optional tests - BerkeleyDB, SPARQL ones using Fuseki etc. - just on the easiest platform (Linux).

Okay, sounds like a good approach, will make this change when I get time. Can merge this if it causes trouble otherwise I will effectuate the suggestion somewhere in the coming week.

nicholascar:
It’s a simple change so let’s merge to keep the pending PR count down!

PR #1522, opened 2021-12-21
Add typing for parsers

aucampia:
This changeset include no runtime changes in rdflib.

I will mark this as non draft after #1515 is handled - and potentially one more PR to make mypy pass on test directory.

### DIFF
--- a/test/test_turtle_w3c.py
+++ b/test/test_turtle_w3c.py
@@ -28,6 +28,7 @@ def turtle(test: RDFTest):
 
         if test.result:  # eval test
             res = Graph()
+            assert not isinstance(test.result, tuple)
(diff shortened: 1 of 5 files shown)


test/test_turtle_w3c.py
```diff
@@ -28,6 +28,7 @@ def turtle(test: RDFTest):
 
         if test.result:  # eval test
             res = Graph()
+            assert not isinstance(test.result, tuple)
```
aucampia, line 31: This assert is here because `test.result` can be a tuple or Identifier, and if it is a tuple then it cannot be use as a parse source.

nicholascar (approved):
Merging without second review as simple PR - no runtime chagnes

PR #1529, opened 2021-12-24
Fix `self.line` typos in call to BadSyntax.

### LINKED ISSUE #821: Invalid URI crashes without BadSyntax error
I am cursed with URIs with % escape codes in them.  If one comes through with an invalid escape code (e.g., `%F` rather than `%2F`) rdflib crashes with a message `AttributeError: 'SinkParser' object has no attribute 'line'`.

I'm not able to install the github code to check if this is still a problem (I'm using rdflib 4.2.2 via python3) but a scan of notation.n3 suggests that this is an condition that ought to generate a BadSyntax error?

eg: `doi:10.1257%2Fjep.27.1.223` accidentally written as `doi:10.1257%2jep.27.1.223`

gjhiggins:
Fixes #821

## Proposed Changes

  - Change incorrect references from `self.line` to `self.lines` in 3 calls to BadSyntax in notation3 parser.

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -1362,7 +1362,7 @@ class SinkParser:
                         if c not in escapeChars:
                             raise BadSyntax(
                                 self._thisDoc,
-                                self.line,
+                                self.lines,
                                 argstr,
                                 i,
                                 "illegal escape " + c,
@@ -1374,7 +1374,7 @@ class SinkParser:
                         ):
                             raise BadSyntax(
                                 self._thisDoc,
-                                self.line,
+                                self.lines,
                                 argstr,
                                 i,
                                 "illegal hex escape " + c,
@@ -1386,7 +1386,7 @@ class SinkParser:
 
             if lastslash:
                 raise BadSyntax(
-                    self._thisDoc, self.line, argstr, i, "qname cannot end with \\"
+                    self._thisDoc, self.lines, argstr, i, "qname cannot end with \\"
                 )
 
             if argstr[i - 1] == ".":


nicholascar (approved):
Approving and merging as a simple PR (but good to fix the bug!)

gjhiggins:
> Approving and merging as a simple PR (but good to fix the bug!)

There's a few similar small-bugfix PRs stacked up, I bravely resisted the temptation to bulk 'em up into a single PR, so sorry for the extra work on your side.

nicholascar:
Definitely the right thing to do to keep ‘em all small! Happy to do the extra resulting reviews.

Really glad you are grinding through all these as these plus all the testing and environment improvements - top etc. - means the codebase is really neatening up. I’m really happy about this as I’d like to see some some major changes for a 7.0.0 early next year but ideally they should be  implemented on top of tidy code rather than having to tidy code within their major changes.

The major changes are: 

* your work on Graph ID
* ConjunctiveGraph / Dataset dedup
* RDF-Star
* JSON-LD 1.1 completion
* Store comparison / alignment
* SPARQLWrapper something or other (fix / retire / integrate / _something_)

gjhiggins:
> Definitely the right thing to do to keep ‘em all small! Happy to do the extra resulting reviews.

Good to know, trying to stay out of your way.

> Really glad you are grinding through all these as these plus all the testing and environment improvements - top etc. - means the codebase is really neatening up. 

To provide a modicum of context - at 70, I'm officially a pensioner but "semi-retired" in that I've still got some skin in the game - I use RDFLib as a back-end for ACME (A Cryptocurrency Metadata Explorer), so I'm quite happy to have something to occupy me other than daytime TV, even better that its purposeful.

> I’m really happy about this as I’d like to see some some major changes for a 7.0.0 early next year but ideally they should be implemented on top of tidy code rather than having to tidy code within their major changes.

As I've observed in the Discussion welcome, 2022 is RDFLib's 20th anniversary. Would be nice to see a neat codebase with a clear set of maintenance tasks and decent documentation. It's fortuitous that Github has integrated Discussions/Issues in such as way as to add a back-and-forth path, that really helps with migrating not-actually-issues out of the issues list and clarifying the remaining issues. 

>     * your work on Graph ID
>     * ConjunctiveGraph / Dataset dedup

I'm not familiar with the dedup intentions and I really should be, is there a pointer? gromgull summarised his understanding of the two in #307, I'm pretty much working to that model.

>     * Store comparison / alignment
I've made some reasonable progress with an ansible/vagrant script for this purpose, it's on the back-burner for the time being.

Issue #944, opened 2019-12-03
Only URIRefs or Paths can be in paths!


charbull:
Hello, 

I am trying to run the example in https://rdflib.readthedocs.io/en/stable/apidocs/rdflib.extras.html

And the following error `Only URIRefs or Paths can be in paths!` keeps occurring at `anonClass = EX.someProp | some | classD `

```
EX = Namespace('http://example.com/')
namespace_manager = NamespaceManager(Graph())
namespace_manager.bind('ex', EX, override=False)
namespace_manager.bind('owl', OWL_NS, override=False)
g = Graph()
g.namespace_manager = namespace_manager
Individual.factoryGraph = g
classD = Class(EX.D)
 anonClass = EX.someProp | some | classD 
```

Thank you

JervenBolleman:
@charbull is it because some is not defined? I expect you want to make an restriction like someValuesFrom D?

gjhiggins:
I've been meaning to look into this for some time. It transpires that the SPARQL paths.py code is triggering on the `|` delimiters used for the Infix class, hence the unexpected Exception. Happily, there is an alternative pair of delimiters: `<<` and `>>` and the examples all work when `| some |` is replaced by `<< some >>`, similarly for `| max |` and `<< max >>`.

PR #1530, opened 2021-12-25
Fix for infixowl issues 1453 and 944

### LINKED ISSUE #1453: infixowl cardinality - 0
`owl.Restriction(property,graph=graph,cardinality=Literal(0))
`Raises an error. 
`assert len(validRestrProps)`
When cardinality value is given as 0.

This comes from:
```
  validRestrProps = [(i, oTerm) for (i, oTerm) in restrTypes if i]
  assert len(validRestrProps)
```

I believe this is an oversight as if i == 0 returns False.
However, I believe cardinality: 0 is valid.

gjhiggins:
Fixes #1453 and #944

## Proposed Changes

  - Change falsy `if i` to `if i is None` to avoid improper Exception when cardinality == 0 (#1453)
  - Change doctest examples to use `<< >>` infix delimiters to avoid `|` clashing with SPARQL paths (#944)
  - Translate/transcribe doctests as pytests in separate file as a foundation for developing some unit tests
  - unskip some (passing) doctests


test/test_infixowl.py
```diff
@@ -0,0 +1,375 @@
...
+    Individual,
+    some,
+    EnumeratedClass,
+    Property,
+    Restriction,
+    max,
+)
+
+
+# @pytest.mark.skip
```
aucampia, line 18: Can be removed.

```suggestion
```

test/test_infixowl.py
```diff
@@ -0,0 +1,375 @@
...
+    )
+
+    Property(exNs.hasParent, graph=g) << max >> Literal(1)
+    assert (
+        repr(Property(exNs.hasParent, graph=g) << max >> Literal(1))
+        == "( ex:hasParent MAX 1 )"
+    )
+
+
+# @pytest.mark.skip
```
aucampia, line 124: Can be removed.
```suggestion
```

test/test_infixowl.py
```diff
@@ -0,0 +1,375 @@
...
+    # Now we can assert rdfs:subClassOf and owl:equivalentClass relationships
+    # (in the underlying graph) with other classes using the 'subClassOf'
+    # and 'equivalentClass' descriptors which can be set to a list
+    # of objects for the corresponding predicates.
+
+    a.subClassOf = [exNs.MusicalWork]
+
+    # We can then access the rdfs:subClassOf relationships
+
+    assert repr(list(a.subClassOf)) == "[Class: ex:MusicalWork ]"
```
aucampia, line 46: I think technically `__repr__` is doing the wrong thing here as far as this is concerned: https://docs.python.org/3/reference/datamodel.html#object.__repr__

Fine to test for what it is doing but probably we should fix it in a major release.

aucampia: If you agree please let me know, I will make an issue.

aucampia (approved):

nicholascar (approved):
Approving but not merging in case @gjhiggins wants to act on @aucampia's suggestions in comments

gjhiggins:
> Approving but not merging in case @gjhiggins wants to act on @aucampia's suggestions in comments

How _diplomatic_ of you both. I got derailed, of _course_ I meant to use `str` ;)

test/test_infixowl.py
gjhiggins: I got derailed, `str` is what I _should_ have been using. Thanks for your unfailingly reliable backstop, it's much appreciated.

aucampia:
I will do final check tomorrow night.

rdflib/extras/infixowl.py
```diff
@@ -195,6 +196,12 @@ def __rlshift__(self, other):
     def __rshift__(self, other):
         return self.function(other)
 
+    def __rmatmul__(self, other):
+        return Infix(lambda x, self=self, other=other: self.function(other, x))
+
+    def __matmul__(self, other):
+        return self.function(other)
+
```
aucampia, line 204: Are these really needed? As far as I understand this makes `A @ B @ C` do the same as `A | B | C` and `A << B >> C`.

I think this Infix OWL has diverged from the [spec](https://www.w3.org/TR/owl-semantics/syntax.html) somewhat but [@ is reserved in there for language tags](https://www.w3.org/TR/owl-semantics/syntax.html#2.2) I think so it may be best to leave it for that.

I'm not an expert at OWL manchester syntax, so I may be missing something here - maybe @ashleysommer or @nicholascar has some input about it.

aucampia (approved):
Change looks good, just want someone else to explicitly confirm the addition of `__matmul__` and `__rmatmul__`.

rdflib/extras/infixowl.py
gjhiggins: As I read it, when gromgull migrated the SPARQL paths processor `paths.py` out of the sparql module back in 2103, the change from monkeypatching Path to directly defining Path methods [definitions](https://github.com/RDFLib/rdflib/blob/43d86224382d64fba2e5e17acedea1704631e097/rdflib/paths.py#L199) overrode the `|` (`__or__` and `__ror__`)  infixOwl delimiters causing exceptions in the infixOwl doctests. Hence my editing the docstrings to use the existing alternative implementation of `<<` and `>>` (`__ rshift` and `__rlshift__`) instead.

I subsequently discovered that unfortunately, SPARQL* uses the Turtle* double-anglebrackets and, anticipating that addition to `Path`, I tried to find an alternative to use with infixOwl. As far as I could ascertain, matmul was the only remaining option but yes, I guess it will  clash with `@` as lang tag so there doesn't appear to be any ready solution for enabling infixOwl to work in RDFLib with the “colloquial programming style” described in the infixOwl docstrings. 

>  think this Infix OWL has diverged from the spec somewhat

TBH, those specific infix operators aren't part of the spec, they were added by Chimezie as part of FuXi and weren't in the RDFLib 2.4 of the time. I'm not going to spend time arguing for the retention of infixOwl if the implementation has aged to a degree that it's now basically out of scope, other than to observe the irony of Harold (funowl) Solbrig's [respect for it](https://github.com/hsolbrig/funowl#other-packages).

aucampia: I'm fine kind of with `@` being used for this, like you said there is already some deviations and using `@` seems like a minor one, though still just want a double check from @nicholascar or someone else on this point.

nicholascar: I think it sensible to avoid `<<` & `>>` as they are indeed very likely to be implemented for Turtle-Start soon, as noted above.

We can always expect to come up with notation/symbol clashes from time-to-time given that we are dealing with:

* Python
* SPARQL
* multiple RDF syntaxes

..all in one.

So, I think it fine to implement `@` here for now and just have to deal with any follow-on issues later, if further syntax conflicts appear.

I am keen to get (@hsolbrig)'s Manchester Syntax into RDFLib (perhaps in 7.0.0??) but patching up preexisting packages first seems sensible.

nicholascar: I replied inline about this elsewhere: TL;DR: fine for now.

nicholascar:
Merging in with `__matmul__` and `__rmatmul__` (use of `@`) adnd will have to review if/when other Manchester syntax implementations, like @hsolbrig's appear.

PR #1607, opened 2021-12-27
Revert error-raising change, enable Exception to be raised.

### LINKED ISSUE #1216: Parsing N3 with a @base IRI which does not include a slash
The issue is illustrated by this gist:

https://gist.github.com/anatoly-scherbakov/9fafb2863b877991f56ac7766b7c1bf0

* I use `@base <local:> .` in an RDF/N3 document, which should convert an RDF term `<Category>` to `<local:Category>`.
* But instead I get an error raised.
* If I try `<local:/>` instead as a `@base` — everything works.

But I was trying to get `<local:>` working, and also, I believe, `<local:Category>` is a perfectly good IRI. Real world examples of such schemas may be `doi` and `mailto`.

This might be related to #816 but I am not certain of that.

rdflib version is 5.0.0. The exception is raised here:

https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/parsers/notation3.py#L139-L144

I would be happy to create a PR removing this check, but I would like first to understand why the check is implemented.

gjhiggins:
Fixes #1216 in terms of not raising an error but the originally-intended Exception.
Doesn't address “But I was trying to get \<local:\> working, and also, I believe, \<local:Category\> is a perfectly good IRI.”

## Proposed Changes

  - Revert error-raising change, enable Exception to be raised.

### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -133,7 +133,7 @@ def join(here, there):
         return here + frag
 
     # join('mid:foo@example', '../foo') bzzt
-    if here[bcolonl + 1] != "/":
+    if here[bcolonl + 1 : bcolonl + 2] != "/":
         raise ValueError(
             "Base <%s> has no slash after "
             "colon - with relative '%s'." % (here, there)


nicholascar (approved):
Approving and merging without second review as it's a straightforward fix

PR #1620, opened 2021-12-27
Compile docs in GitHub Actions CI

### LINKED ISSUE #1521: Add validation for sphinx in CI
Currently document is not being built with sphinx in the validation pipeline, leaving open the possibility that it fails to build. This should be remedied so that any potential failures in sphinx can be picked up before PRs are merged.

aucampia:
This is to make sure that there are no doc build failures introduced by
PRs.

In future I want to expand this check to be more strict. Since it comes up
separately in CI it can be ignored on reviwer discrescion but at least
it will make problems visible.

Also:
- Fix some things in docs that were causing warnings. There are still
  many warnings related to references but I will fix them later.
- Added a `docs` evnrionment for tox which builds the docs, this is also
  what is used from CI.

Fixes #1521

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -9,6 +9,7 @@ on:
 env:
   FORCE_COLOR: 1
   XDG_CACHE_HOME: ${{ github.workspace }}/cache
+  DEFAULT_PYTHON: "3.7"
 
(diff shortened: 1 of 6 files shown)


nicholascar (approved):
Great idea @aucampia

Issue #1524, opened 2021-12-21
TypeErrors from Results do not propagate through list creation


aucampia:
In https://github.com/RDFLib/rdflib/issues/1455 I noticed that `TypeError`s raised from `rdflib.query.Result` get lost when they occur inside the `list` constructor.

I created some tests that reproduce this and marked them with xfail:
- https://github.com/RDFLib/rdflib/pull/1523

As far as i can tell this happens because `Result._genbindings` gets called twice, and the first time it throws the exception, the second time it just behaves like nothing but there are no bindings.

The reason why it get's called twice for TypeError is likely because the list constructor has special handling for TypeError as can be seen here: https://github.com/python/cpython/blob/e9898bf153d26059261ffef11f7643ae991e2a4c/Objects/listobject.c#L2808-L2842

This problem does not occur for RuntimeError.

I think the right fix is to ensure that `Result._genbindings` consistently fails when it gets called again after the first failure, but I'm still looking into it.

PR #1623, opened 2021-12-28
Fix issue1032 error on sparqlstore update

### LINKED ISSUE #1032: Graph.update with SPARQLStore should return Exception
When I instantiate a `Graph` with the `SPARQLStore` and call `query()` on it I get:

    TypeError: update() takes from 2 to 3 positional arguments but 5 were given

while I would expect an Exception.

Reproduce it with:

    from rdflib import Graph
    graph = Graph(store="SPARQLStore")
    graph.update("insert data {graph <urn:ex> {<urn:s> <urn:p> <urn:o>}}")

This is caused by the fact, that the `SPARQLStore` inherits from `SPARQLConnector`:

    class SPARQLStore(SPARQLConnector, Store):

Two issues:

- Why does `SPARQLStore` inherit from `SPARQLConnector`?
- The `SPARQLStore` should define `update` and return en Exception.

gjhiggins:
Fixes #1032

## Proposed Changes

  - Add missing `update()` method which returns "Store not writeable" Exception.

### DIFF
--- a/rdflib/plugins/stores/sparqlstore.py
+++ b/rdflib/plugins/stores/sparqlstore.py
@@ -154,6 +154,9 @@ class SPARQLStore(SPARQLConnector, Store):  # type: ignore[misc]
     def remove(self, _, context):
         raise TypeError("The SPARQL store is read only")
 
+    def update(self, query, initNs={}, initBindings={}, queryGraph=None, DEBUG=False):
+        raise TypeError("The SPARQL store is read only")
+
     def _query(self, *args, **kwargs):
         self._queries += 1
 


nicholascar (approved):
Approving and merging without a second review as it's a straightforward fix

aucampia (approved):
Made a PR to add tests to your banch (https://github.com/gjhiggins/rdflib/pull/1)

Change looks good, `TypeError` may not be the best exception to raise for this situation and I think `NotImplementedError` may be better, but since all other modification methods raise `TypeError` it makes most sense to do the same for `SPARQLStore.update`. We can look at changing all exceptions later versions (noted here https://github.com/RDFLib/rdflib/discussions/1628). As a side note on this, `TypeError` is handled differently from other exceptions in many places: see https://github.com/RDFLib/rdflib/issues/1524 for example.

PR #1624, opened 2021-12-28
Feature prepareupdate

### LINKED ISSUE #272: prepareUpdate method for pre-parsing SPARQL update statements
i.e. matching prepareQuery

gjhiggins:
Fixes #272
Slightly speculative, other option is close #272 with `wontfix`

## Proposed Changes
Basically, follow the pattern established in `prepareQuery`

  - Create an Update class
  - Add a `prepareUpdate` to `processor.py`
  - Propagate pattern
  - Add test.

### DIFF
--- a/rdflib/plugins/sparql/algebra.py
+++ b/rdflib/plugins/sparql/algebra.py
@@ -13,7 +13,7 @@ from functools import reduce
 
 from rdflib import Literal, Variable, URIRef, BNode
 
-from rdflib.plugins.sparql.sparql import Prologue, Query
+from rdflib.plugins.sparql.sparql import Prologue, Query, Update
 from rdflib.plugins.sparql.parserutils import CompValue, Expr
 from rdflib.plugins.sparql.operators import (
     and_,
@@ -761,7 +761,7 @@ def translateUpdate(q, base=None, initNs=None):
 
         res.append(translateUpdate1(u, prologue))
 
-    return res
+    return Update(prologue, res)
 
 
 def translateQuery(q, base=None, initNs=None):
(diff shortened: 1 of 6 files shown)


aucampia (approved):
Looks great.

Notes for other reviewers: I checked that all uses of `rdflib.plugins.sparql.algebra.translateUpdate` will work with new return value, and they will since they all just pass to `rdflib.plugins.sparql.update.evalUpdate`'s second parameter, `update` and this function was updated to access `update.algebra` instead of `update`.

rdflib/plugins/sparql/algebra.py
```diff
@@ -761,7 +761,7 @@ def translateUpdate(q, base=None, initNs=None):
 
         res.append(translateUpdate1(u, prologue))
 
-    return res
+    return Update(prologue, res)
```
aucampia, line 764: Just a technical note, similar to what I raised in #1626 

This technically changes the RDFLib API, I'm not sure that we consider the methods in here part of the public API, but I think if we don't we should make a note in the docs. Happy to do that if there is agreement on it.

gjhiggins: > This technically changes the RDFLib API, I'm not sure that we consider the methods in here part of the public API, but I think if we don't we should make a note in the docs. Happy to do that if there is agreement on it.

I guess from a pragmatic perspective, "public API" is anything that *can* be imported and the SPARQL API is very sparsely documented but yes, agreed.

aucampia: I added it to me my to-do list to note these methods are not considered part of the public interface and may break compatibility within a single major version and will do it in a separate PR.

nicholascar: Perhaps it's breaking the public API but it's making it more consistent so I suppose we just release a slightly more significant version to include this, perhaps 6.2.0.

nicholascar (approved):

Issue #1593, opened 2021-12-26
questions about DefinedNamespaces (OWL, RDFS, etc etc)


jclerman:
I have some questions about the new `DefinedNamespace`:

1. Can we restore a non-private API to get the URI of the namespace itself? Something like the `uri()` method on the now-deprecated `ClosedNamespace`:
```
rdflib.OWL.uri()   # "http://www.w3.org/2002/07/owl#"
```
2. I couldn't find a documented way to get the complete set of URIs within a DefinedNamespace.  It appears necessary to do something like this:
```
from rdflib import OWL, URIRef
owl_uris: Set[URIRef] = {OWL[str(x)] for x in OWL.__annotations__}  # this can be generalized to a function
```
It would be nice to have the above functionality implemented as, say, a method (on `DefinedNamespaceMeta`?).  On `ClosedNamespace`, it looks like it was possible to get this result via `dir(OWL)`, since `__dir__` was defined on that class.

edmondchuc:
> Can we restore a non-private API to get the URI of the namespace itself? Something like the uri() method on the now-deprecated ClosedNamespace

You can get the URI of a `DefinedNamespace` by calling `str()` on it.

```python
>>> import rdflib
>>> str(rdflib.OWL)
'http://www.w3.org/2002/07/owl#'
```

> I couldn't find a documented way to get the complete set of URIs within a DefinedNamespace. It appears necessary to do something like this

I've added this as a feature in this PR https://github.com/RDFLib/rdflib/pull/1626.

jclerman:
> You can get the URI of a DefinedNamespace by calling str() on it.

Thanks @edmondchuc! I realized that I also seem to have figured that out in another part of my brain, since I used it in the 2nd part of my question.

> I've added this [a way to get the complete set of URIs within a DefinedNamespace ] as a feature in this PR #1626.

Nice, thank you.  Regarding that PR, upon further thought, overriding `__dir__` is perhaps not the best approach - it'd break code for anyone relying on the default behavior of `dir(OWL)`.  Is there any reason not to create a method `uris()` on the DefinedNamespace class instead?

Related but separate: I'm not clear why `OWL`, `RDF`, etc aren't designed as instances of the `DefinedNamespace` class, instead of subclasses of it.  I would have thought that making them instances would be more intuitive - is there an advantage to them being classes?

jclerman:
[...]

PR #1626, opened 2021-12-28
Add __dir__ to DefinedNamespaceMeta.

edmondchuc:
Fixes https://github.com/RDFLib/rdflib/issues/1593

## Proposed Changes

  - Add `__dir__` to `rdflib.namespace.DefinedNamespaceMeta`.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -1,7 +1,6 @@
 import logging
 import warnings
-from collections import Iterable
-from typing import TYPE_CHECKING, List, Union
+from typing import TYPE_CHECKING, List, Union, Iterable
 from unicodedata import category
 
 from pathlib import Path
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import warnings
+from collections import Iterable
 from typing import TYPE_CHECKING, List, Union
 from unicodedata import category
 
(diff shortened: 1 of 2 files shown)


aucampia (approved):

rdflib/namespace/__init__.py
```diff
@@ -228,6 +228,10 @@ def __contains__(cls, item):
             if issubclass(c, DefinedNamespace)
         )
 
+    def __dir__(cls) -> Iterable[str]:
```
aucampia, line 231: Technically this breaks the interface, as anything that depended on the default dir implementation would break. Not sure if this is a serious concern, but something we should think about.

```console
$ pipx run --spec rdflib==6.1.1 python3 -c 'from rdflib import RDF; print(dir(RDF))'
['_NS', '__annotations__', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_fail', '_underscore_num']
```

Other option is to provide a different method for this.

nicholascar: > I highly doubt that this would be a problem for anyone though.

Yes, I highly doubt this will be a problem too and I'm in favour of reporposing the well-known `__dict__` for this, rather than adding a new method that people have to learn

nicholascar (approved):

PR #1629, opened 2021-12-28
Add tests for update method on `Graph(store="SPARQLStore")`

aucampia:
This is to confirm that `Graph(store="SPARQLStore").update()` raises an
exception that indicates the Graph is read only.


nicholascar (approved):
Looks good and adding a test only, no major functionality, so merging without second review.

PR #1630, opened 2021-12-29
Add classifier for python 3.10

aucampia:
This is so the version shield ( [PyPI](https://pypi.python.org/pypi/rdflib) ) shows up correctly in future releases.


nicholascar (approved):
Thanks, I couldn't remember what the trick was to add 3.10 in as it's been a long time since I made one of these badges!

nicholascar:
Merging without a second review as this is a simple update

PR #1633, opened 2021-12-29
Add some typing for evaluation related functions in the SPARQL plugin.

aucampia:
Added this while debugging #1524.

This change contains no runtime changes.

### DIFF
--- a/rdflib/plugins/sparql/evaluate.py
+++ b/rdflib/plugins/sparql/evaluate.py
@@ -17,6 +17,7 @@ also return a dict of list of dicts
 import collections
 import itertools
 import re
+from typing import Any, Deque, Dict, List, Union
 from urllib.request import urlopen, Request
 from urllib.parse import urlencode
 import json as j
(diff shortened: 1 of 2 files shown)


rdflib/plugins/sparql/evaluate.py
```diff
@@ -299,7 +301,7 @@ def evalPart(ctx, part):
         raise Exception("I dont know: %s" % part.name)
 
 
-def evalServiceQuery(ctx, part):
+def evalServiceQuery(ctx: QueryContext, part):
```
nicholascar, line 304: Any reason this `part` doesn't have `CompValue` like the others?

nicholascar (approved):
Merging without second review as this PR just adds types

rdflib/plugins/sparql/evaluate.py
aucampia: No particular reason, I was just adding types which were either very clear or part of the code being executed, the untyped values remain `typing.Any`, but should ideally eventually also be typed with gradual typing.

Issue #1451, opened 2021-10-17
Tests for translateAlgebra are not integrated with other unit tests


aucampia:
As far as I can tell `translateAlgebra` tests (stuff in `test/translate_algebra/...`) are not integrated with other unit tests and will not run when `run_tests.py` or `nosetests` run.

```
$ \rm -rv .venv
$ python3.7 -m venv .venv
$ .venv/bin/python3 -m pip install -r requirements.txt -r requirements.dev.txt
$ export PYTHONPATH=$(readlink -f .):$(readlink -f .venv)
$ .venv/bin/python test/translate_algebra/main.py
Traceback (most recent call last):
  File "test/translate_algebra/main.py", line 1, in <module>
    from test_base import Test, TestExecution, format_text
  File "/home/iwana/d.x/github.com/RDFLib/rdflib/test/translate_algebra/test_base.py", line 1, in <module>
    import pandas as pd
ModuleNotFoundError: No module named 'pandas'
$ .venv/bin/python3 -m pip install pandas
$ .venv/bin/python test/translate_algebra/main.py
Traceback (most recent call last):
  File "test/translate_algebra/main.py", line 1, in <module>
    from test_base import Test, TestExecution, format_text
  File "/home/iwana/d.x/github.com/RDFLib/rdflib/test/translate_algebra/test_base.py", line 5, in <module>
    from tabulate import tabulate
ModuleNotFoundError: No module named 'tabulate'
$ .venv/bin/python3 -m pip install tabulate
```

Furthermore it seems these tests need `pandas` and `tabulate` and seems to be expecting a file `test/config.ini` with a section `TEST`

```
$ .venv/bin/python test/translate_algebra/main.py
Executing tests ...
Traceback (most recent call last):
  File "test/translate_algebra/main.py", line 889, in <module>
    t.run_tests()
  File "/home/iwana/d.x/github.com/RDFLib/rdflib/test/translate_algebra/test_base.py", line 107, in run_tests
    logging.getLogger().setLevel(int(self.test_config.get("TEST", "log_level")))
  File "/usr/lib64/python3.7/configparser.py", line 780, in get
    d = self._unify_values(section, vars)
  File "/usr/lib64/python3.7/configparser.py", line 1146, in _unify_values
    raise NoSectionError(section) from None
configparser.NoSectionError: No section: 'TEST'
```

[...]

PR #1635, opened 2021-12-29
Run translate_algebra tests from pytest

aucampia:
**WARNING**: Still in draft - just placing it here to share with @GreenfishK - likely it will be completely different for merging.

This change adds a single pytest test that just runs the tests from
the custom testing framework used in `test/translate_algebra`
and asserts that they all passed.

The next step will be convert those tests into pytest tests.

See also:
- #1451

### DIFF
--- a/requirements.dev-extra.txt
+++ b/requirements.dev-extra.txt
@@ -1,2 +1,7 @@
 berkeleydb
 networkx
+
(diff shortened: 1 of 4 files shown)


aucampia:
superseded by https://github.com/RDFLib/rdflib/pull/1636

PR #1639, opened 2021-12-31
Rename `test/triple_store.py` so pytest picks it up

aucampia:
I'm not sure if this really tests anything of use, but at the moment it
is not being picked up by pytest.

This changeset renames the file to `test/test_triple_store.py` so that
pytest picks it up.

The alternative is to just remove this, I'm sure it is already tested in other places, would happily change this PR to rather do that.


nicholascar:
What about `test/type_check.py` too?

aucampia:
> What about `test/type_check.py` too?

The tests in there will fail, see this for details:

- https://github.com/RDFLib/rdflib/issues/1641

I made a draft PR to remove it together with some other unused code, but it remains draft because as mentioned in #1641 I want to confirm that there is agreement on removing it:

- https://github.com/RDFLib/rdflib/pull/1640

nicholascar (approved):
Merging after only one review as this is a simple PR

PR #1640, opened 2022-01-01
Remove `(TypeCheck|SubjectType|PredicateType|ObjectType)Error` and related

aucampia:
Also remove
`check_(context|subject|predicate|object|statement|pattern)`.

It seems nothing is using these exceptions and functions.

Technically this does remove parts of the public API, but I would argue
they are "buggy" parts as anything that use them would be sorely
disappointed to find that the behaviour is not as expect, and these
functions/exceptions don't serve any function related to the aim of
RDFLib unless they are integrated with the rest of RDFLib.

I'm not sure this is the best approach, so any other options are welcome. Fixes https://github.com/RDFLib/rdflib/issues/1641


nicholascar:
I don't recall ever knowing or caring about these tests so I'm happy for them to be removed.

nicholascar (approved):

gjhiggins (approved):

PR #1642, opened 2022-01-01
Black rdflib except for rdflib/namespace/_GEO.py

aucampia:
Probably we should ingore generated namespaces in black or format them
on generation, or mark them for ignoring by formatter. Will look at that
later.


nicholascar (approved):
Will merge without a second review as it's a simple PR

PR #1646, opened 2022-01-06
Move Store API to work with identifiers, not graphs

### LINKED ISSUE #167: Clarify confusion around type of context element in ConjunctiveGraphs  and context aware stores

gjhiggins:
This is a rebase of PR #958, PR #409 rebased against master to fix #167

## Proposed Changes
As per original PR, all instances of Graph objects as contexts changed to Graph identifiers.
  - graph.py
  - auditable.py
 - notation3.py
 - sparqlstore.py
 - store.py
 - update.py
 - corresponding changes to tests and three annotations of failure.


gjhiggins:
The default assignment of a BNode identifier to Graph is causing some blocking issues:
```py
>>> from rdflib import Graph, URIRef
>>> tarek = URIRef("urn:x-rdflib:tarek")
>>> likes = URIRef("urn:x-rdflib:likes")
>>> pizza = URIRef("urn:x-rdflib:pizza")
>>> g = Graph(store="SPARQLUpdateStore")
>>> g.open(
...     configuration=(
...         "http://localhost:3030/db/query",
...         "http://localhost:3030/db/update",
...     )
... )
>>> g.add((tarek, likes, pizza))
  File ".../rdflib/graph.py", line 431, in add
    self.__store.add((s, p, o), self.identifier, quoted=False)
  File ".../rdflib/plugins/stores/sparqlstore.py", line 628, in add
    q = "INSERT DATA { GRAPH %s { %s } }" % (nts(context), triple)
  File ".../rdflib/plugins/stores/sparqlstore.py", line 33, in _node_to_sparql
    raise Exception(
Exception: SPARQLStore does not support BNodes! See http://www.w3.org/TR/sparql11-query/#BGPsparqlBNodes
```

Although it wasn't included in the original PR, this change:

```diff
diff --git a/rdflib/graph.py b/rdflib/graph.py
index cab825b1..69e31c32 100644
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1940,7 +1940,7 @@ class Dataset(ConjunctiveGraph):
     """
 
     def __init__(self, store="default", default_union=False, default_graph_base=None):
-        super(Dataset, self).__init__(store=store, identifier=None)
+        super(Dataset, self).__init__(store=store, identifier=DATASET_DEFAULT_GRAPH_ID)
 
         if not self.store.graph_aware:
             raise Exception("DataSet must be backed by a graph-aware store!")
```
allows the the otherwise failing tests in [`test_dataset_generators`](https://github.com/RDFLib/rdflib/blob/200cb59d37f579bc6c875a8a0650d85d9a946de7/test/test_dataset_generators.py) and json-ld [`test_named_graphs`](https://github.com/RDFLib/rdflib/blob/200cb59d37f579bc6c875a8a0650d85d9a946de7/test/jsonld/test_named_graphs.py) to succeed.

More generally, niklasl 's observations in #436 remain true “When ConjunctiveGraph.parse is called, it wraps its underlying store in a regular Graph instance. This causes problems for parsers of datasets, e.g. NQuads, TriG and JSON-LD. Specifically, the triples in the default graph of a dataset haphazardly end up in bnode-named contexts.” so while programmatic operations on Datasets are coherent, after this PR is applied, parsing is not.

test/test_graph_generators.py
```diff
@@ -1,13 +1,13 @@
...
 from rdflib import Graph, URIRef
 
-michel = URIRef("urn:michel")
-tarek = URIRef("urn:tarek")
-bob = URIRef("urn:bob")
-likes = URIRef("urn:likes")
-hates = URIRef("urn:hates")
-pizza = URIRef("urn:pizza")
-cheese = URIRef("urn:cheese")
+michel = URIRef("urn:x-rdflib:michel")
```
aucampia, line 4: While `urn:x-rdflib:michel` is significantly better than `urn:michel`, we should rather be using example URN namespace `urn:example:` [[IETF RFC 6963](https://datatracker.ietf.org/doc/html/rfc6963)]  or even the `example:` URI scheme [[IETF RFC 7595](https://datatracker.ietf.org/doc/html/rfc7595#section-8)]. I will add this to development guidelines.

gjhiggins: > While `urn:x-rdflib:michel` is significantly better than `urn:michel`, we should rather be using example URN namespace `urn:example:` [[IETF RFC 6963](https://datatracker.ietf.org/doc/html/rfc6963)] or even the `example:` URI scheme [[IETF RFC 7595](https://datatracker.ietf.org/doc/html/rfc7595#section-8)]. I will add this to development guidelines.

It's still work-in-progress ... I made it all consistent in an effort to try and clarify the parser-wrapping issue and to resolve three or four test failures caused by mismatches in the namespace output. I'll switch them over to `urn:exanple`.

gjhiggins:
Re: https://github.com/RDFLib/rdflib/pull/1646/commits/ea1c2ebaff77631c6a9cc10fdae8bec088426eb0, I'd be happier if I had a better model of _why_ this works but being able to resolve the jsonld test_named_graphs failure seems worth it.

nicholascar:
I just want to check we don't break any current multi-graph functionality.

I sometimes want to get a single graph out of a multi-graph and a function like this will currently do it:

```python
def get_graph(
    self,
    identifier: Union[URIRef, BNode]
) -> Union[Graph, None]:
    """Returns the graph identified by given identifier"""
    return [x for x in self.contexts() if x.identifier == identifier][0]
```

How would a this need to be implemented after this change, because the whole graph object couldn't be returned, could it? Perhaps I would have to copy content out of the multi-graph, tripl-by-triple like this:

```python
def get_graph(
    self,
    identifier: Union[URIRef, BNode]
) -> Union[Graph, None]:
    """Returns the graph identified by given identifier"""
    g2 = Graph()
    for s, p, o in self.triples((None, None, None, identifier)):
        g2.add((s, p, o))
    return g2
```

gjhiggins:
Just a note, *en passant*. Because **all** RDFLib Graphs are named graphs, the Dataset implementation is slightly adrift from the RDF1.1 spec ...
“An RDF dataset is a collection of RDF graphs, and comprises:
 - Exactly one default graph, being an RDF graph. **The default graph does not have a name** and MAY be empty.
 - Zero or more named graphs. Each named graph is a pair consisting of an IRI or a blank node (the graph name), and an RDF graph. Graph names are unique within an RDF dataset.”

nicholascar:
> the Dataset implementation is slightly adrift from the RDF1.1 spec

Yeah, after this PR, my and Ashley's plan is to dedup `ConjunctiveGraph` & `Dataset` with a more spec implementation for an rdflib 7.0.0. I can't think of any benefits for having both classes that are similar but slightly different. It's confusing. I suppose someone added `Dataset` long after `ConjunctiveGraph` and had some specific reasons but I've not found work reasons to be able to differentiate them.

gjhiggins:
> > the Dataset implementation is slightly adrift from the RDF1.1 spec

> Yeah, after this PR, my and Ashley's plan is to dedup `ConjunctiveGraph` & `Dataset` with a more spec implementation for an rdflib 7.0.0. I can't think of any benefits for having both classes that are similar but slightly different. It's confusing. I suppose someone added `Dataset` long after `ConjunctiveGraph` and had some specific reasons but I've not found work reasons to be able to differentiate them.

Gory details of the discussion are collated in [this gist](https://gist.github.com/gjhiggins/41c6e1cbc56a41f72609fd8f585ace41), enjoy :grin:

gjhiggins:
Documentation note. Oracle how-to: [“RDF Datasets, Named and Default Graphs Concepts and How to Query Them.
Emma Thomas. May 5, 2020”](https://www.ateam-oracle.com/post/rdf-datasets-named-and-default-graphs-concepts-and-how-to-query-them)

nicholascar:
Perhaps this PR should be moved to Draft if there are still a series of expected commits @gjhiggins? Just so that I and others don't approve too early or something!

gjhiggins:
> Perhaps this PR should be moved to Draft if there are still a series of expected commits @gjhiggins? Just so that I and others don't approve too early or something!

I meant to set it to Draft status but forgot and there doesn't seem to be any way to change it.

The only remaining task (apart from documentation) is to implement the missing Dataset operators/in-place operators and corresponding tests.

nicholascar:
OK, I can live!

I'm watching this closely so won't be a problem really. Ready to test myself and update documentation and examples like [examples/datasets.py](https://github.com/RDFLib/rdflib/blob/master/examples/datasets.py) on completion.

gjhiggins:
> OK, I can live!
> 
> I'm watching this closely so won't be a problem really. Ready to test myself and update documentation and examples like [examples/datasets.py](https://github.com/RDFLib/rdflib/blob/master/examples/datasets.py) on completion.

Most of the work is now complete so `examples/datasets.py` executes without issue. You might be interested in the contents of [test_issue436](https://github.com/gjhiggins/rdflib/blob/identifier-as-context/test/test_identifier_as_context/test_issues_fixed/test_issue436_ds_capable_parse.py) which (along with the parsed [input trig](https://github.com/gjhiggins/rdflib/blob/identifier-as-context/test/consistent_test_data/testdata01.trig)) is the proof-of-concept response to [your comment](https://github.com/RDFLib/rdflib/issues/436#issuecomment-987435414).

rdflib/graph.py
```diff
@@ -1944,18 +1945,19 @@ class Dataset(ConjunctiveGraph):
...
             raise Exception("DataSet must be backed by a graph-aware store!")
-        self.default_context = Graph(
-            store=self.store,
-            identifier=DATASET_DEFAULT_GRAPH_ID,
-            base=default_graph_base,
-        )
 
         self.default_union = default_union
 
+    def add(self, triple_or_quad):
```
mwatts15, line 1955: I think having `add` return a reference to the graph is dumb, but for consistency with the other `Graph` implementations, it should return the return value of `add` in this method.

rdflib/graph.py
```diff
@@ -1944,18 +1945,19 @@ class Dataset(ConjunctiveGraph):
...
+        if len(triple_or_quad) == 4:
+            super(Dataset, self).add(triple_or_quad)
+        else:
+            super(Dataset, self).add(triple_or_quad + (DATASET_DEFAULT_GRAPH_ID,))
```
mwatts15, line 1959: shouldn't this rather pass in the `self.default_context`?

rdflib/graph.py
```diff
@@ -1944,18 +1945,19 @@ class Dataset(ConjunctiveGraph):
     """
 
     def __init__(self, store="default", default_union=False, default_graph_base=None):
-        super(Dataset, self).__init__(store=store, identifier=None)
+        super(Dataset, self).__init__(store=store, identifier=DATASET_DEFAULT_GRAPH_ID)
 
         if not self.store.graph_aware:
             raise Exception("DataSet must be backed by a graph-aware store!")
-        self.default_context = Graph(
```
mwatts15, line 1951: why delete this??

rdflib/graph.py
```diff
@@ -1983,38 +2003,104 @@ def graph(self, identifier=None, base=None):
...
+
+        See :meth:`rdflib.graph.Graph.parse` for documentation on arguments.
+
+        :Returns:
+
+        The graph into which the source was parsed. In the case of n3
+        it returns the root context.
+        """
+
+        quadformats = ["trig", "trix", "nquads", "hext", "json-ld"]
```
mwatts15, line 2033: I think this is meant to solve the problem of dataset serialization formats that send the default graph to the graph matching the "publicID" (whatever a "public ID" is), but keeping a list of formats RDFLib happens to know have a concept of "quads" or "graphs" seems a very poor way to handle that. More importantly, I think handling *that* problem can be done separately from just changing the store API.

rdflib/graph.py
```diff
@@ -2169,7 +2257,7 @@ class ReadOnlyGraphAggregate(ConjunctiveGraph):
     """Utility class for treating a set of graphs as a single graph
 
     Only read operations are supported (hence the name). Essentially a
-    ConjunctiveGraph over an explicit subset of the entire store.
+    ConjunctiveGraph over an explicit subset of an entire single store.
```
mwatts15, line 2260: isn't "single" redundant here?

rdflib/store.py
```diff
@@ -238,8 +241,15 @@ def addN(self, quads: Iterable[Tuple["Node", "Node", "Node", "Graph"]]):
             )
             self.add((s, p, o), c)
 
-    def remove(self, triple, context=None):
-        """Remove the set of triples matching the pattern from the store"""
+    def remove(
+        self,
+        triple: Tuple["Node", "Node", "Node"],
+        context: Optional[Union["Graph", "Identifier"]],
```
mwatts15, line 247: Why is Graph in this annotation? Isn't the point of this PR to make it always an Identifier?

mwatts15 (changes requested):
Mostly looked at store.py and graph.py.

(Glad to see this finally being ironed out :))

rdflib/graph.py
```diff
@@ -968,6 +967,12 @@ def qname(self, uri):
     def compute_qname(self, uri, generate=True):
         return self.namespace_manager.compute_qname(uri, generate)
 
+    def qname_strict(self, uri):
```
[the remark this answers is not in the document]
nicholascar: Surely the move here is just to reach consistency of API? If you, @mwatts15, want a change in usage patterns, perhaps after this PR you would have to address the existing `qname()` & ``compute_qname()` as well as  `qname_strict`?

So are you saying you use a stand-alone Namespace Manager object do do all your namespaceing, perhaps like:

```python
from rdflib.namespace import NamespaceManager

nm = NamespaceManager()

g = Graph()
g.namespace_manager(nm)

other_gr = Graph()
other_g.namespace_manager(nm)

# etc.
```

rdflib/graph.py
nicholascar: Yeah, what _does_ `publicID` actually mean? Is it just alternate (archaic?) terminology for the graph identifier, i which case we should probably just purge it, or does it have some other meaning?

rdflib/graph.py
nicholascar: I think so...

rdflib/store.py
nicholascar: Looks like a good point...

rdflib/graph.py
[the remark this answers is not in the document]
gjhiggins: > Why add more wrappers to namespace_manager? I don't see any additional usages in this PR.

The other part of this commit details the issue resolved, when serialized as xml an Exception is raised by the XML serialiser: `# 'Dataset' object has no attribute 'compute_qname_strict'`. I'm not minded to _prohibit_ serialisation of a Dataset as triples in XML (albeit, only the triples in the default graph get output) and, given that nicholascar and ashleysommer are considering a ConjunctiveGraph/Dataset dedupe, I thought this approach was the least impactful.

rdflib/graph.py
```diff
@@ -2029,9 +2115,11 @@ def contexts(self, triple=None):
 
     def quads(self, quad):
         for s, p, o, c in super(Dataset, self).quads(quad):
-            if c.identifier == self.default_context:
+            if c.identifier == DATASET_DEFAULT_GRAPH_ID:
                 yield s, p, o, None
-            else:
+            elif len(quad) == 3:
+                yield s, p, o, c.identifier
```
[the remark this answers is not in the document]
gjhiggins: Sorry, I don't understand your point. All three implementations of `quads` are iterators.

rdflib/plugins/stores/auditable.py
```diff
@@ -53,12 +53,7 @@ def add(self, triple, context, quoted=False):
         lock = destructiveOpLocks["add"]
         lock = lock if lock else threading.RLock()
         with lock:
-            context = (
-                context.__class__(self.store, context.identifier)
-                if context is not None
-                else None
-            )
-            ctxId = context.identifier if context is not None else None
+            ctxId = context if context is not None else None
```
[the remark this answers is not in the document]
gjhiggins: Happy to make this change to the original PR if it's generally agreed, otherwise - seeing as it was a rebase of a rebase, I left the original untouched.

rdflib/graph.py
```diff
@@ -1608,13 +1613,12 @@ def _spoc(
         either triples or quads
         """
         if triple_or_quad is None:
-            return (None, None, None, self.default_context if default else None)
+            return (None, None, None, self.identifier if default else None)
```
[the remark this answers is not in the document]
gjhiggins: > This is wrong, isn't it? You want `_spoc` to return a `Graph` for the 4th member.

This specific change is inherited verbatim from the rebased PR [see LoC#1313](https://github.com/RDFLib/rdflib/pull/409/commits/da6837e84235b47a9218a96559a77a33fffe7b89#). Changing it back temporarily doesn't cause any test failures which suggests that if if _is_ wrong, we need a test for it. I added a warning and found the only code that triggers the warning is [`test_conjunctive_graph.py::test_quad_contexts`](https://github.com/RDFLib/rdflib/blob/b2fdaf5a1f45c09694dbd8925ab6b6dee84436b4/test/test_conjunctive_graph.py#L39)

gjhiggins:
I'm getting the sense that this rebase is inadequately-implemented and I am quite prepared to abandon it.

test/test_sparqlupdatestore.py
```diff
@@ -88,12 +89,15 @@ def testSimpleGraph(self):
...
         g.remove((tarek, likes, pizza))
-        r = g.query("SELECT * WHERE { ?s <urn:likes> <urn:pizza> . }")
+        r = g.query("SELECT * WHERE { ?s <urn:example:likes> <urn:example:pizza> . }")
         self.assertEqual(1, len(list(r)), "only bob likes pizza")
 
     def testConjunctiveDefault(self):
         g = self.graph.get_context(graphuri)
         g.add((tarek, likes, pizza))
+
+        isunion = len(self.graph) > 0
```
aucampia, line 99: Won't this always be false? This changes the public API, is this intentional? If so it may be best to note it in the PR description.

test/test_issue1003.py
```diff
@@ -105,8 +106,12 @@ def test_scenarios() -> None:
     ds1.base = base_three
 
     trix = ds1.serialize(format="trix", base=Namespace("http://two.org/"))
-    assert '<graph xml:base="http://one.org/">' in trix
-    assert '<graph xml:base="http://two.org/">' in trix
+
+    with pytest.raises(Exception):
+        assert '<graph xml:base="http://one.org/">' in trix
+    with pytest.raises(Exception):
+        assert '<graph xml:base="http://two.org/">' in trix
```
aucampia, line 113: Should the graphs not have bases set?

gjhiggins: All too likely to be an mistake on my part, I'll change the test and revert the method additions.

test/test_sparqlupdatestore.py
gjhiggins: Aargh, that's overlooked detritus from not being able to replicate GH Action test environment, I've tried and failed to run two instances of fuseki a la `with_fuseki.sh` and didn't want to skip the tests, sigh. Reverted the `isunion` detritus.

rdflib/store.py
```diff
@@ -210,7 +210,7 @@ def gc(self):
     def add(
         self,
         triple: Tuple["Node", "Node", "Node"],
-        context: Optional["Graph"],
+        context: Optional[Union["Graph", "Identifier"]],
```
aucampia, line 213:
```suggestion
        context: Optional["Identifier"],
```
Should accept "Identifier" only I think.

rdflib/store.py
aucampia, line 247:
```suggestion
        context: Optional["Identifier"],
```

as Mark Watts mentioned.

test/test_sparqlupdatestore.py
aucampia: Not sure what OS you are on, but for me I run it like this and it works:

```bash
$ ./with-fuseki.sh .venv/bin/pytest test/test_issue1003.py --log-level DEBUG -rA
declare -- jena_uri="https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-3.17.0.tar.gz"
declare -- jena_archive="/home/iwana/.cache/apache-jena-fuseki-3.17.0.tar.gz"
declare -- XDG_CACHE_HOME="/home/iwana/.cache"
declare -- LOCALSTATEDIR="./var"
apache-jena-fuseki-3.17.0.tar.gz: OK
declare -- FUSEKI_HOME="./var/apache-jena-fuseki-3.17.0"
declare -- fuseki_base_normal="./var/apache-jena-fuseki-3.17.0/run-normal"
declare -- fuseki_base_tdb="./var/apache-jena-fuseki-3.17.0/run-tdb"
declare -- fuseki_pidfile_normal="./var/apache-jena-fuseki-3.17.0/run-normal/server.pid"
declare -- fuseki_pidfile_tdb="./var/apache-jena-fuseki-3.17.0/run-tdb/server.pid"
declare -- fuseki_port_normal="3030"
declare -- fuseki_port_tdb="3031"
declare -- fuseki_log_normal="./var/apache-jena-fuseki-3.17.0/run-normal/out.log"
declare -- fuseki_log_tdb="./var/apache-jena-fuseki-3.17.0/run-tdb/out.log"
INFO: starting fuseki: normal
INFO: starting fuseki: tdb
declare -- fuseki_pid_tdb="1077642"
declare -- fuseki_pid_normal="1077641"
INFO: checking if http://localhost:3030/ (fuseki-normal) is okay
INFO: url http://localhost:3030/ (fuseki-normal) not okay, waiting ...
INFO: url http://localhost:3030/ (fuseki-normal) is okay
INFO: checking if http://localhost:3031/ (fuseki-tdb) is okay
INFO: url http://localhost:3031/ (fuseki-tdb) is okay
INFO: pid 1077642 (fuseki-tdb) running
INFO: pid 1077641 (fuseki-normal) running
running: .venv/bin/pytest test/test_issue1003.py --log-level DEBUG -rA
platform linux -- Python 3.8.12, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/iwana/sw/d/github.com/iafork/rdflib, configfile: tox.ini
plugins: subtests-0.5.0, cov-3.0.0
collected 1 item                                                                                                                                                            

test/test_issue1003.py .                                                                                                                                              [100%]

================================================================================== PASSES ===================================================================================
______________________________________________________________________________ test_scenarios _______________________________________________________________________________
----------------------------------------------------------------------------- Captured log call -----------------------------------------------------------------------------
INFO     root:test_issue1003.py:111 trix = <?xml version="1.0" encoding="utf-8"?>
<TriX xml:base="http://two.org/"
  xmlns:brick="https://brickschema.org/schema/Brick#"
  xmlns:csvw="http://www.w3.org/ns/csvw#"
  xmlns:dc="http://purl.org/dc/elements/1.1/"
  xmlns:dcat="http://www.w3.org/ns/dcat#"
  xmlns:dcmitype="http://purl.org/dc/dcmitype/"
  xmlns:dcterms="http://purl.org/dc/terms/"
  xmlns:dcam="http://purl.org/dc/dcam/"
  xmlns:doap="http://usefulinc.com/ns/doap#"
  xmlns:foaf="http://xmlns.com/foaf/0.1/"
  xmlns:odrl="http://www.w3.org/ns/odrl/2/"
  xmlns:geo="http://www.opengis.net/ont/geosparql#"
  xmlns:org="http://www.w3.org/ns/org#"
  xmlns:owl="http://www.w3.org/2002/07/owl#"
  xmlns:prof="http://www.w3.org/ns/dx/prof/"
  xmlns:prov="http://www.w3.org/ns/prov#"
  xmlns:qb="http://purl.org/linked-data/cube#"
  xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
  xmlns:schema="https://schema.org/"
  xmlns:sh="http://www.w3.org/ns/shacl#"
  xmlns:skos="http://www.w3.org/2004/02/skos/core#"
  xmlns:sosa="http://www.w3.org/ns/sosa/"
  xmlns:ssn="http://www.w3.org/ns/ssn/"
  xmlns:time="http://www.w3.org/2006/time#"
  xmlns:vann="http://purl.org/vocab/vann/"
  xmlns:void="http://rdfs.org/ns/void#"
  xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
  xmlns:xml="http://www.w3.org/XML/1998/namespace"
  xmlns:dct="http://purl.org/dc/terms/"
  xmlns="http://www.w3.org/2004/03/trix/trix-1/"
>
  <graph>
    <uri>http://g9.com/</uri>
    <triple>
      <uri></uri>
      <uri>http://www.w3.org/1999/02/22-rdf-syntax-ns#type</uri>
      <uri>http://www.w3.org/2004/02/skos/core#ConceptScheme</uri>
    </triple>
    <triple>
      <uri></uri>
      <uri>http://purl.org/dc/terms/source</uri>
      <uri>nick</uri>
    </triple>
    <triple>
      <uri></uri>
      <uri>http://purl.org/dc/terms/creator</uri>
      <uri>https://creator.com</uri>
    </triple>
  </graph>
  <graph>
    <uri>http://g8.com/</uri>
    <triple>
      <uri></uri>
      <uri>http://www.w3.org/1999/02/22-rdf-syntax-ns#type</uri>
      <uri>http://www.w3.org/2004/02/skos/core#ConceptScheme</uri>
    </triple>
    <triple>
      <uri></uri>
      <uri>http://purl.org/dc/terms/source</uri>
      <uri>nick</uri>
    </triple>
    <triple>
      <uri></uri>
      <uri>http://purl.org/dc/terms/creator</uri>
      <uri>https://creator.com</uri>
    </triple>
  </graph>
  <graph>
    <uri>urn:x-rdflib:default</uri>
  </graph>
</TriX>
========================================================================== short test summary info ==========================================================================
PASSED test/test_issue1003.py::test_scenarios
============================================================================= 1 passed in 0.03s =============================================================================
declare -- fuseki_pid_normal="1077641"
declare -- fuseki_pid_tdb="1077642"
INFO: killing pid 1077641 (fuseki-normal)
INFO: pid 1077641 (fuseki-normal) is not dead, waiting ...
INFO: pid 1077641 (fuseki-normal) is dead 
INFO: killing pid 1077642 (fuseki-tdb)
INFO: pid 1077642 (fuseki-tdb) is not dead, waiting ...
INFO: pid 1077642 (fuseki-tdb) is dead 
```

Please share the output if it fails, I can fix it.

aucampia: note I added some logging, should have no output on your branch.

rdflib/graph.py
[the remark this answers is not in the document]
mwatts15: The fourth member of the return value of `_spoc` is a `Graph` per the annotation, and it's treated as a `Graph` within this module except in a few cases where I've noted we check whether it is but could more intentionally compare against `None` (and maybe throw a `TypeError` for the case that we don't receive a `Graph` since that is expected)

test/test_sparqlupdatestore.py
gjhiggins: Thanks for that gem. Will make my life much easier.

test/test_sparqlupdatestore.py
```diff
@@ -6,7 +6,7 @@
 from rdflib import ConjunctiveGraph, URIRef, Literal, BNode, Graph
 from urllib.request import urlopen
 
-HOST = "http://localhost:3031"
+HOST = "http://localhost:3030"
```
aucampia, line 9: I don't think this should change, or if it should I'm not sure why.

test/test_graph_context.py
```diff
@@ -18,8 +18,8 @@ def setUp(self):
             self.graph = ConjunctiveGraph(store=self.store)
         except ImportError:
             pytest.skip("Dependencies for store '%s' not available!" % self.store)
-        if self.store == "SQLite":
-            _, self.tmppath = mkstemp(prefix="test", dir="/tmp", suffix=".sqlite")
+        if self.store in ["LevelDB", "SQLiteLSM"]:
+            self.tmppath = os.path.join(gettempdir(), f"test_{self.store.lower()}")
```
aucampia, line 22: I think it is best to use mkstemp, we can drop the `dir="/tmp"` part through.

test/test_graph_context.py
```diff
@@ -18,8 +18,8 @@ def setUp(self):
             self.graph = ConjunctiveGraph(store=self.store)
         except ImportError:
             pytest.skip("Dependencies for store '%s' not available!" % self.store)
-        if self.store == "SQLite":
-            _, self.tmppath = mkstemp(prefix="test", dir="/tmp", suffix=".sqlite")
+        if self.store in ["LevelDB", "SQLiteLSM"]:
```
aucampia, line 21: Why is this changed from `SQLite` to `SQLiteLSM` ?

aucampia:
I don't have that much time now, and I want to look at some other issues still today, but I think we should try and slice of some things here, for example the change to urn example and some of the mordernisation can be merged seperately to reduce the size of this patch.

test/test_graph_context.py
gjhiggins: Probably better to remove it entirely, in the interests of separation of concerns. The SQLite store is pinned to RDFLib 3 via its dependency on the pinned-to-3-and-deprecated rdfextras package, so the above code is never executed. The change restores the original functionality as the SQLiteLMS and the LevelDB store are functional with RDFLib 6. The reason I switched away from mkstemp is to support SQLite's approach of raising an exception when the user configures the app to overwrite an existing store, requesting the user either to delete the store or move it out of the way - mkstemp pre-creates the directory, triggering the exception.

It's just a small convenience for the user, easily dropped if it's an ill fit.

aucampia:
ideally with changes that change API the change should be as small as humanly possible, with everything else merged in, because we are going to have to keep rebasing these things for some weeks, and every change included in here that is not essential to the change makes it harder to keep.

rdflib/graph.py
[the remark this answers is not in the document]
mwatts15: I'm saying that `ConjunctiveGraph.quads` returns a `Graph` for the fourth member of each quad, but `Dataset.quads` returns an `Identifier`. Admittedly, that isn't really the main point of this PR, so that may be a thing to consider later.

gjhiggins:
> I don't have that much time now, and I want to look at some other issues still today, but I think we should try and slice of some things here, for example the change to urn example and some of the mordernisation can be merged seperately to reduce the size of this patch.

> I will do some work on this when I have more time, maybe starting with just the change to `urn:example`.

Hoping to save you the time, I think I can manage that without losing focus :smile: https://github.com/RDFLib/rdflib/pull/1654

gjhiggins:
> ideally with changes that change API the change should be as small as humanly possible, with everything else merged in

Okay, understood and appreciated.

test/test_sparqlupdatestore.py
gjhiggins: Oversight on my part, will revert. (Local test detritus, result of struggles with fuseki, now resolved with your help, thanks.)

rdflib/graph.py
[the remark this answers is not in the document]
gjhiggins: Ah, you did explicitly write "return a Graph for the context", sorry for being obtuse. You're correct that the current [Dataset quads implementation returns an identifier](https://github.com/RDFLib/rdflib/blob/b2fdaf5a1f45c09694dbd8925ab6b6dee84436b4/rdflib/graph.py#L2030) but that behaviour [is documented](https://github.com/RDFLib/rdflib/blob/b2fdaf5a1f45c09694dbd8925ab6b6dee84436b4/rdflib/graph.py#L1884) which is sort of the _de facto_ definition of the public API.

I guess if this is to be changed, it needs to have a period of deprecation.

rdflib/graph.py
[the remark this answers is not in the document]
gjhiggins: > @gjhiggins, from where is `'Dataset' object has no attribute 'compute_qname_strict'` reported? I would think the solution is just to pass `dataset.namespace_manager` instead of `dataset` in such a case.

In [this test](https://github.com/RDFLib/rdflib/blob/1ed5a13df82b25af28f3843e15725f1e4908e8ee/test/test_identifier_as_context/test_dataset_default_graph.py#L132) of serializations of the Dataset default graph.

```py
dataset_default_graph = ds.get_context(DATASET_DEFAULT_GRAPH_ID)
dataset_default_graph.serialize(format="xml")  # <-- results in the Exception
```
Apologies for mistakenly stating earlier that it was a Dataset issue, not a Dataset default graph issue.

rdflib/store.py
gjhiggins: Okay, noted. (I'm not up to speed w.r.t. to typing, so it was a bit of a conjecture)

rdflib/store.py
gjhiggins: Again. noted. Thanks both for the assist.

test/test_sparqlupdatestore.py
gjhiggins: I'm seeing identical output.
However, judging by the above log and the assertions in the test:
```py
    trix = ds1.serialize(format="trix", base=Namespace("http://two.org/"))
    assert '<graph xml:base="http://one.org/">' in trix
    assert '<graph xml:base="http://two.org/">' in trix
    assert '<TriX xml:base="http://two.org/"' in trix
```
I can only see the last of the assertions as being true or is my understanding flawed?

rdflib/graph.py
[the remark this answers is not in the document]
gjhiggins: mypy has just reminded me that if quads are going to have a Graph as a context, then the type signatures of `add` and `remove` need to reflect that - hence my original typing of `Optional[Union["Graph", "Identifier"]]`

rdflib/graph.py
gjhiggins: It's in the [original PR from gromgull](https://github.com/RDFLib/rdflib/pull/409/files#diff-a8bea910ef28d48bce0a390d9ab24c55ac32f1c9a39d9cd215224e5447b70c83L1605) - LoC #1605

rdflib/graph.py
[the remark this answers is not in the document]
gjhiggins: Thanks for the explanation. The collective model of quads seems to have changed somewhat from the original, as documented in the [“Working with multigraphs”](https://github.com/RDFLib/rdflib/blob/master/docs/intro_to_parsing.rst#working-with-multi-graphs) section of the "Introduction to parsing", where the type of the fourth member is explicitly described as an identifier, not a Graph object,

“To read and query multi-graphs, that is RDF data that is context-aware, you need to use rdflib's :class:`rdflib.ConjunctiveGraph` or :class:`rdflib.Dataset` class. These are extensions to :class:`rdflib.Graph` that know all about quads ***(triples + graph IDs)***.” (my emphasis)

The example is equally explicit:
> You could parse the file and query it like this:
```py
from rdflib import Dataset
from rdflib.namespace import RDF

g = Dataset()
g.parse("demo.trig")

for s, p, o, g in g.quads((None, RDF.type, None, None)):
    print(s, g)
```
> This will print out:
```
http://example.com/person/drewp http://example.com/person/graph-1
http://example.com/person/nick http://example.com/person/graph-2
```

gjhiggins:
> Yeah, after this PR, my and Ashley's plan is to dedup `ConjunctiveGraph` & `Dataset` with a more spec implementation for an rdflib 7.0.0. I can't think of any benefits for having both classes that are similar but slightly different. It's confusing. I suppose someone added `Dataset` long after `ConjunctiveGraph` and had some specific reasons but I've not found work reasons to be able to differentiate them.

#### From: FWIW Dept.

Digging back into the commit history reveals that ConjunctiveGraph was introduced back in Oct 2005 in https://github.com/RDFLib/rdflib/commit/750e8b8d762fe1cf68203344e8fd635719166178. Whilst this is contemporaneous with discussions of reification on the www-rdf-interest mailing list (see below), this significantly predates the official WC3 description of Dataset - which wasn't in the RDF1.1 draft of [Aug 2011](https://www.w3.org/TR/2011/WD-rdf11-concepts-20110830/), only appearing in rough form in [June 2012](https://www.w3.org/TR/2012/WD-rdf11-concepts-20120605/#section-dataset).

ConjunctiveGraph was eikeon's implementation of context-as-reification, a concept later refined and formalized by W3C as RDF Datasets. Hence the apparent duplication arises from the common requirements of representing reified triples. Ivan Herman, the author of the RDFLib Dataset implementation was an active member of W3C at the time (and who, tbf, isn't exactly complimentary about his implementation)

The first actual release of RDFLib - [2.0.6](https://github.com/RDFLib/rdflib/releases/tag/release-2.0.6) was a couple of months prior to the publication of the Bizer, Carroll, Hayes and Stickler paper "Named Graphs, Provenance and Trust" at WWW May 2005, Chiba, Japan.

I suspect the fact that all RDFLib Graphs are named graphs is an early implementation of the principles defined therein. In Dec 2005, [Chimezie](https://github.com/chimezie/) committed a [change to the Graph docstring](https://github.com/RDFLib/rdflib/commit/4f5905d376f5a1d53b5d16d9e116a266790d333d) referencing a 2004 paper on Named Graphs and TriX. The notion had been discussed for a couple of years previously, e.g. in this [Feb 2004 paper](https://lists.w3.org/Archives/Public/www-archive/2004Feb/att-0072/swig-bizer-carroll.pdf) after its [airing in Dec 2003 by Chris Bizer](https://lists.w3.org/Archives/Public/www-rdf-interest/2003Dec/0078.html) on the www-rdf-interest mailing list - not that it was the first time reification had been discussed, as Jeremy Carroll observes:

> ... the interpretation of a reified statement is clearly
> defined by RDF Semantics - but this differs from the applications that
> reification is often used for. That was the best the WG could do and it is a
> known limitation with the current round of RDF. That's why things like
> contexts ended up in the postponed pile.

Given that Datasets are at least described in the W3C specs (albeit they declined to present a formal specification of the semantics) then following a clean, robust implementation of Datasets, the use of ConjunctiveGraphs could be deprecated but I imagine that would have profound implications for legacy apps/code.

gjhiggins:
> I meant to set it to Draft status but forgot and there doesn't seem to be any way to change it.

Ah, found the "convert PR to draft" link.

rdflib/graph.py
```diff
@@ -1686,14 +1690,14 @@ def triples(self, triple_or_quad, context=None):
         """
 
         s, p, o, c = self._spoc(triple_or_quad)
-        context = self._graph(context or c)
+        context = context or c
```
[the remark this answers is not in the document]
gjhiggins: Yes, an explicit check would be required, except that issue appears to be addressed in the code that immediately follows the changed LoC:
```py
        if self.default_union:
            if context == self.identifier:
                context = None
        else:
            if context is None:
                context = self.identifier
```

rdflib/graph.py
gjhiggins: > Yeah, what _does_ `publicID` actually mean? Is it just alternate (archaic?) terminology for the graph identifier, i which case we should probably just purge it, or does it have some other meaning?

From the docstring: “The publicID argument is for specifying the logical URI for the case that it's different from the physical source URI”.

It's necessary for correct serialization, e.g. reading the file `examples/foaf.rdf` and parsing it as `data` without a publicID results in URLs being rewritten as the file source, e.g.
```py
rdflib.term.URIRef('file:///home/gjh/dev/rdflib/images/timbl-image-by-Coz-cropped.jpg')
```
whereas using the publicID of `http://dig.csail.mit.edu/2008/webdav/timbl/foaf.rdf` results in the correct serialisation:
```py
rdflib.term.URIRef('http://dig.csail.mit.edu/2008/webdav/timbl/images/timbl-image-by-Coz-cropped.jpg')
```
Just using the CSAIL URL as identifier doesn't result in correct serialization so it does appear that `publicID` has a valid and continued role.

nicholascar: OK, ok, I think I knew that or should have worked that out! So `publicID` stays as is.

nicholascar:
> ...following a clean, robust implementation of Datasets, the use of ConjunctiveGraphs could be deprecated but I imagine that would have profound implications for legacy apps/code.

That's what a 7.0.0 would allow for!

The `ConjunctiveGraph` is dead, all hail the `Dataset`!

gjhiggins:
> > ...following a clean, robust implementation of Datasets, the use of ConjunctiveGraphs could be deprecated but I imagine that would have profound implications for legacy apps/code.
> 
> That's what a 7.0.0 would allow for!
> 
> The `ConjunctiveGraph` is dead, all hail the `Dataset`!

Update. I went the whole hog and implemented the de-dupe. `Dataset` now inherits from `Graph` and the implementation is much cleaner. There are no longer any calls to `ConjunctiveGraph` in the RDFLib code, its presence in `graph.py` is now merely for legacy support. It's not *quite* ready for a PR just yet, even as a Draft, but soon enough (I hope) to be in 6.2 and enable a generous deprecation period for ConjunctiveGraph.

nicholascar:
Great tidying @gjhiggins! Standing by to review/help/release.

gjhiggins:
This work unavoidably breaks the API (in particular, `Dataset.get_context()` which now returns a `URIRef`, not a `Graph`) and so can only be included as part of a version 0.7 release, so I have closed this PR and removed the branch from my fork.

nicholascar:
@gjhiggins are you saying you are expecting to re-propose this PR for a 7.0.0 release?

Issue #225, opened 2012-07-21
Think about __iadd__, __isub__ etc. for ConjunctiveGraph


gromgull:
Currently all operations are done on the default graph, i.e. if you add another graph, even if it's a conjunctive graph, all triples are added to the default graph. 

It may make sense to check if the other thing added is ALSO a conjunctive graph and merge the contexts:

``` python
def __iadd__(self, other):
        """Add all triples in ConjunctiveGraph other to this Graph. 
           BNode IDs are not changed."""
        if isinstance(other, ConjunctiveGraph): 
            self.addN(other.quads((None,None,None)))
        else: 
            self.addN((s,p,o,other.identifier) for s,p,o in other)
        return self

    def __isub__(self, other):
        """Subtract all triples in Graph other from Graph. 
           BNode IDs are not changed."""
        if isinstance(other, ConjunctiveGraph):
            for s,p,o,c in other.quads((None,None,None)):
                self.store.remove((s,p,o),c)
        else:
            for triple in other:
                self.remove(triple)
        return self

    def __add__(self, other) :
        """Set-theoretic union 
           BNode IDs are not changed."""
        retval = ConjunctiveGraph()
        for (prefix, uri) in set(
                list(self.namespaces()) + list(other.namespaces())):
            retval.bind(prefix, uri)
        retval+=self
        retval+=other

        return retval
```

PR #1647, opened 2022-01-06
add test of ConjunctiveGraph operators

gjhiggins:
Although this is a byproduct of the identifier-as-context PR, I'm splitting this off as a separate PR in order to clarify the situation w.r.t issue #225 and establish a pattern for a similar approach to Dataset operators.

## Proposed Changes

  - Add test of ConjunctiveGraph operator combinations

gromgull's ruminations in #225 remain pertinent to ConjunctiveGraph in terms of a set-theoretic perspective as well as Pythonic perspective.

> “Currently all operations are done on the default graph, i.e. if you add another graph, even if it's a conjunctive graph, all triples are added to the default graph. ... It may make sense to check if the other thing added is ALSO a conjunctive graph and merge the contexts:” 

I suspect that we have to read “added” and “merged” in an “amenable-to-operator” sense as  it doesn't seem to make sense for it to mean “copied”  and really it _can't_ because a dbpedia SPARQLStore can be a context. I think it also sheds a little light on the conceptual difference between ConjunctiveGraph and Dataset (in which adding doesn't imply adding to the default graph, except, of course, if your SPARQL endpoint has a different opinion).

Interestingly, the next set of observations (reformatted for re-presentaion) also seem to be based on an assumption that the contexts are all local :

> "all methods work against this default graph" - sounds good, but isn't quite true.
> 
> `len` will give you the length of the conjunction of all graphs; `contains` and `triples` will check for triples in all graphs; `remove` will remove from all graphs; `addN/quads` work on quads directly.
> 
> In fact, the only method to make use of the `default_context` was `add` (and `iadd` gets delegated to `add`). So `add` works on a single graph only, `remove` does not.
> So: `conjgraph += g1` and `conjgraph -= g1` may not be what you want (since `isub` will be delegated to `remove` which removes from ALL graphs).
> 
> I am not really sure what makes the most sense here. Adding `iadd/isub` that works for quads seems sensible, but breaks backwards compatibility in a fairly subtle way. I would postpone until rdflib 4.
> 
> I also do not like that `len` gives you the number of TRIPLES not of quads, but that is another issue.

### DIFF
--- a/test/test_conjunctivegraph_operator_combinations.py
+++ b/test/test_conjunctivegraph_operator_combinations.py
@@ -6,16 +6,16 @@ from rdflib import (
 )
 
 
-michel = URIRef("urn:x-rdflib:michel")
-tarek = URIRef("urn:x-rdflib:tarek")
-bob = URIRef("urn:x-rdflib:bob")
-likes = URIRef("urn:x-rdflib:likes")
-hates = URIRef("urn:x-rdflib:hates")
-pizza = URIRef("urn:x-rdflib:pizza")
-cheese = URIRef("urn:x-rdflib:cheese")
+michel = URIRef("urn:example:michel")
+tarek = URIRef("urn:example:tarek")
+bob = URIRef("urn:example:bob")
+likes = URIRef("urn:example:likes")
+hates = URIRef("urn:example:hates")
+pizza = URIRef("urn:example:pizza")
+cheese = URIRef("urn:example:cheese")
 
-c1 = URIRef("urn:x-rdflib:context-1")
-c2 = URIRef("urn:x-rdflib:context-2")
+c1 = URIRef("urn:example:context-1")
+c2 = URIRef("urn:example:context-2")
 
 sportquadstrig = open(
     os.path.join(os.path.dirname(__file__), "consistent_test_data", "sportquads.trig")
--- a/test/test_conjunctivegraph_operator_combinations.py
+++ b/test/test_conjunctivegraph_operator_combinations.py
@@ -1,0 +1,123 @@
+import os
+from rdflib import (
+    Graph,
+    ConjunctiveGraph,
+    URIRef,
+)
+
+
+michel = URIRef("urn:x-rdflib:michel")
+tarek = URIRef("urn:x-rdflib:tarek")
+bob = URIRef("urn:x-rdflib:bob")
+likes = URIRef("urn:x-rdflib:likes")
+hates = URIRef("urn:x-rdflib:hates")
+pizza = URIRef("urn:x-rdflib:pizza")
+cheese = URIRef("urn:x-rdflib:cheese")
+
+c1 = URIRef("urn:x-rdflib:context-1")
+c2 = URIRef("urn:x-rdflib:context-2")
+
+sportquadstrig = open(
+    os.path.join(os.path.dirname(__file__), "consistent_test_data", "sportquads.trig")
+).read()
+
+
+def test_operators_with_conjunctivegraph_and_graph():
+
+    cg = ConjunctiveGraph()
+    cg.add((tarek, likes, pizza))
+    cg.add((tarek, likes, michel))
+
+    g = Graph()
+    g.add([tarek, likes, pizza])
+    g.add([tarek, likes, cheese])
+
+    assert len(cg + g) == 3  # adds cheese as liking
+
+    assert len(cg - g) == 1  # removes pizza
+
+    assert len(cg * g) == 1  # only pizza
+
+    assert len(cg ^ g) == 2  # removes pizza, adds cheese
+
+
+def test_reversed_operators_with_conjunctivegraph_and_graph():
+
+    cg = ConjunctiveGraph()
+    cg.add((tarek, likes, pizza))
+    cg.add((tarek, likes, michel))
+
+    g = Graph()
+    g.add([tarek, likes, pizza])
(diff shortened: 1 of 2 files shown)


test/test_conjunctivegraph_operator_combinations.py
```diff
@@ -0,0 +1,123 @@
...
+michel = URIRef("urn:x-rdflib:michel")
+tarek = URIRef("urn:x-rdflib:tarek")
+bob = URIRef("urn:x-rdflib:bob")
+likes = URIRef("urn:x-rdflib:likes")
+hates = URIRef("urn:x-rdflib:hates")
+pizza = URIRef("urn:x-rdflib:pizza")
+cheese = URIRef("urn:x-rdflib:cheese")
+
+c1 = URIRef("urn:x-rdflib:context-1")
+c2 = URIRef("urn:x-rdflib:context-2")
```
aucampia, line 18:
```suggestion
michel = URIRef("urn:example:michel")
tarek = URIRef("urn:example:tarek")
bob = URIRef("urn:example:bob")
likes = URIRef("urn:example:likes")
hates = URIRef("urn:example:hates")
pizza = URIRef("urn:example:pizza")
cheese = URIRef("urn:example:cheese")

c1 = URIRef("urn:example:context-1")
c2 = URIRef("urn:example:context-2")
```

aucampia (commented):
Looks good, just a suggestion on using `urn:example:`

test/test_conjunctivegraph_operator_combinations.py
gjhiggins: Will do. Not actually forgotten, just currently distracted by trying to make sense of the chaos :smile:

gjhiggins:
> “Currently all operations are done on the default graph, i.e. if you add another graph, even if it's a conjunctive graph, all triples are added to the default graph.

The tests introduced in this PR include an example of the above, adding a ConjunctiveGraph to a ConjunctiveGraph but it's not a sufficiently thorough test to reveal what gromgull was concerned about:
```py
def test_operators_with_two_conjunctivegraphs():
    cg1 = ConjunctiveGraph()
    cg1.add([tarek, likes, pizza])
    cg1.add([tarek, likes, michel])
    cg2 = ConjunctiveGraph()
    cg2.add([tarek, likes, pizza])
    cg2.add([tarek, likes, cheese])
    assert len(cg1 + cg2) == 3  # adds cheese as liking
    assert len(cg1 - cg2) == 1  # removes pizza from cg1
    assert len(cg1 * cg2) == 1  # only pizza
    assert len(cg1 + cg2) == 3  # adds cheese as liking
    assert len(cg1 ^ cg2) == 2  # removes pizza, adds cheese
```
In the course of neatening the Dataset tests I've just realised an implication from the above that sheds more light on the difference between ConjunctiveGraph and Dataset. It's true that for ConjunctiveGraph all operations are done on the default graph and “all triples are added to the default graph”
```py
def test_cg_with_contexts_add_to_cg():
    cg1 = Dataset()
    assert len(list(cg1.contexts())) == 1
    assert len(list(cg1)) == 0
    cg2 = ConjunctiveGraph()
    cg2.parse(
        data="<urn:x-rdflib:tarek> <urn:x-rdflib:likes> <urn:x-rdflib:pizza> .",
        publicID="urn:x-rdflib:context-a",
        format="ttl",
    )
    cg2.parse(
        data="<urn:x-rdflib:michel> <urn:x-rdflib:likes> <urn:x-rdflib:cheese> .",
        publicID="urn:x-rdflib:context-b",
        format="ttl",
    )
    assert len(list(cg2.contexts())) == 2
    cg1 += cg2
    assert len(list(cg1.contexts())) == 1  # publicID contexts lost
    assert len(list(cg1)) == 2  # Two triples added to default graph
```

Now, according to the Dataset docstring, you get your working Graphs _from_ a Dataset:
```
    >>> # Create a graph in the dataset, if the graph name has already been
    >>> # used, the corresponding graph will be returned
    >>> # (ie, the Dataset keeps track of the constituent graphs)
    >>> g = ds.graph(URIRef("http://www.example.com/gr"))
```
`add_graph` returns a `Graph` object, nominally precluding the use of a `ConjunctiveGraph` object as a `Dataset` context. However, gromgull does describe the situation as "chaos" and I guess this is what he meant:

```py
def test_cg_with_contexts_add_to_ds():
    ds = Dataset()
    assert len(list(ds.contexts())) == 1
    assert len(list(ds)) == 0
    cg = ConjunctiveGraph()
    cg.parse(
        data="<urn:x-rdflib:tarek> <urn:x-rdflib:likes> <urn:x-rdflib:pizza> .",
        publicID="urn:x-rdflib:context-a",
        format="ttl",
    )
    cg.parse(
        data="<urn:x-rdflib:michel> <urn:x-rdflib:likes> <urn:x-rdflib:cheese> .",
        publicID="urn:x-rdflib:context-b",
        format="ttl",
    )
    assert len(list(cg.contexts())) == 2
    ds += cg  # !!
    assert len(list(ds.contexts())) == 1  # publicID contexts lost
    assert len(list(ds)) == 2  # Two triples added to default graph
```
It's difficult to see this as desirable for Dataset (or ConjunctiveGraph for that matter) but any coherent implementation will need to resolve the set-theoretic aspects of Dataset (and ConjunctiveGraph) operators, in-place or otherwise.

aucampia (approved):
Changes look good to me.

nicholascar (approved):

PR #1652, opened 2022-01-08
Don't update `SUFFIX_FORMAT_MAP` in `plugins/parsers/jsonld.py`

aucampia:
`jsonld` will already be in `SUFFIX_FORMAT_MAP`, so the code being
removed here should have no effect.

There are already tests for this and the tests would fail of the removed
code did anything, see:

https://github.com/RDFLib/rdflib/blob/b2fdaf5a1f45c09694dbd8925ab6b6dee84436b4/test/test_parse_file_guess_format.py#L23-L34

### DIFF
--- a/rdflib/plugins/parsers/jsonld.py
+++ b/rdflib/plugins/parsers/jsonld.py
@@ -69,17 +69,6 @@ from ..shared.jsonld.keys import (
 
 __all__ = ["JsonLDParser", "to_rdf"]
 
-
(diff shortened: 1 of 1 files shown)


niklasl (approved):
LGTM! (This was just a hack done by rdflib-jsonld when it was an external package.)

nicholascar (approved):
Good tidy-up

PR #1654, opened 2022-01-08
change tests to use urn:example

gjhiggins:
## Proposed Changes

  - change URIRefs of `urn:` to `urn:example` in existing tests


aucampia (approved):
Looks great, thank you very much for this :+1:

nicholascar (approved):
Thanks for the continued tidy-ups @gjhiggins

Issue #1655, opened 2022-01-08
Some special characters might be parsed wrongly (?)


GreenfishK:
It seems that some special characters in RDF literals are not preserved after parsing them but rather translated into something faulty. So far, I found following ones:

 
n3_test.nt:
```
<http:s> <http:o1> "\n" .
<http:s> <http:o2> "\f" .
<http:s> <http:o3> "\b" .
<http:s> <http:o4> "\\r" .
<http:s> <http:o5> "\\\r" .
```

```
from rdflib import Graph
from rdflib import term

g = Graph()
g.parse("n3_test.nt")

for s, p, o in g:
    assert (type(o) == term.Literal)
    print("{s} {p} {o}".format(s=s.n3(), p=p.n3(), o=o.n3()))
```

Sorted Output:

```
<http:s> <http:o1> """
"""
<http:s> <http:o2> ""
<http:s> <http:o3> "
<http:s> <http:o4> "\\\r"
<http:s> <http:o5> "\\\r"

```

We see e.g. that "\\\\r" and "\\\\\r" result into the same literal and I am not sure if this is the expected behavior. There are some DBPedia logs unfortunately which have such characters and currently I just cannot parse them correctly.

Is there a trick or a proper way how to get around this?

GreenfishK:
I did the same check, just with a .ttl file instead of an .nt file and the result seems to be as expected for case 4 (=<http:s> <http:o4> "\\r" .
) but in all other cases it is the same faulty output as for .nt.

aucampia:
If I take this nt file:

```console
$ cat test/variants/special_chars.nt 
<example:special> <example:newline> "\n" .
<example:special> <example:form_feed> "\f" .
<example:special> <example:backspace> "\b" .
<example:special> <example:carriage_return> "\r" .
<example:special> <example:backslash> "\\" .
<example:special> <example:string-000> "\\r" .
<example:special> <example:string-001> "\\\r" .
```

And round trip it:

```console
$ .venv/bin/python3 -m rdflib.tools.rdfpipe -i nt -o nt test/variants/special_chars.nt
/home/iwana/sw/d/github.com/iafork/rdflib/rdflib/plugins/serializers/nt.py:36: UserWarning: NTSerializer always uses UTF-8 encoding. Given encoding was: None
  warnings.warn(
<example:special> <example:string-000> "\\\r" .
<example:special> <example:backspace> " .
<example:special> <example:backslash> "\\" .
[...]

Issue #1649, opened 2022-01-07
rdf prefix is not emitted when saving to turtle when only rdf:type is used


cmungall:
whenusing rdflib 6.1.1 to convert the following jsonld to turtle:

```json
{
  "slots": [
    {
      "name": "type",
      "slot_uri": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
    }
  ],
  "@context": [
  {
      "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
      "@vocab": "https://w3id.org/linkml/",
      "slot_uri": {
         "@type": "@id"
      }
    }
  ]
}
```

the turtle emitted is:

```turtle
@prefix : <https://w3id.org/linkml/> .

[] :slots [ :name "type" ;
            :slot_uri rdf:type ] .
```

which is invalid, as it lacks a prefix declaration for rdf

The situation is resolved by either

 - adding another node to the graph in the rdf namespace, other than rdf:type
 - saving to a format like nt

This reproduces the error:

```python
JSONLD = """
{
  "slots": [
    {
      "name": "type",
      "slot_uri": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
    }
  ],
  "@context": [
  {
      "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
      "@vocab": "https://w3id.org/linkml/",
      "slot_uri": {
         "@type": "@id"
      }
    }
  ]
}"""

from rdflib import Graph

graph = Graph()
graph.parse(data=JSONLD, format="json-ld", prefix=False)
ttl_str = graph.serialize(format='turtle')
print(ttl_str)
graph.parse(data=ttl_str, format="turtle")
```

This is in contrast to other libs, e.g. Jena, which will produce valid turtle:

```turtle
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .

_:b0    <https://w3id.org/linkml/slots>  _:b1 .

_:b1    <https://w3id.org/linkml/name>  "type" ;
        <https://w3id.org/linkml/slot_uri>  rdf:type .

```

nicholascar:
I think that the use of `rdf:type` as an object, as opposed to the commonly seen predicate, is somehow confusing RDFlib. This use is perfectly valid, so this is an RDFLib bug. Tagging as such.

[...]

PR #1656, opened 2022-01-09
Allow hext to participate in RDF format roundtripping

nicholascar:
Until now, HexTuples parsing & serializing could not participate in RDF format roundtripping due to numerous failures. well, I've solved them by:

* doing better JSON serializing
* allowing the test_roundtrip.py functions to conflate "" with ""^^xsd:string, for hext only
* skipping only 2 tests
* allowing HexT parsing to treat a value "" as value, not None

PR #1658, opened 2022-01-09
Add xfail tests for a couple of issues

aucampia:
This includes xfails for the following issues:

- https://github.com/RDFLib/rdflib/issues/1216
- https://github.com/RDFLib/rdflib/issues/1655
- https://github.com/RDFLib/rdflib/issues/1649

Also:

- Add graph variant test scaffolding. Multiple files representing the
  same graph can now easily be tested to be isomorphic by just adding
  them in `test/variants`.
- Add more things to `testutils.GraphHelper`, including some methods that does
  asserts with better messages. Also include some tests for GraphHelper.
- Add some extra files to test_roundtrip, set the default identifier
  when parsing, and change verbose flag to rather be based on debug
  logging.
- move one test from `test/test_issue247.py` to variants.
- Fix problems with `.editorconfig` which prevents it from working
  properly.


aucampia:
I will try fix #1655 - but still think it may be worth merging this before then, and the other two issues will take some time to fix.

I'm marking this as draft until #1656 is fixed, as this will need to be rebased against that, and that fixes some of the xfails added in this.

aucampia:
Also need to fix for windows.

aucampia:
Closing pull request temporarility to prevent builds until windows CI is fixed.

PR #1659, opened 2022-01-10
Make unregister_custom_function idempotent

### LINKED ISSUE #1492: Redundant deregistering of custom functions?
I'm running into an issue with a complex ontology project where a Javascript-based SPARQL function is being "unregistered" twice and causing the following stack trace when I run my unit tests:

```
../venv/lib/python3.9/site-packages/pyshacl/validate.py:419: in validate
    conforms, report_graph, report_text = validator.run()
../venv/lib/python3.9/site-packages/pyshacl/validate.py:265: in run
    unapply_functions(advanced['functions'], g)
../venv/lib/python3.9/site-packages/pyshacl/functions/__init__.py:108: in unapply_functions
    f.unapply(data_graph)
../venv/lib/python3.9/site-packages/pyshacl/functions/shacl_function.py:197: in unapply
    unregister_custom_function(self.node, self.execute_from_sparql)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

uri = rdflib.term.URIRef('http://datashapes.org/dash#toString'), func = <bound method SPARQLFunction.execute_from_sparql of <pyshacl.functions.shacl_function.SPARQLFunction object at 0x7f5b80852c40>>

    def unregister_custom_function(uri, func):
        if _CUSTOM_FUNCTIONS.get(uri, (None, None))[0] != func:
>           raise ValueError("This function is not registered as %s" % uri.n3())
E           ValueError: This function is not registered as <http://datashapes.org/dash#toString>

../venv/lib/python3.9/site-packages/rdflib/plugins/sparql/operators.py:616: ValueError
```

I believe what is happening is the `dash` graph is getting loaded multiple times because of redundant imports. This loads the `dash:toString` function multiple times, which is (correctly) idempotent. However, [during cleanup](https://github.com/RDFLib/pySHACL/blob/master/pyshacl/validate.py#L249-L265) (I am using pySHACL's implementation of SHACL-AF), pySHACL attempts to unregister the `dash:toString` function multiple times. After a function is unregistered the first time, the `ValueError` in https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/sparql/operators.py#L614-L617 gets thrown. I believe this should be idempotent as well --- if the function does not exist in the function lookup table, then a warning should be thrown rather than an exception. I could also see pySHACL catching the exception to avoid halting the cleanup.

I am happy to introduce a change PR --- either to RDFlib or pySHACL --- to address this, but I wanted to ask if my intended fix is appropriate. I'll see what I can do about creating a minimal reproducible example

gtfierro:
Fixes #1492 

## Proposed Changes

  - Replace the raised exception with a new warning; `unregister_custom_function` can now be called multiple times on the same function (idempotent)
  - Adjust the `issue 274` test by checking that the warning is raised, rather than the exception

### DIFF
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -612,11 +612,17 @@ def custom_function(uri, override=False, raw=False):
     return decorator
 
 
-def unregister_custom_function(uri, func):
-    if _CUSTOM_FUNCTIONS.get(uri, (None, None))[0] != func:
-        warnings.warn("This function is not registered as %s" % uri.n3())
-    else:
+def unregister_custom_function(uri, func=None):
+    """
+    The 'func' argument is included for compatibility with existing code.
+    A previous implementation checked that the function associated with
+    the given uri was actually 'func', but this is not necessary as the
+    uri should uniquely identify the function.
+    """
+    if _CUSTOM_FUNCTIONS.get(uri):
         del _CUSTOM_FUNCTIONS[uri]
+    else:
+        warnings.warn("This function is not registered as %s" % uri.n3())
 
 
 def Function(e, ctx):
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -13,6 +13,7 @@ import random
 import uuid
 import hashlib
 import datetime as py_datetime  # naming conflict with function within this module
+import warnings
 
 from functools import reduce
 
@@ -613,8 +614,9 @@ def custom_function(uri, override=False, raw=False):
 
 def unregister_custom_function(uri, func):
     if _CUSTOM_FUNCTIONS.get(uri, (None, None))[0] != func:
-        raise ValueError("This function is not registered as %s" % uri.n3())
-    del _CUSTOM_FUNCTIONS[uri]
+        warnings.warn("This function is not registered as %s" % uri.n3())
+    else:
+        del _CUSTOM_FUNCTIONS[uri]
 
 
 def Function(e, ctx):
--- a/test/test_issue274.py
+++ b/test/test_issue274.py
@@ -189,7 +189,7 @@ class TestCustom(TestCase):
 
     def test_wrong_unregister_warns(self):
         with pytest.warns(UserWarning):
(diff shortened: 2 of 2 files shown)


aucampia (approved):
The change looks good, this changes public API, but I don't really think it is a massive problem in this case. I think the behaviour with this patch is more desirable.

rdflib/plugins/sparql/operators.py
```diff
@@ -613,8 +614,9 @@ def decorator(func):
 
 def unregister_custom_function(uri, func):
```
aucampia, line 615: One thing that maybe bothers me a bit, what if you just want to unregister a custom function regardless of the action function identity. Maybe if func is none the check for function identity should not be performed.

gtfierro: We still have to check for membership inside the dictionary, else we will get a `KeyError` when deleting a non-existent function. Maybe something like this would work?

```python
def unregister_custom_function(uri, func):
    if func is not None and _CUSTOM_FUNCTIONS.get(uri, (None, None))[0] != func:
        warnings.warn("This function is not registered as %s" % uri.n3())
    elif uri in _CUSTOM_FUNCTIONS:
        del _CUSTOM_FUNCTIONS[uri]
```

aucampia: yes, but lets wait for @nicholascar to give some input, I tagged him for review. I still the the change as is makes sense, but if we are going to change the function we can maybe go a bit further, even if not it is good to discuss some options.

gtfierro: I agree, and thanks for the perspective. I've put your simplified implementation in my PR and adjust the tests to match the new behavior

nicholascar (approved):

PR #1660, opened 2022-01-10
Remove narrow build detection

aucampia:
Before Python 3.3 and [PEP 393](https://www.python.org/dev/peps/pep-0393/)
(which came in with Python 3.3) there was a distinction between narrow
and wide python builds.

However this is no longer the situation, and `chr(0x10FFFF)` should not
raise `ValueError` on any python build.

From [PEP 393](https://www.python.org/dev/peps/pep-0393/):

> The Unicode string type is changed to support multiple internal
> representations, depending on the character with the largest Unicode
> ordinal (1, 2, or 4 bytes). This will allow a space-efficient
> representation in common cases, but give access to full UCS-4 on all
> systems. For compatibility with existing APIs, several representations
> may exist in parallel; over time, this compatibility should be phased
> out. The distinction between narrow and wide Unicode builds is dropped.

From [Python 3.7 docs for Unicode Objects and Codecs](https://docs.python.org/3.7/c-api/unicode.html)
docs regarding `Py_UNICODE`:

> Changed in version 3.3: In previous versions, this was a 16-bit type
> or a 32-bit type depending on whether you selected a “narrow” or “wide”
> Unicode version of Python at build time.


aucampia:
Noticed while working on fix for #1655

nicholascar:
Just removed a test for narrow builds that I remembered was there

nicholascar (approved):

PR #1662, opened 2022-01-12
Add xfail tests for a couple of issues

aucampia:
This includes xfails for the following issues:

- https://github.com/RDFLib/rdflib/issues/1216
- https://github.com/RDFLib/rdflib/issues/1655
- https://github.com/RDFLib/rdflib/issues/1649

Also:

- Add graph variant test scaffolding. Multiple files representing the
  same graph can now easily be tested to be isomorphic by just adding
  them in `test/variants`.
- Add more things to `testutils.GraphHelper`, including some methods that does
  asserts with better messages. Also include some tests for GraphHelper.
- Add some extra files to test_roundtrip, set the default identifier
  when parsing, and change verbose flag to rather be based on debug
  logging.
- move one test from `test/test_issue247.py` to variants.
- Fix problems with `.editorconfig` which prevents it from working
  properly.


aucampia:
Won't merge this as many things are fixed already, and it needs rebase, was just testing CI.

PR #1667, opened 2022-01-14
Eliminate star import

aucampia:
This breaks pycln which I'm trying to integrate with pre-commit
in a seperate branch.


nicholascar:
Looks like a good improvement and all tests pass so merging immediately!

nicholascar (approved):

PR #1670, opened 2022-01-14
Move pytest config from `tox.ini` to `setup.cfg`

aucampia:
`tox.ini` is pretty noisy and I'm adding more stuff there for
pre-commit so I think to make things more manageable it is better ot
keep unrleated config in `setup.cfg`.


aucampia:
Not a thing that is going to work sadly:

- https://github.com/pypa/pip/issues/5182

```
configparser.InterpolationMissingOptionError: Bad value substitution: option 'log_cli_format' in section 'tool:pytest' contains an interpolation key 'asctime' which is not a valid option name. Raw value: '%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)'
```

Will be gone eventually, but for now the config stays in `tox.ini`.

Issue #1664, opened 2022-01-13
Fails to parse statement where prefixed name with no local name immediately followed by dot


pduchesne:
The following turtle sample fails to be parsed, because of the lack of whitespace between `object:` and the terminal dot: 
```
@prefix test: <http://example.org/>.
@prefix object: <http://example.org/object>.
test:subject test:predicate object:.
```

In (my understanding of) the Turtle grammar, this is valid.
Turtle validator such as http://ttl.summerofcode.be concurs.

To the question : "why such a construct", the turtle serializer of [linkeddata/rdflib.js ](https://github.com/linkeddata/rdflib.js/) f.i. does produce this.

nicholascar:
@niklasl yet another Turtle niggle...

niklasl:
Yes, this is valid Turtle. But it appears that recent versions of RDFLib does handle this correctly. @pduchesne which version of RDFLib do you use? The following code works with RDFLib 6.1.1:
```python
import rdflib

g = rdflib.Graph().parse(data= """
    @prefix test: <http://example.org/>.
    @prefix object: <http://example.org/object>.
    test:subject test:predicate object:.
    """, format="text/turtle")

assert len(g) == 1
print(g.serialize(format="text/turtle"))
```

aucampia:
Confirming this works with #1677

nicholascar:
OK, this will be closed with the merger of #1667

PR #1677, opened 2022-01-18
Add a test case for a prefix followed by dot in Turtle format

aucampia:
Tests that #1664 is fixed.


nicholascar (approved):
Approving and merging since the only fail is the precommit hook fail which isn't fully implemented yet

PR #1680, opened 2022-01-18
Add IdentifiedNode abstract intermediary class

### LINKED ISSUE #1526: Name for union of URIRef and BNode?
I frequently need to do type checks for `Node`s that are not `Literal`s, mainly to catch programming errors where the "Thing vs String" logic may have been lax.  I'm casting around for a name that either could be defined in the `Node` subclass hierarchy as an abstract class, or instead could be defined as a `Union`.  My recollection is there isn't an obvious name from RDFS for this.

@aucampia suggested `Subject` in [this comment](https://github.com/RDFLib/rdflib/pull/1515#issuecomment-997446099):

```python
Subject = typing.Union[rdflib.BNode, rdflib.URIRef]
```

I have a mental TODO item to review the definition of RDF Resource to see if that meets the need, but I'd like to see if someone has already found an appropriate term for "not-`Literal`".

ajnelson-nist:
Fixes #1526

## Proposed Changes

  - Add `IdentifiedNode` to class hierarchy in `term.py`.
  - Cite RDF section for class name, found from following up on @nicholascar 's name sleuthing.
  - Expose `IdentifiedNode` symbol from `/rdflib/__init__.py`.
  - Update unit test to use new symbol.

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -51,6 +51,7 @@ __date__ = "2021-12-20"
 __all__ = [
     "URIRef",
     "BNode",
+    "IdentifiedNode",
     "Literal",
     "Variable",
     "Namespace",
@@ -155,7 +156,7 @@ In particular, this determines how the rich comparison operators for
 Literal work, eq, __neq__, __lt__, etc.
 """
 
-from rdflib.term import URIRef, BNode, Literal, Variable
+from rdflib.term import URIRef, BNode, IdentifiedNode, Literal, Variable
 
 from rdflib.graph import Dataset, Graph, ConjunctiveGraph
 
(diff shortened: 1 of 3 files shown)


aucampia (approved):
Looks good to me, thanks for the change.

nicholascar (approved):
Looks good, and I think that the only tests failing are ok as these are the not finalised pre-commit hooks.

aucampia:
> Looks good, and I think that the only tests failing are ok as these are the not finalised pre-commit hooks.

yes, I disabled them now and will re-enable once that PR is merged to avoid confusion from others, the PR for pre-commits has links to my fork which has it enabled for illustrative purposes.

PR #1681, opened 2022-01-20
Add typing to rdflib.namespace

aucampia:
Everything is as typed as it can be in here, there is maybe some
improvement that can be made to `__new__` in some places but I think
there may be some runtime changes needed there so I don't want to change
that now, the inferred return types are close enough.

Some of the type hints are based on pytest-monkeytype output which was
slightly tuned for simplicity and with some additional typing for methods
that are not being tested.

This changeset contains type changes only and does not change runtime
behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ Marked the following functions as deprecated:
   rdflib anymore and the utility that it does provide is not implemented
   correctly. It will be removed in RDFLib 7.0.0
 
+### Type Hints
+
+Type hints were added to the following modules:
+
+- `rdflib.namespace`: Nearly everything in this module has type hints now.
+
 PRs merged since last release:
 ------------------------------
 
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -1,6 +1,6 @@
 import logging
 import warnings
-from typing import TYPE_CHECKING, List, Union, Iterable
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union, Iterable
 from unicodedata import category
 
 from pathlib import Path
@@ -8,6 +8,10 @@ from urllib.parse import urldefrag
 from urllib.parse import urljoin
 
 from rdflib.term import URIRef, Variable, _is_valid_uri
+
+if TYPE_CHECKING:
+    from rdflib.graph import Graph
+    from rdflib.store import Store
 
 __doc__ = """
 ===================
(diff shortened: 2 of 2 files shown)


nicholascar (approved):
Thanks @aucampia, looks good. I need to get my head across the type hinting exclusions, all the various `# type: ignore[REASON] and when it's appropriate to use them...

nicholascar:
Merging after one review as there's no runtime change (essentially documentation).

aucampia:
> Thanks @aucampia, looks good. I need to get my head across the type hinting exclusions, all the various `# type: ignore[REASON] and when it's appropriate to use them...

In most cases they are either bugs or places where we kind of break inheritance, for example, in case of `rdflib.namespace.Namespace.__get_item__`:

https://github.com/RDFLib/rdflib/blob/c6eccdcb414634399e89ab607a92a16c72ff61f4/rdflib/namespace/__init__.py#L125-L126

`rdflib.namespace.Namespace` inherits from [`str`](https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L356),  A string's slice method has a signature of:

https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L427

```python
    def __getitem__(self, __i: SupportsIndex | slice) -> str: ...
```

So if someone is using our `Namespace` as a str, then indexing will behave differently from what a normal `str` would provide.

In most other cases they are either bugs or sometimes to avoid doing a cast where the type inferencing is not clever enough.

The ignores in function declarations most of the time are because of inheritence issues, in function bodies they are more likely to mask bugs.

My general approach for adding typing is to try and avoid runtime changes, they can be made separately, We have mypy configured to warn for unused ignores, so all ignores in the code are needed for typing to be consistent, and if we fix something later on that makes that an ignore is no longer needed we will get mypy failure telling us to remove the ignore.

Further I always try and include the code I am ignoring, which I get from mypy execution output (should also show up in pip output), e.g.

```console
$ .venv/bin/python3 -m mypy --show-error-context --show-error-codes
rdflib/namespace/__init__.py: note: In member "__new__" of class "ClosedNamespace":
rdflib/namespace/__init__.py:263: error: "Namespace" has no attribute "__uris"  [attr-defined]
```

That way if some other type error creeps in for the line it still results in a mypy failure.

nicholascar:
@aucampia thanks for the explanation

aucampia:
> @aucampia thanks for the explanation

I actually thought about the `__get_item__` behaviour, and in this case we should actually fix the places where we override it so that it also does the right thing if it gets `SupportsIndex` (e.g. `int`) or slice, it won't be that difficult, so even in that case it does kind of represent something we should fix, but again this is a runtime change so I would rather keep it separate.

PR #1683, opened 2022-01-22
Add typing to rdflib.term

aucampia:
This adds as much typing as possible to `rdflib.term`.

Other changes:
- Added back `warn_unused_ignores`. I actually thought this was enabled
  but I forgot I disabled it because of some issue on python 3.10.
- Disabled `warn_unused_ignores` only for `rdflib.plugin`. There is an
  ignore in this module which is not needed on python 3.10, this is the
  most targetted way to avoid having that fail the type checking that I
  can think of for now.
- Removed unused type ignores.

This changeset includes no runtime changes.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1223,7 +1223,7 @@ class Graph(Node):
         parser = plugin.get(format, Parser)()
         try:
             # TODO FIXME: Parser.parse should have **kwargs argument.
-            parser.parse(source, self, **args)  # type: ignore[call-arg]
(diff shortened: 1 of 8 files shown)


aucampia:
Label is so that it shows up in global search for [`is:pr is:open label:"review wanted" `](https://github.com/search?q=is%3Apr+is%3Aopen+label%3A%22review+wanted%22+&type=)

nicholascar (approved):
Looks fine to me and perhaps we only need one review since this includes no runtime changes

PR #1686, opened 2022-01-23
Bind prefixes choices

### LINKED ISSUE #1679: Replace long list of namespaces with list of prefixes used when using serialize.
I am wondering why the latest version of rdflib gives me a long list of namespaces when serializing a Graph to JSON-LD. It didn't used to be like that.
```
            return self.graph.serialize(
                format=return_format,
                context=dict(self.graph.namespaces()),
                auto_compact=True
            )
```
The return_format is ```'json-ld'```. The context in the result is:
```
  "@context": {
    "brick": "https://brickschema.org/schema/Brick#",
    "csvw": "http://www.w3.org/ns/csvw#",
    "dc": "http://purl.org/dc/elements/1.1/",
    "dcam": "http://purl.org/dc/dcam/",
    "dcat": "http://www.w3.org/ns/dcat#",
    "dcmitype": "http://purl.org/dc/dcmitype/",
    "dcterms": "http://purl.org/dc/terms/",
    "doap": "http://usefulinc.com/ns/doap#",
    "foaf": "http://xmlns.com/foaf/0.1/",
    "gtaa": "http://data.beeldengeluid.nl/gtaa/",
    "non-gtaa": "http://data.beeldengeluid.nl/nongtaa/",
    "odrl": "http://www.w3.org/ns/odrl/2/",
    "org": "http://www.w3.org/ns/org#",
    "owl": "http://www.w3.org/2002/07/owl#",
    "prof": "http://www.w3.org/ns/dx/prof/",
    "prov": "http://www.w3.org/ns/prov#",
    "qb": "http://purl.org/linked-data/cube#",
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
    "schema": "https://schema.org/",
    "sdo": "https://schema.org/",
    "sh": "http://www.w3.org/ns/shacl#",
    "skos": "http://www.w3.org/2004/02/skos/core#",
    "sosa": "http://www.w3.org/ns/sosa/",
    "ssn": "http://www.w3.org/ns/ssn/",
    "time": "http://www.w3.org/2006/time#",
    "vann": "http://purl.org/vocab/vann/",
    "void": "http://rdfs.org/ns/void#",
    "xml": "http://www.w3.org/XML/1998/namespace",
    "xsd": "http://www.w3.org/2001/XMLSchema#"
  },
```
This is where the graph and namespace bindings (including some custom ones) were created:
```
        self.graph = Graph()
        self.graph.namespace_manager.bind("skos", SKOS)
        self.graph.namespace_manager.bind("gtaa", Namespace(self._model.GTAA_NAMESPACE))
        self.graph.namespace_manager.bind("non-gtaa", Namespace(self._model.NON_GTAA_NAMESPACE))
```
Further, in the custom class I add another namespace and triples. Here's a fragment:
```
        self.graph.namespace_manager.bind('sdo', Namespace(self._model.SCHEMA_DOT_ORG_NAMESPACE))
        # create a node for the record
        self.itemNode = URIRef(self.get_uri(concept_type, metadata["id"]))

        # get the RDF class URI for this type
        self.classUri = self._model.CLASS_URIS_FOR_DAAN_LEVELS[concept_type]

        # add the type
        self.graph.add((self.itemNode, RDF.type, URIRef(self.classUri)))
```
The custom Class is used to read some JSON from a backend system, interpret this and generate RDF for the item. You could see this as a wrapper pattern. 

As I wrote in comments to [this issue](https://github.com/RDFLib/rdflib/issues/794), I had to create some custom function to remove unused prefixes from the context, but that code is not so dynamic:
```
    def remove_unused_prefixes(self):
        """ Clean up the long list of namespaces.
        """
        context = dict(self.graph.namespaces())
        used_prefixes = ['gtaa', 'non-gtaa', 'rdf', 'rdfs', 'sdo', 'skos', 'xml', 'xsd']
        return {p: context[p] for p in used_prefixes}
```
and this is used here: 
```
            context_used = self.remove_unused_prefixes()
            return self.graph.serialize(
                format=return_format,
                context=context_used,
                auto_compact=True
            )
```
Now I just discovered that the context argument can be left out. This is probably because of recent improvements and integration of json-ld. Well done. But it still gives me the long list. Also, when omitting the context argument the ```auto_compact=True``` argument finally gives me a short representation that I wanted, for example: ```"sdo:datePublished": "2006-02-19",```. This is not the case when using this ```context = dict(self.graph.namespaces())```. But, after all I still get the long list of namespace that I aren't used.

Another discovery: when further reducing the number of arguments, I still get JSON-LD, but **no context** at all in the results.
```           
           return self.graph.serialize(
                format=return_format
            )
```
To summarize this issue: I would like to get JSON-LD serialization including context, but with a minimal list of (used) prefixes/namespaces in response to this request:
```
            return self.graph.serialize(
                format='json-ld',
                auto_compact=True
            )
```
I hope provided examples will help.

nicholascar:
Fixes #1679

## Proposed Changes

  - a bunch of reST documentation updates
  - updates to inline docc for term.py classes (RDF 1.1, not 1.0 explanations)
  - new param for `Graph()`: `bind_namespaces` 
      - with options for none, core prefixes (default), all RDFLib Namespaces or use of prefix.cc

Since this PR is really needed - see the comments - I'm pushing it up to review before the prefix.cc option is implemented.


docs/index.rst
```diff
@@ -8,18 +8,19 @@ RDFLib is a pure Python package for working with `RDF <http://www.w3.org/RDF/>`_
...
 * **Parsers & Serializers**
 
-  * for RDF/XML, N3, NTriples, N-Quads, Turtle, TriX, JSON-LD, RDFa and Microdata
+  * for RDF/XML, N3, NTriples, N-Quads, Turtle, TriX, JSON-LD, HexTuples, RDFa and Microdata
 
 
 * **Store implementations**
 
-  * for in-memory and persistent RDF storage, including remote SPARQL endpoints
+  * for memory stores
```
aucampia, line 16: Maybe it was meant to read "four"?

```suggestion
  * four memory stores
```

aucampia: Otherwise:

```suggestion
  * memory stores
```

docs/rdf_terms.rst
```diff
@@ -109,65 +119,112 @@ identifier -- `W3 RDF Concepts`__
...
+.. __: http://www.w3.org/TR/rdf-concepts/#section-Graph-URIref
+
+Common XSD datatypes
+^^^^^^^^^^^^^^^^^^^^
+
+Most simple literals such as *string* or *integer* have XML Schema (XSD) datatypes defined for them, see the figure 
+below. Additionally, these XSD datatypes are listed in the :class:`XSD Namespace class <rdflib.XSD>` that 
+ships with RDFLib, so many Python code editors will prompt you with autocomplete for them when using it.
+
+Remember, you don't *have* to use XSD datatype and can always make up your own, as GeoSPARQL does, as described above.
```
aucampia, line 212:
```suggestion
Remember, you don't *have* to use XSD datatypes and can always make up your own, as GeoSPARQL does, as described above.
```

rdflib/namespace/__init__.py
```diff
@@ -346,37 +346,27 @@ def __init__(self, graph: "Graph"):
...
+            for prefix, ns in NAMESPACE_PREFIXES_CORE.items():
+                self.bind(prefix, ns)
+        elif bind_namespaces == "cc":
+            # bind any prefix that can be found with lookups to prefix.cc
+            # first bind core and rdflib ones
+            # work out remainder - namespaces without prefixes
+            # only look those ones up
+            raise NotImplementedError("Haven't got to this option yet")
+        else:  # bind_namespaces is None
+            pass  # bind nothing
```
aucampia, line 369: May be a bit better to rather distinguish between explicit `None` and invalid/unknown string so that sending the wrong string results in a failure, will make a PR against your branch to show you what I mean.

nicholascar: I think I get it. I haven't finished all the implementation here of course, so was going to add checks for `None` and other strings too but just haven't got to it so far.

Feel free to put in a PR or commits to the branch!

aucampia: PR here: https://github.com/RDFLib/rdflib/pull/1693

wmelder:
Noticed this dependency in the PR: ``This is in DRAFT until I implement the prexif.cc lookup option``
For me, that lookup option is a nice to have. It would already be an improvement if the options: ``none, core prefixes (default), all RDFLib Namespaces ..`` can be chosen. Using 'none' I can add all namespaces I need and no other defaults would bloat my RDF.

cmungall:
I agree with @wmelder's comments - I don't think having prefix.cc lookup should be core functionality. It would be really awesome if we could get this PR merged as currently these additional prefixes are causing a lot of problems for us.

nicholascar:
I'm held up in completing this PR for, due to my testing, I've think I'm seeing that `bind` with `replace=True` doesn't actually work! This test fails:

```python
def test_rebinding():
    g = Graph()  # 'core' bind_namespaces (default)
    print()
    # 'owl' should be bound
    assert "owl" in [x for x, y in list(g.namespaces())]

    # replace 'owl' with 'sowa'
    # 'sowa' should be bound
    # 'owl' should not be bound
    g.bind("sowa", OWL, replace=True)

    assert "sowa" in [x for x, y in list(g.namespaces())]
    assert "owl" not in [x for x, y in list(g.namespaces())]
```

gjhiggins:
> I'm held up in completing this PR for, due to my testing, I've think I'm seeing that `bind` with `replace=True` doesn't actually work! This test fails:
> ```py
>     assert "owl" not in [x for x, y in list(g.namespaces())]
> ```

I believe that's a _correctly_-failing test according to the implementation, if not according to the API spec. It rather looks like `replace` should be read as "replace the namespace bound to this prefix" rather than "replace the prefix bound to this namespace".

In [`namespace/__init.py__::bind()`](https://github.com/RDFLib/rdflib/blob/03c5e79298b35f29087ff82002becaaba0e3a656/rdflib/namespace/__init__.py#L543), LoC#566 results in `bound_namespace=None` and so the `if bound_namespace and bound_namespace != namespace:` of LoC#573 fails and the execution skips to the `else` at Loc#598, where it skips to the `else` on Loc#674 - resulting in the new binding of sowa to the OWL namespace, that gets added to the trie and returned.

I'll hazard a guess that for `replace` to work as originally intended, the old binding needs to be deleted from the trie.

nicholascar:
> I'll hazard a guess that for replace to work as originally intended, the old binding needs to be deleted from the trie.
> the SimpleMemory and Memory [implementation](https://github.com/RDFLib/rdflib/blob/03c5e79298b35f29087ff82002becaaba0e3a656/rdflib/plugins/stores/memory.py#L402) ignores any already-bound prefix and so the test fails

That's what I though I was seeing, that it was Store-specific and that the Memory stores were failing! I was hoping to be able to restrict my PR changes to the namespace module but clearly it's not possible to sort this our without also touching the Memory stores.

OK, I'll make the updates as outlined above later today.

aucampia:
> That's what I though I was seeing, that it was Store-specific and that the Memory stores were failing! I was hoping to be able to restrict my PR changes to the namespace module but clearly it's not possible to sort this our without also touching the Memory stores.
> 
> OK, I'll make the updates as outlined above later today.

Happy to help here if you don't find time, can try make a separate PR to fix it tomorrow if you want, just ping me.

nicholascar:
Thanks for the offer @aucampia. I think I'd like to give this a go initially. The fix for Memory stores is easy, even though I was trying to avoid having to fiddle with the Stores! - but I'm keen to test out other stores too, especially the SPARQL stores. So I'll give it a shot shortly and reach out to you if I need a hand.

nicholascar:
@aucampia any ideas as to why everything's failing? It looks like a system fault with errors such as `ERROR:   py38: commands failed` rather than application code errors. 

Running locally, everything passes so I think I'm done with the fixes for Stores.

gjhiggins:
Just FTR, there's a relevant extant issue from a while ago - #543 (which contains some informative observations by gromgull of the intended `replace` and `override` functionality) and an unfinished open PR #1094, both of which can be closed when this PR is merged.

aucampia:
> @aucampia any ideas as to why everything's failing? It looks like a system fault with errors such as `ERROR: py38: commands failed` rather than application code errors.
> 
> Running locally, everything passes so I think I'm done with the fixes for Stores.

image

mypy is failing, will make a fix to your branch in a moment.

aucampia:
Another error now:

```
  ==================================== ERRORS ====================================
  ________________ ERROR collecting test/test_namespacemanager.py ________________
  test/test_namespacemanager.py:47: in <module>
      test_rebinding()
  test/test_namespacemanager.py:43: in test_rebinding
      assert "owl" not in [x for x, y in list(g.namespaces())]
  E   AssertionError: assert 'owl' not in ['owl', 'rdf', 'rdfs', 'xsd', 'xml', 'sowa']
  ------------------------------- Captured stdout --------------------------------
  
  =============================== warnings summary ===============================
  .tox/py37-extensive/lib/python3.7/site-packages/_pytest/fixtures.py:228
```

aucampia:
Not sure I understand why this error occurs, will double check tonight.

aucampia:
If the tox output collapsing is annoying (see below) I can also disable it, just let me know.

image

aucampia:
> Not sure I understand why this error occurs, will double check tonight.

Okay the problem is that bind(..., override=True) is also not handled correctly. Fixing it now.

rdflib/namespace/__init__.py
```diff
@@ -564,25 +570,27 @@ def bind(
...
-            if replace:
-                self.store.bind(prefix, namespace)
-                insert_trie(self.__trie, str(namespace))
-                return
+            self.store.bind(prefix, namespace, replace=replace)
+            insert_trie(self.__trie, str(namespace))
 
+            return
+
+        if bound_namespace and bound_namespace != namespace:
```
aucampia, line 586: This branch is unreachable, because if bound_namespace is truthish then the previous if will have matched and function would have returned.

I think @gjhiggins  may have noted this

test/test_namespacemanager.py
```diff
@@ -0,0 +1,47 @@
...
+def test_rebinding():
+    g = Graph()  # 'core' bind_namespaces (default)
+    print()
+    # 'owl' should be bound
+    assert "owl" in [x for x, y in list(g.namespaces())]
+
+    # replace 'owl' with 'sowa'
+    # 'sowa' should be bound
+    # 'owl' should not be bound
+    g.bind("sowa", OWL, replace=True)
```
aucampia, line 40: I think this should be testing with `override` instead of `replace`, as the namespace's prefix is being changed to a new prefix

```
        If override, rebind, even if the given namespace is already
        bound to another prefix.
```

as opposed to changing the namespace associated with a prefix, which is what `replace=True` should be for.

```
        If replace, replace any existing prefix with the new namespace
```

Similarly I think the variable you should be passing is override, not replace.

gjhiggins: Coupla points. 
1. There's an implicit legacy issue in https://github.com/RDFLib/rdflib/blob/6b356a4eacde1280f78b4ed966e8a499a727189a/rdflib/graph.py#L641 where `Graph.__add__()` accumulates namespace bindings which `Graph.__sub__()` ignores.
2. I'll copy [gromgull's observations](https://github.com/RDFLib/rdflib/issues/543) here, for convenience:
> The confusion here is the two arguments `replace` and `override`
> 
> Say they release FOAF 2.0 (Now based on google+? :) ) and you want to change `foaf` to point to `http://xmlns.com/foaf/2.0/`. Up to now you had `foaf=>http://xmlns.com/foaf/0.1` and `foaf2=>http://xmlns.com/foaf/2.0`
> 
> Now you want to "upgrade" to FOAF 2.0 :
> 
> ```py
> g.namespace_manager.bind('foaf', Namespace('http://xmlns.com/foaf/2.0'), .... )
> ```
> 
> Adding:
> 
>     `override` will do what you want, and replace the current binding for foaf to be the new URI.
> 
>     `replace` will replace *any other bindings pointing to the same namespace*, so the 
>      old foaf2 binding will be removed.
> 
> In the FOAF2 example, you probably want both.

and [further](https://github.com/RDFLib/rdflib/issues/543#issuecomment-158754466)

> The problem is (maybe) in the individual implementation, for instance, here:
> 
> https://github.com/RDFLib/rdflib/blob/master/rdflib/plugins/memory.py#L150
> 
> that when replacing, the `uri=>prefix` is updated correctly (i.e. the 
> http://xmlns.com/foaf/2.0 now points to `foaf3`) but the `prefix=>uri` map of course 
> has both. This should be cleaned up if the prefix was already used.
> 
> Also, an unbind method would be handy. Sometimes you want to remove a prefix without providing a new one.

aucampia:
@nicholascar I created a PR against your branch (see https://github.com/RDFLib/rdflib/pull/1786) - this fixes the issues as far as I can tell but it is quite a big change, happy to merge/commit directly to your branch if you are happy with it.

aucampia:
Going to rebase and fix issues

aucampia:
pre-commit.ci autofix

docs/namespaces_and_bindings.rst
```diff
@@ -10,37 +10,103 @@ The :mod:`rdflib.namespace` defines the :class:`rdflib.namespace.Namespace` clas
...
+
+Note that regardless of the strategy employed, prefixes for namespaces can be overwritten with users preferred prefixes,
+for example::
+
+    from rdflib import Graph
+    from rdflib.namespace import GEO  # imports GeoSPARQL's namespace
+    
+    g = Graph(bind_namespaces="rdflib")  # binds GeoSPARQL's namespace to prefix 'geo'
+    
+    g.bind('geosp', GEO)
```
aucampia, line 94: Should this not be:

```suggestion
    g.bind('geosp', GEO, override=True) 
```

rdflib/namespace/__init__.py
```diff
@@ -336,48 +354,43 @@ class NamespaceManager(object):
...
-        self.bind("vann", VANN)
-        self.bind("void", VOID)
-        self.bind("wgs", WGS)
-        self.bind("xsd", XSD)
-
-        # Namespace bindings.
-        self.bind("xml", XMLNS)
+        # for p, n in self.namespaces():  # self.bind is not always called
+        #     insert_trie(self.__trie, str(n))
+
```
aucampia, line 366:
```suggestion
```

I'm guessing this can be removed now.

aucampia (approved):
Looks good to me, I will make a PR once this is merged to maybe change default behaviour slightly though, as currently this will treat any invalid string as the default case which may be surprising to users and could also lead to silent errors where someone things they are using the right option but it gets treated as default and then does not do what they want.

test/test_namespacemanager.py
nicholascar: @gjhiggins could you perhaps provide a test for what you see as correct behaviour? We could add that test and the resolution for it here before merging. Happy to get it right one time.

nicholascar:
Just one more review by @gjhiggins please, including a response to my comment about providing a test for expected behaviour.

test/test_namespacemanager.py
gjhiggins: Sorry to be holding everyone up. I'm trying to get a very clear model of the apparently-intended behaviour but I'm not there yet:
```py
def test_binding_strategy():

    # The confusion here is in the two arguments “override” and “replace” and
    # how they serve two different purposes.

    # 1. Replacing a binding:
    # Say they release FOAF 2.0 and you want to change `foaf` to point to `http://xmlns.com/foaf/2.0/`.

    # Up to now you had `foaf=>http://xmlns.com/foaf/0.1/` and `foaf2=>http://xmlns.com/foaf/2.0/`
    g = Graph(bind_namespaces="none")
    assert len(list(g.namespaces())) == 0

    foaf1_uri = URIRef('http://xmlns.com/foaf/0.1/')
    foaf2_uri = URIRef("http://xmlns.com/foaf/2.0/")

    FOAF1 = Namespace(foaf1_uri)
    FOAF2 = Namespace(foaf2_uri)

    g.bind("foaf", FOAF1)
    g.bind("foaf2", FOAF2)
    assert len(list(g.namespaces())) == 2
    assert list(g.namespaces()) == [('foaf', foaf1_uri), ("foaf2", foaf2_uri)]

    # Now you want to "upgrade" to foaf=>FOAF 2.0.
    g.bind("foaf", FOAF2)

    # But that doesn't happen, instead a different prefix is bound:
    assert list(g.namespaces()) == [("foaf", foaf1_uri), ("foaf1", foaf2_uri)]

    # The prefix "foaf" was already bound to the FOAF1 namespace, so a new
    # prefix for FOAF2 was created.

    # You can achieve the intended result by setting `replace` to `True`:
    g.bind("foaf", FOAF2, replace=True)

    # Because the FOAF2 namespace was rebound to a different prefix, the old 
    # "foaf2" prefix has gone and because the "foaf" prefix was rebound to a
    # different namespace, the old FOAF1 namespace has gone, leaving just:

    assert list(g.namespaces()) == [("foaf", URIRef("http://xmlns.com/foaf/2.0/"))]

    # Now you can re-bind the FOAF1.0 namespace
    g.bind("oldfoaf", FOAF1)

        assert list(g.namespaces()) == [("foaf", foaf2_uri), ('oldfoaf', foaf1_uri)]

    assert list(g.namespaces()) == [('oldfoaf', foaf1_uri)]
```

I need to work through the code to figure out why the last `g.bind("oldfoaf", FOAF1)` causes the `foaf` binding to be lost.

gjhiggins: Oookaayy.
> Just one more review by @gjhiggins please, including a response to my comment about providing a test for expected behaviour.

Kinda lost in the weeds here w.r.t. committing to branches/PRs, the `default_prefixes` branch doesn't appear in the list of branches in my GH clone and I don't seem to be able to change that and in my ignorance, I'm reduced to posting a patch:

```diff
--- rdflib/rdflib/plugins/stores/memory.py  2022-04-08 18:28:03.656450160 +0100
+++ rdflib/rdflib/plugins/stores/memory.py  2022-04-08 16:38:54.877989383 +0100
@@ -403,11 +403,19 @@
                             yield triple, self.__contexts(triple)
 
     def bind(self, prefix, namespace, override=True):
-        bound_prefix = self.__prefix.get(namespace)
-        if override and bound_prefix:
-            del self.__namespace[bound_prefix]
-        self.__prefix[namespace] = prefix
-        self.__namespace[prefix] = namespace
+        bound_namespace = self.__namespace.get(prefix)
+        bound_prefix = self.__prefix.get(namespace) or self.__prefix.get(bound_namespace)
+        if override:
+            if bound_prefix:
+                del self.__namespace[bound_prefix]
+            if bound_namespace:
+                del self.__prefix[bound_namespace]
+
+            self.__prefix[namespace] = prefix
+            self.__namespace[prefix] = namespace
+        else:
+            self.__prefix[bound_namespace or namespace] = bound_prefix or prefix
+            self.__namespace[bound_prefix or prefix] = bound_namespace or namespace
 
     def namespace(self, prefix):
         return self.__namespace.get(prefix, None)
```
I'm unsure whether there's a more elegant means of implementing the above change, so I've refrained from propagating it to the other stores until it's been reviewed.

gjhiggins: Again, apologies for dumping stuff here rather than via a PR to a PR. I worked up an extended set of tests, the first of which (`test_binding_strategy`) I intend to transcribe to doctest format and migrate into the ["Namespaces and bindings" document](https://github.com/RDFLib/rdflib/blob/master/docs/namespaces_and_bindings.rst) (I'll probably do the same with `test_multigraph_bindings` because it's more of an illustration than a test).

```py
import sys
from pathlib import Path
import pytest

from rdflib.term import URIRef
from rdflib import ConjunctiveGraph, Graph, Literal
from rdflib.plugins.stores.memory import Memory
from rdflib.namespace import Namespace, NAMESPACE_PREFIXES_CORE, NAMESPACE_PREFIXES_RDFLIB
from rdflib.namespace import OWL, RDFS
from test.data import tarek, michel, context0, context1, context2

foaf1_uri = URIRef("http://xmlns.com/foaf/0.1/")
foaf2_uri = URIRef("http://xmlns.com/foaf/2.0/")

FOAF1 = Namespace(foaf1_uri)
FOAF2 = Namespace(foaf2_uri)

def test_binding_strategy():

    # The confusion here is in the two arguments “override” and “replace” and
    # how they serve two different purposes --- changing a prefix already bound
    # to a namespace versus changing a namespace already bound to a prefix.

    g = Graph(bind_namespaces="none")
    assert len(list(g.namespaces())) == 0

    # 1. Changing the namespace of an existing namespace=>prefix binding:

    # Say they release FOAF 2.0 and you want to change the binding of
    # `http://xmlns.com/foaf/2.0/` to `foaf`.

    # Up to now you had `http://xmlns.com/foaf/0.1/=>foaf` and `http://xmlns.com/foaf/2.0/=>foaf2`

    # We'll just set up those two bindings ...
    g.bind("foaf", FOAF1)
    g.bind("foaf2", FOAF2)
    assert len(list(g.namespaces())) == 2
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri),
        ("foaf2", foaf2_uri)
    ]

    # Now you want to "upgrade" to FOAF2=>foaf and try the obvious:
    g.bind("foaf", FOAF2)

    # But that doesn't happen, instead a different prefix, `foaf1` is bound:
    assert list(g.namespaces()) == [
        ("foaf", foaf1_uri),
        ("foaf1", foaf2_uri)
    ]

    # The rationale behind this behaviour is that the prefix "foaf" was already
    # bound to the FOAF1 namespace, so a different prefix for the FOAF2 namespace 
    # was automatically created.

    # You can achieve the intended result by setting `replace` to `True`:
    g.bind("foaf", FOAF2, replace=True)

    # Because the FOAF2 namespace was rebound to a different prefix, the old 
    # "foaf2" prefix has gone and because the "foaf" prefix was rebound to a
    # different namespace, the old FOAF1 namespace has gone, leaving just:

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri)
    ]

    # Now, if you wish, you can re-bind the FOAF1.0 namespace to a prefix
    # of your choice
    g.bind("oldfoaf", FOAF1)

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ('oldfoaf', foaf1_uri)
    ]

    # 2. Changing the prefix of an existing namespace=>prefix binding:

    # The namespace manager preserves the existing bindings from any
    # subsequent unwanted programmatic rebindings such as:
    g.bind("foaf", FOAF1)

    # Which, as things stand, results in:

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ('foaf1', foaf1_uri)
    ]

    # In which the attempted change from `oldfoaf` to (the already 
    # bound-to-a different-namespace `foaf`) was intercepted and
    # changed to a non-clashing prefix of `foaf1`

    # So, after undoing the above unwanted rebinding ..
    g.bind("oldfoaf", FOAF1, replace=True)

    # The bindings are again as desired 
    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ('oldfoaf', foaf1_uri)
    ]

    # Next time through, set override to False
    g.bind("foaf", FOAF1, override=False)

    # And the bindings will remain as desired
    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ('oldfoaf', foaf1_uri)
    ]

    # 3. Parsing data with prefix=>namespace bindings 
    # Let's see the situation regarding namespace bindings
    # in parsed data - it can be a bit confusing ...

    # Starting with a very likely example where `foaf` is a
    # prefix for `FOAF1`

    data = """\
    @prefix foaf: <http://xmlns.com/foaf/0.1/> .

    <urn:example:a> foaf:name "Bob" ."""

    g.parse(data=data, format="n3")

    # The FOAF2 namespace is already bound to `foaf` so a
    # non-clashing prefix of `foaf1` is rebound to FOAF1 in
    # place of the existing `oldfoaf` prefix

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ('foaf1', foaf1_uri)
    ]

    # Yes, namespace bindings in parsed data replace existing
    # bindings (i.e. `oldfoaf` was replaced by `foaf1`), just
    # live with it ...

    # A different example of the same principle where `foaf2`
    # replaces the exsting `foaf`

    data = """\
    @prefix foaf2: <http://xmlns.com/foaf/2.0/> .

    <urn:example:b> foaf2:name "Alice" ."""

    g.parse(data=data, format="n3")

    # The already-bound namespace of `foaf=>FOAF2` is replaced
    # by the `foaf2=>FOAF2` binding specified in the N3 source

    assert list(g.namespaces()) == [
        ('foaf1', foaf1_uri),
        ("foaf2", foaf2_uri),
    ]

    # Where a prefix-namespace binding in the data does
    # not clash with the existing binding ...

    data = """\
    @prefix foaf1: <http://xmlns.com/foaf/0.1/> .

    <urn:example:a> foaf1:name "Bob" ."""

    g.parse(data=data, format="n3")

    # The prefix `foaf1`, already bound to FOAF1 is
    # used

    assert list(g.namespaces()) == [
        ('foaf1', foaf1_uri),
        ("foaf2", foaf2_uri),
    ]

    # Otherwise, the existing prefix binding is replaced

    data = """\
    @prefix foaf: <http://xmlns.com/foaf/0.1/> .

    <urn:example:b> foaf:name "Alice" ."""

    g.parse(data=data, format="n3")

    assert list(g.namespaces()) == [
        ("foaf2", foaf2_uri),
        ('foaf', foaf1_uri),
    ]

    # Prefixes are bound to namespaces which in turn have a URIRef
    # representation - so a different prefix=>namespace binding
    # means a different predicate and thus a different statement:

    expected = """
        @prefix foaf: <http://xmlns.com/foaf/0.1/> .
        @prefix foaf2: <http://xmlns.com/foaf/2.0/> .

        <urn:example:a> foaf:name "Bob" .

        <urn:example:b> foaf:name "Alice" ;
            foaf2:name "Alice" .

        """

    s = g.serialize(format="n3")

    for l in expected.split():
        assert l in s


def test_prefix_alias_disallowed():

    # CANNOT BIND A PREFIX ALIASED TO AN EXISTING BOUND NAMESPACE

    g = Graph(bind_namespaces="none")
    g.bind("owl", OWL)
    assert ('owl', URIRef('http://www.w3.org/2002/07/owl#')) in list(g.namespaces())

    g.bind("wol", OWL, override=False)
    assert ('owl', URIRef('http://www.w3.org/2002/07/owl#')) in list(g.namespaces())
    assert ('wol', URIRef('http://www.w3.org/2002/07/owl#')) not in list(g.namespaces())


def test_rebind_prefix_succeeds():

    # CAN REPLACE A PREFIX-NAMESPACE BINDING

    g = Graph(bind_namespaces="none")
    g.bind("owl", OWL)
    assert ('owl', URIRef('http://www.w3.org/2002/07/owl#')) in list(g.namespaces())

    g.bind("wol", OWL)
    assert ('wol', URIRef('http://www.w3.org/2002/07/owl#')) in list(g.namespaces())
    assert ('owl', URIRef('http://www.w3.org/2002/07/owl#')) not in list(g.namespaces())


def test_parse_rebinds_prefix():

    # PARSED PREFIX-NAMESPACE BINDINGS REPLACE EXISTING BINDINGS

    data = """@prefix friend-of-a-friend: <http://xmlns.com/foaf/0.1/> .

    <urn:example:a> friend-of-a-friend:name "Bob" .

    """

    # Use full set of namespace bindings, including foaf 
    g = Graph(bind_namespaces="none")
    g.bind('foaf', FOAF1)
    assert ('foaf', foaf1_uri) in list(g.namespaces())

    g.parse(data=data, format="n3")

    # foaf no longer in set of namespace bindings
    assert ('foaf', foaf1_uri) not in list(g.namespaces())
    assert ('friend-of-a-friend', foaf1_uri) in list(g.namespaces())


@pytest.mark.xfail(reason="Automatic handling of unknown predicates not automatically registered with namespace manager")
def test_automatic_handling_of_unknown_predicates():

    # AUTOMATIC HANDLING OF UNKNOWN PREDICATES

    """
    Automatic handling of unknown predicates
    -----------------------------------------

    As a programming convenience, a namespace binding is automatically
    created when :class:`rdflib.term.URIRef` predicates are added to the graph.
    """

    g = Graph(bind_namespaces="none")

    g.add((tarek, URIRef('http://xmlns.com/foaf/0.1/name'), Literal("Tarek")))

    assert len(list(g.namespaces())) > 0


def test_automatic_handling_of_unknown_predicates_only_effected_after_serialization():

    g = Graph(bind_namespaces="none")

    g.add((tarek, URIRef('http://xmlns.com/foaf/0.1/name'), Literal("Tarek")))

    assert "@prefix ns1: <http://xmlns.com/foaf/0.1/> ." in g.serialize(format="n3")

    assert len(list(g.namespaces())) > 0
    assert ("ns1", URIRef("http://xmlns.com/foaf/0.1/")) in list(g.namespaces())


def test_multigraph_bindings():
    data = """@prefix friend-of-a-friend: <http://xmlns.com/foaf/0.1/> .

    <urn:example:a> friend-of-a-friend:name "Bob" .

    """

    store = Memory()

    g1 = Graph(store, identifier=context1, bind_namespaces="none")

    g1.bind('foaf', FOAF1)
    assert list(g1.namespaces()) == [
        ('foaf', foaf1_uri)
    ]
    assert list(store.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    g1.add((tarek, FOAF1.name, Literal("tarek")))

    assert list(store.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    g2 = Graph(store, identifier=context2, bind_namespaces="none")
    g2.parse(data=data, format="n3")

    # The parser-caused rebind is in the underlying store and all objects
    # that use the store see the changed binding:
    assert list(store.namespaces()) == [
        ('friend-of-a-friend', foaf1_uri)
    ]
    assert list(g1.namespaces()) == [
        ('friend-of-a-friend', foaf1_uri)
    ]

    # Including newly-created objects that use the store
    cg = ConjunctiveGraph(store=store)

    assert ('foaf', foaf1_uri) not in list(cg.namespaces())
    assert ('friend-of-a-friend', foaf1_uri) in list(cg.namespaces())

    assert len(list(g1.namespaces())) == 6
    assert len(list(g2.namespaces())) == 6
    assert len(list(cg.namespaces())) == 6
    assert len(list(store.namespaces())) == 6

    cg.store.add_graph(g1)
    cg.store.add_graph(g2)

    assert "@prefix friend-of-a-friend: <http://xmlns.com/foaf/0.1/>" in cg.serialize(format="n3")

    # In the notation3 format, the statement asserting tarek's name
    # now references the changed prefix:
    assert '<urn:example:tarek> friend-of-a-friend:name "tarek" .' in cg.serialize(format="n3")

    # Add foaf2 binding if not already bound
    cg.bind("foaf2", FOAF2, override=False)
    assert ('foaf2', foaf2_uri) in list(cg.namespaces())

    # Impose foaf binding ... if not already bound
    cg.bind("foaf", FOAF1, override=False)
    assert ('foaf', foaf1_uri) not in list(cg.namespaces())


def test_new_namespace_new_prefix():
    g = Graph(bind_namespaces="none")
    g.bind("foaf", FOAF1)
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri)
    ]


def test_change_prefix_override_true():
    g = Graph(bind_namespaces="none")

    g.bind("foaf", FOAF1)
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    g.bind("oldfoaf", FOAF1)
    # Changed
    assert list(g.namespaces()) == [
        ("oldfoaf", foaf1_uri)
    ]


def test_change_prefix_override_false():
    g = Graph(bind_namespaces="none")

    g.bind("foaf", FOAF1)
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    g.bind("oldfoaf", FOAF1, override=False)
    # No change
    assert list(g.namespaces()) == [
        ("foaf", foaf1_uri)
    ]


def test_change_namespace_override_true():
    g = Graph(bind_namespaces="none")

    g.bind("foaf", FOAF1)
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    g.bind("foaf", FOAF2)
    # Different prefix used
    assert list(g.namespaces()) == [
        ("foaf", foaf1_uri),
        ("foaf1", foaf2_uri)
    ]


def test_change_namespace_override_false():
    g = Graph(bind_namespaces="none")

    g.bind("foaf", FOAF1)
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    # Different namespace so override is irrelevant in this case
    g.bind("foaf", FOAF2, override=False)
    # Different prefix used
    assert list(g.namespaces()) == [
        ("foaf", foaf1_uri),
        ("foaf1", foaf2_uri)
    ]


def test_new_namespace_override_false():
    g = Graph(bind_namespaces="none")

    g.bind("foaf", FOAF2)
    assert list(g.namespaces()) == [
        ('foaf', foaf2_uri)
    ]

    # Namespace not already bound so override is irrelevant in this case
    g.bind("owl", OWL, override=False)
    # Given prefix used
    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ('owl', URIRef('http://www.w3.org/2002/07/owl#')),
    ]


def test_change_namespace_and_prefix():

    # A more extensive illustration of attempted rebinds of
    # `foaf` being intercepted and changed to a non-clashing
    # prefix of `foafN` (where N is an incrementing integer) ...

    g = Graph(bind_namespaces="none")

    g.bind("foaf", FOAF1)
    assert list(g.namespaces()) == [
        ('foaf', foaf1_uri)
    ]

    g.bind("foaf", FOAF2, replace=True)
    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri)
    ]

    g.bind("foaf1", FOAF1)

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ("foaf1", foaf1_uri)
    ]

    foaf3_uri = URIRef("http://xmlns.com/foaf/3.0/")
    FOAF3 = Namespace("http://xmlns.com/foaf/3.0/")

    g.bind("foaf", FOAF3)

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ("foaf1", foaf1_uri),
        ("foaf2", foaf3_uri),
    ]

    foaf4_uri = URIRef("http://xmlns.com/foaf/4.0/")
    FOAF4 = Namespace("http://xmlns.com/foaf/4.0/")

    g.bind("foaf", FOAF4)

    assert list(g.namespaces()) == [
        ("foaf", foaf2_uri),
        ("foaf1", foaf1_uri),
        ("foaf2", foaf3_uri),
        ("foaf3", foaf4_uri),
    ]
```

Phew! Sorry for taking so long to build a solid mental model.

aucampia: @gjhiggins  should these tests be added here? If so I will make a PR to add them, and see if they pass and if not discuss it with you.

gjhiggins: I guess so, ideally they should go into a file in the `test/test_namespace` folder but that folder doesn't yet exist in this branch.

I've (hopefully, helpfully) decanted the trimmed-down test code into a putative `test_namespace_rebinding.py` in [this gist](https://gist.github.com/gjhiggins/b4b59bffc18aa080ce466b01eb0e5d70).

Without the patch (posted above) the `test_binding_replace` (marked xfail) will xfail but will xpass if the patch is applied.

The other xfail pertains to the auto-creation of a namespace from a URIRef predicate.

aucampia: I made a PR with @gjhiggins changes here https://github.com/RDFLib/rdflib/pull/1800.

This is basically what you put in the comment here @gjhiggins - doctests should be fixed ideally  but I think our primary means of testing should always be unit tests, and doctests should be seen as mainly a mechanism to ensure that code in docs is correct.

I think it makes sense to merge https://github.com/RDFLib/rdflib/pull/1800 into this branch as is, as it fixes a bug, though I think the same issue occurs in other stores, so they should also be tested and fixed ideally. I will make another branch to do that.

Ideally this issue should be addressed in another branch, as we are putting a lot of eggs in one basket and basically playing on hard mode, but I guess the best we can do now is just push through and try and get things in from here.

aucampia:
I would actually merge this as is, the problem with putting additional fixes for other issues on this just further delays fixing problems. https://github.com/RDFLib/rdflib/pull/1800 fixes more issues, but that does not make this a regression. But anyway, I will work on https://github.com/RDFLib/rdflib/pull/1800 to fix more issues as I get time, but it may take some days.

aucampia:
@nicholascar let me know if you want me to split things off from this PR, can easily use rebase to keep your commits the same just put them in a seperate branch with you as author, I think a lot of these doc changes are very helpful and very unobjectionable. Up to you though.

aucampia:
Will look at tying up all loose ends here next on 2022-04-18 if there is no further feedback.

gjhiggins (approved):
Well in hand now.

aucampia:
Will rebase and fix the docs warning.

nicholascar:
pre-commit.ci autofix

aucampia:
> This is only failing the documentation building check, but that is being addressed elsewhere, so I'm going to merge this in now.

Happy with that, I figured out what the problem is and will make a PR to fix on master shortly.

cmungall:
it would be awesome to have a new release with this PR in - I'm currently blocked on upgrading rdflib in a few projects as the namespace pollution is causing issues

nicholascar:
@cmungall we're going to get a release out ASAP. Yes, I'm in the same boat as you!

cmungall:
Just a friendly ping to see if there was a timeline for a new release with the namespace pollution removed?

Issue #1159, opened 2020-09-07
code formatting guidelines


FlorianLudwig:
Hey there,

i noticed that there is `black` in the `requirements.dev.txt`. Does it mean that all merge requests are supposed to be formatted with black? The current code in master is not fully formatted by black though, making it harder to run black since it results in unrelated changes.

white-gecko:
With travis we execute flake8 which does some basic style checks. I think this should be the minimum requirement. So far I do not have experience with black.

FlorianLudwig:
Hey @white-gecko,

flake8 uses pycodestyle to check code formatting, which does conflict with black in some cases. I would recommend to either remove black from the `requirements.dev.txt` and by that stop suggesting it as formatter or use it to check the code as well (`black --check` in travis).

Having a code formatter as guide instead of a checker makes it easier for everyone imho ... but there is a certain pain involved when it reformats the code someone formatted by hand and disagrees with the decision made by the formatter. In the end, formatting does not matter that much to argue to much about it.

ashleysommer:
Hi @FlorianLudwig 

Yes, we recommend that all PRs are formatted using Black. That is why it is in `requirements.dev.txt`. [See note below].

You mention:
> The current code in master is not fully formatted by black though

Please refer to these commits:
2020-05-16 [Black all python files](https://github.com/RDFLib/rdflib/commit/0be6f6039479ce29cf71b11e76be54e186130036)
2020-05-25 [Black container.py and other fixes](https://github.com/RDFLib/rdflib/commit/bc213dc048454ceb2d01865a6111b0766277aa8b)
2020-08-19 [More black formatting](https://github.com/RDFLib/rdflib/commit/c1e68b83fa32190888e288b8b9a4d2bc74a0ece1)
2020-08-27 [Re-Run black](https://github.com/RDFLib/rdflib/commit/dca850e9f649689b73c685148f7f01d7448bf4d8)

The whole codebase (well, at lease everything in the `rdflib/` directory) has been Blacked 4 times since May.

I believe the inconsistencies are from merging very old PRs, this causes non-blacked code to come in, necessitating re-running Black occasionally on the whole codebase. Can you let us know which files you believe to be not fully black formatted?

I also find sometimes when working on a small change within one file, when I run Black before I commit the change, Black will make a small change in an unrelated file. I don't understand why it does that, because the other files should already be blacked, but I do still occasionally see it happen.

[...]

PR #1687, opened 2022-01-24
Format whole `*.py` codes

eggplants:
Ref: #1159, https://github.com/RDFLib/rdflib/discussions/1589

## Proposed Changes

I just ran this in the project root: `$ black . && isort .`

### DIFF
--- a/admin/get_merged_prs.py
+++ b/admin/get_merged_prs.py
@@ -1,9 +1,9 @@
 """Get all merged PRs since last release, save them to a JSON file"""
 
-import httpx
 import json
 from datetime import datetime
 
+import httpx
 
 r = httpx.get(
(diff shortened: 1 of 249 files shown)


ashleysommer:
Hi @eggplants 
We usually encourage all PRs to run `black` on their code before merging, and we regularly check the whole codebase with the black style check in our github actions. So the whole codebase should already be formatted according to our `black` rules, and should not need to be changed.

We do not currently use `isort` in this project. I agree it is a good tool, and I use it in several of my personal projects. If we started using it in RDFLib, it would be gradually, starting with only new code and new PRs. There is a lot of _very_ old code in RDFLib (going back to 2001) that has not been touched in years, there is no need to run `isort` over every source file.

eggplants:
@ashleysommer OK, I resubmitted PR.

PR #1672, opened 2022-01-16
pre-commit and pre-commit-ci

aucampia:
Add pre-commit and pre-commit-ci with the following hooks:

- black for formatting code.

For more info see `docs/developers.rst`.

PR #1688, opened 2022-01-24
Fix black

eggplants:
Ref: #1687 

## Proposed Changes

  - fix black command argument in CI (`./rdflib` -> `.`)
  - format with `black --config black.toml .`
    - formatted:
      - rdflib/namespace/_GEO.py 
      - rdflib/plugins/parsers/hext.py
      - rdflib/plugins/serializers/hext.py
  - sort dev dependencies in `requirements.dev.txt`

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -87,7 +87,7 @@ jobs:
             1>&2 echo "Running with fuseki"
             test_harness+="./with-fuseki.sh"
           fi
-          black --config black.toml --check . || true
+          black --config black.toml --check --diff . || true
           flake8 --exit-zero rdflib
           mypy --show-error-context --show-error-codes
           "${test_harness[@]}" pytest -ra --cov
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -87,7 +87,7 @@ jobs:
             1>&2 echo "Running with fuseki"
             test_harness+="./with-fuseki.sh"
           fi
-          black --config black.toml --check ./rdflib || true
+          black --config black.toml --check . || true
           flake8 --exit-zero rdflib
           mypy --show-error-context --show-error-codes
           "${test_harness[@]}" pytest -ra --cov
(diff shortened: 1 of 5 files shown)


ashleysommer:
> fix black command argument in CI (./rdflib -> .)

It was intentional that we only check for `black` formatting in the `/rdflib` subdirectory.

@nicholascar 
Do you know why the `_GEO.py`, hext-parser, and hext-serializer were not Blacked before they were merged?

nicholascar:
> Do you know why the _GEO.py, hext-parser, and hext-serializer were not Blacked before they were merged?

My mistake? They should have been! Clearly I'm responsible for those particular files...

aucampia (approved):

aucampia:
> Do you know why the `_GEO.py`, hext-parser, and hext-serializer were not Blacked before they were merged?

I think with nothing doing any check as it stands it is easy to forget, this may help:

- https://github.com/RDFLib/rdflib/pull/1672

aucampia:
> > fix black command argument in CI (./rdflib -> .)
> 
> It was intentional that we only check for `black` formatting in the `/rdflib` subdirectory.

The current config should exclude everything else and also relying on config simplifies integration with other tooling. But should be same behavior.

.github/workflows/validate.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
             1>&2 echo "Running with fuseki"
             test_harness+="./with-fuseki.sh"
           fi
-          black --config black.toml --check ./rdflib || true
+          black --config black.toml --check --diff . || true
```
aucampia, line 90:
```suggestion
          black --config black.toml --check --diff ./rdflib || true
```

Let's rather keep it as is for now, but ideally we should try and rely solely on config for these tools.

aucampia (approved):
looks good, thanks.

aucampia:
actually, `black.toml` has no effect: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file - so actually for the time being it is quite critical to keep the specific directory specification. I will make a separate pr to rename it to pyproject.toml.

aucampia:
Fix for black config: #1692

nicholascar (approved):

PR #1691, opened 2022-01-24
Improved tox config

aucampia:
- Removed some redundant options.
- Created a covreport environment for reporting coverage for all
environments.
- Make it possible to run a custom command instead of pytest.
- Make it possible to run an extra command
- Make it possible to change the mypy command
- Use extras instead of using requirements files.
- Improve rebuild speed of docs environment by not using an ephemeral doctree dir.

Also:

- Update setup.py
  - Remove berkleydb and networkx from tests_require and tests extra
    group as they are not always needed for tests.
  - Added separate extra groups for berkleydb and networkx
  - Added a dev extras group with dev tools.
- Update github actions
  - Use tox to run tests.
  - Run flake8 only once instead of for every OS and python version.
  - Change docs job to rather be more generic for potential future tox
    environments.

NOTE: Split off from #1672

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -20,29 +20,39 @@ jobs:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-latest, macos-latest, windows-latest]
         include:
-          - {
-              python-version: "3.7",
-              os: ubuntu-latest,
-              extensive-tests: true,
-            }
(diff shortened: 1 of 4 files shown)


nicholascar (approved):
Looks good to me. I matched the black version to PR #1763 but otherwise unchanged

white-gecko (approved):
Looks good to me.

PR #1692, opened 2022-01-24
Rename black.toml to pyproject.toml

aucampia:
Black does not read config from black.toml, and without this change the
config has no effect and tools like pre-commit ci will not do what we
want.


nicholascar (approved):
Approving & merging without second review as a simple PR

PR #1693, opened 2022-01-24
Make `bind_namespaces` optional

aucampia:
Use `None` to indicate no-namespace binding so that invalid values are
treated as errors.


nicholascar:
@aucampia once the main `bind_namespaces` PR is through, we can revisit this one as there's double-up here with the allowance of `None` v. my addition of "none". Happy to go with this version instead of "none".

PR #1697, opened 2022-01-25
Apply IdentifiedNode to Graph iterators

### LINKED ISSUE #1696: IdentifiedNode should be used in Graph.triples() and related iterators
I started a patch to try to strengthen the type signature of `Graph.triples()`.  The gist is, the signature should be changed from:

```python
def triples(
    self, triple: Tuple[Optional[Node], Union[None, Path, Node], Optional[Node]]
):
```

to

```python
def triples(
    self, triple: Tuple[Optional[IdentifiedNode], Union[None, Path, IdentifiedNode], Optional[Node]]
) -> Iterable[Tuple[IdentifiedNode, IdentifiedNode, Node]]:
```

to let the type system catch "thing-or-string" errors from trying to put `Literal`s into the wrong triple position.

I implemented the above change (draft PR coming in a minute, I need the Issue number first), and followed through with the analogous signatures for `Graph.subjects`, `Graph.subject_predicates`, and alike for the SPO combinations.  This revealed some deeper architectural issues that prevent immediate adoption of the new signature.  Issues were found via `mypy`.  (I'm also aware the middle triple-member of the return type needs some review with respect to `Path`.)

* `/rdflib/graph.py` and `/rdflib/paths.py` have circular dependencies that seem to necessitate a higher in-common module.  The issues boil down to `evalPath()` needing the signature of `Graph.triples()`, but `graph.py` imports `paths.py` and defines `Graph.triples` using `Path.eval()`, which in all subclass implementations calls `evalPath()`, closing the loop.
* `/rdflib/compare.py` has some type signatures that assumed general `Node`s, and from a 2-minute glance it isn't immediately obvious to me whether a `IdentifiedNode` can be used in place of `Node` usage from before.  Patch 2 in the PR fixes these, and IMO looks good enough to merge independently except for the dangling `Path` question.
* There are typing issues in `rdflib/plugins/serializers/rdfxml.py` that I think I managed to work around, using `IdentifiedNode` instead of `Identifier`.  Patch 3 in the PR fixes these, and (also) IMO looks good enough to merge independently except for the dangling `Path` question.

So, this type signature revision ended up blossoming into a rearchitecting project.  If the maintainers are interested in the signature on `Graph.triples()` being refined to enable some safety around `Literal` placement, what would be the next steps, broadly speaking?  Moving the circular type dependencies into a higher inner module?

ajnelson-nist:
Fixes #1696 

## Proposed Changes

  - Use `IdentifiedNode` on `Graph.triples()` and related methods
  - Apply type signature revisions to `compare.py` and `rdf-xml` serializer.
  - (More to come after discussion on #1696 settles.)

Note the `Graph` signature updates should also be propagated to `Graph` subclasses before this is merged.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -18,7 +18,7 @@ import random
 from rdflib.namespace import Namespace, RDF
 from rdflib import plugin, exceptions, query, namespace
 import rdflib.term
-from rdflib.term import BNode, Node, URIRef, Literal, Genid
+from rdflib.term import BNode, IdentifiedNode, Node, URIRef, Literal, Genid
 from rdflib.paths import Path
 from rdflib.store import Store
 from rdflib.serializer import Serializer
@@ -453,8 +453,8 @@ class Graph(Node):
         return self
 
     def triples(
-        self, triple: Tuple[Optional[Node], Union[None, Path, Node], Optional[Node]]
-    ):
+        self, triple: Tuple[Optional[IdentifiedNode], Union[None, Path, IdentifiedNode], Optional[Node]]
+    ) -> Iterable[Tuple[IdentifiedNode, IdentifiedNode, Node]]:
         """Generator over the triple store
 
         Returns triples that match the given triple pattern. If triple pattern
@@ -672,7 +672,7 @@ class Graph(Node):
         self.add((subject, predicate, object_))
         return self
 
-    def subjects(self, predicate=None, object=None, unique=False) -> Iterable[Node]:
+    def subjects(self, predicate: Union[None, Path, IdentifiedNode]=None, object: Optional[Node]=None, unique: bool=False) -> Iterable[IdentifiedNode]:
         """A generator of (optionally unique) subjects with the given
         predicate and object"""
         if not unique:
@@ -691,7 +691,7 @@ class Graph(Node):
                         )
                         raise
 
-    def predicates(self, subject=None, object=None, unique=False) -> Iterable[Node]:
+    def predicates(self, subject: Optional[IdentifiedNode]=None, object: Optional[Node]=None, unique: bool=False) -> Iterable[IdentifiedNode]:
         """A generator of (optionally unique) predicates with the given
         subject and object"""
         if not unique:
@@ -710,7 +710,7 @@ class Graph(Node):
                         )
                         raise
 
-    def objects(self, subject=None, predicate=None, unique=False) -> Iterable[Node]:
+    def objects(self, subject: Optional[IdentifiedNode]=None, predicate: Union[None, Path, IdentifiedNode]=None, unique: bool=False) -> Iterable[Node]:
         """A generator of (optionally unique) objects with the given
         subject and predicate"""
         if not unique:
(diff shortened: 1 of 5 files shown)


ajnelson-nist:
This is ready for look-overs.  Let's keep discussion to #1696.

aucampia:
I will have a look later this week, but this is quite critical to do, so thanks. I will also see if I can make CI pipeline pass.

rdflib/graph.py
```diff
@@ -453,20 +453,23 @@ def remove(self, triple):
         return self
 
     def triples(
-        self, triple: Tuple[Optional[Node], Union[None, Path, Node], Optional[Node]]
-    ):
+        self,
+        triple: Tuple[
+            Optional[IdentifiedNode], Union[None, Path, IdentifiedNode], Optional[Node]
+        ],
+    ) -> Iterable[Tuple[IdentifiedNode, Union[IdentifiedNode, Path], Node]]:
```
aucampia, line 460: I think this is more correct than it was, but technically this should be:

```suggestion
        self,
        triple: Tuple[
            Optional[IdentifiedNode], Union[None, Path, URIRef], Optional[Node]
        ],
    ) -> Iterable[Tuple[IdentifiedNode, Union[URIRef, Path], Node]]:
```

Fine with it as is though, we can narrow it more later, this narrow already makes things better.

Same applies other in places also, won't label them all, just a note.

aucampia:
Change looks very good, I made some minor fixes in https://github.com/ajnelson-nist/rdflib/pull/14 to get the type checking to pass, and I also roll back some changes which is not entirely essential for this PR's aim, but it's up to you what to do with those.

aucampia:
For some reason mypy does not complain about unused ignores on my computer, will have to figure out why, but I removed the ignores that are not needed now and type validation seems to pass.

ajnelson-nist:
Thanks!  I just hit that too, and I'm dealing with a weirdness in `rdflib/plugins/serializers/rdfxml.py`, line 190:

```
rdflib/plugins/serializers/rdfxml.py:190: error: Incompatible types in assignment (expression has type "Set[IdentifiedNode]", variable has type "Set[Node]")
rdflib/plugins/serializers/rdfxml.py:190: error: Argument 1 to "union" of "set" has incompatible type "Iterable[Node]"; expected "Iterable[IdentifiedNode]"
```

Partially reverting your PR to split that up again is...almost helping.  I'm close to pushing a patch once I figure out the spirit of the `possibles` variable.

aucampia:
This validates on github actions now, I think maybe I have an older version of mypy than is being used on github actions

ajnelson-nist:
Hm, seems I had an old-mypy issue too.  I was on 0.920, upgraded to 0.931, and now that curious issue with `possibles` has gone away.  (Though, squinting at that line, I think it'll be worth an independent PR.)

I'm going to take this PR out of Draft mode now.  It seems ready for any last review and merge to me.

ajnelson-nist:
...I spoke a little soon, I missed `black` wanted to reformat a bit.  NOW I think it's ready.

aucampia:
pre-commit.ci autofix

aucampia (approved):
This looks good to me, thanks again @ajnelson-nist

This only changes type annotations with one small exception, which is a change in variable names in well tested code path, this change in variable names is needed for type checking to pass without error/ignores as mypy does not like that the same variable has different types in the same scope:

https://github.com/RDFLib/rdflib/blob/06feef03ec071a265f79ea08d95b91be62ea2f70/rdflib/graph.py#L495-L497

This PR also narrows some type annotations, but this narrowing brings things more in line with RDF 1.1 spec so I think it is good, if someone wants to use rdflib with non compliant types and type checking they will have to use `# type: ignore` or `cast` - but on the other hand, if we try and cater for non compliant RDF it also makes our lives harder.

The narrower types should however help users of this library detect errors in their code where they are potentially creating invalid RDF graphs.

I will add a page in docs regarding type annotations which covers some of the questions in #1696 and #1447

PR #1703, opened 2022-02-01
Use the `property` built-in as a decorator

aucampia:
The use of the `property` built-in as a function does not allow for
the same type annotations as using `property` as a decorator.

This change should not affect runtime behaviour, but I added tests to
cover all properties that were changed to decorators explicitly just to
make sure they work as they should.

There are still other places where the `property` built-in is used as a
function but this is outside of the core parts of RDFLib.

---

For some context, I'm busy adding typing to store/queries and without this things get a bit hairy. I'm doing this separately as this goes beyond simple addition of type annotations, and as such want this to stand on it's own to make it easier to review and asses.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -350,29 +350,26 @@ class Graph(Node):
         self.formula_aware = False
         self.default_union = False
 
-    def __get_store(self):
+    @property
+    def store(self):
         return self.__store
 
-    store = property(__get_store)  # read-only attr
-
-    def __get_identifier(self):
+    @property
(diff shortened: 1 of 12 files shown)


aucampia:
Actually I'm going to add more tests first, all lines are covered by tests but there are no explicit tests for these properties.

aucampia:
> Actually I'm going to add more tests first, all lines are covered by tests but there are no explicit tests for these properties.

added now.

nicholascar (approved):
Seems reasonable to me!

PR #1718, opened 2022-02-17
`DefinedNamespaceMeta.__getitem__` is slow

anatoly-scherbakov:
Fixes #

## Proposed Changes

  - Cache the `__getitem__` method to improve performance.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import warnings
+from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union, Iterable
 from unicodedata import category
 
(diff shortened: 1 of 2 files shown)


aucampia (approved):
Looks good to me, thanks for the PR.

white-gecko (commented):
Looks good to me, just a question, why not just use `@cache`, according to https://docs.python.org/3/library/functools.html#functools.cache it is the same as `lru_cache(maxsize=None)` and would be easier to read.

aucampia:
> Looks good to me, just a question, why not just use `@cache`, according to https://docs.python.org/3/library/functools.html#functools.cache it is the same as `lru_cache(maxsize=None)` and would be easier to read.

[`@cache`](https://docs.python.org/3/library/functools.html#functools.cache) would be better but it is only available from 3.9.

aucampia:
Will merge this weekend if there is no further feedback.

white-gecko:
> > Looks good to me, just a question, why not just use `@cache`, according to https://docs.python.org/3/library/functools.html#functools.cache it is the same as `lru_cache(maxsize=None)` and would be easier to read.
> 
> [`@cache`](https://docs.python.org/3/library/functools.html#functools.cache) would be better but it is only available from 3.9.

Thank you for the clarification. I think we can merge this now.

PR #1719, opened 2022-02-20
Use the correct warnings module

aucampia:
Accidentally used the one from sphinx instead of the builtin one, this
fixes the situation.


nicholascar:
Ooo, this is an interesting find! Is it just a typo that's included the wrong warnings here in the past?

aucampia:
> Ooo, this is an interesting find! Is it just a typo that's included the wrong warnings here in the past?

It came from vscode auto import, I noticed it when rebasing the tox pr.

nicholascar (approved):

PR #1721, opened 2022-02-22
Update reference to black.toml

ajnelson-nist:
Updates #1692 

## Proposed Changes

  - Update Github workflow to no longer refer to black.toml

This could be reduced further and cut the `--config` flag.  I personally prefer the explicit reference, for the ease of finding configuration parameters.


aucampia (approved):

aucampia:
Thank you very much for the fix @ajnelson-nist

PR #1759, opened 2022-03-14
Remove redundant type ignores and fix typing errors

aucampia:
This seems to be related to changes in typeshed and mypy.

Also fixed comparison done by `_isEqualXMLNode` for
`node.nodeType == XMLNode.DOCUMENT_TYPE_NODE`,
it was comparing `node.systemId == other.system.Id`
instead of `node.systemId == other.systemId`.


aucampia:
Just want to add a test for `_isEqualXMLNode`

aucampia:
Okay tests added now, issue reproduced here: https://github.com/RDFLib/rdflib/runs/5546115541

white-gecko (approved):
Everything fine for me.

Issue #1760, opened 2022-03-15
Issues with installing rdflib 


SubaSiva95:
Hi, I am trying to install rdflib through pip to integrate with the dynamo platform. But, I am getting an error message,

"Warning: AttributeError : ‘NoneType’ object has no attribute ‘isatty’ [’ File “”, line 4, in \n’, ’ File “----------------\rdflib\init.py”, line 99, in \n and sys.stderr.isatty()\n’]".

The "init.py" is one of the default files existing within the rdflib folder.

Could anyone help me to resolve this issue?
Thanks

eggplants:
Related issue and its fix:
https://github.com/tornadoweb/tornado/issues/920
https://github.com/tornadoweb/tornado/commit/689d9c8e9a48e4695a8b172ff5b02a6ba41fd7cb

aucampia:
> Hi, I am trying to install rdflib through pip to integrate with the dynamo platform. But, I am getting an error message,
> 
> "Warning: AttributeError : ‘NoneType’ object has no attribute ‘isatty’ [’ File “”, line 4, in \n’, ’ File “----------------\rdflib\init.py”, line 99, in \n and sys.stderr.isatty()\n’]".
> 
> The "init.py" is one of the default files existing within the rdflib folder.
> 
> Could anyone help me to resolve this issue? Thanks

Please share the command you use to install it and some details of your environment? Technically according to python docs `sys.stderr` should not be `None`, so I suspect there is something wrong with your environment, thought he code that is causing this error should likely be removed.

SubaSiva95:
I am using dynamo 2.10 (which is built-in with Revit 2022). I followed "https://github.com/DynamoDS/Dynamo/wiki/Customizing-Dynamo's-Python-3-installation" and installed rdflib through pip. 

Based on the previous comments, I added a line in the following section of the (init.py) file.

  if (
        not hasattr(__main__, "__file__")
        and sys.stdout is not None
	**and hasattr(sys.stderr, 'isatty')**
        and sys.stderr.isatty()
    ):

Now, I am free from errors, but unfortunately, the results that I get from the script is always null.

semanticmachinelearning:
Hi I use pycharm and pip to install rdflib but waiting for the whole day still not able to accessing it. I tested on Jupiter notebook, the same thing. I believe there are something wrong to receive the installation package at the moment.

PR #1761, opened 2022-03-15
Check if sys.stderr has isatty

eggplants:
Closes: #1760 

### Proposed Changes:

- Check if `sys.stderr` has `isatty` in `rdflib/__init__.py`.


aucampia:
Please include tests.

aucampia:
> Please include tests.



@eggplants Looking at the issue this is trying to I think I can accept this without tests, however tests always makes it an easier choice to approve.

I would like to see this code being removed though at some point in time, logging setup should be left to the user of the library, as with this approach it is likely that the setup is not as users want it, but I guess while it is there best to make it not break.

aucampia:
> Please include tests.



@eggplants Looking at the issue this is trying to I think I can accept this without tests, however tests always makes it an easier choice to approve.

I would like to see this code being removed though at some point in time, logging setup should be left to the user of the library, as with this approach it is likely that the setup is not as users want it, but I guess while it is there best to make it not break.

aucampia (approved):
Approving but will wait for at least one more review and will really appreciate some tests here to make sure we can actually reproduce the issue and ensure it is fixed.

eggplants:
> Do I get it right, only in the case that sys.stderr has isatty() the logging is enabled?
Yes.

aucampia:
Test works fine even without the change.

eggplants:
Is there a way to overwrite `sys.stderr` with `None` from the outside without changes to `__init__.py`? `sys.stderr = None` in the test seem to be reflected. Sorry for my stupidity.

aucampia:
I think the best way to test it is to create a new python process with stdout closed, which will result in stdout being null as far as I understand, though I'm not entirely sure. 

https://github.com/python/cpython/blob/d4a93e4de17ed8c2334fc72ca4cee737c3b738ec/Python/bltinmodule.c#L1871-L1873

```c
        /* sys.stdout may be None when FILE* stdout isn't connected */
        if (file == Py_None)
            Py_RETURN_NONE;
```

aucampia:
The other option is to use one of:

- https://docs.pytest.org/en/stable/how-to/monkeypatch.html
- https://docs.python.org/3/library/unittest.mock.html

nicholascar (approved):
Approving but can't merge until Drone is working again and all tests are seen to pass.

aucampia:
Removed the test file as it is out of sync with master branch organization and it passes even without this PR, can be added in another PR if there is something else it tests for.

aucampia:
CI passed now, again the test is removed as it passes with or without this change, I will merge tonight still if there are no objections. I think this change is safe enough.

PR #1762, opened 2022-03-15
Fix flake8 errors

eggplants:
Fixes flake8 errors for passing CI.

PR #1763, opened 2022-03-15
Fix black

eggplants:
Ref: #1762 

I changed to install and use black in the environment created by pre-commit instead of the local one when running pre-commit hooks.

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -10,8 +10,7 @@ This document describes the process and conventions to follow when
 developing RDFLib code.
 
 * Please be as Pythonic as possible (:pep:`8`).
-* Code should be formatted using `black <https://github.com/psf/black>`_
-  and we use Black v21.9b0, with the black.toml config file provided.
+* Code should be formatted using `black <https://github.com/psf/black>`_.
 * Code should also pass `flake8 <https://github.com/psf/black>`_ linting
   and `mypy <http://mypy-lang.org/>`_ type checking.
 * You must supply tests for new code
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -10,7 +10,7 @@ This document describes the process and conventions to follow when
 developing RDFLib code.
 
 * Please be as Pythonic as possible (:pep:`8`).
-* Code should be formatted using `black <https://github.com/psf/black>`_.
+* Code should be formatted using `black <https://github.com/psf/black>`_  and we use Black v21.9b0, with the black config in ``pyproject.toml``.
 * Code should also pass `flake8 <https://github.com/psf/black>`_ linting
   and `mypy <http://mypy-lang.org/>`_ type checking.
 * You must supply tests for new code
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -10,7 +10,7 @@ This document describes the process and conventions to follow when
 developing RDFLib code.
 
 * Please be as Pythonic as possible (:pep:`8`).
-* Code should be formatted using `black <https://github.com/psf/black>`_  and we use Black v21.9b0, with the black config in ``pyproject.toml``.
+* Code should be formatted using `black <https://github.com/psf/black>`_  and we use Black v22.1.0, with the black config in ``pyproject.toml``.
 * Code should also pass `flake8 <https://github.com/psf/black>`_ linting
   and `mypy <http://mypy-lang.org/>`_ type checking.
 * You must supply tests for new code
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [tool.black]
-required-version = "22"
+required-version = "22.1.0"
 line-length = "88"
 skip-string-normalization = true
 target-version = ['py37']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,5 @@
 [tool.black]
+required-version = "22"
 line-length = "88"
 skip-string-normalization = true
 target-version = ['py37']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,4 @@
 [tool.black]
-# WARNING: This should be the same as in `.pre-commit-config.yaml`.
-required-version = "21.9b0"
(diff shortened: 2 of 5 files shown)


requirements.dev.txt
```diff
@@ -1,8 +1,7 @@
-black==21.9b0
+black
 coverage
 doctest-ignore-unicode==0.1.2
 flake8
-flake8-black
```
aucampia, line 5: Why remove this?

aucampia (commented):
Thanks for the PR @eggplants it looks good. I am in favor of not pinning the version of black, though I think if possible we should try and figure out why this was done to begin with.

Also flake8-black is quite useful as it integrates with VSCode, I would not remove it unless there is a specific problem it causes.

requirements.dev.txt
```diff
@@ -1,8 +1,7 @@
-black==21.9b0
+black
 coverage
 doctest-ignore-unicode==0.1.2
 flake8
```
aucampia, line 4:
```suggestion
flake8
flake8-black
```

eggplants:
As you can see of this:
https://github.com/RDFLib/rdflib/runs/5561399664?check_suite_focus=true

flake8-black 0.3.2 (latest) outputs many unexpected and useless error like this:

```
rdflib/compare.py:0:1: BLK999 Unexpected exception: int object expected; got str
```

eggplants:
The problem I reported some time ago seems to be recurring.
https://github.com/peterjc/flake8-black/issues/47

nicholascar:
Re: pinning of black version

> I think if possible we should try and figure out why this was done to begin with. 

This was done because black was, a year or two ago, undergoing rapid development and its outputs were different between versions which caused Git diffs as people used different versions over time. So we pinned to one version to lock things into being exactly the same.

I think we can pin to black per major or semi-major release of RDFLib so we can update black over time, just not with every PR!

ashleysommer:
> its outputs were different between versions which caused Git diffs as people used different versions over time

Thats right. Eg, if I had v21.2b0 installed in my development environment, ran `black` before my PR, my PR got merged. But you have `black` v21.4b0 on your system, now when you run `black` some of my formatted code gets changed again, causing diff noise in almost every pull request.
Then when we introduced black checking on the CI test pipeline, it used v21.6b0, and now PRs are flagged as non-conformant to Black, even if you formatted it with your version before pushing.
So we decided to settle on "this is the one black version we use for RDFLib", we put that in the documentation, and pinned that in the requirements.
In v21.6b0 black even introduced the `--required-version` option when running `black`, so that your check would fail if you were not using the exact given version f the tool, that is to prevent the issues we've seen.

Now Black is out of the recent January release, `black v22.1.0` is finally out of Beta. So that means we should definitely update to using that as "the black version we use for RDFLib". However I'm not sure about unpinning it, and I still think we should have `--required-version` set to match, eg `22.1.0`.

nicholascar:
> I'm not sure about unpinning it

I agree that we should keep pinning it, but I think every so often we can update what we have pinned to, like every major release. 

So let's now pin to 22.1.0.

ashleysommer:
@nicholascar @eggplants 
I may have to adjust my recommendation. I just saw the new `Black Stability Policy`: https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy

So that means the output of Black will be the same for the whole calendar year. So we can set the requirements-dev.txt line to `>=22.0,<23` and the `--required-version` to "22", and _all will be fine_ (famous last words).

nicholascar:
> I may have to adjust my recommendation...

Makes sense that they have such a policy! Thanks for finding it. So let's do this.

pyproject.toml
```diff
@@ -1,6 +1,4 @@
 [tool.black]
-# WARNING: This should be the same as in `.pre-commit-config.yaml`.
-required-version = "21.9b0"
```
ashleysommer, line 3: Can we change this now to `required-version = "22"`

nicholascar: Just re-added. Do we need to re-add stuff to the pre-commit hook too, or is that covered by this?

aucampia:
Partial version spec does not work for me:

```console
$ .venv/bin/black --version
black, 22.1.0 (compiled: yes)
$ .venv/bin/black rdflib
Oh no! 💥 💔 💥 The required version `22` does not match the running version `22.1.0`!
$ .venv/bin/black --required-version 22 rdflib
Oh no! 💥 💔 💥 The required version `22` does not match the running version `22.1.0`!
$ .venv/bin/black --required-version 22.1.0 rdflib
All done! ✨ 🍰 ✨
116 files left unchanged.
```

Also, if we want to use a fixed version with pre-commit we sadly have to keep the local hook as it was before this PR.

I will push changes to your branch @eggplants - I would like it if there was some way to make a partial version spec (i.e. version range) work, but I can't find any.

nicholascar (approved):
OK, looks like it might be as good as we can get it and at least we've updated to a major stable black version

eggplants:
@ashleysommer @aucampia @nicholascar Thank you very much for the correction to my poor PR! I hope this makes CI better.

nicholascar:
Thanks you @eggplants for kicking off this improvement!

Will merge as soon as I see a confirmation of the final PR from @ashleysommer or @aucampia

docs/developers.rst
```diff
@@ -10,8 +10,7 @@ This document describes the process and conventions to follow when
 developing RDFLib code.
 
 * Please be as Pythonic as possible (:pep:`8`).
-* Code should be formatted using `black <https://github.com/psf/black>`_
-  and we use Black v21.9b0, with the black.toml config file provided.
+* Code should be formatted using `black <https://github.com/psf/black>`_  and we use Black v21.9b0, with the black config in ``pyproject.toml``.
```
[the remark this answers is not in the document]
eggplants: 9abbbcb35fc99558616091490c7c9c15d5b7a3ee

white-gecko (approved):
Looks good to me.

PR #1765, opened 2022-03-17
Move test_issue*.py into test_issues folder

gjhiggins:
## Proposed Changes

  - Move tests of issues into their own folder.


nicholascar (approved):
Looks good to me: I'm happy to shorten the list of test files a developer is confronted with!

I've slowely tried to add test_parser_* for parser testing and test_serializer* for serializers, but, I suppose, better would be to have a parsers/ and serializers/ folder too, so we can drop the _parser & _serializer bit in the file names.  @gjhiggins if you can, you could make that change here! If not, I'll do it after this PR is merged.

PR #1771, opened 2022-03-23
Fix for issue1769

gjhiggins:
# Summary of changes
Fix for issue #1769, suggested extension of `from_n3` to handle “numeric shortcut” strings containing (otherwise valid) characters that cause `isdigit` to return `False`, add test illustrating previously-failing examples.

**Or** continue with the current “fail-safe” approach, reject this PR, resolve issue #1769 as “wontfix” and explicitly note the use of `isdigit` in the documentation along with the implications for numeric shortcut strings containing negative numbers and scientific notation --- the non-shortcut representation (e.g. '"-5"^^xsd:integer') is properly handled:
```py
assert rdflib.util.from_n3('"-5"^^xsd:integer') == rdflib.term.Literal(
    '-5', datatype=rdflib.term.URIRef('http://www.w3.org/2001/XMLSchema#integer'))
```

# Checklist

  the same change.
- [x] Checked that all tests and type checking passes. (but see ¹)

¹
```
$ mypy rdflib/util.py 
rdflib/term.py:1909: error: Item "Element" of "Union[Document, Element]" has no attribute "attributes"
rdflib/term.py:1914: error: Item "Element" of "Union[Document, Element]" has no attribute "attributes"
Found 2 errors in 1 file (checked 1 source file)
```


aucampia:
I'm somewhat hesitant to maintain `rdflib.util.from_n3` as it is not used in actual parsing and has issues that is not present in actual parsers (see https://github.com/RDFLib/rdflib/pull/1778), so I would prefer we deprecate it and rather advise people to use the real parsers.

However if we do plan to keep it we should fix it so it does work for n3 as it's name suggests, I would prefer we somehow hook into the actual n3 parser, but otherwise any fix that makes it more conformant is good.

rdflib/util.py
```diff
@@ -203,8 +203,25 @@ def from_n3(s: str, default=None, backend=None, nsm=None):
...
         return Literal(s == "true")
-    elif s.isdigit():
-        return Literal(int(s))
+    elif (
+        s.lower()
+        .replace('.', '', 1)
+        .replace('-', '', 1)
+        .replace('e', '', 1)
+        .isnumeric()
+    ):
```
nicholascar, line 212: Couldn't this be better written as:

"""python
if "e" in s.lower():
    return Literal(s, datatype=XSD.double)
else:
    return Literal(float(s), datatype=XSD.decimal) if "." in s else Literal(int(s), datatype=XSD.integer)
"""

nicholascar (approved):
Fine other than my one style only change request in a comment

aucampia:
@gjhiggins would you mind if I integrate the tests here into https://github.com/RDFLib/rdflib/blob/8a309774665182df135be9f2707303571764320a/test/test_parsers/test_parser_turtlelike.py ?

Would prefer that as they should pass for many formats, and we likely have many other tests we want to perform on both from_n3 and on the other formats.

gjhiggins:
> @gjhiggins would you be okay if I integrate the tests here into https://github.com/RDFLib/rdflib/blob/8a309774665182df135be9f2707303571764320a/test/test_parsers/test_parser_turtlelike.py ?

Oh, absolutely. I have zero sense of "ownership" - except perhaps for stuff other people wouldn't touch with a bargepole :smile: but thanks for asking.

> Would prefer that as they should pass for many formats, and we likely have many other tests we want to perform on both from_n3 and on the other formats.

There's more to come in the general context of the test matrices area w.r.t. the dataset rework, I've been handling it thusly:

```py
example2_root = os.path.join(
    CONSISTENT_DATA_DIR, "example-2-default-and-two-named-graphs."
)

@pytest.fixture
def example2(request):
    d = Dataset()
    d.bind("ex", Namespace("http://example.org/"))

    alice = BNode()  # Alice
    bob = BNode()  # Bob

    alice_graph = d.graph(alice_uri)
    bob_graph = d.graph(bob_uri)

    d.add((alice_uri, DCTERMS.publisher, Literal("Alice")))
    d.add((bob_uri, DCTERMS.publisher, Literal("Bob")))

    alice_graph.add((alice, FOAF.mbox, URIRef("mailto:alice@work.example.org")))
    alice_graph.add((alice, FOAF.name, Literal("Alice")))

    bob_graph.add((bob, FOAF.name, Literal("Bob")))
    bob_graph.add((bob, FOAF.mbox, URIRef("mailto:bob@oldcorp.example.org")))
    bob_graph.add((bob, FOAF.knows, alice))

    yield d, alice_graph, bob_graph

context_formats = [
    "json-ld",
    "trix",
    "nquads",
    "trig",
    "hext",
]

@pytest.fixture
def xfail_selected_context_parse_data_formats(request):
    fmt = request.getfixturevalue("fmt")
    expected_failures = [
        # "json-ld",
        # "trix",
        # "nquads",
        # "trig",
        # "hext",
    ]
    if fmt in expected_failures:
        request.node.add_marker(
            pytest.mark.xfail(reason=f"Expected failure with {fmt}")
        )

@pytest.mark.parametrize("fmt", context_formats)
@pytest.mark.usefixtures("xfail_selected_context_parse_data_formats")
def test_parse_example2_from_data(fmt, example2):
    # Use programmatically-created graphs as standard to be checked against
    d1, alice_graph, bob_graph = example2

    d2 = Dataset()
    d2.bind("ex", Namespace("http://example.org/"))
    with open(example2_root + fmt, "r") as fp:
        d2.parse(data=fp.read(), format=fmt)

    assert len(list(d2.contexts())) == 2
    assert len(list(d2.graphs())) == 2
    assert len(d2) == 2
    assert alice_uri in d2.contexts()
    assert bob_uri in d2.contexts()
    assert len(list(d1.quads((None, None, None, None)))) == len(
        list(d2.quads((None, None, None, None)))
    )
    assert len(d1.store) == len(d2.store)
    assert isomorphic(d1, d2)
```

rdflib/util.py
aucampia: Changed this as per your suggestion.

aucampia:
Tests are moved now, and I added a bunch more tests. I also applied @nicholascar 's suggestion.

aucampia (approved):
Looks good, thanks @gjhiggins

There are still some corner cases that won't be handled so I'm still somewhat hesitant to invest much more into this function, I think the next step is to rewrite it to using the actual parser or deprecate it, but I think it is a marked improvement and I there are no regressions as far as I can tell.

aucampia:
Will merge later tonight if there is no more feedback.

Issue #737, opened 2017-04-22
ill-formed literals in SPARQL


pfps:
It appears that ill-formed literals are not treated correctly

SELECT ?result WHERE {  BIND ( COALESCE("999"^^xsd:byte>0,"OK") AS ?result )}

should return "OK" but instead returns true

joernhees:
hmm, i guess you're referring to 999 not fitting into `xsd:byte` (so a signed 8 bit int)...

- jena seems to agree with you:
  ```
  $ sparql -v --data=empty.n3 --query=<(echo 'SELECT ?result WHERE { BIND ( COALESCE("999"^^<http://www.w3.org/2001/XMLSchema#byte> >0,"OK") AS ?result )}')
  18:40:13 WARN  NodeValue            :: Datatype format exception: "999"^^xsd:byte
    1 SELECT  ?result
    2 WHERE
    3   { BIND(coalesce(( "999"^^<http://www.w3.org/2001/XMLSchema#byte> > 0 ), "OK") AS ?result) }
  
  18:40:14 INFO  exec                 :: QUERY
    SELECT  ?result
    WHERE
      { BIND(coalesce(( "999"^^<http://www.w3.org/2001/XMLSchema#byte> > 0 ), "OK") AS ?result) }
  18:40:14 INFO  exec                 :: ALGEBRA
    (project (?result)
      (extend ((?result (coalesce (> "999"^^<http://www.w3.org/2001/XMLSchema#byte> 0) "OK")))
        (table unit)))
  ----------
  | result |
  ==========
  | "OK"   |
  ----------
  ```

- stardog seems to agree with us :
  `true`

- virtuoso 😂:
  result:
  ```
  Virtuoso 37000 Error SP031: SPARQL compiler: The list of return values contains '*' but the pattern does not contain variables

  SPARQL query:
  define sql:big-data-const 0 
  #output-format:text/html
  define sql:signal-void-variables 1 define input:default-graph-uri <http://dbpedia.org> SELECT ?result WHERE { BIND ( COALESCE("999"^^<http://www.w3.org/2001/XMLSchema#byte> >0,"OK") AS ?result )}
  ```

  `SELECT ( COALESCE("999"^^<http://www.w3.org/2001/XMLSchema#byte> >0,"OK") AS ?result ) {}`
  result:
  `0`

  `SELECT ( "999"^^<http://www.w3.org/2001/XMLSchema#byte> > 0 AS ?result ) {}`
  result:
  `0`

[...]

Issue #1326, opened 2021-05-21
Literal should always have a datatype


white-gecko:
According to RDF 1.1 (https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal)

> A literal in an RDF graph consists of two or three elements:
> 
> - a lexical form, being a Unicode [UNICODE] string, which SHOULD be in Normal Form C [NFC],
> - a datatype IRI, being an IRI identifying a datatype that determines how the lexical form maps to a literal value, and
> - if and only if the datatype IRI is http://www.w3.org/1999/02/22-rdf-syntax-ns#langString, a non-empty language tag as defined by [BCP47]. The language tag MUST be well-formed according to section 2.2.9 of [BCP47].
> 
> A literal is a language-tagged string if the third element is present. Lexical representations of language tags MAY be converted to lower case. The value space of language tags is always in lower case.
> 
> Please note that concrete syntaxes MAY support simple literals consisting of only a lexical form without any datatype IRI or language tag. Simple literals are syntactic sugar for abstract syntax literals with the datatype IRI http://www.w3.org/2001/XMLSchema#string. Similarly, most concrete syntaxes represent language-tagged strings without the datatype IRI because it always equals http://www.w3.org/1999/02/22-rdf-syntax-ns#langString.

So the datatype of Literals should always be set.

If a language is specified and a datatype it has to be ensured that the datatype is http://www.w3.org/1999/02/22-rdf-syntax-ns#langString

See also #670

kouralex:
I presume always setting the datatype will fix the issue I just encountered with duplicate (implicit and explicit) literal entries gathered from multiple files. Minimalistic example (RDFLib version: 5.0.0):

```
import sys 
from rdflib import Graph
g = Graph().parse(format='ttl', data='<http://a> <http://b> ""^^<http://www.w3.org/2001/XMLSchema#string>,  "" .')
g.serialize(format="ttl", destination=sys.stdout.buffer)
```

Output:
```
@prefix ns1: <http://> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ns1:a ns1:b "",
        ""^^xsd:string .
```
Expected:
```
@prefix ns1: <http://> .

ns1:a ns1:b "" .
```

For now, I just normalized all triples (dropped explicit `^^xsd:string` part with sed) in the data before parsing them with RDFLib.

[...]

PR #1773, opened 2022-03-24
Add ability to detect and mark ill-typed literals

### LINKED ISSUE #848: rdflib.Literals need a well-formed / ill-formed status flag
I encountered this issue while working on [pySHACL](https://github.com/RDFLib/pySHACL).
Specifically, this bug is causing a failure in several of the tests in the standard data-shapes-test-suite here [or-datatypes-001.ttl](https://github.com/w3c/data-shapes/blob/gh-pages/data-shapes-test-suite/tests/core/property/or-datatypes-001.ttl) and [datatype-ill-formed.ttl](https://github.com/w3c/data-shapes/blob/gh-pages/data-shapes-test-suite/tests/core/property/datatype-ill-formed.ttl) 
This test relies on the assertion that literals such as `"none"^^xsd:boolean` and `"300"^^xsd:byte` should be considered by the validator to be ill-formed literals, so that when checking rules such as `sh:datatype`, if validating that this property value is a well-formed Literal of type `xsd:boolean` or `xsd:byte` (respectively) this should Fail.

Currently, `"none"^^xsd:boolean` is parsed to a Literal with `value=False, datatype=xsd:boolean` and `"300"^^xsd:byte` is parsed to a Literal with `value=int(300) and datatype=xsd:byte`, so the validation checks which should fail actually pass.

An ideal solution would be at Literal-creation time _before_ converting the lexical value to a Python value, check if it is ill-formed first, store that as an `ill_formed` flag on the Literal itself, then do the conversion as normal.

### LINKED ISSUE #1757: Polar-integer datatypes not verifying polarity of lexical value
I recently encountered an issue using pySHACL to validate a `xsd:positiveInteger` and found it was accepting `"0"^^xsd:positiveInteger`.

From review of today's `term.py` and the influence of the members of `_NUMERIC_LITERAL_TYPES`, and from a test-patch (linking in a few moments, I need this Issue number first), this is currently an acceptable rdflib statement:

```python
value = Literal("-1", datatype=XSD.positiveInteger)
```

What should this behavior be instead?  A runtime error (something like `raise ValueError`) would be correct to me, but I appreciate it'd potentially be a jarring change for many applications.

ashleysommer:
Add ability to mark ill-typed literals where the lexical value does not match the datatype given if known

Parameterized test is included to test feature

Fixes #1757
Fixes #848

Related to, but doesn't quite relate to #737


gjhiggins:
Not wishing to add to workload but ... would this be a good opportunity to also address #1326 “Literal should always have a datatype”?

RDFLib‘s current stance on Literal datatypes (according to the error message text) dates from the long-superseded [W3C Recommendation 10 February 2004](http://www.w3.org/TR/rdf-concepts/#section-Graph-Literal)

```py
if lang is not None and datatype is not None:
    raise TypeError(
        "A Literal can only have one of lang or datatype, "
    )
```
In #1326, @white-gecko summarises [the entailments of the current spec](https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal) as: 
1. The datatype of a Literal should always be set
2. If a language tag is specified, the datatype must be `http://www.w3.org/1999/02/22-rdf-syntax-ns#langString`

AIUI, both of the above could be flagged via the “ill-formed” property.

rdflib/term.py
```diff
@@ -662,6 +670,10 @@ def normalize(self) -> "Literal":
         else:
             return self
 
+    @property
+    def ill_formed(self) -> bool:
+        return self._ill_formed
+
```
aucampia, line 676: It may be worth considering to make this a three valued property:

```suggestion
    @property
    def ill_formed(self) -> Optional[bool]:
        return self._ill_formed
```

And only return True/False if it can be established conclusively, the current approach will yield `ill_formed` for values which may actually be well formed but just for a datatype that is not registered.

rdflib/term.py
```diff
@@ -606,7 +608,12 @@ def __new__(
             # passed a string
             # try parsing lexical form of datatyped literal
             value = _castLexicalToPython(lexical_or_value, datatype)
-
+            if datatype:
+                dt_uri: URIRef = URIRef(datatype)
+                if dt_uri in _check_well_formed_types:
+                    checker = _check_well_formed_types[dt_uri]
+                    well_formed = checker(lexical_or_value, value)
+                    ill_formed = ill_formed or (not well_formed)
```
aucampia, line 616: are there cases where this value will be different from:

```suggestion
                    ill_formed = not well_formed
```

rdflib/term.py
```diff
@@ -1484,6 +1501,94 @@ def _parseBoolean(value: str) -> bool:
...
+    """
+    xsd:unsignedInt has a 32bit value of between 0 and 4294967295
+    """
+    return len(lexical) > 0 and isinstance(value, int) and (0 <= value <= 4294967295)
+
+
+def _well_formed_short(lexical: Union[str, bytes], value: Any) -> bool:
+    """
+    The value space of xs:short is the set of common short integers (16 bits),
+    i.e., the integers between -2147483648 and 2147483647,
```
aucampia, line 1547:
```suggestion
    i.e., the integers between -32768 and 32767,
```

rdflib/term.py
aucampia: Also, should this not rather be "ill_typed" instead of ill_formed?

https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal

> Otherwise, the literal is ill-typed and no literal value can be associated with the literal. Such a case produces a semantic inconsistency but is not syntactically ill-formed. Implementations must accept ill-typed literals and produce RDF graphs from them. Implementations may produce warnings when encountering ill-typed literals.

aucampia (approved):
Patch looks good, just the concern that technically `Literal.ill_formed == False` implies it is either not lexically well-typed/formed, or that we just don't have a checker registered for this.

Also may be better to call it ill_typed rather than ill_formed, though I'm not entirely sure about this.

rdflib/term.py
ashleysommer: No, for now there are no instances where it will be different. But this is a coding pattern I've picked up from other projects, where other code may be added in the future that will set the boolean before this point. Probably not needed for now though.

ashleysommer:
When I started this patch, my assumption was that if we cannot prove the Literal is ill-formed, then it must be well-formed. Thus ill_formed is False until proven to the true by the checks. But thinking about it since then, you're right, maybe it should be a three valued property, with `None` being the default. If there are built-in tests, then it can be marked as true/false. otherwise it is None, which is unknown.

Regarding naming.. I'm using "ill-formed" because the term "well-formed" is used in the PySHACL spec document, in the context of describing a "well-formed literal". A Literal that is not "well-formed" is one where the the lexical does not match the datatype. Therefore, I concluded a Literal that is not "well-formed" could be called "ill-formed". But maybe "ill-typed" is a better description.

aucampia:
> Regarding naming.. I'm using "ill-formed" because the term "well-formed" is used in the PySHACL spec document, in the context of describing a "well-formed literal". A Literal that is not "well-formed" is one where the the lexical does not match the datatype. Therefore, I concluded a Literal that is not "well-formed" could be called "ill-formed". But maybe "ill-typed" is a better description.
> 
> EDIT: I just saw that indeed the language in the [W3C SHACL test suite](https://github.com/w3c/data-shapes/blob/gh-pages/data-shapes-test-suite/tests/core/property/datatype-ill-formed.ttl) refers to these literals as "ill-formed".

Okay, happy with name then.



> When I started this patch, my assumption was that if we cannot prove the Literal is ill-formed, then it must be well-formed. Thus ill_formed is False until proven to the true by the checks. But thinking about it since then, you're right, maybe it should be a three valued property, with `None` being the default. If there are built-in tests, then it can be marked as true/false. otherwise it is None, which is unknown.

This is actually even a bit more hairy now that I think about it, we have the following scenarios as I interpret [RDF 1.1 Concepts and Abstract Syntax / 3.3 Literals](https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal)

1. The Literal's datatype is not a [recognized datatype IRIs](https://www.w3.org/TR/rdf11-concepts/#dfn-recognized-datatype-iris): In this case the concept is not well defined, as it is only defined for recognized datatype IRIs

    In this case I would say the reasonable value for `Literal.ill_formed` is `None`.

2.  The Literal's datatype is a [recognized datatype IRIs](https://www.w3.org/TR/rdf11-concepts/#dfn-recognized-datatype-iris): In this case the value should ideally only ever be true or false, and I think in this case it is maybe reasonable to conclude that it is NOT `ill_typed` unless we can show it is `ill_typed`, however maybe there are cases here which we are not yet sure for which `Literal.ill_formed` could also be None.

I made a PR to change `ill_typed` to be three valued and follow this logic: https://github.com/ashleysommer/rdflib/pull/7

Happy to make any further changes to it if you want, just let me know.

aucampia:
@ashleysommer let me know if you need help on this, happy to work to get this in.

aucampia:
I will rebase this next week and merge https://github.com/ashleysommer/rdflib/pull/7 so we can merge this.

aucampia:
Merged https://github.com/ashleysommer/rdflib/pull/7 and merged master. I also did another review. To me it looks like a fairly good addition, one concern is that this will make parsing slower, maybe not that much, but somewhat. We should maybe consider setting up something to measure performance, not that familiar with the various solutions for it.

Just a heads up to @rchateauneu as you are very focused on performance, probably good to get some input from you here.

aucampia:
@ajnelson-nist as this addresses or relates to an issue you raised (https://github.com/RDFLib/rdflib/issues/1757) please check to see if this makes sense, at least on the interface side, as we can fix the implementation easier than the interface.

aucampia:
Planning to merge this on or after 2022-04-24.

rdflib/term.py
```diff
@@ -1485,6 +1513,94 @@ def _parseBoolean(value: str) -> bool:
     return False
 
 
+def _well_formed_by_value(lexical: Union[str, bytes], value: Any) -> bool:
+    return value is not None
```
ajnelson-nist, line 1517: This looked like an error on first glance, like it was intended to review `self._value` somehow.  After seeing how it's called, I see this is intended to be a fallback method.  Can it get a short documentation string summarizing this behavior?

test/test_literal/test_literal.py
```diff
@@ -80,6 +83,60 @@ def test_cant_pass_invalid_lang(
...
+            ("a", XSD.double, True),
+            ("0", XSD.double, False),
+            ("0.1", XSD.double, False),
+            ("0.1", XSD.decimal, False),
+            ("0.g", XSD.decimal, True),
+            ("b", XSD.integer, True),
+            ("2147483647", XSD.int, False),
+            ("2147483648", XSD.int, True),
+            ("2147483648", XSD.integer, False),
+            ("valid ASCII", XSD.string, False),
```
ajnelson-nist, line 113: I promise this is only 50% a digression:

Is there any temptation to try non-ASCII strings within this test series?  Or does the waiver embedded in the `pytest` option [`disable_test_id_escaping_and_forfeit_all_rights_to_community_support`](https://docs.pytest.org/en/6.2.x/parametrize.html) kinda nix that plan?

In more seriousness, non-ASCII data handling has taken a non-trivial amount of attention from several communities I've worked with, and it's always nice to see handling guidance encoded.

test/test_literal/test_literal.py
```diff
@@ -80,6 +83,60 @@ def test_cant_pass_invalid_lang(
...
+            ("Not a valid time", XSD.time, True),
+            ("Not a valid date", XSD.date, True),
+            ("7264666c6962", XSD.hexBinary, False),
+
+            # RDF.langString is not a recognized datatype IRI as we assing no literal value to it, though this should likely change.
```
ajnelson-nist, line 118: Whoops, typo, looks like "assign" was meant.

test/test_literal/test_literal.py
aucampia: I'm happy to include non ascii-strings, we can easily set the test ID manually if we don't like the inferred test ID [[ref](https://docs.pytest.org/en/6.2.x/reference.html#pytest-param)] if that is what you mean. If you have any ideas please do add them.

aucampia: I will google for some phrases from different languages and add them.

ajnelson-nist: I've encountered a few different pain points:
* Purported support for unicode that could not withstand the appearance of the unicode snowman: ☃.
* Different languages being transcoded to UTF-8, but with an incorrect guess at the original encoding (IIRC, Latin-accented characters guessed to be ISO-8859-1 but being something else).
* UTF-16BE being dumped as ASCII without transcoding, so every other byte is `\x00`.

ajnelson-nist: Also, truly, I defer to you on the best way to handle `parameterize`.  I'm a bit too new to that method to have any ideas about best practices.

aucampia: I will include the snowman, and some non latin script phrases, probably we should have more extensive testing for this but it is easy to include here.

rdflib/term.py
```diff
@@ -663,6 +672,20 @@ def normalize(self) -> "Literal":
         else:
             return self
 
+    @property
+    def ill_formed(self) -> Optional[bool]:
+        """
+        For `recognized datatype IRIs
+        <https://www.w3.org/TR/rdf11-concepts/#dfn-recognized-datatype-iris>`_,
+        this value will be `True` if the literal is ill formed, otherwise it
+        will be `False`. `Literal.value` (i.e. the `literal value <https://www.w3.org/TR/rdf11-concepts/#dfn-literal-value>`_) should always be defined if this value is `True`, but should not be considered reliable if this property is not `True`.
```
ajnelson-nist, line 681: Can we double-check the wording at the very end of this line?  I would consider the `.value` not-reliable if this property IS `True`.

aucampia: I think you are right, I will double check and fix it.

ajnelson-nist:
Summarizing with my latecomer's perspective to this thread (I didn't get a notice about this issue until @aucampia pinged me):

I appreciate @gjhiggins 's last remark on #1757 , with external data sources being potentially significant problems.  I'm personally a fan of programs that give a `--strict` flag, but it's unacceptable for data someone else hands me to halt my program if it includes a bad restricted-integer.

So, I agree with this implementation of a `.ill_formed` property, deferring enforcement decisions but enabling a surveyable property.

The next question, which we could agree to leave out of scope of this PR, is how to review the `Literal`s for where this property is `True`.  Should `rdflib.Graph` provide a convenience method/property for that?

aucampia:
> Should `rdflib.Graph` provide a convenience method/property for that?

I'm open to having a flag on Graph to make it strict, in which case we would check all literals for `ill_typed is not True`, or even possibly `ill_typed is False`, a relatively easy interim measure is possibly to just extend the store you are using to only accept literals if they have `ill_typed is not True`

gjhiggins:
> > Should `rdflib.Graph` provide a convenience method/property for that?
> 
> I'm open to having a flag on Graph to make it strict, in which case we would check all literals for `ill_typed is not True`, or even possibly `ill_typed is False`, a relatively easy interim measure is possibly to just extend the store you are using to only accept literals if they have `ill_typed is not True`

I would argue for a state-reporting method rather than a state-asserting property because (unless it's a property of the store as @aucampia suggests), it could get quite messy in the context of operators, e.g. `flaggedgraph += unflaggedgraph`.

rdflib/term.py
aucampia: Yes the documentation's explanation was inverted, should be correct now.

test/test_literal/test_literal.py
aucampia: Fixed.

test/test_literal/test_literal.py
aucampia: Added a string with arabic script and a snowman, but this does not do much to test this handling.

rdflib/term.py
aucampia: Added a docstring.

ajnelson-nist: Thanks!

aucampia:
@ajnelson-nist thank you very much for the review, I addressed all the issues you raised I think. Will merge this tomorrow morning (CET) if there is no further feedback.

ajnelson-nist:
> > > Should `rdflib.Graph` provide a convenience method/property for that?
> > 
> > 
> > I'm open to having a flag on Graph to make it strict, in which case we would check all literals for `ill_typed is not True`, or even possibly `ill_typed is False`, a relatively easy interim measure is possibly to just extend the store you are using to only accept literals if they have `ill_typed is not True`
> 
> I would argue for a state-reporting method rather than a state-asserting property because (unless it's a property of the store as @aucampia suggests), it could get quite messy in the context of operators, e.g. `flaggedgraph += unflaggedgraph`.

I'm getting the feeling the matter of a first broad-consumption interface should be raised as a separate PR.  I'm partial to a strategy that lets me most quickly find where the problem is.  From that use case, I'd be happiest with the `Graph` keeping a dictionary of all problematic literals, with the dictionary's value being a set of all triples that have the literal.  That implementation helps me fix data problems, but introduces a cache design pattern that has to reconcile with graph-graph operations (at least adding and subtracting), and might also have performance implications.

I don't think this discussion should hold up making the ill-typedness property available.

aucampia:
> I'm getting the feeling the matter of a first broad-consumption interface should be raised as a separate PR.

> I don't think this discussion should hold up making the ill-typedness property available.

I agree with this, was mainly a suggestion, PR scope should be clamped to the `ill_typed` property. We can make a separate issue for further matters.

PR #1780, opened 2022-03-29
Update to black 22.3.0 because of issue with click

aucampia:
See this https://github.com/psf/black/pull/2966

Also simplified pre-commit config of black based on a suggestion from
https://github.com/jack1142 in https://github.com/psf/black/issues/2493#issuecomment-1081987650


ashleysommer (approved):
Thanks!

PR #1786, opened 2022-03-31
Fix issues with NamespaceManager.bind

aucampia:
replace and override handling is now correct as far as I can tell.

Also added some tests.


aucampia:
For fixing issues in https://github.com/RDFLib/rdflib/pull/1686

nicholascar (approved):

PR #1788, opened 2022-04-02
Test reorg (continued)

gjhiggins:
Further test re-org, all but a couple of changes (noted below) restricted to moving tests into sub-folders and relocating w.r.t import of `test.testutils` / `test.helpers` and test fixture data.

Of note:
- Removed unused `store_performance_test.py`
- Removed unused `test_rules.py` and `a.n3` (test imports pychinko, unavailable for a decade or more!)

Migration of some more tests to pytest to follow.


test/test_rules.py
```diff
@@ -1,77 +0,0 @@
-import unittest
```
aucampia, line 1: where was this file moved to?

test/store_performance.py
```diff
@@ -1,103 +0,0 @@
-import unittest
```
aucampia, line 1: Don't see where this moved to.

test/test_rules.py
gjhiggins: > Of note:
> 
>     Removed unused test_rules.py and a.n3 (test imports pychinko, unavailable for a decade or more!)

test/store_performance.py
gjhiggins: > Of note:
> 
>     Removed unused store_performance_test.py

Oh, I mistyped the test file name, sry

test/test_rules.py
aucampia: Would it not be better to just fix it so it becomes a test instead?

gjhiggins: A swift foray into RDFLib's dev history reveals that [`test/rules.py`](https://github.com/RDFLib/rdflib/blob/6ed2ef48ed38679bcdafe7cae250a2ef4b315e7b/test/test_rules.py) was introduced somewhere between 2.2.3 (2005) and 2.4.0 (2007). I'll hazard a guess that it's _never_ been executed because it imports [pychinko](https://web.archive.org/web/20111127205750/http://www.mindswap.org/~katz/pychinko/)
```py
try:
    from pychinko import terms
    from pychinko.interpreter import Interpreter
    ...
except ImportError as e:
    print("Could not test Pychinko: %s" % e)
```

Although pychinko was integrated into [swap](https://github.com/linkeddata/swap/tree/master/pychinko) back in 2007 or so, I see two issues acting to prevent a refurb: i) I doubt we'd want to introduce a dependency on the 15-year-old `swap` package and ii) in order to function, pychinko [relies on a local installation of the Java-implemented Pellet reasoner](https://github.com/linkeddata/swap/blob/7a71cfd27e48aa79b1c111053d2ca58c2051a782/pychinko/builtins.py#L298).

(_If we want to test RDF rule-processing, refurbishing FuXi is possibly the more feasible option - as noted on the Pychinko web page, at one time FuXi used Pychinko but Chimezie subsequently factored out that dependency._)

aucampia (approved):
Thanks @gjhiggins - looks good to me - should have taken a closer look at removed files, they look fine now that I took a closer look.

nicholascar (approved):

PR #1795, opened 2022-04-09
Convert `test/test_util.py` to `pytest`

aucampia:
Fairly straight forward conversion from unittest to pytest.

This also fixes some issues that were present in the old tests:

- `assertTrue` was used instead of `assertEqual` inside:
  - `test_util_from_n3_expectliteralwithdatatypefromint`
  - `test_util_from_n3_expectliteralwithdatatypefrombool`
  - `test_util_from_n3_expectliteralmultiline`
  - `test_util_from_n3_expectliteralwithescapedquote`
- `test_util_from_n3_expectliteralwithescapedquote` was not comparing
  against the right literal, but this was missed because it was using
  `assertTrue` instead of `assertEqual`
- `test_util_from_n3_not_escapes_xf` had `("\\I")` as a value but this
  was not failing, so it has been moved to `test_util_from_n3_not_escapes`.

# Checklist

  the same change.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 
-import unittest
 import time
 from unittest.case import expectedFailure
+
+import pytest
 from rdflib.graph import Graph
 from rdflib.graph import QuotedGraph
 from rdflib.graph import ConjunctiveGraph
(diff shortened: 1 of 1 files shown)


aucampia:
I kept this very conservative, there are much more that can be done to improve all tests, but I also want to make it easy to review, so this is just doing a 1:1 conversion except for places where there are actual problems not just stylistic concerns.

gjhiggins (approved):

PR #1800, opened 2022-04-09
Fix issue with namespace rebinding

aucampia:
Also add additional tests for namespace rebinding.

## Note

Commits with @gjhiggins as author are taken from:
- https://github.com/RDFLib/rdflib/pull/1686#discussion_r846407575
- https://github.com/RDFLib/rdflib/pull/1686#discussion_r846413277
- https://gist.github.com/gjhiggins/b4b59bffc18aa080ce466b01eb0e5d70
- https://github.com/RDFLib/rdflib/pull/1800#issuecomment-1094106019

### DIFF
--- a/rdflib/plugins/stores/berkeleydb.py
+++ b/rdflib/plugins/stores/berkeleydb.py
@@ -470,10 +470,17 @@ class BerkeleyDB(Store):
         prefix = prefix.encode("utf-8")
         namespace = namespace.encode("utf-8")
         bound_prefix = self.__prefix.get(namespace)
-        if override and bound_prefix:
-            self.__namespace.delete(bound_prefix)
-        self.__prefix[namespace] = prefix
-        self.__namespace[prefix] = namespace
+        bound_namespace = self.__namespace.get(prefix)
+        if override:
+            if bound_prefix:
+                self.__namespace.delete(bound_prefix)
+            if bound_namespace:
+                self.__prefix.delete(bound_namespace)
+            self.__prefix[namespace] = prefix
+            self.__namespace[prefix] = namespace
+        else:
+            self.__prefix[bound_namespace or namespace] = bound_prefix or prefix
+            self.__namespace[bound_prefix or prefix] = bound_namespace or namespace
 
     def namespace(self, prefix):
         prefix = prefix.encode("utf-8")
--- a/rdflib/plugins/stores/memory.py
+++ b/rdflib/plugins/stores/memory.py
@@ -403,11 +403,19 @@ class Memory(Store):
                             yield triple, self.__contexts(triple)
 
     def bind(self, prefix, namespace, override=True):
-        bound_prefix = self.__prefix.get(namespace)
-        if override and bound_prefix:
-            del self.__namespace[bound_prefix]
-        self.__prefix[namespace] = prefix
-        self.__namespace[prefix] = namespace
+        bound_namespace = self.__namespace.get(prefix)
+        bound_prefix = self.__prefix.get(namespace) or self.__prefix.get(bound_namespace)
+        if override:
+            if bound_prefix:
+                del self.__namespace[bound_prefix]
+            if bound_namespace:
+                del self.__prefix[bound_namespace]
+
+            self.__prefix[namespace] = prefix
+            self.__namespace[prefix] = namespace
+        else:
+            self.__prefix[bound_namespace or namespace] = bound_prefix or prefix
+            self.__namespace[bound_prefix or prefix] = bound_namespace or namespace
 
     def namespace(self, prefix):
         return self.__namespace.get(prefix, None)
--- a/rdflib/plugins/stores/memory.py
+++ b/rdflib/plugins/stores/memory.py
@@ -404,7 +404,9 @@ class Memory(Store):
 
     def bind(self, prefix, namespace, override=True):
         bound_namespace = self.__namespace.get(prefix)
-        bound_prefix = self.__prefix.get(namespace) or self.__prefix.get(bound_namespace)
+        bound_prefix = self.__prefix.get(namespace) or self.__prefix.get(
+            bound_namespace
+        )
         if override:
             if bound_prefix:
                 del self.__namespace[bound_prefix]
--- a/test/test_namespace_rebinding.py
+++ b/test/test_namespace_rebinding.py
@@ -1,0 +1,488 @@
+import pytest
+
+from rdflib.term import URIRef
+from rdflib import ConjunctiveGraph, Graph, Literal
+from rdflib.plugins.stores.memory import Memory
+from rdflib.namespace import Namespace
+from rdflib.namespace import OWL
+from test.data import tarek, context1, context2
+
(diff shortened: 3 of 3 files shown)


aucampia:
pre-commit.ci autofix

aucampia:
@gjhiggins Good spotting on this - I think we need to fix it for other stores also though, I will add an additional test to check this against other stores and then add fixes for that.

gjhiggins (approved):
looks good, thank you

gjhiggins:
> @gjhiggins Good spotting on this - I think we need to fix it for other stores also though, I will add an additional test to check this against other stores and then add fixes for that.

Maybe save you some time, here's a patch for `berkeleydb.py`

```diff
diff --git a/rdflib/plugins/stores/berkeleydb.py b/rdflib/plugins/stores/berkeleydb.py
index 9ca4380b..b5f6e561 100644
--- a/rdflib/plugins/stores/berkeleydb.py
+++ b/rdflib/plugins/stores/berkeleydb.py
@@ -470,10 +470,17 @@ class BerkeleyDB(Store):
         prefix = prefix.encode("utf-8")
         namespace = namespace.encode("utf-8")
         bound_prefix = self.__prefix.get(namespace)
-        if override and bound_prefix:
-            self.__namespace.delete(bound_prefix)
-        self.__prefix[namespace] = prefix
-        self.__namespace[prefix] = namespace
+        bound_namespace = self.__namespace.get(prefix)
+        if override:
+            if bound_prefix:
+                self.__namespace.delete(bound_prefix)
+            if bound_namespace:
+                del self.__prefix[bound_namespace]
+            self.__prefix[namespace] = prefix
+            self.__namespace[prefix] = namespace
+        else:
+            self.__prefix.put(bound_namespace or namespace, bound_prefix or prefix)
+            self.__namespace.put(bound_prefix or prefix, bound_namespace or namespace)
 
     def namespace(self, prefix):
         prefix = prefix.encode("utf-8")
```

aucampia:
@nicholascar @gjhiggins I'm going to throw caution to the wind with this and fix some things which I would have liked to keep to separate PRs in this PR.

aucampia:
There are 4 stores doing about the same thing (`rdflib.plugins.store.{memory.SimpleMemory,memory.Memory,memory.BerkeleyDB,}`), I think the best option is to move the code for `Store.{bind,prefix,namespace,namespaces}` to a separate class.

aucampia:
Will look at this further tomorrow, won't work that well with memory store, but there is also many cases where strings are checked for turthyness instead of `is not None`, will try fix them also.

aucampia:
Actually, scratch the last comments, I will fix further issues and expand tests once this is in master. But beware, there are still at least one store with this issue and then this should be checking is None instead of turthyness, and it should also use a proper null coalescing method, and I'm not entirely sure of all the consequences, but I will figure it out once this and https://github.com/RDFLib/rdflib/pull/1686 is merged.

rdflib/plugins/stores/berkeleydb.py
```diff
@@ -470,10 +470,17 @@ def bind(self, prefix, namespace, override=True):
         prefix = prefix.encode("utf-8")
         namespace = namespace.encode("utf-8")
         bound_prefix = self.__prefix.get(namespace)
-        if override and bound_prefix:
-            self.__namespace.delete(bound_prefix)
-        self.__prefix[namespace] = prefix
-        self.__namespace[prefix] = namespace
+        bound_namespace = self.__namespace.get(prefix)
+        if override:
+            if bound_prefix:
```
aucampia, line 475:
```suggestion
            if bound_prefix is not None:
```

rdflib/plugins/stores/berkeleydb.py
```diff
@@ -470,10 +470,17 @@ def bind(self, prefix, namespace, override=True):
...
+                self.__namespace.delete(bound_prefix)
+            if bound_namespace:
```
aucampia, line 477:
```suggestion
            if bound_namespace is not None:
```

rdflib/plugins/stores/berkeleydb.py
```diff
@@ -470,10 +470,17 @@ def bind(self, prefix, namespace, override=True):
...
+                self.__prefix.delete(bound_namespace)
+            self.__prefix[namespace] = prefix
+            self.__namespace[prefix] = namespace
+        else:
+            self.__prefix[bound_namespace or namespace] = bound_prefix or prefix
```
aucampia, line 482: This will not do the right thing if bound_namespace is `""`

rdflib/plugins/stores/berkeleydb.py
```diff
@@ -470,10 +470,17 @@ def bind(self, prefix, namespace, override=True):
...
+            self.__namespace[bound_prefix or prefix] = bound_namespace or namespace
```
aucampia, line 483: This will not do the right thing if bound_prefix is `""`

rdflib/plugins/stores/memory.py
```diff
@@ -403,11 +403,21 @@ def triples(self, triple_pattern, context=None):
...
-        if override and bound_prefix:
-            del self.__namespace[bound_prefix]
-        self.__prefix[namespace] = prefix
-        self.__namespace[prefix] = namespace
+        bound_namespace = self.__namespace.get(prefix)
+        bound_prefix = self.__prefix.get(namespace) or self.__prefix.get(
+            bound_namespace
+        )
+        if override:
+            if bound_prefix:
```
aucampia, line 411:
```suggestion
            if bound_prefix is not None:
```

rdflib/plugins/stores/memory.py
```diff
@@ -403,11 +403,21 @@ def triples(self, triple_pattern, context=None):
...
+                del self.__namespace[bound_prefix]
+            if bound_namespace:
```
aucampia, line 413:
```suggestion
            if bound_namespace is not None:
```

rdflib/plugins/stores/memory.py
```diff
@@ -403,11 +403,21 @@ def triples(self, triple_pattern, context=None):
...
+                del self.__prefix[bound_namespace]
+
+            self.__prefix[namespace] = prefix
+            self.__namespace[prefix] = namespace
+        else:
+            self.__prefix[bound_namespace or namespace] = bound_prefix or prefix
```
aucampia, line 419: This will not do the right thing if `bound_namespace` is `""`

rdflib/plugins/stores/memory.py
```diff
@@ -403,11 +403,21 @@ def triples(self, triple_pattern, context=None):
...
+            self.__namespace[bound_prefix or prefix] = bound_namespace or namespace
```
aucampia, line 420: This will not do the right thing if `bound_prefix` is `""`

aucampia: Should use null coalesing, something like this:


```python
def _coalesce(*args: Optional[_AnyT]) -> Optional[_AnyT]:
    for arg in args:
        if arg is not None:
            return arg
    return None
```

aucampia:
@gjhiggins I will redo this on master.

PR #1806, opened 2022-04-12
Migrate more tests to pytest

gjhiggins:
# Summary of changes
Continued migration of unittests to pytest. I've changed my approach to follow @aucampia's lead, acknowledging that retaining the UnitTest class structure makes changes far more readily apparent. We're not done with test re-org yet (blacking in prospect) so the UnitTest structure can be removed at a later stage.

# Checklist
-  [x ] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),

### DIFF
--- a/test/test_literal/test_duration.py
+++ b/test/test_literal/test_duration.py
@@ -10,37 +10,30 @@ from rdflib.term import Literal
 class TestDuration(unittest.TestCase):
     def test_to_python_timedelta(self):
         l = Literal("P4DT5H6M7S", datatype=XSD.dayTimeDuration)
-        self.assertTrue(isinstance(l.toPython(), timedelta))
-        self.assertEqual(l.toPython(), parse_duration("P4DT5H6M7S"))
+        assert isinstance(l.toPython(), timedelta)
+        assert l.toPython() == parse_duration("P4DT5H6M7S")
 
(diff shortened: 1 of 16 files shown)


test/test_literal/test_duration.py
```diff
@@ -10,37 +10,30 @@
 class TestDuration(unittest.TestCase):
```
aucampia, line 10: Not that critical, but you can safely remove this as pytest picks up all classes that start with `Test`
```suggestion
class TestDuration:
```

aucampia: Can of course be done in subsequent PRs, and you may already know, just in case. Think PR is fine as is, already switching to asserts make it much more readable and then makes subsequent PRs smaller and simpler.

aucampia (approved):
Thanks @gjhiggins  - PR looks great, and as easy to review as such a thing can be. Will merge tomorrow if there is no further feedback.

aucampia (approved):

aucampia (approved):
Checked latest changes to remove unittest, looks good, also double checked test coverage and count and all looks fine.

PR #1810, opened 2022-04-14
test: Add tests for SPARQL parsing and serialization

aucampia:
# Summary of changes

This patch adds some tests for SPARQL serialization and deserialization
with checks based on round tripping where possible.

Some of these tests are expected failures in that they seem to pass fo
some parsers or serializers, but not others. If there is agreement on
what parameters should be supported we can fix the parsers and
serializers to actually support them.

There are also xfails for the handling of string paths on windows, as
this seems to fail for all serializers.

Other changes:
 - style: apply black on `test/test_graph/test_graph.py` and `test/test_n3.py`. There are no other changes to these files other than applying black, and it is done by pre-commit.ci bot.

# Checklist

  the same change.

### DIFF
--- a/test/test_graph/test_graph.py
+++ b/test/test_graph/test_graph.py
@@ -32,12 +32,14 @@ def test_property_store() -> None:
     graph = Graph()
     assert isinstance(graph.store, Store)
 
+
 def test_property_identifier_default() -> None:
     """
     The default identifier for a graph is a `rdflib.term.BNode`.
     """
     graph = Graph()
     assert isinstance(graph.identifier, BNode)
+
 
 def test_property_identifier() -> None:
     """
@@ -46,6 +48,7 @@ def test_property_identifier() -> None:
     id = URIRef("example:a")
     graph = Graph(identifier=id)
     assert id == graph.identifier
+
 
 def test_property_namespace_manager() -> None:
     """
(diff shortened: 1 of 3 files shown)


aucampia:
The most controversial thing here, and what is most in need of review, is the expected failures:

https://github.com/RDFLib/rdflib/blob/9b5ac57fa12a2cb1cf39ad1fae75bdfa75785873/test/test_sparql/test_result.py#L186-L201

Some parsers support the encodings, and some parsers support TEXT_IO, and some support BINARY_IO. In my view all parsers should support all of them, and all of those tests should pass, so the expected failures are correct in that they denote known issues, and we can fix the issues without breaking interfaces. I'm happy to extend support once we are in agreement there.

aucampia:
Okay and it seems path handling completely fails on windows, adding xfails for that also, and will fix those also in a future patch.

gjhiggins (approved):
Solid stuff, useful to have.

aucampia:
pre-commit.ci autofix

aucampia:
rebased now onto master, two files in diff are result of pre-commit.ci applying black.

aucampia:
Merged as this is only adding tests, not changing any tests. The xfail marked tests also seem valid as they pass for some parsers and some operating systems. Happy to adjust them in future PRs though.

PR #1814, opened 2022-04-14
changes to core library code

gjhiggins:
# Summary of changes
Complete replacement of ConjunctiveGraph by Dataset , imposition of identifier-as-context throughout.

# Checklist
- [x] Checked that all tests and type checking passes. (all tests passing prior to `test_variants.py`)
  - [x] Updated relevant documentation to avoid inaccuracies. (in progess)
  - [x] Considered adding additional documentation. (in progress)
  - [x] Considered adding an example in `./examples` for new features. (in progress)
  - [x] Considered updating our changelog (`CHANGELOG.md`). (TBD)

### DIFF
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -57,7 +57,6 @@ __all__ = [
     "Namespace",
     "Dataset",
     "Graph",
-    "ConjunctiveGraph",
     "BRICK",
     "CSVW",
     "DC",
@@ -159,7 +158,7 @@ Literal work, eq, __neq__, __lt__, etc.
 
 from rdflib.term import URIRef, BNode, IdentifiedNode, Literal, Variable
 
-from rdflib.graph import Dataset, Graph, ConjunctiveGraph
+from rdflib.graph import Dataset, Graph
 
 from rdflib import plugin
 from rdflib import query
--- a/rdflib/compare.py
+++ b/rdflib/compare.py
@@ -92,7 +92,7 @@ __all__ = [
     "similar",
 ]
 
-from rdflib.graph import Graph, ConjunctiveGraph, ReadOnlyGraphAggregate
+from rdflib.graph import Graph, Dataset, ReadOnlyGraphAggregate
 from rdflib.term import BNode, Node, URIRef, IdentifiedNode
 from hashlib import sha256
 
@@ -159,7 +159,16 @@ class _call_count(object):
         return wrapped_f
 
 
-class IsomorphicGraph(ConjunctiveGraph):
+def _yield_triples(graph):
+    if isinstance(graph, Dataset):
+        for s, p, o, c in graph:
+            yield s, p, o
+    else:
+        for s, p, o in graph:
+            yield s, p, o
+
+
+class IsomorphicGraph(Graph):
     """An implementation of the RGDA1 graph digest algorithm.
 
     An implementation of RGDA1 (publication below),
@@ -314,7 +323,7 @@ class _TripleCanonicalizer(object):
         bnodes: Set[BNode] = set()
         others = set()
         self._neighbors = defaultdict(set)
-        for s, p, o in self.graph:
+        for s, p, o in _yield_triples(self.graph):
             nodes = set([s, p, o])
             b = set([x for x in nodes if isinstance(x, BNode)])
             if len(b) > 0:
@@ -518,7 +527,7 @@ class _TripleCanonicalizer(object):
             stats["canonicalize_triples_runtime"] = _total_seconds(
                 datetime.now() - start_coloring
             )
-        for triple in self.graph:
+        for triple in _yield_triples(self.graph):
             result = tuple(self._canonicalize_bnodes(triple, bnode_labels))
             yield result
 
@@ -540,7 +549,8 @@ def to_isomorphic(graph):
     result = IsomorphicGraph()
     if hasattr(graph, "identifier"):
         result = IsomorphicGraph(identifier=graph.identifier)
-    result += graph
+    for t in _yield_triples(graph):
+        result.add(t)
     return result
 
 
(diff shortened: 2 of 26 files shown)


gjhiggins:
# Comments on changes:

WORK-IN-PROGRESS: Re-working of "identifier-as-context" PR and removing ConjunctiveGraph, re-positioned as a contribution to release 7.0 as the removal of ConjunctiveGraph comprehensively breaks the public API.

I've been working on this more complete implementation since January, hence the large number of changes. Actually there's relatively little impact on core library files, what's different from the last rebased PR is a few changed lines here and there in the parsers and serializers to get a coherent implementation. The bulk of the changes are additional tests and test fixture data.

1. ConjunctiveGraph completely replaced by Dataset which now inherits directly from Graph.
2. Identifier-as-context referencing imposed throughout.
3. Store changed to use identifier-as-context, API remains the same (for now).
4. Parsers and serializers changed to use Dataset instead of ConjunctiveGraph..
5. Tests, tests and more tests.

Isomorphism is an unresolved issue, as are operators. The W3 declined to publish a formal semantics for Datasets and the pertinent issue was resolved as "won't fix"

>     https://www.w3.org/2011/rdf-wg/track/issues/17
> 
>     RESOLVED: close issue-17 -- there is no general purpose way to merge datasets;
>     it can only be done with external knowledge.
> 
>     David Wood, 29 Oct 2012, 13:42:58

I've retained my attempted implementation of operators so that reviewers can more easily get an idea of the representational issues without having to code them up but atm I'm inclined to believe that RDFLib should _not_ support operators for Datasets because it seems semantically vacuous to me and that the implementation of Dataset operators can be removed.

Isomorphic functionality is currently sematically suspect, having been casually extended to use Datasets. There are only a couple of _feasible_ changes that can be made without wrecking the implementation of the RGDA1 graph digest algorithm and I'm currently thrashing through the issues attendant on QuotedGraphs, formulae and the consequent implications for Dataset conicalization and isomorphism. One issue presenting difficulty is that `IsomorphicGraph` was a subclass of ConjunctiveGraph but can now only be a subclass of Dataset and atm I'm not at all sure that's a trivial change.

aucampia:
@gjhiggins not sure exactly what your plan is to merge this, but I would be very happy to slice a bunch of things off from this into seperate PRs - for example most of the changes to change tests from ConjunctiveGraph to DataSet would possibly make sense independently of this PR.

test/test_serializers/test_prettyxml.py
```diff
@@ -43,19 +43,19 @@ def _assert_equal_graphs(g1, g2):
...
-    for s, p, o in g:
+    for s, p, o, c in g:
         if isbnode(s):
             s = _blank
         if isbnode(p):
             p = _blank
         if isbnode(o):
             o = _blank
-        gcopy.add((s, p, o))
+        gcopy.add((s, p, o, DATASET_DEFAULT_GRAPH_ID if c is None else c))
```
aucampia, line 58: I know this is still in draft, just thinking, would it not be better to associate None with the default Graph ID? That would make the interface a bit easier to use I think.

aucampia:
@gjhiggins happy to do rebases on this, just let me know either here or on gitter and I will rebase with priority. Also happy to fix any other CI problems with this.

gjhiggins:
## Uneasy about the migration to Dataset

The more I think about it, the more uncomfortable I am about this move as it necessitates imposing some arbitrary decisions about the (undefined) semantics of named graphs. Just last night @aucampia and I had to resolve a mismatch in expectations about the syntactic-vs-semantic content of named graphs in the context of checking the accuracy of round-tripping.

For a quad-aware store (i.e. all of the current RDFLib stores) the `Graph.parse()` method arbitrarily changes the content of the stored quads depending on whether the `publicID` parameter is set and/or whether there is a detectable document location: 

> "`publicID`: the logical URI to use as the document base. If None specified the document location is used (at least in the case where there is a document location)".

In practice, it can result in test failures, e.g. the mismatch between a triple parsed in the context of a document base:
```py
(
    rdflib.term.URIRef('http://example.org/subject'),
    rdflib.term.URIRef('http://example.org/predicate'),
    rdflib.term.URIRef('http://example.org/object'),
    rdflib.term.URIRef('example:variants/simple_triple')
)
```
and one created as a result of parsing the same triple but without a document location:

```py
(
    rdflib.term.URIRef('http://example.org/subject'),
    rdflib.term.URIRef('http://example.org/predicate'),
    rdflib.term.URIRef('http://example.org/object'),
    None
)
```
Given two Datasets with quad content as above, resolving the question of whether the two Datasets are _isomorphic_ is arbitrary - either the names of graphs are treated as i) semantic tokens, in which case the datasets are not considered to be isomorphic or ii) merely syntactic tokens in which case the datasets can be considered to be isomorphic.

Currently, graph identifiers are ignored by `isomorphic` and the fact that named graphs can have BNode identifiers is worrying in this context, to say the least.  I do appreciate that the implementation of `ConjunctiveGraph`'s set-theoretic operators blithely ignores these considerations and this (arguably casual) approach to the semantics seems not to present users with any (reported) issues - I'm left wondering whether this eminently pragmatic approach to "working with RDF" isn't actually offering users some critically-useful advantage.

The W3 declined to publish a formal semantics for Datasets and whilst there was discussion of the implications, the issue of merging datasets was closed as "can't fix" (my interpretation):

>     https://www.w3.org/2011/rdf-wg/track/issues/17
> 
>     RESOLVED: close issue-17 -- there is no general purpose way to merge datasets;
>     it can only be done with external knowledge.
> 
>     David Wood, 29 Oct 2012, 13:42:58

If one can't provide semantically-sound support for UNION, then that must extend to all of the other set-theoretic operators. My concern here is that if Datasets can't be merged without external knowledge then it would seem that computing isomorphism must be similarly constrained - or am I missing something?

PR #1818, opened 2022-04-14
docs: fix sphinx warnings

aucampia:
# Summary of changes
    
This patch fixes all sphinx warnings and also changes sphinx building to
treat warnings as errors so that we don't introduce new warnings.

Also:
- ci: fail on sphinx warnings.
- ci: don't use GitHub actions to build docs as this is being done by
  readthedocs.org.
- build: supress "more than one target found for cross-reference"
  warnings from sphinx autodoc.

# Checklist

  the same change.
  
  **NOTE**: This is fixing a warning which is also fixed in #1686 - though it should merge fine hopefully as the fixed source should be the same.


gjhiggins (approved):
Thanks for cleaning this up, makes working on the docs easier.

nicholascar (approved):

PR #1845, opened 2022-04-18
fix: Raise ValueError for unsupported `bind_namespace` values

aucampia:
This patch changes `NamespaceManager.__init__` to raise a `ValueError`
when it receives an unsupported value for `bind_namespaces`.

Other changes:
- Added a Literal type for valid `bind_namespaces` values, with this
  type checkers will raise an error if an invalid string is supplied as
  an argument.
- Rename some variables/properties to have a underscore in front to make
  it clear they are not part of the RDFLib public API.


gjhiggins (approved):
Nice to see this tidied up.

nicholascar (approved):

PR #1846, opened 2022-04-18
text: fix pytest config

aucampia:
# Summary of changes
    
pytest was using config from `tox.ini` preferentially and ignoring config
from `setup.cfg`, as a side-effect doctests were not running on
code/docstrings in `rdflib/`.

The reason why some pytest config was in `tox.ini` instead of `setup.cfg`
was because of these issues:
- https://github.com/pypa/pip/issues/5182
- https://github.com/pytest-dev/pytest/issues/3062

As a compromise to fix this I have opted for moving all pytest config to
`pyproject.toml`:
- https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml

This seems sensible as `pyproject.toml` is standarized by PEPs and
eventually most things from `setup.cfg` will end up in there anyway.

Also:
- remove the pytest ignore on `test/translate_algebra` as tests in there
  have been running and passing for some time.
- fixed path to test data in `rdflib/plugins/parsers/nquads.py`.

# Checklist

  the same change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,3 +28,22 @@ exclude = '''
   )/
 )
 '''
+
+[tool.pytest.ini_options]
+addopts = [
+   "--doctest-modules",
+   "--ignore=admin",
+   "--ignore=rdflib/extras/external_graph_libs.py",
+   "--ignore-glob=docs/*.py",
+]
+doctest_optionflags = "ALLOW_UNICODE"
+filterwarnings = [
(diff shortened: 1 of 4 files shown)


gjhiggins (approved):
> The reason why some pytest config was in tox.ini instead of setup.cfg
was because of these issues:

Ahh, so _that's_ why. Definitely worth moving to `pyproject.toml`

aucampia:
Thanks for the review @gjhiggins

PR #1849, opened 2022-04-18
Remove testing and debug code from rdflib

aucampia:
# Summary of changes

This patch removes code from `rdflib/` that does not seem like it belongs
in `rdflib/`, most of it is related to doctest, some of it belongs in
`test/` and was moved to `test/test_misc/test_collection.py`, and yet
more of it seems to just be there for debugging purposes, though it
would possibly be better to put that in a separate place if it is needed
again or to debug using tests if possible.

Other changes:
- Removed an invocation of `rdflib.util.test` from `test_util.py`. This
  seems like an attempt to invoke doctest however pytest takes care of
  that so this is not needed.

# Checklist

  the same change.


aucampia:
Probably best to only merge this after https://github.com/RDFLib/rdflib/pull/1846 - but it is not entirely related and can be merged independently also.

gjhiggins (approved):
Agreed, this approach is redundant/unused legacy which is ripe for cleanup.

nicholascar (approved):

PR #1850, opened 2022-04-18
[DO NOT REVIEW] Add a bunch of typing

aucampia:
This is a branch I'm maintaing with a bunch of typing, I'm planning to
integrate it with master in parts, anyone is welcome to have a look but
most of what is happening here is subject to change.

Creating this pull request mainly go get docs built and verifying everything is working on all platforms.

### DIFF
--- a/rdflib/plugins/stores/concurrent.py
+++ b/rdflib/plugins/stores/concurrent.py
@@ -20,6 +20,7 @@ class ResponsibleGenerator:
         return next(self.gen)
 
 
+# NOTE: This does not seem to be tested and probably should inherit from Store.
 class ConcurrentStore:
     def __init__(self, store):
         self.store = store
--- a/examples/resource_example.py
+++ b/examples/resource_example.py
@@ -25,7 +25,8 @@ if __name__ == "__main__":
     bill.add(RDF.type, FOAF.Agent)
     bill.set(RDFS.label, Literal("Bill"))
 
-    bill.add(FOAF.knows, bob)
+    # type error: Argument 2 to "add" of "Resource" has incompatible type "Resource"; expected "Node"  [arg-type]
+    bill.add(FOAF.knows, bob)  # type: ignore[arg-type]
 
     # Resources returned when querying are 'auto-boxed' as resources:
     print(f"Bill knows: {bill.value(FOAF.knows).value(FOAF.name)}")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -255,6 +255,11 @@ no_implicit_optional = false
 implicit_reexport = false
 
 
+[[tool.mypy.overrides]]
+module = "rdflib.*"
+check_untyped_defs = true
+
+
 [tool.coverage.run]
 branch = true
 source = ["rdflib"]
--- a/rdflib/_typing.py
+++ b/rdflib/_typing.py
@@ -1,0 +1,23 @@
+# import sys
+# from typing import TYPE_CHECKING, Optional, Tuple, TypeVar
+
+# if sys.version_info >= (3, 10):
+#     from typing import TypeAlias
+# else:
+#     from typing_extensions import TypeAlias
+
+# if TYPE_CHECKING:
+#     from rdflib.graph import Graph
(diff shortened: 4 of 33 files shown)


rdflib/plugins/stores/concurrent.py
```diff
@@ -20,6 +20,7 @@ def __next__(self):
         return next(self.gen)
 
 
+# NOTE: This does not seem to be tested and probably should inherit from Store.
```
gjhiggins, line 23: It could - and probably should - be removed completely - it is neither context- nor formula-aware so can only be of extremely limited use.

rdflib/store.py
```diff
@@ -169,10 +203,10 @@ def node_pickler(self):
         return self.__node_pickler
 
     # Database management methods
-    def create(self, configuration):
+    def create(self, configuration: Optional[Any]) -> None:
         self.dispatcher.dispatch(StoreCreatedEvent(configuration=configuration))
 
-    def open(self, configuration, create: bool = False):
+    def open(self, configuration: Any, create: bool = False) -> Optional[int]:
```
gjhiggins, line 209: I think the `configuration` param can be safely particularised as `str`, given the document text "Opens the store specified by the ***configuration string***"

rdflib/store.py
```diff
@@ -184,20 +218,20 @@ def open(self, configuration, create: bool = False):
...
-    def close(self, commit_pending_transaction=False):
+    def close(self, commit_pending_transaction: bool = False) -> None:
         """
         This closes the database connection. The commit_pending_transaction
         parameter specifies whether to commit all pending transactions before
         closing (if the store is transactional).
         """
 
-    def destroy(self, configuration):
+    def destroy(self, configuration: Any) -> None:
```
gjhiggins, line 228: `configuration` param can safely be paricularised as `str` the documentation states "configuration string"

rdflib/store.py
aucampia: There is at least one store that takes a tuple and uses it to do something that cannot be done via a string:

https://github.com/RDFLib/rdflib/blob/3e2987bfed12bc7c76ad373097739029e3fc56e5/rdflib/plugins/stores/sparqlstore.py#L554-L568

https://github.com/RDFLib/rdflib/blob/3e2987bfed12bc7c76ad373097739029e3fc56e5/rdflib/plugins/stores/sparqlstore.py#L591-L609

aucampia: But fair enough, I think given docstrings say string I will make it a string.

aucampia: changed to string.

rdflib/store.py
aucampia: changed to string.

mr-c:
This is great to see @aucampia ; we've got partial type stubs for the bits of rdflib we use in `cwltool` at https://github.com/common-workflow-language/cwltool/tree/main/mypy-stubs/rdflib

Would you like me to send a PR to this PR with the added bits, or would you prefer to review them yourself? Already [I'm copying some of your new type hints to our stubs](https://github.com/common-workflow-language/cwltool/pull/1702), so thanks!

aucampia:
> This is great to see @aucampia ; we've got partial type stubs for the bits of rdflib we use in `cwltool` at https://github.com/common-workflow-language/cwltool/tree/main/mypy-stubs/rdflib
> 
> Would you like me to send a PR to this PR with the added bits, or would you prefer to review them yourself? Already [I'm copying some of your new type hints to our stubs](https://github.com/common-workflow-language/cwltool/pull/1702), so thanks!

I'm somewhat slow walking the type hints to make it easier on reviewers. I break things off from here into small chunks and then merge it and use a script to strip non runtime parts from python to make it easy to verify there are no runtime changes, and then merge it in parts.

Most likely a good part of Graph will be typed by the time of the next release which is going to be somewhere in October.

I will have a look at your stubs if there are some things that are ambiguous though, and we are always open to PRs, though preferably as small as possible, and for type hints preferably only changing type hints and not trying to do more than that.

Fixing all the bugs the type hints are highlighting will come next but that should also be accompanied with tests in most cases.

Issue #1701, opened 2022-01-31
Strange bug of notation3 serialization that occur with “probability”


blackwint3r:
Version:
  rdflib 6.1.1
  python 3.8.10

I have a `test.py`:
```
import rdflib
data1='''
@prefix : <http://www.example.com/dev#>.
:a :b :c.
'''
data2='''
@prefix : <http://www.example.com/dev#>.
{:a :b :c}=>{:d :e :f}.
'''
g = rdflib.Graph()
g.parse(data=data1, format = 'n3')
g.parse(data=data2, format = 'n3')
r = g.serialize(format='n3')
print(r)
```
The output of this `test.py` has 2 possibilities.
The first one (which is correct):

```
@prefix : <http://www.example.com/dev#> .

:a :b :c .

{
    :a :b :c .

} => {
        :d :e :f .

    } .
```
The second one (missing the triple parsed at the first time):
```
@prefix : <http://www.example.com/dev#> .

{
    :a :b :c .

} => {
        :d :e :f .

    } .
```

This strange bug appears in a completely random way. In order to count the "probability" of its occurrence, I wrote a loop like this：

```
num = 10000
error =0
for i in range(num):
    g = rdflib.Graph()
    g.parse(data=data1, format = 'n3')
    g.parse(data=data2, format = 'n3')
    r = g.serialize(format='n3')
    #print(g.serialize(format='n3'))
    index = (r.find(":a :b :c"))
    #print(index)
    if index > 43:  #second case occur
        error +=1
print("error =" ,error)
```

Each execution gives a different result, like error= 8533, 5693, 437...  As the `num` increase, the error rate do not converge to a specific value. Can't find any patterns.

aucampia:
Seems to be the same issue as https://github.com/RDFLib/rdflib/issues/1807

aucampia:
This seems to happen every time the first element in `subjects_list` here is the quoted graph:

https://github.com/RDFLib/rdflib/blob/e9f5c22c3fb24ecc7f073af3e65483599f8d6c17/rdflib/plugins/serializers/turtle.py#L239

I'm guessing this may be because because the serializer thinks it already serialized the triple as it has been serialized inside the quoted graph. Checking further.

aucampia:
Indeed it is, debugging inside https://github.com/RDFLib/rdflib/compare/master...aucampia:iwana-20220419T2343-n3_serialize_quoted_graph

```console
[...]

PR #1853, opened 2022-04-18
More type hints for `rdflib.graph` and related

aucampia:
# Summary of changes

This patch primarily adds more type hints for `rdflib.graph`, but also
adds type hints to some related modules in order to work with the new
type hints for `rdflib.graph`.

I'm mainly doing this as a baseline for adding type hints to
`rdflib.store`.

I have created type aliases to make it easier to type everything
cosnsitently and to make type hints easier easier to change in the
future. The type aliases are private however (i.e. `_`-prefixed) and
should be kept as such for now.

This patch only contains typing changes and does not change runtime
behaviour.

Broken off from https://github.com/RDFLib/rdflib/pull/1850

# Checklist

  the same change.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -486,7 +486,7 @@ class Graph(Node):
     def triples(
         self,
         triple: Tuple[Optional["_SubjectType"], Path, Optional["_ObjectType"]],
-    ) -> Generator[Tuple["_SubjectType", "_PredicateType", "_ObjectType"], None, None]:
+    ) -> Generator[Tuple["_SubjectType", Path, "_ObjectType"], None, None]:
         ...
 
     @overload
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -16,6 +16,7 @@ from typing import (
     TextIO,
     Tuple,
     Type,
+    TypeVar,
     Union,
     cast,
     overload,
@@ -36,7 +37,27 @@ from rdflib.paths import Path
 from rdflib.resource import Resource
 from rdflib.serializer import Serializer
 from rdflib.store import Store
-from rdflib.term import BNode, Genid, IdentifiedNode, Literal, Node, RDFLibGenid, URIRef
+from rdflib.term import (
+    BNode,
+    Genid,
+    IdentifiedNode,
+    Identifier,
+    Literal,
+    Node,
+    RDFLibGenid,
+    URIRef,
+)
+
+_SubjectType = IdentifiedNode
+_PredicateType = IdentifiedNode
+_ObjectType = Identifier
+
+_TripleType = Tuple["_SubjectType", "_PredicateType", "_ObjectType"]
+_QuadType = Tuple["_SubjectType", "_PredicateType", "_ObjectType", "Graph"]
+_TriplePatternType = Tuple[
+    Optional["_SubjectType"], Optional["_PredicateType"], Optional["_ObjectType"]
+]
+_GraphT = TypeVar("_GraphT", bound="Graph")
 
 assert Literal  # avoid warning
 assert Namespace  # avoid warning
@@ -356,15 +377,15 @@ class Graph(Node):
         self.default_union = False
 
     @property
-    def store(self):
+    def store(self) -> Store:
         return self.__store
 
     @property
-    def identifier(self):
+    def identifier(self) -> Node:
         return self.__identifier
 
     @property
-    def namespace_manager(self):
+    def namespace_manager(self) -> NamespaceManager:
         """
         this graph's namespace-manager
         """
@@ -373,7 +394,7 @@ class Graph(Node):
         return self.__namespace_manager
 
     @namespace_manager.setter
-    def namespace_manager(self, nm):
+    def namespace_manager(self, nm: NamespaceManager):
         self.__namespace_manager = nm
 
     def __repr__(self):
@@ -424,7 +445,7 @@ class Graph(Node):
         """
         return self.__store.close(commit_pending_transaction=commit_pending_transaction)
 
-    def add(self, triple: Tuple[Node, Node, Node]):
+    def add(self, triple: "_TripleType"):
         """Add a triple with self as context"""
         s, p, o = triple
         assert isinstance(s, Node), "Subject %s must be an rdflib term" % (s,)
@@ -433,7 +454,7 @@ class Graph(Node):
         self.__store.add((s, p, o), self, quoted=False)
         return self
 
-    def addN(self, quads: Iterable[Tuple[Node, Node, Node, Any]]):  # noqa: N802
+    def addN(self, quads: Iterable["_QuadType"]):  # noqa: N802
         """Add a sequence of triple with context"""
 
         self.__store.addN(
@@ -457,34 +478,40 @@ class Graph(Node):
     @overload
     def triples(
         self,
-        triple: Tuple[
-            Optional[IdentifiedNode], Optional[IdentifiedNode], Optional[Node]
-        ],
-    ) -> Iterable[Tuple[IdentifiedNode, IdentifiedNode, Node]]:
+        triple: "_TriplePatternType",
+    ) -> Generator["_TripleType", None, None]:
         ...
 
     @overload
     def triples(
         self,
-        triple: Tuple[Optional[IdentifiedNode], Path, Optional[Node]],
-    ) -> Iterable[Tuple[IdentifiedNode, Path, Node]]:
+        triple: Tuple[Optional["_SubjectType"], Path, Optional["_ObjectType"]],
+    ) -> Generator[Tuple["_SubjectType", "_PredicateType", "_ObjectType"], None, None]:
         ...
 
     @overload
     def triples(
         self,
         triple: Tuple[
-            Optional[IdentifiedNode], Union[None, Path, IdentifiedNode], Optional[Node]
+            Optional["_SubjectType"],
+            Union[None, Path, "_PredicateType"],
+            Optional["_ObjectType"],
         ],
-    ) -> Iterable[Tuple[IdentifiedNode, Union[IdentifiedNode, Path], Node]]:
+    ) -> Generator[
+        Tuple["_SubjectType", Union["_PredicateType", Path], "_ObjectType"], None, None
+    ]:
         ...
 
     def triples(
         self,
         triple: Tuple[
-            Optional[IdentifiedNode], Union[None, Path, IdentifiedNode], Optional[Node]
+            Optional["_SubjectType"],
+            Union[None, Path, "_PredicateType"],
+            Optional["_ObjectType"],
         ],
-    ) -> Iterable[Tuple[IdentifiedNode, Union[IdentifiedNode, Path], Node]]:
+    ) -> Generator[
+        Tuple["_SubjectType", Union["_PredicateType", Path], "_ObjectType"], None, None
+    ]:
         """Generator over the triple store
 
         Returns triples that match the given triple pattern. If triple pattern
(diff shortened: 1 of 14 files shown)


aucampia:
@ajnelson-nist please have a look if you can spare some time and feel comfortable doing a review.

ajnelson-nist:
It seems a shame the `_TriplePatternType` type isn't used in the one spot I most expected it to be used: The definition of `Graph.triples`.  At least it did get used in the first `@overload` signature for `Graph.triples`.

I'm also surprised so many of the type references need to be quoted strings, but I'll trust this was an effect of some need of the type-checking system.

Overall, this looks like good work in the right direction.  I'm happier with the distinct type names for the subject, predicate, and object positions.

aucampia:
> It seems a shame the `_TriplePatternType` type isn't used in the one spot I most expected it to be used: The definition of `Graph.triples`. At least it did get used in the first `@overload` signature for `Graph.triples`.

I double checked now, and the second overload was actually wrong and I fixed it in https://github.com/RDFLib/rdflib/pull/1853/commits/faf95a9489623e4cc3063643254c96d6b8d59a7c, overloads are now as follow:

https://github.com/RDFLib/rdflib/blob/faf95a9489623e4cc3063643254c96d6b8d59a7c/rdflib/graph.py#L473-L498

The third overload is identical to the implementation, and a catch all for cases where it receives an undifferentiated predicate which is needed for the implementations of `Graph.{subjects,predicates,objects,...}` which passes predicate of type `Union[None, Path, "_PredicateType"]`. 


> I'm also surprised so many of the type references need to be quoted strings, but I'll trust this was an effect of some need of the type-checking system.

They don't really, but I had to move it in and out of `if TYPE_CHECKING:` while trying to figure out how to get sphinx to still correctly document types with aliases (see https://github.com/RDFLib/rdflib/pull/1854) guards a couple of times, and I think I will likely have to do it in places in future again also so I figured it is best to just leave it in quotes as it is safer anyway. We can (should?) maybe look at switching to `from __future__ import annotations` in which case it is not needed, but to me it is about the same, this does not really affect users of RDFLib though.

> Overall, this looks like good work in the right direction. I'm happier with the distinct type names for the subject, predicate, and object positions.

In part I went for this because I could not really decide what the best restrictions are for them. Currently it is:

https://github.com/RDFLib/rdflib/blob/71a14465e39bf0e7c793bd4a6438ab44f3c57827/rdflib/graph.py#L49-L58

However, `_PredicateType` should technically be `URIRef` - as per the [spec](https://www.w3.org/TR/rdf11-concepts/#section-triples) - but I think N3 actually does allow blank node predicates, and then `_ObjectType` should technically be `Union[Literal, URIRef, BNode]` - because variables should not be used in graphs I think - but this makes things needlessly fiddly also. Using type aliases makes it easy to change without having to go adjust each place.

Thanks for having a look.

aucampia:
@eggplants any feedback from you would also be appreciated.

gjhiggins:
N3 goes well beyond the RDF spec:

```py
def test_triple_types():
    from rdflib import _typing
    from rdflib.graph import QuotedGraph
    from rdflib.paths import SequencePath
    from rdflib.term import Variable

    # In Notation3, predicates can be one of a variety of Path subclasses

    g = Graph()

    g = g.parse(data='''
@prefix : <ex:> .

:a :p1 :c ; :p2 :f .
:c :p2 :e ; :p3 :g .
:g :p3 :h ; :p2 :j .
:h :p3 :a ; :p2 :g .

:q :px :q .

''', format='n3')

    e = Namespace('ex:')

    triple = (e.a, e.p1/e.p2, e.e)

    assert triple in g

        assert type(triple) == _typing._TripleType

    assert type(triple[1]) == SequencePath


    # In Notation3 subjects and objects can be QuotedGraphs

    g = Graph()
    g.parse(data="""
@prefix log: <http://www.w3.org/2000/10/swap/log#> .
:bob :hasmother :alice .
:john :hasmother :alice .

@forAll :x, :y, :z .
{ :x :hasmother :z . :y :hasmother :z } log:implies { :x :issibling :y } .
""", format="n3")

    triple = list(g.triples(
        (None, URIRef("http://www.w3.org/2000/10/swap/log#implies"), None)))[0]

        assert type(triple) == _typing._TripleType

    assert type(triple[0]) == QuotedGraph


    # In QuotedGraphs, subjects and objects can be Variable

    triple = list(sorted([t for t in triple[0]]))[0]

        assert type(triple) == _typing._TripleType

    assert type(triple[0]) == Variable
```

aucampia:
@gjhiggins good point, I was actually just noticing this when working on https://github.com/RDFLib/rdflib/issues/1701

I think we have to expand the types as follow to accommodate N3:

```python
 _SubjectType = Union[Identifier, "QuotedGraph"]
 _PredicateType = Union[Identifier, "QuotedGraph"] 
 _ObjectType = Union[Identifier, "QuotedGraph"] 
```

aucampia:
A bit of a pity that there is no abstract syntax spec for N3's graph model.

aucampia:
> ```python
> QuotedGraphs
> ```

To clarify why:

AFAIU, variables can be parts of graphs in N3, and Literals and BNodes can be any of subject, predicate or object. Possibly we could create a helper [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) that exposes an interface which only allows RDF 1.1 compliant triples but as Graph can contain much more we have to keep the typing broad enough.

aucampia:
Marking this as draft, will be still be good to get some more input here but I will add some more tests for quotedgraph first and ensure that the typing actually accommodates this.

eggplants:
LTGM. I can't wait to get type annotations on all of the rdflib!
By the way, `from __future__ import annotations` is a good way to reduce the amount of description by eliminating the cumbersome imports and quotations related to types. Only recently have I begun to adopt it.

tips: I recommend `pyupgrade` to help you rewrite it.

gjhiggins:
> A bit of a pity that there is no abstract syntax spec for N3's graph model.

There sort of is. There's an EBNF expression in the more extensive [description of Notation3](https://w3c.github.io/N3/spec/#grammar) which references a [downloadable ebnf file](https://w3c.github.io/N3/grammar/n3.ebnf).

IIRC you mentioned the possibility of using the Lark parser, fwiw I took a pass by pymantic's use of the Lark parser (in which they [feed lark with a string containing the EBNF expression](https://github.com/norcalrdf/pymantic/blob/43125e320076ca86074a41c0f02bce943a760c80/pymantic/parsers/lark/ntriples.py#L34)) in the context of the EBNF descriptions provided in the [RDF Star spec](https://www.w3.org/2021/12/rdf-star.html).

rdflib/graph.py
```diff
@@ -33,7 +35,27 @@
...
+    RDFLibGenid,
+    URIRef,
+)
+
+_SubjectType = IdentifiedNode
+_PredicateType = IdentifiedNode
+_ObjectType = Identifier
+
+_TripleType = Tuple["_SubjectType", "_PredicateType", "_ObjectType"]
+_QuadType = Tuple["_SubjectType", "_PredicateType", "_ObjectType", "Graph"]
```
nicholascar, line 54: This may have to be updated if @gjhiggins PR's come through!

nicholascar (approved):

aucampia:
I will republish this after https://github.com/RDFLib/rdflib/pull/1858 is merged, likely the types will become:

```python
_SubjectType = Node
_PredicateType = Node
_ObjectType = Node
```

This is mainly because N3 allows a much broader range of subjects, predicates and objects.

aucampia:
https://github.com/RDFLib/rdflib/pull/1858 is merged and I updated this to support N3, the typing is now broader.

The highlight is likely this:

https://github.com/RDFLib/rdflib/blob/5cb10c69d67b3f561735cf3a5034f5599f65c108/rdflib/graph.py#L43-L63

gjhiggins (approved):
lgtm

ajnelson-nist:
> #1858 is merged and I updated this to support N3, the typing is now broader.
> 
> The highlight is likely this:
> 
> https://github.com/RDFLib/rdflib/blob/5cb10c69d67b3f561735cf3a5034f5599f65c108/rdflib/graph.py#L43-L63

I'm a little disappointed (I hope this doesn't come across as being stern! or upset! Just bummed.) with this line:

```python
_SubjectType = Node
```

My original goal with nudging type review was to assist with catching thing-or-string errors, particularly easy to do through various context dictionary and/or typo scenarios in JSON-LD.

Moving `_SubjectType` to `Node` once again permits a `Literal` to be a subject.  I appreciate rdflib's need to provide broad support, and N3 makes this requirement.

Meanwhile, I've thought about this remark earlier, and I'm not sure it will work:

> AFAIU, variables can be parts of graphs in N3, and Literals and BNodes can be any of subject, predicate or object. Possibly we could create a helper [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) that exposes an interface which only allows RDF 1.1 compliant triples but as Graph can contain much more we have to keep the typing broad enough.

I'm not so sure about the `typing.Protocol` strategy because I don't know how hard it would be to provide just alternative signatures for things like the `Graph.triples()` methods....but, I don't have a lot of experience implementing `typing.Protocol`.

However, I am aware that subclassing `Graph` to e.g. `Graph11` (starter name for a `Graph` meeting RDF 1.1 requirements) and tightening the function signatures with e.g. `Graph11.triples(s: IdentifiedNode, p: IdentifiedNode, o: Node)` will fail `mypy` with a violation of the Liskov Substitution Principle.  If `Graph` says the `subject` parameter can be any `Node` in the type signature, all `Graph` subclasses' *type signatures* also have to support any `Node` for the `subject` parameter.

So, there's potentially a significant (to the type system) downstream consequence to supporting `Graph.triples(s: Node, p: Node, o: Node)`.  I'm aware that `assert isinstance()` can implement runtime narrower type assurances in subclasses, but that's not the same as catching type errors with static review before code deployment.

One solution to this (caveat - given not too much thought) might be implementing everything but the iterators in an abstract base class `_Graph`, and requiring users to instantiate a subclass, `GraphN3` or `Graph11`, both of which handle the appropriate iterators' type signatures.  My reflex on writing that sentence is it feels like a significant intrusion of the type system on user experience.  I hope I've overthought this but missed something that makes it a non-issue.

aucampia:
> I'm not so sure about the `typing.Protocol` strategy because I don't know how hard it would be to provide just alternative signatures for things like the `Graph.triples()` methods....but, I don't have a lot of experience implementing `typing.Protocol`.

Thinking about it, I'm not so sure it will work either without an error, but it could still work using `cast(...)` - but I will have to test it.

> However, I am aware that subclassing `Graph` to e.g. `Graph11` (starter name for a `Graph` meeting RDF 1.1 requirements) and tightening the function signatures with e.g. `Graph11.triples(s: IdentifiedNode, p: IdentifiedNode, o: Node)` will fail `mypy` with a violation of the Liskov Substitution Principle.

On this, there are quite a lot of Liskov Substitution Principle violations already, we should not add more, but fixing them will take time, my primary goal with contributions at the moment is to get things into the best state we can have it with the current design and interface, and have as much validation as possible, both static and dynamic, and only then start re-architecting.

Your suggestion of having a more narrow/constained Graph base class makes a lot of sense, and we should try and head towards that so we can fix all the type violations. If we could move towards that while maintaining backwards compatibility it will be ideal.

There may also be some options with generics, something like this:

```
RDFGraph = Graph[IdentifiedNode, URIRef, Identifier]
N3Graph = Graph[Node, Node, Node]
```

Ideally in such cases the Graph should do some strict validation also, but I can't say I have the perfect answer as it is now.

In short term I think the easiest will be a Protocol that one can cast to, but even if I make this I will only keep it in `rdflib._experimental` as a start, because I'm not sure we want to commit ourselves to that route.

For now I have little choice but to go with it as is, but it also does not sit right with me, as this significantly reduces the level of static validation offered to users.

PR #1825, opened 2022-04-15
fix: DefinedNamespace: fixed handling of control attributes.

aucampia:
# Summary of changes

This patch changes `DefinedNamespace` to always raise `AttributeError` for
the name `_NS` and other "control attributes" from `__getattr__`, and to
also not consider them as part of the `__dir__`.

Without doign this `inspect.signature` recurses
infinitely when inspecting `rdflib.namespace.DefinedNamespace` and `dir`
results in an `AttributeError`.

One situation in which this occurs is when sphinx autodoc is generating
documentation from type hints:

```
WARNING: error while formatting signature for rdflib.namespace.DefinedNamespace: Handler <function record_typehints at 0x7fbf2696dd40> for event 'autodoc-process-signature' threw an exception (exception: maximum recursion depth exceeded while calling a Python object)
```

Also:
- Changed `DefinedNamespace.__repr__` to use repr for formatting
  the URI string instead of quoting by hand. This probably has no real effect,
  as the namespace shoudl not have a double or single quote in it, but
  it is still correct to use repr.

The main reason for this patch is to eliminate a warning for sphinx
which is blocking the build since `sphinx.fail_on_warning` was enabled.

There is also two cases in `DefinedNamespaceMeta` where an undefined
method is being called, these were causing type errors before, but I
added notes now to indicate they are real bugs, because if they are
reached they just result in `AttributeError` exceptions. This should be
fixed sometime, not sure what the reasoning behind it was.

# Checklist

- [x] Checked that there aren't other open pull requests for
  the same change.
- [x] Added tests for any changes that have a runtime impact.
- [x] Checked that all tests and type checking passes.
- [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

PR #1854, opened 2022-04-18
docs: switch to sphinx-autodoc-typehints

aucampia:
# Summary of changes

Currently rdflib is relying on sphinx-autodoc's built in handling for type
hints, this however has some deficiencies as it does not handle stuff
inside `if TYPE_CHECKING:` and thus only work in cases where there are no
potential for circular dependencies.

This patch changes sphinx to use sphinx-autodoc-typehints instead, the
documentation generated by this plugin looks slightly different but as a
positive it correctly works with things defined inside `if TYPE_CHECKING:`
guards.

Also:
- moved the `_NamespaceSetString` type alias to `rdflib._type_checking`
  This is so that `sphinx-autodoc-typehints` recognizes it up, as it
  does not handle type aliases that are defined inside
  `if TYPE_CHECKING:` guards in imported files.
- remove sphinx requirements from `requirements.dev.txt` and
  replaced it with `-r docs/sphinx-requirements.txt` to reduce redundancy. 

# Checklist

  the same change.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,6 +30,7 @@ extensions = [
     "sphinxcontrib.apidoc",
     "sphinx.ext.autodoc",
     #'sphinx.ext.autosummary',
+    "sphinx_autodoc_typehints",
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
     "sphinx.ext.todo",
@@ -42,8 +43,12 @@ extensions = [
 
 apidoc_module_dir = "../rdflib"
 apidoc_output_dir = "apidocs"
+
+# https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
 autodoc_default_options = {"special-members": True}
-autodoc_typehints = "both"
+
+# https://github.com/tox-dev/sphinx-autodoc-typehints
+always_document_param_types = True
 
 autosummary_generate = True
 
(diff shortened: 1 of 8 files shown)


aucampia:
Will rebase and publish after merging https://github.com/RDFLib/rdflib/pull/1825

aucampia:
marking as draft so I can fix warnings.

aucampia:
Difference illustrated below:

- master branch: https://rdflib.readthedocs.io/en/latest/apidocs/rdflib.html#rdflib.graph.Graph.add
    
  image


- PR branch: https://rdflib--1854.org.readthedocs.build/en/1854/apidocs/rdflib.html#rdflib.graph.Graph.add
    
  image

Again, the reason why `sphinx-autodoc-typehints` is mainly because it also resolves type aliases, and then shows the underlying types instead of the alias name, so if the type hint was `_TripleType`, and `_TripleType = Tuple[Node, Node, Node]`, then `sphinx-autodoc-typehints` will put `Tuple[Node, Node, Node]` in the docs instead of `_TripleType`. This is primarily so that https://github.com/RDFLib/rdflib/pull/1853 does not result in confusing documentation, once this is merged I will rebase that so that the docs can be reviewed with this in place.

gjhiggins:
PR branch looks very nice and closer to what I conceive of as a "typical" programming model, i.e. what one would reference when documenting or authoring HOW-TOs.

nicholascar (approved):
good plug in use!

PR #1856, opened 2022-04-21
Generate random string suffixes for unique prefixes

gtfierro:
The grammar of Turtle does not allow numbers in the prefixes (https://www.w3.org/TR/turtle/#grammar-production-PN_PREFIX). The current generation of prefixes in the case of prefix collision uses numbers which are incremented until an unused prefix is found. Unfortunately these generated prefixes fail to parse on stricter Turtle parsers (such as oxigraph's rio). The new method should generate random Turtle-compliant prefixes.

Before:

```
# data.ttl
@prefix1: <urn:abcdef/> .
```

After:

```
# data.ttl
@prefixAGHLEISPAW: <urn:abcdef/>
```

Two questions:
1. I am open to any alternative method for generating unique prefixes --- I just chose something that I thought would work.
2. What tests should I write for this, if any? The current Turtle parser is fairly permissive which I don't really feel needs to change, so what is there to verify in the tests other than the fact that a unique prefix is still generated? I can't find any tests that verify this behavior though

# Summary of changes

Changed the algorithm for generating unique prefixes to use random 10-char strings rather than an incrementing integer.

# Checklist

  the same change.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -1,7 +1,7 @@
 import json
-import string
-import random
 import logging
+import random
+import string
 import sys
 import warnings
 from functools import lru_cache
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -1,4 +1,6 @@
 import json
+import string
+import random
 import logging
 import sys
 import warnings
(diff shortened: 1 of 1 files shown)


gtfierro:
I see a problem now with the tests -- I'll see if I can find a nice way to make the ID generation deterministic

gtfierro:
Actually, I believe I've misinterpreted the Turtle spec --- I have other prefixes with numbers in them that parse fine. I will re-open once I determine the actual issue

aucampia:
@gtfierro don't hesitate to ask on gitter either:

- https://gitter.im/RDFLib/rdflib
- https://gitter.im/linkeddata/chat

gtfierro:
Thanks! Appreciate it

PR #1867, opened 2022-04-24
build: add Taskfile with development tasks

aucampia:
# Summary of changes

This patch adds a Taskfile with various tasks that is useful for
RDFLib development. Details of the available tasks can be seen from
`task help` output and some more information is included in the RDFLib
developers guide.

For more information about Taskfiles see https://taskfile.dev/#/usage.

The GitHub action pipeline was also changed to now use tasks from the
Taskfile which simplifies the pipeline quite a bit.

This patch also adds development container which can be used to run
various development task and which can also be used with development
container enabled tools such as VSCode or GitHub codespaces.

This is all being done in preperation for moving coveralls reporting to
the GitHub actions workflow.

Other changes:
- Removed the `Makefile` and associated `test/Dockerfile` as it is now replaced by a `Taskfile.yml`.

# Checklist

  the same change.


edmondchuc (approved):

Dockerfile.devcontainer
```diff
@@ -0,0 +1,72 @@
+FROM mcr.microsoft.com/vscode/devcontainers/base:ubuntu-20.04
```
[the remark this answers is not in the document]
aucampia: I will give it a try, I don't see any problem with in in principle.

.github/workflows/validate.yaml
```diff
@@ -69,38 +64,18 @@ jobs:
...
-          elif [ "${{ matrix.os }}" == "macos-latest" ]
-          then
-              brew install berkeley-db@4
-              export BERKELEYDB_DIR=$(brew --prefix berkeley-db@4)
-          fi
-      - name: Validate
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
white-gecko, line 70: I don't think, we need the github secrets here.

white-gecko (commented):
In general it looks good to me, even though I could not test all parts to be functional.

Only the one secret variable stuck out to me. Other then that I'm fine with it.

.github/workflows/validate.yaml
aucampia: I have changed the repository settings to assign restrictive permissions to the token:

image

And I have now set the token permission explicitly in the workflow:

https://github.com/aucampia/rdflib/blob/dd1b97a2684eca863fec18b614cad6a0cbe66ddb/.github/workflows/validate.yaml#L16-L17

```yaml
    permissions:
      contents: read
```

The reason for having this though is that without it the following error occurs:

https://github.com/RDFLib/rdflib/runs/6149683404?check_suite_focus=true

```
Run arduino/setup-task@v1
  with:
    version: 3.x
  env:
    FORCE_COLOR: 1
    XDG_CACHE_HOME: /Users/runner/work/rdflib/rdflib/cache
    DEFAULT_PYTHON: 3.7
    pythonLocation: /Users/runner/hostedtoolcache/Python/3.10.4/x64
Error: API rate limit exceeded for 199.19.85.[2](https://github.com/RDFLib/rdflib/runs/6149683404?check_suite_focus=true#step:8:2)2[6](https://github.com/RDFLib/rdflib/runs/6149683404?check_suite_focus=true#step:8:6). (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

From the docs:

> repo-token
> (Optional) GitHub access token used for GitHub API requests. Heavy usage of the action can result in workflow run failures caused by rate limiting. GitHub provides a more generous allowance for Authenticated API requests.
> 
> It will be convenient to use [${{ secrets.GITHUB_TOKEN }}](https://docs.github.com/en/actions/reference/authentication-in-a-workflow).

Dockerfile.devcontainer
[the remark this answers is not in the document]
aucampia: I have switched to `mcr.microsoft.com/vscode/devcontainers/base:ubuntu-22.04` which was released about 4 days ago, and it seems to work fine, [deadsnakes](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) is sadly only available on LTS releases, (18.04, 20.04, 22.04). I don't really mind using Debian either, but if Ubuntu 22.04 works I think that is fine, if there is some problem with it I will consider switching to Debian.

As far as I can tell 22.04 does support ARM64:

```console
$ docker buildx imagetools inspect mcr.microsoft.com/vscode/devcontainers/base:ubuntu-22.04
Name:      mcr.microsoft.com/vscode/devcontainers/base:ubuntu-22.04
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:1d2c69ebde66dc6fa86b0d88e48660f6795763fe70d4a6be6fc47d7209a65591
           
Manifests: 
  Name:      mcr.microsoft.com/vscode/devcontainers/base:ubuntu-22.04@sha256:ee1c53071f9c04326860b7e4f33288005fa3e50827ec7aa614cd48a2c93ce010
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64
             
  Name:      mcr.microsoft.com/vscode/devcontainers/base:ubuntu-22.04@sha256:063d4c8574e232ddc022143fc0f4cac07bc962f5200cdb5c398a7b3d5c24bcef
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64
```

aucampia:
Planning to merge before 2022-05-01 if there is no further feedback.

.github/workflows/validate.yaml
white-gecko: Ok, I understand

white-gecko (approved):
Fine for me

Issue #1868, opened 2022-04-25
Add a method to convert qname back to URI


dany-nonstop:
`rdflib.Graph` already has a very helpful method `qname()` which basically converts `http://example.com/ns#object` to `ns:obj` if the namespace prefix `ns` has already been bound by `bind('ns', Namespace('http://example.com/ns'))`. 

I sometimes need use the reverse of `qname()` to make the URI input more human-friendly. It would be great if we have something like `resolve_qname()` which would translate a qname `ns:obj` back to `http://example.com/ns#object`. A neat solution is already provided here by somebody on stackoverflow.

https://stackoverflow.com/questions/63088670/rdflib-python-how-to-get-a-uriref-from-a-string-such-as-nsxxx

gjhiggins:
Resolved in https://github.com/RDFLib/rdflib/pull/1869

PR #1869, opened 2022-04-25
resolve issue1868, add a method to expand qname to URI

gjhiggins:
# Summary of changes
In response to issue #1868:
> “I sometimes need use the reverse of qname() to make the URI input more human-friendly. It would be great if we have something like resolve_qname() which would translate a qname ns:obj back to http://example.com/ns#object”

Seems a reasonable enough request and trivial to resolve... add NamespaceManager method `expand_qname` and tests.

# Checklist
  the same change.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -590,6 +590,26 @@ class NamespaceManager(object):
 
             return self.__cache_strict[uri]
 
+    def expand_qname(self, curie):
+        """
+        Expand a qname (aka CURIE) of the form <prefix:element>, e.g. "rdf:type"
+        into its full expression:
+
+        >>> g = Graph()
+        >>> g.namespace_manager.expand_qname("rdf:type")
+        http://www.w3.org/1999/02/22-rdf-syntax-ns#type
+
+        Returns None if a namespace is not bound to the prefix or the prefix
+        is malformed.
+
+        """
+        if (
+            isinstance(curie, str)
+            and (prefix := curie.split(":")[0])
+            and (ns := self.store.namespace(prefix))
+        ):
+            return str(ns) + curie.split(":")[1]
+
     def bind(
         self,
         prefix: Optional[str],
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -603,12 +603,20 @@ class NamespaceManager(object):
         is malformed.
 
         """
-        if (
-            isinstance(curie, str)
-            and (prefix := curie.split(":")[0])
-            and (ns := self.store.namespace(prefix))
-        ):
-            return str(ns) + curie.split(":")[1]
+        # TODO: When we drop support for 3.7 ...
+        # if (
+        #     isinstance(curie, str)
+        #     and (prefix := curie.split(":")[0])
+        #     and (ns := self.store.namespace(prefix))
+        # ):
+        #     return str(ns) + curie.split(":")[1]
+
+        if isinstance(curie, str) and ":" in curie:
+            ns = self.store.namespace(curie.split(":")[0])
+            if ns is not None:
+                return (
+                    str(self.store.namespace(curie.split(":")[0])) + curie.split(":")[1]
+                )
 
     def bind(
         self,
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -590,29 +590,24 @@ class NamespaceManager(object):
 
             return self.__cache_strict[uri]
 
-    def expand_qname(self, curie: str) -> Union[str, None]:
+    def expand_curie(self, curie: str) -> Union[str, None]:
         """
         Expand a qname (aka CURIE) of the form <prefix:element>, e.g. "rdf:type"
         into its full expression:
 
         >>> import rdflib
         >>> g = rdflib.Graph()
-        >>> g.namespace_manager.expand_qname("rdf:type")
+        >>> g.namespace_manager.expand_curie("rdf:type")
         'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'
 
-        Returns None if a namespace is not bound to the prefix or the prefix
-        is malformed.
+        Returns None if a namespace is not bound to the prefix.
 
         """
-        # TODO: When we drop support for 3.7 ...
-        # if (
-        #     isinstance(curie, str)
-        #     and (prefix := curie.split(":")[0])
-        #     and (ns := self.store.namespace(prefix))
-        # ):
-        #     return str(ns) + curie.split(":")[1]
-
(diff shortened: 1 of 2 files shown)


rdflib/namespace/__init__.py
```diff
@@ -590,6 +590,34 @@ def compute_qname_strict(
...
+        Expand a qname (aka CURIE) of the form <prefix:element>, e.g. "rdf:type"
+        into its full expression:
+
+        >>> import rdflib
+        >>> g = rdflib.Graph()
+        >>> g.namespace_manager.expand_qname("rdf:type")
+        'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'
+
+        Returns None if a namespace is not bound to the prefix or the prefix
+        is malformed.
```
aucampia, line 604: I think it is best to raise an error n such cases, the error can maybe be differentiated but for example if it is not a string I would raise a type error, if it does not contain ":" I would raise ValueError.

rdflib/namespace/__init__.py
```diff
@@ -590,6 +590,34 @@ def compute_qname_strict(
 
             return self.__cache_strict[uri]
 
+    def expand_qname(self, curie: str) -> Union[str, None]:
```
aucampia, line 593: I would rather have this expand_curie, as curie is a superset of QName:

https://www.w3.org/TR/curie/

> This specification addresses the problem by creating a new data type whose purpose is specifically to allow for the definition of scoped names that map to URIs in exactly this way. This type is called a "CURIE" or a "Compact URI". Syntactically, CURIEs are a superset of QNames.

aucampia (commented):
Happy to make the suggested changes if you want.

rdflib/namespace/__init__.py
gjhiggins: Claro, implemented.

rdflib/namespace/__init__.py
gjhiggins: Ah so. np, changed.

rdflib/namespace/__init__.py
```diff
@@ -590,6 +590,31 @@ def compute_qname_strict(
 
             return self.__cache_strict[uri]
 
+    def expand_curie(self, curie: str) -> Union[str, None]:
```
white-gecko, line 593: I would expect an `URIRef` to be returned rather then a string.

rdflib/namespace/__init__.py
```diff
@@ -590,6 +590,31 @@ def compute_qname_strict(
...
+        >>> import rdflib
+        >>> g = rdflib.Graph()
+        >>> g.namespace_manager.expand_curie("rdf:type")
+        'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'
+
+        Returns None if a namespace is not bound to the prefix.
+
+        """
+        if not type(curie) is str:
+            raise TypeError("Argument must be a string.")
```
aucampia, line 607: I'm understand somewhat why you are doing this, to raise an error if someone accidentally passes in a  URIRef, but I'm not sure it is worth it, because this will then break other subclasses of string also. Would like to hear other opinions though.

rdflib/namespace/__init__.py
```diff
@@ -590,6 +590,31 @@ def compute_qname_strict(
...
+        elif ":" not in curie or curie.split(":")[0] == "" or curie.split(":")[1] == "":
+            raise ValueError(
+                "Malformed curie argument, format should be e.g. “foaf:name”."
+            )
```
aucampia, line 611: The follow are valid curies which will not be be accepted, and resolve as follow given `@prefix ex: <https://example.com/>.`:

- `ex:` -> `https://example.com/`
- `ex:a:b:c` -> `https://example.com/a:b:c`

Checked with riot:

```console
$ echo "@prefix ex: <https://example.com/>. ex: ex:a ex:a:b:c." | riot --syntax TURTLE --out NTRIPLES
<https://example.com/> <https://example.com/a> <https://example.com/a:b:c> .
```

I will make a PR to your branch shortly to add this, and to also reduce the amount of times split is called.

rdflib/namespace/__init__.py
```diff
@@ -590,6 +590,31 @@ def compute_qname_strict(
...
+        else:
+            ns = self.store.namespace(curie.split(":")[0])
+            if ns is not None:
+                return URIRef(str(ns) + curie.split(":")[1])
+        return None
```
aucampia, line 616: I'm also a bit on the fence on returning None, I would just raise an error, but I would not say this is wrong, would like to hear some other opinions on the matter.

CC: @RDFLib/core-reviewers

rdflib/namespace/__init__.py
aucampia: Will also fix test in it, I simplified them a bit.

aucampia: Pull request here: https://github.com/gjhiggins/rdflib/pull/6/files

I would be very happy to apply this to your branch directly, just let me know.

aucampia:
Made a PR that changes logic slightly to be more compliant with CURIE spec:

- https://github.com/gjhiggins/rdflib/pull/6

rdflib/namespace/__init__.py
aucampia: Applied now.

rdflib/namespace/__init__.py
aucampia: Maybe the best compromise here is to rather do:

```suggestion
        if not isinstance(curie, str) or isinstance(curie, Identifier):
            raise TypeError("Argument must be a string but should not be an `rdflib.term.Identifier`")
```

That way if someone passes in something like a URIRef, BNode or Literal accidentally they will get an error. It may still be going too far to try and save someone, those are strings, and if this takes a string I think it may be best to just accept that if someone does it they mean to do it.

gjhiggins: I did start with `isinstance(curie, str)` but on testing, that proved too lax a check, hence the use of the more restrictive `type` and specific admonition to use a string. But making the error message more informative is worth doing and I added some accompanying tests and yes, in the case of an unbound prefix, it should return an informative exception rather than just `None`.

rdflib/namespace/__init__.py
nicholascar: I prefer raising an error: if the thig ased for cannot be calculated, an error should be returned, not None which just indicates it doesn't exist.

aucampia (approved):
Thanks @gjhiggins, very useful method to have.

edmondchuc (approved):
Nice, looks good.

nicholascar (approved):

dany-nonstop:
many thanks @gjhiggins for answering the feature request and adding this new feature. It works beautifully.

PR #1879, opened 2022-05-02
Move test utility modules into `test/utils/`

aucampia:
# Summary of changes

This patch moves all test utility modules into `test/utils/` and the
tests for these modules into `test/utils/test/`.

This is to make test utilities more organized and in preperation for
adding more test utilities related to test suites.

# Checklist

  the same change.


gjhiggins (approved):
makes sense to me

aucampia:
Merged with only 1 review as this only affects tests, and trivially so, also coverage remained the same.

PR #1822, opened 2022-04-15
graph modified

Priyansh19077:
# Summary of changes

# Checklist

- [ ] Checked that there aren't other open pull requests for
  the same change.
- [ ] Added tests for any changes that have a runtime impact.
- [ ] Checked that all tests and type checking passes.
- For changes that have a potential impact on users of this project:
  - [ ] Updated relevant documentation to avoid inaccuracies.
  - [ ] Considered adding additional documentation.
  - [ ] Considered adding an example in `./examples` for new features.
  - [ ] Considered updating our changelog (`CHANGELOG.md`).
- [ ] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

Issue #1424, opened 2021-09-29
RecursionError when serializing long linked list to turtle or ntriples


recalcitrantsupplant:
I have some graphs that contain long linked lists, which look like this when serialized as turtle:

```data:node-599b2326-0610-5dab-a1cb-54286acabe94 a doco:Section ;
    rdfs:label "section label" ;
    co:firstItem [ co:itemContent _:N4176d04ca148447794cfd0119164bde5 ;
            co:nextItem [ co:itemContent _:N222f957ef2564e9dbb2d5381e51616c5 ;
                    co:nextItem [ co:itemContent _:N0324dae7c1984ee7955df3cb0d3ba273 ;
                            co:nextItem [ co:itemContent _:N26f0049d1ae94c0c92a1dd1f6e2a0d07 ;
                                    co:nextItem [ co:itemContent _:N055c69a24c884b78bcf1b1e32096d4d3 ;
                                            co:nextItem [ co:itemContent _:Nd4cb0645ffcb4a96a2548437198791a7 ;
                                                    co:nextItem [ co:itemContent _:N6e0eceea0de54dd5a40fde45d4233025 ;
                                                            ...
```

For some of the longer lists I have, RDFLib fails to serialize these to either turtle or ntriples. I can however serialize the graphs to RDF/XML.
I have produced a graph which is only a linked list of length 500 to reproduce the issue [here](https://gist.github.com/recalcitrantsupplant/5b5916f5de2fccd47252c3b54cc064e2).

The code to reproduce the error is:
```
g = Graph().parse('https://gist.githubusercontent.com/recalcitrantsupplant/5b5916f5de2fccd47252c3b54cc064e2/raw/3982bb2b4129607572a1226fef907a18de42d089/long_linked_list.xml', format='xml')
g.serialize(format='turtle')
```
For me this raises:
`RecursionError: maximum recursion depth exceeded while calling a Python object`
The recursion appears to be on these lines in the stack trace (which are repeated many times):
```
  File ".../venv/lib/python3.8/site-packages/rdflib/plugins/serializers/turtle.py", line 335, in path
    self.p_squared(node, position, newline)
  File ".../venv/lib/python3.8/site-packages/rdflib/plugins/serializers/turtle.py", line 387, in p_squared
    self.predicateList(node, newline=False)
  File ".../venv/lib/python3.8/site-packages/rdflib/plugins/serializers/turtle.py", line 422, in predicateList
    self.verb(propList[0], newline=newline)
```

[...]

PR #1883, opened 2022-05-03
Issue 1424 Resolution as suggested

vahsek300501:
This is a resolution for the issue 1424. This resolution has been suggested by the David Habgood. We have added a parameter in the Graph() constructor for the recursion_depth so that the user can specify the desired recursion depth according to their needs. This way the user won't run into a stack overflow and even if they do they can come back to the code and set a bigger recursion limit to prevent it. 

However, setting the recursion depth to a particular value actually sets the default recursion depth of the system to that value. We have fixed this issue too by adding a destructor __del__() in the Graph class so that when the program ends, the recursion depth is set back to the original default value.


vahsek300501:
Issue 1424 Resolution

rdflib/graph.py
```diff
@@ -329,6 +330,7 @@ class Graph(Node):
 
     def __init__(
         self,
+        recursion_depth = 1000,
```
aucampia, line 333: This will break our API and I'm not clear it is for that a good reason. At the very least the parameter should go at the end but this seems entirely too arbitrary. Without understanding what problem you are trying to fix this is difficult to see the rationale.

.gitignore
```diff
@@ -151,3 +151,5 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+#Remove DS_Store: Mac
+**/.DS_Store
```
aucampia, line 155: This should go in `$XDG_CONFIG_HOME/git/ignore` ([ref](https://git-scm.com/docs/gitignore), this is ~/.config/git/ignore on Linux by default)

aucampia:
The right solution here seems to be to just avoid recursion wherever it is happening.

aucampia:
The first step in fixing this is reproducing it, we should have a test that fails for this, and then I think we should consider changing the parsing algorithm to not be recursive.

rdflib/graph.py
Priyansh19077: We can move the parameter to the very end. 
As far as the problem is concerned, in the issue (1424: https://github.com/RDFLib/rdflib/issues/1424) it is stated that in case the size of the graph is very large the program runs into a stack overflow. We have solved this by allowing the user to give an input of the value of recursion depth they would like for their code. As suggested by you, the last time we had raised a PR (1822: https://github.com/RDFLib/rdflib/pull/1822#discussion_r851310616), we have made the same changes.

Priyansh19077:
> The first step in fixing this is reproducing it, we should have a test that fails for this, and then I think we should consider changing the parsing algorithm to not be recursive.

The issue can be reproduced as follows: 
from rdflib import Graph
g=Graph().parse('https://gist.githubusercontent.com/recalcitrantsupplant/5b5916f5de2fccd47252c3b54cc064e2/raw/3982bb2b4129607572a1226fef907a18de42d089/long_linked_list.xml', format='xml')
g.serialize(format='turtle')

This gets resolved if we run pass the recursion depth as 100000
from rdflib import Graph
g=Graph(recursion_depth=100000).parse('https://gist.githubusercontent.com/recalcitrantsupplant/5b5916f5de2fccd47252c3b54cc064e2/raw/3982bb2b4129607572a1226fef907a18de42d089/long_linked_list.xml', format='xml')
g.serialize(format='turtle')

rdflib/graph.py
aucampia: Maybe you misunderstood my point, anyone can at any point call `sys.setrecursionlimit()` with any value they want, and if someone wants a different recursion limit they should set it appropriately. The underlying issue though is that python does not really work well with indefinite recursion, and if we have some algorithm that does do indefinite recursion then using `sys.setrecursionlimit()` will just be a stopgap anyway, as you can only make it so large. The only real solution that is appropriate in RDFLib to this issue from my point of view is to change to a non recursive parsing implementation.

aucampia:
By reproducing it I mean in RDFLib tests. In general we expect that all fixes are accompanied with test that would fail without the fix: 

https://rdflib.readthedocs.io/en/latest/developers.html#pull-requests-guidelines

> Changes that have a runtime impact are covered by unit tests.
> 
> There should either be existing tests that cover the changed code and behaviour, or the PR should include tests. For more information about what is considered adequate testing see the [Tests section](https://rdflib.readthedocs.io/en/latest/developers.html#tests).

https://rdflib.readthedocs.io/en/latest/developers.html#tests

> Tests that you add should show how your new feature or bug fix is doing what you say it is doing: if you remove your enhancement, your new tests should fail!

rdflib/graph.py
aucampia: To be clear though, by non recursive I mean in terms python function calls - the actual parsing algorithm can still be recursive as long as it is implemented using loops instead of python recursion.

aucampia (changes requested):
@Priyansh19077 there may be other opinions from other maintainers, but I would not be comfortable with approving a fix for this issue that is based on playing around with `sys.setrecursionlimit`. I could possibly consider a PR that adds a way to set `sys.setrecursionlimit` for our command line tools (e.g. rdfpipe).

Further, this PR does not add a tests for the issue, so there is little to salvage. 

Happy to review once tests are in place with a solution not based on `sys.setrecursionlimit`.

aucampia:
Will close this PR if there is nothing further by 2022-05-15

aucampia:
Closing as the solution proposed in this PR can already easily be implemented in code by users of this library if they want using functionality already present in python.

PR #1894, opened 2022-05-08
Fix simple literals returned as NULL using SERVICE (issue #1278)

### LINKED ISSUE #1278: In using SERVICE, "string" variables get retrieved as NULL
I was testing the federated SPARQL queries using "service". There are two issues:
* only low-case "service" is accepted, i.e., SPARQL with "SERVICE" would fail. This was fixed for next release. 
*  "string" variables get retrieved as null. This is new.

Sorry, I do not have time to create a small reproducible example. 
Below, two examples of querying the same SPARQL endpoint, one with   

from pandas import DataFrame
from rdflib.plugins.sparql.processor import SPARQLResult

def sparql_results_to_df(results: SPARQLResult) -> DataFrame:
    return DataFrame(
        data=([None if x is None else x.toPython() for x in row] for row in results),
        columns=[str(x) for x in results.vars],
    )

qf="""
    PREFIX tsmodel: <http://siemens.com/tsmodel/>
    PREFIX vobj: <http://siemens.com/tsmodel/VAROBJECTS#>
                  
     SELECT ?x ?node_name ?idx 
     {      
        service <http://localhost:8080/sparql>
        {
            ?x a tsmodel:VAROBJECTS .
            ?x vobj:idx ?idx .
            ?x vobj:node_name ?node_name .
        }
      }
"""

results=graph.query(qf)
df=sparql_results_to_df(results)
df.head(20)

x	node_name	idx
0	http://siemens.com/tsmodel/VAROBJECTS/idx=1	None	1
1	http://siemens.com/tsmodel/VAROBJECTS/idx=2	None	2
2	http://siemens.com/tsmodel/VAROBJECTS/idx=3	None	3
3	http://siemens.com/tsmodel/VAROBJECTS/idx=4	None	4
4	http://siemens.com/tsmodel/VAROBJECTS/idx=5	None	5

import sparql_dataframe

endpoint = "http://localhost:8080/sparql"

q = """
PREFIX tsmodel: <http://siemens.com/tsmodel/>
PREFIX vobj: <http://siemens.com/tsmodel/VAROBJECTS#>

SELECT ?x ?node_name ?idx { 
        ?x a tsmodel:VAROBJECTS .
        ?x vobj:idx ?idx .
        ?x vobj:node_name ?node_name .
    }
"""
        
df = sparql_dataframe.get(endpoint, q)
df.head()

	x	node_name	idx
0	http://siemens.com/tsmodel/VAROBJECTS/idx=1	ns=3;s="ultrasonicLeft"	1
1	http://siemens.com/tsmodel/VAROBJECTS/idx=2	ns=3;s="voltageL1-N"	2
2	http://siemens.com/tsmodel/VAROBJECTS/idx=3	ns=3;s="voltageL2-N"	3
3	http://siemens.com/tsmodel/VAROBJECTS/idx=4	ns=3;s="voltageL3-N"	4
4	http://siemens.com/tsmodel/VAROBJECTS/idx=5	ns=3;s="currentL1"	5

I do have more examples like this. As I have checked, "int", "float", "datetime" work fine. But, "string" variables are always get "NULL".

gitmpje:
# Summary of changes

Fixes #1278 simple literals returned as NULL. The resolution uses same logic as here: https://github.com/RDFLib/rdflib/blob/6f2c11cd2c549d6410f9a1c948ab3a8dbf77ca00/rdflib/plugins/sparql/results/jsonresults.py#L89-L107

# Checklist

  the same change.


aucampia (dismissed):
Thanks for the PR, a bit swamped fixing bugs and reviewing/fixing older PRs right now so I won't be able to take a detailed look until later, but it would be good to expand testing for `SERVICE` directive to ensure that each branch here is covered, which I'm not sure it will be, though I may be wrong.

aucampia:
Also it seems your test is failing: https://github.com/RDFLib/rdflib/runs/6341689732?check_suite_focus=true#step:9:372

```console
def test():
          """Test service returns simple literals not as NULL.
      
          Issue: https://github.com/RDFLib/rdflib/issues/1278
          """
      
          g = Graph()
          q = """SELECT ?s ?p ?o
      WHERE {
          SERVICE <https://dbpedia.org/sparql> {
              VALUES (?s ?p ?o) {(<http://example.org/a> <http://example.org/b> "c")}
          }
      }"""
  >       assert results.bindings[0].get(Variable("o")) == Literal("c")
  E       NameError: name 'results' is not defined
```

gitmpje:
Thanks for your support, the failing test should be fixed now. This test uses a similar approach as the existing SERVICE tests: https://github.com/RDFLib/rdflib/blob/eba13739e1d24b3e52697a1ec9545a361116951f/test/test_sparql/test_service.py. When I have time I will have a look at your suggestions for improving the test(s).

aucampia:
> Thanks for your support, the failing test should be fixed now. This test uses a similar approach as the existing SERVICE tests: https://github.com/RDFLib/rdflib/blob/eba13739e1d24b3e52697a1ec9545a361116951f/test/test_sparql/test_service.py. When I have time I will have a look at your suggestions for improving the test(s).

Okay in that case I think it is fine then, but maybe it would be good to add this test to `test_service.py`.

gitmpje:
I added a more general test for different node types returned by a SERVICE clause to test_service.py.

aucampia:
@gitmpje will review tonight, I don't think there are any issues with the code, but I may move tests around slightly, thank you for the PR.

rdflib/plugins/sparql/evaluate.py
```diff
@@ -381,18 +381,22 @@ def _yieldBindingsFromServiceCallResult(ctx: QueryContext, r, variables):
...
-            elif r[var]["type"] == "literal" and "xml:lang" in r[var]:
+            elif t == "typed-literal":
                 res_dict[Variable(var)] = Literal(
-                    r[var]["value"], lang=r[var]["xml:lang"]
+                    d["value"], datatype=URIRef(d["datatype"])
                 )
+            elif t == "bnode":
+                res_dict[Variable(var)] = BNode(d["value"])
+            else:
+                raise NotImplementedError("json term type %r" % t)
```
aucampia, line 399: I know there is a lot of places where this is used like this, but I think `NotImplementedError` is not appropriate in this case as it is described as

https://docs.python.org/3/library/exceptions.html#NotImplementedError

> This exception is derived from [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError). In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

I would rather go for [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError):
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

```suggestion
                raise ValueError(f"invalid type {t!r} for variable binding {var!r}")
```

Don't apply this suggestion though, I will made a commit on your branch to fix this and some other issues.

rdflib/plugins/sparql/evaluate.py
```diff
@@ -381,18 +381,22 @@ def _yieldBindingsFromServiceCallResult(ctx: QueryContext, r, variables):
...
+            t = d["type"]
+            if t == "uri":
+                res_dict[Variable(var)] = URIRef(d["value"])
+            elif t == "literal":
                 res_dict[Variable(var)] = Literal(
-                    r[var]["value"], datatype=r[var]["datatype"]
+                    d["value"], datatype=d.get("datatype"), lang=d.get("xml:lang")
                 )
-            elif r[var]["type"] == "literal" and "xml:lang" in r[var]:
+            elif t == "typed-literal":
```
aucampia, line 392: I'm don't think `typed-literal` is a valid type:

https://www.w3.org/TR/sparql11-results-json/#select-encode-terms

image

The reason it is here is because of https://www.w3.org/TR/2006/NOTE-rdf-sparql-json-res-20061004/#variable-binding-results AFAIK - so best to add a comment here.

gitmpje: @aucampia To my opinion both are valid comments. I reused the logic from the JSON results parser: https://github.com/RDFLib/rdflib/blob/6f2c11cd2c549d6410f9a1c948ab3a8dbf77ca00/rdflib/plugins/sparql/results/jsonresults.py#L89-L107
Maybe it is a good idea to apply the same changes there too.

aucampia: I will do that in a separate PR, trying to figure out how to reproduce a BNode in response.

gitmpje: I also considered adding BNode to the test, but could not come up with a good way to do it, so it would be nice if you can figure this out!

aucampia: I kind of gave up, I will go with using the mock HTTP server that is now in master and check it in a separate test.

aucampia: Also I will add this comment shortly, will tell you when done but will get everything to a state where I approve tonight still.

aucampia: Added mock based tests now.

rdflib/plugins/sparql/evaluate.py
aucampia: Fixed on your branch now.

aucampia (approved):
@gitmpje thanks for the PR. I added more tests to your branch and made some minor mostly cosmetic improvements, let me know if there are any concerns you have with the changes.

gitmpje:
> @gitmpje thanks for the PR. I added more tests to your branch and made some minor mostly cosmetic improvements, let me know if there are any concerns you have with the changes.

No concerns from my side, thank you for your comments and support!

aucampia:
@RDFLib/core I will merge this by 2022-05-17 if there is no further feedback.

gjhiggins (approved):

PR #1902, opened 2022-05-09
Fixes #1429, add `iri2uri`

### LINKED ISSUE #1429: Graph parse module does not handle URI / IRI with non-ascii characters

gjhiggins:
Fixes #1429

https://github.com/RDFLib/rdflib/issues/1429#issuecomment-994199483)
> why park this in the "restore-pyrdfa" branch? This fix will work for non-HTML sources with non-ASCII chars in the IRI, perhaps something like `http://example.com/thing/Almería.ttl` for instance. So this is a generic fix for `parse()`

# Summary of changes

Add an iri-to-uri conversion utility to encode IRIs to URIs for `Graph.parse()` sources. Added a couple of tests because feeding it with a suite of IRIs to check seems overkill (not that I could find one).

# Checklist


rdflib/parser.py
```diff
@@ -47,9 +47,43 @@
     "URLInputSource",
     "FileInputSource",
     "PythonInputSource",
+    "iri2uri",
 ]
 
 
+def iri2uri(iri: str) -> str:
```
aucampia, line 54: I'm a bit ambivalent to having this as a public function, it may be useful but it is not really that important for working with RDF and in this case it only fixes an issue which is not specific to RDF but more specific to network interaction.

Also I would somewhat favor having this in `rdflib.utils` instead of `rdflib.graph` - as I think we will likely end up using it in JSON-LD and SPARQL code in future.

So my suggestion would be `rdflib.util._iri2uri` instead of `rdflib.graph.iri2uri`.

aucampia: Happy to make the change on request.

rdflib/parser.py
```diff
@@ -47,9 +47,43 @@
...
+    (scheme, netloc, path, query, fragment) = urlsplit(iri)
+
+    # Just support http/https, otherwise return the iri unmolested
+    if scheme not in ["http", "https"]:
+        return iri
+
+    scheme = quote(scheme)
+    netloc = netloc.encode("idna").decode("utf-8")
+    path = quote(path)
+    query = unquote(quote(query))
```
aucampia, line 77: I'm busy adding some more tests, but as far as I can tell this should not unquote after quoting, as then the following tests fails:

```console
_________________________________________________________ test_iri2uri[http://example.com/?\xe9=1-expected_result5] _________________________________________________________
  File "/home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/test_util.py", line 626, in test_iri2uri
    assert result in expected_result
AssertionError: assert 'http://example.com/?é=1' in {'http://example.com/?%C3%A9%3D1', 'http://example.com/?%C3%A9=1'}
```

I would say it should be:

```suggestion
    query = quote(query)
```

aucampia: I will make the change in a commit shortly, if you disagree let me know, I may be missing something.

rdflib/parser.py
aucampia: Done now.

aucampia:
Made some changes now to your branch, in brief:

> - Move `rdflib.parser.iri2uri` to `rdflib.util._iri2uri`
> 
>   In part this is because iri2uri does not operate on all IRIs, just
>   http and https IRIs, but furthermore it is because this is not
>   something essential to working with RDF.
> 
> - Expanded tests for `iri2uri` to cover each part of the URI.
> 
> - Convert the test that connects to dbpedia to using httpmock.
> 
>   This is so we have more control over response and can explicitly
>   verify the request path.
> 
> - Some changes to imports to reduce the probability of circular imports.

aucampia (approved):
LGTM

rdflib/parser.py
gjhiggins: idk, looks like a typo,  so long ago, I've forgotten.
lgtm

aucampia: okay, resolving.

edmondchuc (approved):
Nice! Looks good.

PR #1918, opened 2022-05-13
Made the N-triples parser more robust. Resolved #724

### LINKED ISSUE #724: Please make the parser more robust

karish-grover:
Fixes #724 

# Summary of changes

Consider the given set of ill-formatted n-triples:

```
<http://example.org/#ThreeMemberList> <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <http://example.org/#p> .
<http://example.org/#ThreeMemberList> <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> _:list2 \t
_:list2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> "false" defgvcsxa
_:list2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> _:
_:list3 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <http://example.org/#r> .
_:list3 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil> .
```

There are formatting errors in triples 2, 3, and 4, and the current state of the NTriples parser would raise the below exception `rdflib.exceptions.ParserError: Invalid line:` **without specifying** the triple which was incorrect or the line number. 

We have made a series of changes to the existing state of the parser:
- Added a keyword argument `ignore_errors` to the `g.parse()` function in the file [ntriples.py](https://github.com/RDFLib/rdflib/compare/master...karish-grover:master#diff-91378596f5461879529f6f18f2e33714ab8787397df361260c3cb86c9f9465de).
- If ignore_`errors == False`, specified the triplet for user convenience, as shown below:-

`rdflib.exceptions.ParserError: Invalid line: <http://example.org/#ThreeMemberList> <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> _:list2`

- If `ignore_errors == True`, then we added a few more functionalities that could be convenient to the users:- 
    - We ignore the faulted triples and add the triples which are in the correct format to the graph. 
    - This further logs the line number and triplet which is faulty as shown:

`WARNING  root:ntriples.py:179 Invalid triple at line #5: _:list2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> _:; IGNORED`

We have also added a mechanism to correct some of the faulted triples in the case when the trailing character after `(s, p, o)` is either multiple dots `.` or no dots, or any other characters (like `\t`). In this case, we correct the triples and add the corrected triples to the graph. This further logs the triple which was corrected as shown:-

`WARNING  root:ntriples.py:237 Invalid triple at line #4: _:list2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> "false" defgvcsxa (Extra characters detected at the end of triple); CORRECTED`

We have also added a test file [test_nt.py](https://github.com/RDFLib/rdflib/compare/master...karish-grover:master#diff-459ef51a7fb3e0a6c0f21be8b3e7ff779a657bdc639bf9038e6db76a7cd8c3ad) containing tests for the same.

In addition to these changes, there can be a few more formatting faults that might be rectifiable during the runtime and we urge the reviewers to kindly let us know these changes so we can add them to this PR.

### DIFF
--- a/rdflib/plugins/parsers/ntriples.py
+++ b/rdflib/plugins/parsers/ntriples.py
@@ -7,10 +7,11 @@ Author: Sean B. Palmer, inamidst.com
 """
 
 import codecs
+import logging
 import re
 from io import BytesIO, StringIO, TextIOBase
 from typing import IO, TYPE_CHECKING, Optional, Pattern, TextIO, Union
-import logging
+
 from rdflib.compat import _string_escape_map, decodeUnicodeEscape
 from rdflib.exceptions import ParserError as ParseError
 from rdflib.parser import InputSource, Parser
@@ -116,7 +117,16 @@ class W3CNTriplesParser(object):
     `NTriplesParser`.
     """
 
-    __slots__ = ("_bnode_ids", "sink", "buffer", "file", "line", 'ignore_errors', 'line_number', 'current_line')
+    __slots__ = (
+        "_bnode_ids",
+        "sink",
+        "buffer",
+        "file",
+        "line",
+        'ignore_errors',
+        'line_number',
+        'current_line',
+    )
 
     def __init__(
         self, sink: Optional[Union[DummySink, "NTGraphSink"]] = None, bnode_context=None
@@ -138,7 +148,10 @@ class W3CNTriplesParser(object):
         self.ignore_errors = False
 
     def parse(
-        self, f: Union[TextIO, IO[bytes], codecs.StreamReader], bnode_context=None, ignore_errors = False,
+        self,
+        f: Union[TextIO, IO[bytes], codecs.StreamReader],
+        bnode_context=None,
+        ignore_errors=False,
     ):
 
         """
@@ -166,17 +179,21 @@ class W3CNTriplesParser(object):
         self.line_number = 0
         while True:
             self.line = self.readline()
-            self.line_number+=1
+            self.line_number += 1
             self.current_line = self.line
             if self.line is None:
                 break
             try:
                 self.parseline(bnode_context=bnode_context)
             except ParseError:
-            
+
                 if ignore_errors:
-                    #logging the invalid triples in the input file and ignoring them while creating the final graph
-                    logging.warning("Invalid triple at line #{}: {}; IGNORED".format(self.line_number, self.current_line))
+                    # logging the invalid triples in the input file and ignoring them while creating the final graph
+                    logging.warning(
+                        "Invalid triple at line #{}: {}; IGNORED".format(
+                            self.line_number, self.current_line
+                        )
+                    )
                     continue
                 else:
                     raise ParseError("Invalid line: {}".format(self.current_line))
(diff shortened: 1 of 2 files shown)


karish-grover:
pre-commit.ci autofix

rdflib/plugins/parsers/ntriples.py
```diff
@@ -135,10 +145,15 @@ def __init__(
         self.buffer: Optional[str] = None
         self.file: Optional[Union[TextIO, codecs.StreamReader]] = None
         self.line: Optional[str] = ""
+        self.ignore_errors = False
```
aucampia, line 148: should not be a public variable.

rdflib/plugins/parsers/ntriples.py
```diff
@@ -158,16 +173,30 @@ def parse(
             # someone still using a bytestream here?
             f = codecs.getreader("utf-8")(f)  # type: ignore[arg-type]
 
+        self.ignore_errors = ignore_errors
```
aucampia, line 176: Why do you need this as an instance variable?

test/test_parsers/test_nt.py
```diff
@@ -0,0 +1,43 @@
...
+_:list3 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil> .
+"""
+
+
+class Test_NT_Parsing:
+    def test_nt_1(self):
+        ignore_errors = True
+        g = Graph().parse(data=DATA_1, format="nt", ignore_errors=ignore_errors)
+        if ignore_errors:
+            assert len(g) == 9
```
aucampia, line 35: This to me is not really sufficent validation, it should use `GraphHelper.assert_isomorphic` and/or `GraphHelper.assert_sets_equal`.

aucampia: I would actually prefer it be added to `rdflib/test/data/variants` and that variants processor be extended to read parse options from something like `-options.json`. Will need to change [test_variants.py](https://github.com/RDFLib/rdflib/blob/master/test/test_graph/test_variants.py)

aucampia: I would also actually just use an existing variant, e.g. `test/data/variants/diverse_triples.nt`

So

```
test/data/variants/diverse_triples.nt
test/data/variants/diverse_triples-variant-blank_line.nt
test/data/variants/diverse_triples-variant-blank_line.nt-opt.json
```

And then in  `test/data/variants/diverse_triples-variant-blank_line.nt-opt.json`

```json
{
  "ignore_failures": true
}
```

aucampia:
I'm somewhat ambivalent of this approach, I think a better approach may be to rather provide a way to do stream parsing, the problem with this is that it establishes a contract to provide a broken parser, and now if we want to replace our parser we have to similarly change it to accept this broken input.

rdflib/plugins/parsers/ntriples.py
karish-grover: Hey, thanks a lot for your reviews. 

Can we make it a global variable instead? Would that suffice?

gjhiggins:
I agree with your ambivalence. 

> .. it establishes a contract to provide a broken parser, and now **if we want to replace our parser** ...

This might even be on the cards - I note that the RDF Star parser imports lark which suggests RDFLib might well be adding lark as a dependency in future and that opens up the potential for a lark-based re-implementation of at least some of the RDFLib parsers.

aucampia:
Please rebase this branch onto master or merge master into this branch.

aucampia:
If there is nothing further on this PR I will close it.

aucampia:
Closing this PR due to inactivity, feel free to open it again if you continue working on it again.

Issue #1427, opened 2021-10-06
Allow prefixed names to be used as the endpoint in a SERVICE clause


mgberg:
Federated queries do not work when the provided endpoint in a SERVICE clause is a prefixed name. I have tracked the error down to the [`evalServiceQuery` function](https://github.com/RDFLib/rdflib/blob/038af4547e799845e08986867b24acafcbe72d48/rdflib/plugins/sparql/evaluate.py#L302).

The regular expression used (`"^service <(.*)>[ \n]*{(.*)}[ \n]*$"`) will not match a prefixed name so the query will not be evaluated. That regex should be adjusted to include prefixed names.

Also, note that if a prefixed name is currently used as the endpoint, `part.term` already contains the correctly expanded `URIRef` of the endpoint that should be able to be used as the `service_url` a few lines later; the prefixed name matched by the regular expression shouldn't need to be expanded.

shashankdaima:
can you assign this to me? @mgberg

shashankdaima:
@mgberg can you please give me the example of query that will cause error?

mgberg:
In rdflib 6.1.1, the following works as expected:
```python
list(
    g.query("""\
        SELECT *
        WHERE {
          SERVICE <https://dbpedia.org/sparql> {
            {
              SELECT ?s ?p ?o
              WHERE {
                ?s ?p ?o
              }
              LIMIT 10
            }
          }
        }"""
    )
)
```
However, the following does not work:
```python
g.bind("dbp", "https://dbpedia.org/")
list(
    g.query("""\
        SELECT *
        WHERE {
          SERVICE dbp:sparql {
            {
              SELECT ?s ?p ?o
              WHERE {
                ?s ?p ?o
              }
              LIMIT 10
            }
          }
        }"""
    )
)
```

shashankdaima:
@mgberg  and prefix can entered like this as well? ```PREFIX dbp: <http://dbpedia.org/>
        SELECT *
        WHERE {
          SERVICE dbp:sparql {
            {
              SELECT ?s ?p ?o
              WHERE {
                ?s ?p ?o
              }
              LIMIT 1
            }
          }
        }```.

mgberg:
Explicitly specifying the prefix does not help. This snippet doesn't work for me either: g.bind("dbp", "https://dbpedia.org/")
list(
    g.query("""\
        PREFIX dbp: <http://dbpedia.org/>
        SELECT *
        WHERE {
[...]

PR #1920, opened 2022-05-13
Resolve Issue 1427, allow prefixed names to be used as the endpoint in a SERVICE clause

shashankdaima:
Fixed : #1427
## Summary of this PR:
Previously Federated queries(Remote Queries) don't have support for `BaseUrl` to be included as `Namespace`. So, queries that do have `URLRef` as `namespace` don't show up the result. 
#### Code:
``` python
from rdflib import Graph

if __name__ == "__main__":
    g = Graph()
    g.bind("dbp", "https://dbpedia.org/")
    qres = g.query(
        """
        SELECT ?s
            WHERE {
                SERVICE dbp:sparql {
                    ?s a ?o .
                }
            }
        LIMIT 3
        """
    )
    for row in qres:
        print(row.s)

```
#### Result:
```
http://www.openlinksw.com/virtrdf-data-formats#default-iid
http://www.openlinksw.com/virtrdf-data-formats#default-iid-nullable
http://www.openlinksw.com/virtrdf-data-formats#default-iid-blank 
```
## Checklist of this PR:

  the same change.
- [x] Added example query that I used as sample for this code (file name =`examples/sparql_service_query_example.py`).

### DIFF
--- a/rdflib/plugins/sparql/evaluate.py
+++ b/rdflib/plugins/sparql/evaluate.py
@@ -302,9 +302,24 @@ def evalPart(ctx: QueryContext, part: CompValue):
 
 def evalServiceQuery(ctx: QueryContext, part):
     res = {}
+    service_string = part.get("service_string", "")
+    match = re.match(
+        "^service (.*):(.*) [ \n]*{(.*)}[ \n]*$",
+        service_string,
+        re.DOTALL | re.I,
+    )
+    if match:
+        _prefix = match.group(1)
+        for prefix, namespace in ctx.graph.namespace_manager.namespaces():
+            if prefix == _prefix:
+                service_string = (
+                    f"SERVICE <{namespace}{match.group(2)}> {{{match.group(3)}}}"
+                )
+                break
+
     match = re.match(
         "^service <(.*)>[ \n]*{(.*)}[ \n]*$",
-        part.get("service_string", ""),
+        service_string,
         re.DOTALL | re.I,
     )
 
--- a/examples/sparql_service_query_example.py
+++ b/examples/sparql_service_query_example.py
@@ -1,0 +1,18 @@
+from rdflib import Graph
+
+if __name__ == "__main__":
+    g = Graph()
+    g.bind("dbp", "https://dbpedia.org/")
+    qres = g.query(
+        """
+        SELECT ?s
+            WHERE {
+                SERVICE dbp:sparql {
+                    ?s a ?o .
+                }
+            }
+        LIMIT 3
+        """
+    )
+    for row in qres:
+        print(row.s)


shashankdaima:
thanks @gjhiggins for helping me out with my queries. I will contribute more to the library.

aucampia:
This should include tests.

gjhiggins:
> This should include tests.

Agreed. The code in your [sparql_service_query_example.py](https://github.com/RDFLib/rdflib/pull/1920/files#diff-098f1fea954c54e787ebf2722a7e9020063843b7b49549e13a0ba5a800c64b0d) should be re-worked to fit in with the other service tests in https://github.com/RDFLib/rdflib/blob/master/test/test_sparql/test_service.py

Also, as [@mgberg noted](https://github.com/RDFLib/rdflib/issues/1427#issuecomment-1126237154) the context of federated queries is relevant here and should be part of the test.

rdflib/plugins/sparql/evaluate.py
```diff
@@ -302,9 +302,24 @@ def evalPart(ctx: QueryContext, part: CompValue):
...
     res = {}
+    service_string = part.get("service_string", "")
+    match = re.match(
+        "^service (.*):(.*) [ \n]*{(.*)}[ \n]*$",
+        service_string,
+        re.DOTALL | re.I,
+    )
+    if match:
+        _prefix = match.group(1)
+        for prefix, namespace in ctx.graph.namespace_manager.namespaces():
```
mgberg, line 313: Will this work in the case when the prefix declaration is in the query itself and not bound to the graph? Or is this graph different than the graph the query is being executed against?

rdflib/plugins/sparql/evaluate.py
```diff
@@ -302,9 +302,24 @@ def evalPart(ctx: QueryContext, part: CompValue):
 
 def evalServiceQuery(ctx: QueryContext, part):
     res = {}
+    service_string = part.get("service_string", "")
+    match = re.match(
+        "^service (.*):(.*) [ \n]*{(.*)}[ \n]*$",
+        service_string,
+        re.DOTALL | re.I,
+    )
```
mgberg, line 310: Again, `term.part` already has the expanded URI so it may not be necessary to do an additional regex match to figure it out. I'm just not familiar with the query processor enough to know if `term.part` _always and reliably_ has the expanded URI.

rdflib/plugins/sparql/evaluate.py
shashankdaima: this will work if namespace provided in form by `bind` function. The normal PREFIX eg:<http://www.example.org/>; will not be able to work as the original query is not looking for PREFIX in `query` method.

rdflib/plugins/sparql/evaluate.py
shashankdaima: i am not very sure of what `term.part` is. But i will make further changes and add tests. So, this pr is still active. I just want to break first PR asap. So, that i will not hesitate contribute to it more. thanks for mentioning this issue.

rdflib/plugins/sparql/evaluate.py
shashankdaima: btw should i create a new issue of `query` function is not analysing for PREFIX commands? i think it should be a separate issue.

rdflib/plugins/sparql/evaluate.py
mgberg: I'm not either, which is why I was hoping someone who was more familiar with the query processor would know how to fix this the best way.

rdflib/plugins/sparql/evaluate.py
mgberg: No, there is not a problem with prefixes specified in a query not getting bound to a graph. That's perfectly valid, as in general prefixes are scoped to a query and the ability to bind prefixes to a graph is just a convenience. So I don't this change is a generalized solution.

That being said, the query parser _does_ parse PREFIX clauses and _does_ expand prefixed qnames where the prefix is specified in a PREFIX clause. In the case of the SERVICE clause, the expanded service URI seems to appear in `part.term`. [This example I posted](https://github.com/RDFLib/rdflib/issues/1427#issuecomment-1126072997) shows this working correctly. Note in that example I did NOT bind the prefix to the graph first.

gjhiggins: > I don't think that this change is a generalized solution.

Agreed. I distilled it down to:
```diff
diff --git a/rdflib/plugins/sparql/evaluate.py b/rdflib/plugins/sparql/evaluate.py
index 6cdbe754..5c0792e3 100644
--- a/rdflib/plugins/sparql/evaluate.py
+++ b/rdflib/plugins/sparql/evaluate.py
@@ -300,13 +300,19 @@ def evalPart(ctx: QueryContext, part: CompValue):
         raise Exception("I dont know: %s" % part.name)
 
 
+prefix_match_re = re.compile(
+    "^service [^<](.*):(.*)[^>] [ \n]*{(.*)}[ \n]*$", re.DOTALL | re.I
+)
+match_re = re.compile("^service <(.*)>[ \n]*{(.*)}[ \n]*$", re.DOTALL | re.I)
+
+
 def evalServiceQuery(ctx: QueryContext, part):
     res = {}
-    match = re.match(
-        "^service <(.*)>[ \n]*{(.*)}[ \n]*$",
-        part.get("service_string", ""),
-        re.DOTALL | re.I,
-    )
+    service_string = part.get("service_string", "")
+    prefix_match = prefix_match_re.match(service_string)
+    if prefix_match:
+        service_string = f"SERVICE <{part.term}> {{{prefix_match.group(3)}}}"
+    match = match_re.match(service_string)
 
     if match:
         service_url = match.group(1)
```

And added these two (passing) tests to `test/test_sparql/test_service.py`:
```py
def test_service_with_graph_bound_prefix():
    g = Graph()
    g.bind("dbp", "https://dbpedia.org/")
    q = """SELECT ?s
    WHERE {
        SERVICE dbp:sparql {
            ?s a ?o .
        }
    }
    LIMIT 3"""
    results = helper.query_with_retry(g, q)
    assert len(results) == 3


def test_service_with_query_bound_prefix():
    g = Graph()
    q = """PREFIX dbp: <https://dbpedia.org/>
    SELECT ?s
    WHERE {
        SERVICE dbp:sparql {
            ?s ?p ?o .  
        }
    }
    LIMIT 3"""
    results = helper.query_with_retry(g, q)
    assert len(results) == 3
```

gjhiggins:
Even if we have a *pro tem* workaround for simple cases, **in general**, string-matching is not an appropriate technique to apply to this issue.

There's a formidable example in the [SPARQL Federated Query docs](https://www.w3.org/TR/sparql11-federated-query/#optionalTwoServices) that I shall reproduce here so we can contemplate the full implications:

### 2.2 SPARQL query with OPTIONAL to two remote SPARQL endpoints
Imagine we want to query people and optionally obtain their interests  and the names of people they know. Imagine for instance, two endpoints  containing data about people:

Data in the default graph at remote SPARQL endpoint: `http://people.example.org/sparql` 
```turtle
@prefix foaf:  <http://xmlns.com/foaf/0.1/> .
@prefix : <http://example.org/> .

:people15  foaf:name     "Alice" .
:people16  foaf:name     "Bob" .
:people17  foaf:name     "Charles" .
:people17  foaf:interest     <http://www.w3.org/2001/sw/rdb2rdf/> .
```

and data in the default graph the remote SPARQL endpoint: `http://people2.example.org/sparql`
```turtle
@prefix foaf:  <http://xmlns.com/foaf/0.1/> .
@prefix : <http://example.org/> .

:people15  foaf:knows    :people18 . 
:people18  foaf:name     "Mike" .
:people17  foaf:knows    :people19 . 
:people19  foaf:name     "Daisy" .
```

##### Query:
```sparql
PREFIX foaf:   <http://xmlns.com/foaf/0.1/>
SELECT ?person ?interest ?known
WHERE
{
  SERVICE <http://people.example.org/sparql> { 
    ?person foaf:name ?name .  
    OPTIONAL { 
      ?person foaf:interest ?interest .
      SERVICE <http://people2.example.org/sparql> { 
        ?person foaf:knows ?known . } }
  }    
}
```
This query, on the data above, has three solutions:

Query Result:
```
person | interest | known
-- | -- | --
"Alice" |   |  
"Bob" |   |  
"Charles" | <http://www.w3.org/2001/sw/rdb2rdf/> | <http://example.org/people19>
```

Notice that in the query above there is **a nested `SERVICE`** in the `OPTIONAL` clause. This query requires the SPARQL query service at `http://people.example.org/sparql` to support basic federated query.

aucampia:
> Even if we have a _pro tem_ workaround for simple cases, **in general**, string-matching is not an appropriate technique to apply to this issue.

Agreed

aucampia (changes requested):
- This will need tests.
- In most cases string matching is not acceptable.
- Please make your PR title more descriptive.

shashankdaima:
okay @aucampia . sorry, I was busy in some academic related work. Thanks for making this list of stuff i need to do.  Will start working on it.

aucampia:
Please rebase this branch onto master or merge master into this branch.

shashankdaima:
> Please rebase this branch onto master or merge master into this branch.
if i may ask why? would making PR from specific branch other than master, will make branch inside the main repository.

aucampia:
> > Please rebase this branch onto master or merge master into this branch.
> 
> if i may ask why? would making PR from specific branch other than master, will make branch inside the main repository?

I enabled [flakeheaven](https://github.com/flakeheaven/flakeheaven) on master branch and we have added more tests and typing since your PR was made and it mainly to ensure this is compliant with our flake8 rules and does not violate tests or typing.

aucampia:
You can do it via the web user interface also: 

image

gjhiggins:
> You can do it via the web user interface also:

What's your preference, merge or rebase? I'm nursing a number of to-be-PRs and am accumulating a fair number of merge-commits which I feel may well make reviewing more effortful. Is there an acceptable strategy to follow?

aucampia:
> > You can do it via the web user interface also:
> 
> What's your preference, merge or rebase? I'm nursing a number of to-be-PRs and am accumulating a fair number of merge-commits which I feel may well make reviewing more effortful. Is there an acceptable strategy to follow?

My preference is for rebasing, it was a bit weird when I started doing it but now I'm pretty used to it and it feels right, even though you end up doing `git push --force` a lot. I don't think either is wrong exactly and it does not bother me that much what others use, it is just with rebase you can keep the commit history cleaner and easier to understand, while with merge it becomes quite noisy but it is also sometimes easier to notice what exactly happened during merge, so there are tradeoffs either way.

I would not change anything of existing branches though, and if you have already used merging on a branch it may be best to continue using it.

gjhiggins:
> My preference is for rebasing ... while with merge .. it is also sometimes easier to notice what exactly happened during merge

Ah, I get it. Slightly depends on what the branch is addressing.

> I would not change anything of existing branches though, and if you have already used merging on a branch it may be best to continue using it.

Thank you. Noted, will migrate to rebase for future work.

mgberg:
> Even if we have a _pro tem_ workaround for simple cases, **in general**, string-matching is not an appropriate technique to apply to this issue.
> 
> There's a formidable example in the [SPARQL Federated Query docs](https://www.w3.org/TR/sparql11-federated-query/#optionalTwoServices) that I shall reproduce here so we can contemplate the full implications:

@gjhiggins Are you trying to say that more work needs to be done to make the federation safer? As far as I'm aware, if someone is nesting `SERVICE` clauses they are responsible for knowing that the outermost endpoint supports federation.

That being said, I could imagine that [`_buildQueryStringForServiceCall`](https://github.com/RDFLib/rdflib/blob/038af4547e799845e08986867b24acafcbe72d48/rdflib/plugins/sparql/evaluate.py#L355) could be made a bit more robust to make sure what getting sent out is valid.

gjhiggins:
> > Even if we have a _pro tem_ workaround for simple cases, **in general**, string-matching is not an appropriate technique to apply to this issue.

> @gjhiggins Are you trying to say that more work needs to be done to make the federation safer? As far as I'm aware, if someone is nesting `SERVICE` clauses they are responsible for knowing that the outermost endpoint supports federation.

It's more the case that the [distillation I posted](https://github.com/RDFLib/rdflib/pull/1920#discussion_r872980786) is basically just a workaround if the issue is an operational prroblem for the simple use case that you just want to “go away” for the interim.

I wouldn't aspire to suggest it as a PR because it's essentially *wrong*  - not only it is incomplete but it's also ill-founded in its employment of string-matching - the core bug remains unaddressed. It's not a good idea to add temporary incomplete fixes because having to track down the code later and strip it out actually adds to the cost of fixing the core bug.

aucampia:
Closing this PR due to inactivity, feel free to open it again if you continue working on it again.

PR #1928, opened 2022-05-14
style: enable and baseline flakeheaven

aucampia:
# Summary of changes

This patch adds flakeheaven with a baseline, the baseline ignores all
existing errors and only reports on new errors.

The benefit with this approach is that we can get full flake8 on new
changes without having to first address the flake8 warnings with our
existing codebase.

Baselined errors can still be seen with:
- `flakeheaven lint --baseline /dev/null`
- `flake8`

This also adds a flakeheaven to GitHub Actions in a way which will
result in independent reporting.

Other changes:
- Simplified caching behaviour in GitHub Actions
- Removed flake8-black because it does not work well

  https://github.com/peterjc/flake8-black/issues/59
- Remove redundant/duplicated config for flake8

See also:
- https://github.com/RDFLib/rdflib/discussions/1927

# Checklist

  the same change.


gjhiggins (approved):

gjhiggins (approved):
still good :smile:

.github/workflows/validate.yaml
```diff
@@ -81,6 +75,36 @@ jobs:
...
+        - task: "gha:flake8"
+          python-version: 3.8
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{env.DEFAULT_PYTHON}}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Cache pip
+        uses: actions/cache@v2
```
cclauss, line 93:
```suggestion
        uses: actions/cache@v3
```

.github/workflows/validate.yaml
```diff
@@ -81,6 +75,36 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - task: "gha:flake8"
+          python-version: 3.8
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{env.DEFAULT_PYTHON}}
+        uses: actions/setup-python@v2
```
cclauss, line 89:
```suggestion
        uses: actions/setup-python@v3
```

.github/workflows/validate.yaml
```diff
@@ -81,6 +75,36 @@ jobs:
...
+  extra-tasks:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - task: "gha:flake8"
+          python-version: 3.8
+    steps:
+      - uses: actions/checkout@v2
```
cclauss, line 87:
```suggestion
      - uses: actions/checkout@v3
```

requirements.dev.txt
```diff
@@ -11,3 +9,4 @@ pytest-cov
 pytest-subtests
 types-setuptools
 -r docs/sphinx-requirements.txt
+-r requirements.flake8.txt
```
cclauss, line 12:
```suggestion
-r requirements-flake8.txt
```

cclauss (approved):

requirements.dev.txt
aucampia: While I agree that `requirements-???.txt` is more conventional, I will leave this for now so it is in line with the rest of the files, I will maybe make a separate PR to change them all when I can.

aucampia:
Thanks for the review and suggestions @cclauss.

PR #1944, opened 2022-05-15
Logging exceptions from Literal value converters

mwatts15:
# Summary of changes

Adds logging at warning level for exceptions raised when trying to convert a Literal's lexical form into Python-value space. Also refactored to not have `convFunc` be a boolean -- I think it should resolve the type-checking issue ~~, but I don't know how to run the type checker~~. 

# Checklist

  the same change.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -1962,6 +1962,8 @@ _GenericPythonToXSDRules: List[
     (Fraction, (None, _OWL_RATIONAL)),
 ]
 
+_OriginalGenericPythonToXSDRules = list(_GenericPythonToXSDRules)
+
 _SpecificPythonToXSDRules: List[
     Tuple[Tuple[Type[Any], str], Optional[Callable[[Any], Union[str, bytes]]]]
 ] = [
@@ -1972,6 +1974,8 @@ _SpecificPythonToXSDRules: List[
     ((str, _XSD_B64BINARY), b64encode),
     ((bytes, _XSD_B64BINARY), b64encode),
 ]
+
+_OriginalSpecificPythonToXSDRules = list(_SpecificPythonToXSDRules)
 
 XSDToPython: Dict[Optional[str], Optional[Callable[[str], Any]]] = {
     None: None,  # plain literals map directly to value space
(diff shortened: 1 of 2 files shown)


aucampia:
Some options for running type checking:


With docker compose:

```
docker compose build
docker compose run --rm devcontainer mypy --show-error-context --show-error-codes
```

With tox - this will run mypy and unit tests after:

```
tox -e py37
```

With [go-task](https://taskfile.dev/#/) in a virtual environment:

```
task WITH_VENV=1 venv:install validate:static
```

rdflib/term.py
```diff
@@ -1858,30 +1862,50 @@ def _castPythonToLiteral(
 _toPythonMapping.update(XSDToPython)
 
 
+def reset_bindings() -> None:
```
aucampia, line 1865: Is there a need for this to be public? Would somewhat prefer to make it private if possible.

mwatts15: sure, why not

mwatts15:
@aucampia what's the policy on updating function/variable names per the "extra-tasks" job? They're pre-existing names, so I didn't see fit to update them.

aucampia:
> @aucampia what's the policy on updating function/variable names per the "extra-tasks" job? They're pre-existing names, so I didn't see fit to update them.

It is okay if you don't for now, flakeheaven is the best option we have for gradually getting the codebase to pass with flake8, it is not perfect though, as the baseline kind of gets confused when a file is edited, so sometimes you get errors like you saw now. I will fix them after merging, the most important thing is that we want to catch new code that causes flake8 errors.

mwatts15:
win 3.10 failure is an unrelated socket error. could someone re-run?

aucampia:
> win 3.10 failure is an unrelated socket error. could someone re-run?

Will do, just waiting for others to complete.

aucampia (approved):
Your change looks good, I'm a bit on the fence of using `logging.warning` though, it has the benefit over `warnings.warn` that you can see the stack trace, but it has the downside that it is less easy for users to handle programmatically.

The rest of the change makes some much needed improvement.

On a related note, you may be interested in the recently added https://rdflib.readthedocs.io/en/latest/apidocs/rdflib.html#rdflib.term.Literal.ill_formed

Which is basically what this warning will be for:

https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal

> Otherwise, the literal is ill-typed and no literal value can be associated with the literal. Such a case produces a semantic inconsistency but is not syntactically ill-formed. Implementations must accept ill-typed literals and produce RDF graphs from them. Implementations may produce warnings when encountering ill-typed literals.

mwatts15:
@aucampia thanks. I hit this because I didn't have html5lib installed and parsing for the HTML literal failed (probably should have mentioned in the description...)

Regarding warnings vs logging, main thing I wanted was a stack trace. I tend to see logs from other modules even when I don't see warnings -- not exactly conclusive, but it's pretty consistent for the applications I've worked on.

aucampia:
> Regarding warnings vs logging, main thing I wanted was a stack trace. I tend to see logs from other modules even when I don't see warnings -- not exactly conclusive, but it's pretty consistent for the applications I've worked on.

I think this is reasonable, without a stack trace the warnings are of limited use, and while getting stack traces from `warnings.warn` is technically possible, it certainly is not convenient.

gjhiggins (approved):

PR #1959, opened 2022-05-19
Fix for issue1957 sparql parser percent encoded reserved chars

gjhiggins:
# Summary of changes
Seems like [`_hexExpand`](https://github.com/RDFLib/rdflib/blob/6ed2ef48ed38679bcdafe7cae250a2ef4b315e7b/rdflib/plugins/sparql/parser.py#L230) internal SPARQL parser function inappropriately expands [percent-encoded reserved characters](https://en.wikipedia.org/wiki/Percent-encoding). Added an exclusionary regexp to disable this behaviour and a parameterized test which checks SPARQL parser processing of the set of percent-encoded reserved chars

# Checklist

### DIFF
--- a/rdflib/plugins/sparql/parser.py
+++ b/rdflib/plugins/sparql/parser.py
@@ -227,8 +227,17 @@ PN_LOCAL = Regex(
 )
 
 
+percent_encoded_reserved_chars_re = re.compile(
+    """%2[0-13-9a-cfA-CF]|%3[abdfABDF]|%40|%5[bdBD]"""
+)
+
+
 def _hexExpand(match):
-    return chr(int(match.group(0)[1:], 16))
+    return (
+        match.group(0)
+        if percent_encoded_reserved_chars_re.match(match.group(0))
+        else chr(int(match.group(0)[1:], 16))
+    )
 
 
 PN_LOCAL.setParseAction(lambda x: re.sub("(%s)" % PERCENT_re, _hexExpand, x[0]))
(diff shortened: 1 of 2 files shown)


gjhiggins:
pre-commit.ci autofix

rdflib/plugins/sparql/parser.py
```diff
@@ -227,8 +227,17 @@ def expandCollection(terms):
 )
 
 
+percent_encoded_reserved_chars_re = re.compile(
+    """%2[0-13-9a-cfA-CF]|%3[abdfABDF]|%40|%5[bdBD]"""
+)
+
+
 def _hexExpand(match):
```
aucampia, line 235: I was thinking about this a bit - should we even be doing `_hexExpand` at all on URIs?

aucampia: Will do some testing on it.

aucampia: I tested this now, works fine if I just remove all of this:

https://github.com/RDFLib/rdflib/blob/ba855c233310afee65abd67f2f77239f909cd895/rdflib/plugins/sparql/parser.py#L230-L234

And I'm not sure if there is a good reason to not remove it.

Will make a commit to your PR to remove this and also move the tests to `test/test_sparql/test_prefixed_name.py` together with the addition of some more tests.

gjhiggins: np.

aucampia: Done now.

aucampia:
@gjhiggins  made a commit on your branch:

> Remove unnecessary hex expansion for PN_LOCAL in SPARQL parser
> 
> There is no need to expand percentage encoded parts of PN_LOCAL, as they
> should just be directly be used in the resulting IRI, and escaping them
> just risks creating an invalid IRI.
> 
> Also:
> - Move tests to `test/test_sparql/test_prefixed_name.py`
> - Add some more general tests for prefixed names.

aucampia (approved):
LGTM, good to have this fixed.

PR #1961, opened 2022-05-21
test: convert `test/jsonld/test_util.py` to pytest

aucampia:
# Summary of changes

Not the most straight forward conversion, but it simplifies it somewhat.

Main reason for doing this is to remove the use of unittest.subTest as
we need `pytest-subtest` to process that with pytest and that dependency
is causing some complications in dependency solving for pip sometimes.

Cleaner diff can be obtained by hiding whitespaces: https://github.com/RDFLib/rdflib/pull/1961/files?diff=split&w=1

# Checklist

  the same change.

### DIFF
--- a/test/jsonld/test_util.py
+++ b/test/jsonld/test_util.py
@@ -1,89 +1,74 @@
-import unittest
-from typing import NamedTuple
+import pytest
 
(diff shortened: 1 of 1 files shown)


gjhiggins (approved):
lgtm

PR #1966, opened 2022-05-21
Fix trix parser

gjhiggins:
# Summary of changes
Changed TriX parser to allow `trix` and `TriX`

The RDFLib TriX parser currently only accepts TriX documents conforming to the the [Nokia-published XSD](https://web.archive.org/web/20040821093939/http://swdev.nokia.com/trix/trix-1.0.xsd) which specifies (the mixed-case) `TriX`, contradicting the [W3C-published XSD spec for TriX](https://www.w3.org/2004/03/trix/trix-1/trix-1.0.xsd) which specifies (the lower-case) `trix`. We should accept both.

We were also a bit light on a TriX test suite for exercising the parser, so I recruited some TriX test fixtures from NG4J and Jena and fleshed out the TriX test suite, using the W3C "Manifest"-style approach.

# Checklist


test/test_w3c_spec/test_trix_w3c.py
```diff
@@ -0,0 +1,129 @@
+"""This runs the nt tests for the W3C RDF Working Group's N-Quads
+test suite."""
```
aucampia, line 2:
```suggestion
```

test/data/suites/trix/README
```diff
@@ -0,0 +1,41 @@
+This README is taken from the README of the W3C RDF Working Group's
```
aucampia, line 1: I think the references in here to w3c should be removed and mentions of TriG should be replaced with TriX

test/test_w3c_spec/test_trix_w3c.py
```diff
@@ -0,0 +1,129 @@
+"""This runs the nt tests for the W3C RDF Working Group's N-Quads
```
aucampia, line 1: I think maybe this should rather go into `test/test_suites/test_trix.py` rather than `test/test_w3c_spec/test_trix_w3c.py` as this is not a w3c suite as far as I understand.

aucampia (commented):
Fix looks good and nice to have the additional tests, just some minor notes on naming.

test/test_w3c_spec/test_trix_w3c.py
gjhiggins: I changed it to:
`"""This runs the TriX tests for RDFLib's informally-assembled TriX
test suite."""
`
might as well be explicit.

test/test_w3c_spec/test_trix_w3c.py
gjhiggins: Sure.

aucampia (approved):
Looks good to me.

test/test_w3c_spec/test_trix_w3c.py
aucampia: Happy with the new name.

test/data/suites/trix/README
aucampia: README looks good now.

aucampia:
Will merge by 2022-05-25 if there is no further feedback. This is a very simple fix.

Issue #812, opened 2018-02-22
rdflib.tools.graphisomorphism doesn't work


Aaronepower:
Hello, I was trying to use `rdflib.tools.graphisomorphism` and it repeatedly gives the following error when trying to use it from either the command line or as a module and directly using `IsomorphicTestableGraph`.

**Python Version:** 3.6
**System Version:** macOS Sierra/10.13.3

```
INFO:rdflib:RDFLib Version: 4.2.2
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/site-packages/rdflib/tools/graphisomorphism.py", line 127, in <module>
    main()
  File "/usr/local/lib/python3.6/site-packages/rdflib/tools/graphisomorphism.py", line 117, in main
    fn, format=options.inputFormat)
  File "/usr/local/lib/python3.6/site-packages/rdflib/graph.py", line 1043, in parse
    parser.parse(source, self, **args)
  File "/usr/local/lib/python3.6/site-packages/rdflib/plugins/parsers/nt.py", line 26, in parse
    parser.parse(f)
  File "/usr/local/lib/python3.6/site-packages/rdflib/plugins/parsers/ntriples.py", line 140, in parse
    self.parseline()
  File "/usr/local/lib/python3.6/site-packages/rdflib/plugins/parsers/ntriples.py", line 200, in parseline
    self.sink.triple(subject, predicate, object)
  File "/usr/local/lib/python3.6/site-packages/rdflib/plugins/parsers/nt.py", line 12, in triple
    self.graph.add((s, p, o))
  File "/usr/local/lib/python3.6/site-packages/rdflib/graph.py", line 397, in add
    self.__store.add((s, p, o), self, quoted=False)
  File "/usr/local/lib/python3.6/site-packages/rdflib/plugins/memory.py", line 248, in add
    self.__all_contexts.add(context)
TypeError: unhashable type: 'IsomorphicTestableGraph'
```

ajnelson-nist:
I have encountered the same issue.

Python Version: 3.8
[...]

PR #1968, opened 2022-05-22
add chunk serializer & tests

nicholascar:
# Summary of changes

This file provides a single function `serialize_in_chunks()` which can serialize a 
Graph into a number of NT files with a maximum number of triples or maximum file size.

There is an option to preserve any prefixes declared for the original graph in the first
file, which will be a Turtle file.

# Checklist

  the same change.


aucampia:
I will have a look at the windows test failures on Monday.

nicholascar:
@gjhiggins @ashleysommer @aucampia what do you think of the approach here? 

The need to chunk serialize files is a small one - a project I'm working on needs it - and I thought it interesting enough to make an RDFLib tool for, rather than just keeping the code within the project. 

I've tried to be efficient in terms of memory usage - no duplicate graph objects etc. - and to faithfully serialize the graph but there may be smarter approaches.

aucampia:
pre-commit.ci autofix

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+__all__ = ["serialize_in_chunks"]
+
+
+def serialize_in_chunks(
+    g: Graph,
+    max_triples: int = 10000,
+    max_file_size_kb: int = None,
+    file_name_stem: str = "chunk",
+    output_dir: Path = Path(__file__).parent,
+    first_file_contains_prefixes: bool = False,
```
aucampia, line 29: I think it may be best to just have this:

```suggestion
    write_prefixes: bool = False,
```

And then use `_prefixes.ttl` instead for the file name.

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+):
+    """
+    Serializes a given Graph into a series of n-triples with a given length
+
+    max_file_size_kb:
```
aucampia, line 34: One challenge with using kB in the naming is that it is a bit ambiguous in whether this is 1000 bytes or 1024 bytes. It may be slightly less ambiguous to just use size in bytes, but bytes would be too small for most cases.

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+            return "%s@%s" % (encoded, l_.language)
+        elif l_.datatype:
+            return "%s^^<%s>" % (encoded, l_.datatype)
+        else:
+            return "%s" % encoded
+
+    def _quote_encode(l_):
+        return '"%s"' % l_.replace("\\", "\\\\").replace("\n", "\\n").replace(
+            '"', '\\"'
+        ).replace("\r", "\\r")
```
aucampia, line 90: Is there any differences between these are the functions in https://github.com/RDFLib/rdflib/blob/958b9a11129f79523d76524802629f34ec649198/rdflib/plugins/serializers/nt.py#L56-L87

maybe it is best to just re-use those.

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+
+            fh.write(_nt_row(t))
+    else:
+        # count the triples in the graph
+        graph_length = len(g)
+
+        if graph_length <= max_triples:
+            # the graph is less than max so just NT serialize the whole thing
+            g.serialize(
+                destination=Path(output_dir) / f"{file_name_stem}_all.nt", format="nt"
```
aucampia, line 125: It may be easier on consumers to name the file as follow:

```suggestion
                destination=Path(output_dir) / f"{file_name_stem}_000000.nt", format="nt"
```

aucampia: Actually, maybe not,  `_all` carries some additional semantic meaning which may be helpful.

test/test_tools/test_chunk_serializer.py
```diff
@@ -0,0 +1,67 @@
...
+    from rdflib.tools.chunk_serializer import serialize_in_chunks
+except Exception:
+    px = str(Path(__file__).absolute().parent.parent.parent)
+    import sys
+
+    sys.path.insert(0, px)
+    from test.data import TEST_DATA_DIR
+
+    from rdflib import Graph
+    from rdflib.tools.chunk_serializer import serialize_in_chunks
```
aucampia, line 18: Not that this hurts exactly, but it should not be needed, how do you run test locally? Maybe I could add something to `Taskfile.yml` that could alleviate the problems you experience here.

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+
+__all__ = ["serialize_in_chunks"]
+
+
+def serialize_in_chunks(
+    g: Graph,
+    max_triples: int = 10000,
+    max_file_size_kb: int = None,
+    file_name_stem: str = "chunk",
+    output_dir: Path = Path(__file__).parent,
```
aucampia, line 28: It may be best to rather default to `Path.cwd()`

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+    sys.path.insert(0, str(Path(__file__).parent.parent.parent.absolute()))
+    from rdflib import Graph, Literal
+
+__all__ = ["serialize_in_chunks"]
+
+
+def serialize_in_chunks(
+    g: Graph,
+    max_triples: int = 10000,
+    max_file_size_kb: int = None,
```
aucampia, line 26:
```suggestion
    max_file_size_kb: Optional[int] = None,
```

aucampia, line 26: What should happen if someone specifies both `max_triples` and `max_file_size_kb`?

aucampia:
> I've tried to be efficient in terms of memory usage - no duplicate graph objects etc. - and to faithfully serialize the graph but there may be smarter approaches.

I think it is a fairly reasonable approach. It would have been nice if we had some way to do stream serialization to some text sink, but that is quite a big change and should probably be done with an abundance of caution and this seems like a reasonable approach in the interim.

I will maybe add some more tests on your branch if that is okay with you. Also happy to address comments I made, just let me know on the comment if you agree or disagree.

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+    file_name_stem: str = "chunk",
+    output_dir: Path = Path(__file__).parent,
+    first_file_contains_prefixes: bool = False,
+):
+    """
+    Serializes a given Graph into a series of n-triples with a given length
+
+    max_file_size_kb:
+        Maximum size per NT file in MB
```
aucampia, line 35:
```suggestion
        Maximum size per NT file in kB
```

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,136 @@
...
+        Equivalent to ~6,000 triples
+
+    max_triples:
+        Maximum size per NT file in triples
+        Equivalent to lines in file
+
+    file_name_stem:
+        Prefix of each file name
+        e.g. "chunk" = chunk_001.nt, chunk_002.nt...
```
aucampia, line 44:
```suggestion
        e.g. "chunk" = chunk_000001.nt, chunk_000002.nt...
```

gjhiggins:
> @gjhiggins @ashleysommer @aucampia what do you think of the approach here?

Minimally, it should indicate clearly that it is restricted to `Graph` serialization because, as the test below shows, context information is not preserved:

```py
@pytest.mark.xfail(reason="Context information not preserved")
def test_chunking_of_conjunctivegraph():
    nquads = """\
<http://example.org/alice> <http://purl.org/dc/terms/publisher> "Alice" .
<http://example.org/bob> <http://purl.org/dc/terms/publisher> "Bob" .
_:harry <http://purl.org/dc/terms/publisher> "Harry" .
_:harry <http://xmlns.com/foaf/0.1/name> "Harry" _:harry .
_:harry <http://xmlns.com/foaf/0.1/mbox> <mailto:harry@work.example.org> _:harry .
_:alice <http://xmlns.com/foaf/0.1/name> "Alice" <http://example.org/alice> .
_:alice <http://xmlns.com/foaf/0.1/mbox> <mailto:alice@work.example.org> <http://example.org/alice> .
_:bob <http://xmlns.com/foaf/0.1/name> "Bob" <http://example.org/bob> .
_:bob <http://xmlns.com/foaf/0.1/mbox> <mailto:bob@oldcorp.example.org> <http://example.org/bob> .
_:bob <http://xmlns.com/foaf/0.1/knows> _:alice <http://example.org/bob> ."""
    g = ConjunctiveGraph()
    g.parse(data=nquads, format="nquads")

    # make a temp dir to work with
    temp_dir_path = Path(tempfile.TemporaryDirectory().name)
    Path(temp_dir_path).mkdir()

    # serialize into chunks file with 100 triples each
    serialize_in_chunks(
        g, max_triples=100, file_name_stem="chunk_100", output_dir=temp_dir_path
    )

    # check, when a graph is made from the chunk files, it's isomorphic with original
    g2 = ConjunctiveGraph()
    for f in Path(temp_dir_path).glob("*.nt"):
        g2.parse(f, format="nt")

    assert len(list(g.contexts())) == len(list(g2.contexts()))
```

> The need to chunk serialize files is a small one - a project I'm working on needs it - and I thought it interesting enough to make an RDFLib tool for, rather than just keeping the code within the project.

RDFLib has traditionally been ambivalent about what's perceived as core vs non-core. Additional functionality appears to inevitably accrete, up to a point where it gets migrated out *en masse* into [a separate package](https://rdfextras.readthedocs.io/), the contents of which gradually become obsolete as they either fall out of use or are subsequently integrated into core library functionality.

Additional non-core functionality does have a regrettable tendency to languish in an untended and unkempt state. For instance, there's [`tools/graphisomorphism.py`](https://github.com/RDFLib/rdflib/blob/6ed2ef48ed38679bcdafe7cae250a2ef4b315e7b/rdflib/tools/graphisomorphism.py) which is
1. currently broken (and [has been since 2018](https://github.com/RDFLib/rdflib/issues/812))
2. long-obsolete, refererring as it does to `RDFa` as a supported format and based on [Sean B. Palmers's 2004 `rdfdiff.py` implementation](https://web.archive.org/web/20060522193057/http://www.w3.org/2001/sw/DataAccess/proto-tests/tools/)
3. Is subject to the same triples-only limitation.
4. Is obsoleted in functionality by both [`rdflib.compare.isomorphic`](https://github.com/RDFLib/rdflib/blob/a3a46114ccbeec873b77a8f746bfc0282a8545f0/rdflib/compare.py#L544) and the weaker [`Graph.isomorphic()`](https://github.com/RDFLib/rdflib/blob/a3a46114ccbeec873b77a8f746bfc0282a8545f0/rdflib/graph.py#L1382)

Is it even worth bothering with a relatively trivial fix/update ...
```diff
diff --git a/rdflib/tools/graphisomorphism.py b/rdflib/tools/graphisomorphism.py
index 004b567b..75462eb9 100644
--- a/rdflib/tools/graphisomorphism.py
+++ b/rdflib/tools/graphisomorphism.py
@@ -27,6 +27,10 @@ class IsomorphicTestableGraph(Graph):
         """
         return hash(tuple(sorted(self.hashtriples())))
 
+    def __hash__(self): 
+        # return hash(tuple(sorted(self.hashtriples())))
+        return self.internal_hash()
+
     def hashtriples(self):
         for triple in self:
             g = ((isinstance(t, BNode) and self.vhash(t)) or t for t in triple)
@@ -49,19 +53,19 @@ class IsomorphicTestableGraph(Graph):
             else:
                 yield self.vhash(triple[p], done=True)
 
-    def __eq__(self, G):
+    def __eq__(self, g):
         """Graph isomorphism testing."""
-        if not isinstance(G, IsomorphicTestableGraph):
+        if not isinstance(g, IsomorphicTestableGraph):
             return False
-        elif len(self) != len(G):
+        elif len(self) != len(g):
             return False
-        elif list.__eq__(list(self), list(G)):
+        elif list.__eq__(list(self), list(g)):
             return True  # @@
-        return self.internal_hash() == G.internal_hash()
+        return self.internal_hash() == g.internal_hash()
 
-    def __ne__(self, G):
+    def __ne__(self, g):
         """Negative graph isomorphism testing."""
-        return not self.__eq__(G)
+        return not self.__eq__(g)
 
 
 def main():
@@ -82,10 +86,10 @@ def main():
         default="xml",
         dest="inputFormat",
         metavar="RDF_FORMAT",
-        choices=["xml", "trix", "n3", "nt", "rdfa"],
+        choices=["xml", "n3", "nt", "turtle", "trix", "trig", "nquads", "json-ld", "hext"],
         help="The format of the RDF document(s) to compare"
-        + "One of 'xml','n3','trix', 'nt', "
-        + "or 'rdfa'.  The default is %default",
+        + "One of 'xml', 'turtle', 'n3', 'nt', 'trix', 'trig', 'nquads', 'json-ld'"
+        + "or 'hext'.  The default is %default",
     )
 
     (options, args) = op.parse_args()
```

when its appearance in `tools` is unlikely to persist for much longer?

There are a few non-core contributions in the closed PRs which I'm recruiting for preservaition in the cookbook. I'm guessing that a command-line version of graphisomorphism will ultimately end up there.

aucampia:
NOTE: I still have not had time to look at windows issue, will try tomorrow.

aucampia:
@nicholascar made some changes to your branch:

- Use bytes written as size instead of using `os.path.size()`.
  The second option here is very dependent on OS behaviour and what is
  on disk.

- Add all open files to an exit stack so they are closed by the time the
  function returns.

- Set encoding explicitly to utf-8 on opened files.

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _quote_encode(l_):
+        return '"%s"' % l_.replace("\\", "\\\\").replace("\n", "\\n").replace(
+            '"', '\\"'
+        ).replace("\r", "\\r")
+
+    @contextmanager
+    def _start_new_file(file_no):
+        fp = Path(output_dir) / f"{file_name_stem}_{str(file_no).zfill(6)}.nt"
+        with open(fp, "a", encoding="utf-8") as fh:
```
aucampia, line 95: Should this not rather overwrite than append?
```suggestion
        with open(fp, "w", encoding="utf-8") as fh:
```

rdflib/tools/chunk_serializer.py
```diff
@@ -0,0 +1,145 @@
...
+
+    bytes_written = 0
+    with ExitStack() as xstack:
+        if max_file_size_kb is not None:
+            file_no = 1 if first_file_contains_prefixes else 0
+            for i, t in enumerate(g.triples((None, None, None))):
+                if i == 0:
+                    fp, fh = xstack.enter_context(_start_new_file(file_no))
+                    bytes_written = 0
+                elif bytes_written >= max_file_size_kb * 1000:
```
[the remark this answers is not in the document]
aucampia: Actually I fixed this in latest commit to your branch.

aucampia:
Another commit:

- Very that writing triple won't exceed max file size before writing
  instead of after writing.

  This also necesitates using binary mode for file IO so that an
  accurate byte count can be obtained.

test/test_tools/test_chunk_serializer.py
nicholascar: I'm not good with pytest and path issues, so any improvements you can make here... I accept!

rdflib/tools/chunk_serializer.py
nicholascar: I know what you mean: I wrestled with B, kB, MB etc. I think kB is "ok", as in there are no perfect solutions and it's not likely critical what the value is, e.g. 1000 or 1024, so long as it's approximate.

rdflib/tools/chunk_serializer.py
nicholascar: I was tossing up having all functions self-contained, like now, and importing them. I guess it is better to import really, so I will change this.

rdflib/tools/chunk_serializer.py
nicholascar: Good idea, thanks, adopted.

rdflib/tools/chunk_serializer.py
nicholascar: Done

rdflib/tools/chunk_serializer.py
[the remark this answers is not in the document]
nicholascar: Thanks

rdflib/tools/chunk_serializer.py
nicholascar: max_file_size_kb wins! Added note about this to docco

rdflib/tools/chunk_serializer.py
nicholascar: Fixed

rdflib/tools/chunk_serializer.py
nicholascar: Done

aucampia:
@nicholascar will finish this up in W32

aucampia:
pre-commit.ci autofix

rdflib/tools/chunk_serializer.py
aucampia: changing this to "w"

rdflib/tools/chunk_serializer.py
aucampia: Changing this now.

aucampia: Done

test/test_tools/test_chunk_serializer.py
aucampia: Changed this now

rdflib/tools/chunk_serializer.py
aucampia: Changed

aucampia (approved):
I closed all remaining open comments, added some more tests and changed to using functions from `rdflib.plugins.serializers.nt`.

I think this is good to merge now.

aucampia:
> I think this is good to merge now.

I will merge this later this week if there is no further feedback.

PR #1976, opened 2022-05-30
escape literal * symbol in `rdflib.paths` docs

EFord36:
# Summary of changes

Currently, there are unescaped `*`s in the rdflib.paths docs, which get interpreted as italics markup in ReStructured Text when generating the docs.

This PR escapes them so that the `*` characters are visible and the text isn't in italics.

# Checklist

  the same change.

# more details

The original reStructured Text in the `__doc__` for the module includes (see [here](https://github.com/RDFLib/rdflib/blob/4ded2eb380e15ed7babbf8a632fc502d75cae040/rdflib/paths.py#L67)):

```
Modifiers (?, *, +) are done using * (the multiplication operator) and
the strings '*', '?', '+', also defined as constants in this file.
```

But in the rendered html, the literal `*`s are missing, and the text is italicised between where two `*`s should be. See the red box below:

Screenshot 2022-05-30 at 14 42 07

This PR escapes them so that the `*` characters are visible and the text isn't in italics:

Screenshot 2022-05-30 at 14 45 40

Happy to discuss the fix more if necessary even though it's a small change since:
* There are multiple ways to escape/include literal `*` characters in Restructured text. Notably ` ``*`` ` also gives the desired effect. I chose to use backslash since this is used elsewhere in the rdflib docs to escape things, e.g. in the same file [here](https://github.com/RDFLib/rdflib/blob/4ded2eb380e15ed7babbf8a632fc502d75cae040/rdflib/paths.py#L32)
* There are other unescaped `*`, for example [here](https://github.com/RDFLib/rdflib/blob/4ded2eb380e15ed7babbf8a632fc502d75cae040/rdflib/paths.py#L22) in the same file, but it looks like because they don't occur in a pair, ReStructured Text doesn't interpret them as italics markers, so the desired end behaviour in the html render is fine. Unclear if it's still worth checking through the docs and escaping all literal `*`s.
* My PR doesn't do anything to address this re-occuring in future. I don't think there's a 'rdflib apidoc styleguide' or anything that it belongs in as human-readable, and I don't think an automatic check makes sense here since someone could want to use `*`s to italicise something in the docs, so it really needs a human to check/catch this.

After making the change, I generated the tests locally and checked the output to generate the 'after' screenshot above. I think this is all the testing needed for this change, but let me know if I should do anything else, or need to do anything related to contributing/releasing - sorry if I've missed a step here!

# Fly-by thank you for RDF

Thank you for RDF, it's a great tool that is incredibly useful in my daily job. It's one of my favourite python packages in terms of how easy it is to use and the effort it saves me and the docs are great. The `rdflib.paths` feature is also a really cool feature that I only came across relatively recently (hence the PR).

### DIFF
--- a/rdflib/paths.py
+++ b/rdflib/paths.py
@@ -64,8 +64,8 @@ Path(http://xmlns.com/foaf/0.1/knows / http://xmlns.com/foaf/0.1/name)
 >>> FOAF.name|FOAF.givenName
 Path(http://xmlns.com/foaf/0.1/name | http://xmlns.com/foaf/0.1/givenName)
 
-Modifiers (?, *, +) are done using * (the multiplication operator) and
-the strings '*', '?', '+', also defined as constants in this file.
+Modifiers (?, \*, +) are done using \* (the multiplication operator) and
+the strings '\*', '?', '+', also defined as constants in this file.
 
 >>> FOAF.knows*OneOrMore
 Path(http://xmlns.com/foaf/0.1/knows+)


aucampia (approved):
LGTM, thanks for the fix.

gjhiggins (approved):
a fine eye for detail, thank you.

PR #1977, opened 2022-05-30
correct italic markup in plugin stores docs

EFord36:
# Summary of changes

Currently the doc file plugin_stores.rst uses `_` in a way that appears to be intended to produce italics in the final html render, which doesn't appear to be valid ReST markup for italics. It is Markdown markup for italics (hence me reading the intention of italics in the presence of the `_`s).

I have changed this to use (unescaped) `*`s for the italics, which works as seemed to be the desired behaviour.

# Checklist

  the same change.

# More Detail

Discovered this while working on PR #1976 , looking at other places where there might be that issue, but found this related one instead.

Before (i.e. live docs [here](https://rdflib.readthedocs.io/en/stable/plugin_stores.html?highlight=HDT#external)):

Screenshot 2022-05-30 at 15 31 42

After (local build of docs after change):

Screenshot 2022-05-30 at 15 33 33

Similar caveats to PR #1976:

* My PR doesn't do anything to address this re-occuring in future. I don't think there's a 'rdflib apidoc styleguide' or anything that it belongs in as human-readable, and I don't think an automatic check makes sense here since someone could want to use *s to italicise something in the docs, so it really needs a human to check/catch this.
* After making the change, I generated the tests locally and checked the output to generate the 'after' screenshot above. I think this is all the testing needed for this change, but let me know if I should do anything else, or need to do anything related to contributing/releasing - sorry if I've missed a step here!

### DIFF
--- a/docs/plugin_stores.rst
+++ b/docs/plugin_stores.rst
@@ -38,7 +38,7 @@ HDT               `<https://github.com/RDFLib/rdflib-hdt>`_            A Store b
 Oxigraph          `<https://github.com/oxigraph/oxrdflib>`_            Works with the `Pyoxigraph <https://oxigraph.org/pyoxigraph>`_ Python graph database library
 ================= ==================================================== =============================================================================================
 
-_If you have, or know of a Store implementation and would like it listed here, please submit a Pull Request!_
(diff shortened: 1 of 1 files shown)


aucampia (approved):
LGMT, thanks for the correction.

PR #1979, opened 2022-06-04
Fix SPARQL CSV result serialization of blank nodes

aucampia:
# Summary of changes

This patch fixes SPARQL CSV result serialization of blank nodes.
Previously blank nodes were serialized the same as normal nodes, but
blank nodes should be prefixed with `_:` as per
<https://www.w3.org/TR/sparql11-results-csv-tsv/#csv-terms>:

> Blank nodes use the `_:label` form from Turtle and SPARQL.
> Use of the same label indicates the same blank node within
> the results but has no significance outside the results.

# Checklist

  the same change.


gjhiggins (approved):
lgtm

PR #1983, opened 2022-06-04
Add more tests for graph_diff

aucampia:
# Summary of changes

This is to confirm that `graph_diff` does not work with quads.

Other changes:
- Added support to GraphHelper for collapsing BNodes to a specifc IRI
  when coverting to triple sets, this is in addition to the current
  functionality for ignoring bnodes. This makes it possible to more
  accurately compare Graphs that contain blank nodes.
- Added `assert_cgraph_isomorphic` to GraphHelper.
  This method adds some rudementary support for checking if conjunctive
  graphs are isomorphic, it will ignore named graphs that have blank
  nodes as name.

# Checklist

  the same change.

### DIFF
--- a/test/test_graph/test_diff.py
+++ b/test/test_graph/test_diff.py
@@ -1,12 +1,23 @@
-from test.utils import GraphHelper
-from typing import TYPE_CHECKING, Set
+from dataclasses import dataclass, field
+from test.utils import (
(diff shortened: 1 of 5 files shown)


gjhiggins (approved):
very necessary

PR #1986, opened 2022-06-05
Rework turtle-like test suites

aucampia:
This patch reworks the turtle-like test suites (N-Quads, N-Triples,
Turtle, TriG). The changes includes:

- A new DAWG Manifest processor:
  - Incorporates pytest expected failure support in test parameter generation.
  - Supports URI mapping so that local filesystem URIs can be mapped to the
    correct remote base URIs and still work correctly with local files.
  - Supports custom entry types so that suite specific handling can be
    implemented separately from the core manifest processor. This will
    be used for SPARQL test suite and JSON test suite processing.
- Updated EARL reporter:
  - Support for the new DAWG Manifest processor.
  - Writing of EARL reports by default if a report prefix is
    specified in the DAWG Manifest processor.
  - Reports output is sorted so that it can be compared with text based
    diff tools.
  - Fixed reporting of xfail tests. These were reported as skipped before,
    now they are reported as failed. 

The testing is also more accurate, and now tests which were reported as
passing before are now correctly reported as failing. With one exception
all these failures are related to leniency in the parser, as our parser
accepts input that should raise an error when parsing.

Other changes:
- Moved IRI related test utilities to a separate module.
- Added generated reports for turtle-like formats to git with `-HEAD`
  suffix.

Subsequent changes will use this new processor for SPARQL and RDF/XML,
the draft for these can be seen in https://github.com/RDFLib/rdflib/pull/1969.

# Summary of changes

# Checklist

  the same change.

### DIFF
--- a/test/README.rst
+++ b/test/README.rst
@@ -50,56 +50,56 @@ EARL test reports can be generated using the EARL reporter plugin from ``earl.py
 
 When this plugin is enabled it will create an ``earl:Assertion`` for every test that has a ``rdf_test_uri`` parameter which can be either a string or an ``URIRef``.
 
-To enable the EARL reporter plugin an output file path must be supplied to pytest with ``--earl-report``. The report will be written to this location in turtle format.
+To enable the EARL reporter plugin an output file path must be supplied to pytest with ``--earl-output-file``. The report will be written to this location in turtle format.
 
 Some examples of generating test reports:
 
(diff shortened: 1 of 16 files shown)


gjhiggins (approved):
impressive, much-needed improvement to the coherence of the RDFLib test suite

PR #1991, opened 2022-06-10
Add fragment property to URIRef

ottokruse:
# Summary of changes

Added a property `fragment` to `URIRef` so it is easy to access the URI's fragment:

```python
>>> URIRef("http://example.com/some/path/#some-fragment").fragment
'some-fragment'
>>> URIRef("http://example.com/some/path/").fragment
''
```

## Why is this useful?

The non-fragment part of the URI often just serves a namespace purpose, e.g. `https://brickschema.org/schema/Brick#`.

Actual "things" use that namespace, and have the "thing" name as the fragment, e.g. `https://brickschema.org/schema/Brick#Valve_Position_Sensor`

For display purposes, it's nice to get to the "thing" name, i.e. the fragment easily.

TODO: update docs and tests. I'll do that if you deem this PR is useful.

# Checklist

  the same change.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -313,6 +313,18 @@ class URIRef(IdentifiedNode):
             return URIRef(url)
         else:
             return self
+    
+    @property
+    def fragment(self) -> str:
+        """
+        Return the URL Fragment
+        
+        >>> URIRef("http://example.com/some/path/#some-fragment").fragment
+        'some-fragment'
+        >>> URIRef("http://example.com/some/path/").fragment
+        ''
+        """
+        return urlparse(self).fragment
 
     def __reduce__(self) -> Tuple[Type["URIRef"], Tuple[str]]:
         return (URIRef, (str(self),))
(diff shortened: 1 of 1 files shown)


nicholascar (approved):
I'm ok with this without tests as it's just passing on an existing function from urlparse().

I try to never use hash URIs with fragments but if you do use them, I can see this might be nice.

gjhiggins (approved):
yup, straightforward enough to merge.

aucampia:
pre-commit.ci autofix

gjhiggins:
> yup, straightforward enough to merge.

aaaand ... nice work, btw :smile:

aucampia (commented):
While this is useful, it is also a bit arbitrary, as we don't have scheme, netloc, path, params or query. But I think on balance it does not really do any harm to include it, and we can just deprecate it if it does cause problems down the line.

aucampia:
@ottokruse thanks for the PR, we plan to make the next release around the middle of July.

PR #1995, opened 2022-06-20
rebase of Alex Dutton's Draft PR #1385

gjhiggins:
# Summary of changes

Rebase of Alex Dutton's Draft PR #1385, adding a custom resolver to fix issue #1844.

> This adds a new Resolver framework, and updates the JSON-LD parser to use it.
> As a result, implementors have much more control over which URLs referenced in JSON-LD documents will be resolved.
> This PR implements a default-deny approach to URL resolution, which will ***need explaining in release notes and documentation***.
> Implementors can revert to the old behaviour via environment variable if they are happy with the risks.
> This also provides a base for future extensibility, e.g. caching or using locally-held copies of external resources.
> No tests or documentation yet, but all existing tests pass when using the new PermissiveResolver.

I'm unsure whether this is "just" a bugfix or whether it qualifies as a break in the API because even when using the PermissiveResolver, [`test_parse_file_guess_format.py`](https://github.com/RDFLib/rdflib/compare/master...gjhiggins:fix-issue1844-resolver?expand=1#diff-fc44cfdf16cebd34a168aa917bb4beca371379e52ec0d70997f3f3b3548a92e2) fails which suggests that other, currently-working code might be similarly impacted.

I added a minimal test, needs expanding.

# Checklist

### DIFF
--- a/rdflib/exceptions.py
+++ b/rdflib/exceptions.py
@@ -5,6 +5,8 @@ TODO:
 __all__ = [
     "Error",
     "ParserError",
+    "ResolutionError",
+    "ResolutionForbiddenError",
 ]
 
 
@@ -25,6 +27,31 @@ class ParserError(Error):
 
     def __str__(self):
         return self.msg
+
+
+class ResolutionError(Error):
+    """Resolution error.
+
+    This includes the case when the URI scheme is not supported.
+    """
+
+    def __init__(self, msg):
+        Error.__init__(self, msg)
+        self.msg = msg
+
+    def __str__(self):
+        return self.msg
+
+
+class ResolutionForbiddenError(Error):
+    """Resolution of the given URL is not allowed."""
+
+    def __init__(self, url):
+        Error.__init__(self, self.__doc__)
+        self.url = url
+
+    def __str__(self):
+        return f"Resolution of '{self.url}' is not allowed."
 
 
 class UniquenessError(Error):
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1327,6 +1327,8 @@ class Graph(Node):
         parser = plugin.get(format, Parser)()
         try:
             # TODO FIXME: Parser.parse should have **kwargs argument.
+            if format not in ["application/ld+json", "json-ld"] and "resolver" in args:
+                del args["resolver"]
             parser.parse(source, self, **args)
         except SyntaxError as se:
             if could_not_guess_format:
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1327,8 +1327,6 @@ class Graph(Node):
         parser = plugin.get(format, Parser)()
         try:
             # TODO FIXME: Parser.parse should have **kwargs argument.
-            if format not in ["application/ld+json", "json-ld"] and "resolver" in args:
-                del args["resolver"]
             parser.parse(source, self, **args)
         except SyntaxError as se:
             if could_not_guess_format:
--- a/rdflib/plugins/shared/jsonld/context.py
+++ b/rdflib/plugins/shared/jsonld/context.py
@@ -6,7 +6,8 @@ Implementation of the JSON-LD Context structure. See:
 
 """
 # https://github.com/RDFLib/rdflib-jsonld/blob/feature/json-ld-1.1/rdflib_jsonld/context.py
-
+import contextlib
+import json
 from collections import namedtuple
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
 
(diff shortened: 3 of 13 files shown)


aucampia:
Working on this now.

rdflib/graph.py
```diff
@@ -1327,6 +1327,8 @@ def parse(
         parser = plugin.get(format, Parser)()
         try:
             # TODO FIXME: Parser.parse should have **kwargs argument.
+            if format not in ["application/ld+json", "json-ld"] and "resolver" in args:
+                del args["resolver"]
```
aucampia, line 1331: I don't like this, as the kwargs passed to parse should not be given any special meaning by RDFLib, they are entirely for the plugins to use. While this may have some design flaws it is the way it works now and this breaks that.

aucampia: I'm going to remove this.

rdflib/exceptions.py
```diff
@@ -27,6 +29,31 @@ def __str__(self):
...
+
+class ResolutionForbiddenError(Error):
+    """Resolution of the given URL is not allowed."""
+
+    def __init__(self, url):
+        Error.__init__(self, self.__doc__)
+        self.url = url
+
+    def __str__(self):
+        return f"Resolution of '{self.url}' is not allowed."
```
aucampia, line 54: These should inherit from ValueError also I think so I will change it to do that.

rdflib/resolver.py
```diff
@@ -0,0 +1,234 @@
...
+
+    @url_resolver(schemes={"file"})
+    def resolve_file(self, url: str, format: str, scheme: str) -> InputSource:
+        filename = url2pathname(url.replace("file:///", "/"))
+        file = open(filename, "rb")
+        return FileInputSource(file)
+
+    @url_resolver(schemes={"http", "https"})
+    def resolve_http(self, url: str, format: str, scheme: str) -> InputSource:
+        return URLInputSource(url, format)
```
aucampia, line 144: I don't see a good reason why this should be part of the core resolver, if someone wants to implement some deferral logic for different schemes they can do it in their own code, having one method for resolving seems much simpler.

rdflib/resolver.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+DEFAULT_RESOLVABLE_URL_ALLOWLIST = {
+    "https://w3c.github.io/json-ld-rc/context.jsonld",
+}
+
+_unspecified = object()
+
+
+class Resolver:
```
aucampia, line 84: There should be an abstract base class for this so someone can implement it without all the baggage from the default implementation.

rdflib/resolver.py
```diff
@@ -0,0 +1,234 @@
...
+    resolvable_url_schemes = os.environ.get("RDFLIB_RESOLVABLE_URL_SCHEMES", "").split()
+    resolvable_url_allowlist = os.environ.get(
+        "RDFLIB_RESOLVABLE_URL_ALLOWLIST",
+        " ".join(DEFAULT_RESOLVABLE_URL_ALLOWLIST),
+    ).split()
+
+    def __init__(
+        self,
+        resolvable_url_schemes=_unspecified,
+        resolvable_url_allowlist=_unspecified,
```
aucampia, line 163: Why use _unspecified instead of None?

rdflib/resolver.py
```diff
@@ -0,0 +1,234 @@
...
+                try:
+                    mod_name, cls_name = os.environ[
+                        "RDFLIB_DEFAULT_RESOLVER_CLASS"
+                    ].split(":")
+                except KeyError:
+                    _default_resolver = DefaultResolver()
+                else:
+                    mod = importlib.import_module(mod_name)
+                    resolver = getattr(mod, cls_name)
+                    assert isinstance(resolver, Resolver)
```
aucampia, line 231: Should be issubclass instead of isinstance, which makes me fairly sure this code is not being tested, type checker does complain about it, but this code is a bit anemic on the typing.

rdflib/plugins/shared/jsonld/context.py
```diff
@@ -437,7 +441,9 @@ def _fetch_context(self, source, base, referenced_contexts):
         if source_url in self._context_cache:
             return self._context_cache[source_url]
 
-        source = source_to_json(source_url)
+        input_source = self.resolver.resolve(source_url, format="json-ld")
+        with contextlib.closing(input_source.getByteStream()) as stream:
+            source = json.load(stream)
```
aucampia, line 446: This will end poorly, resolve returns a tuple, this seems to expect an input_source

aucampia:
@gjhiggins not sure if you want to keep this open. I think if the original author of the PR is willing to take this to completion he should, but otherwise this can basically go into a footnote into the issue it is trying to address.

gjhiggins:
> @gjhiggins not sure if you want to keep this open. I think if the original author of the PR is willing to take this to completion he should, but otherwise this can basically go into a footnote into the issue it is trying to address.

Quite relaxed about letting it lie. I originally took my cue from the discussion about the original still-open PR (with the the aim of addressing the majority of lingering open PRs) and rebased it in order to be able to check it with current master.

PR #2000, opened 2022-06-23
Update test README

aucampia:
# Summary of changes

This removes the test suite specific commands as output is now written
by default to the right files.

Other changes:
- Write EARL report to file suffixed with `-HEAD` without timestamps by
  default. This is to make it easier to keep the checked in test reports
  accurate.
- Remove XFAILs for tests that are now passing.

# Checklist

  the same change.


gjhiggins (approved):
Sterling work. Nice to see this for RDFLib

Issue #1873, opened 2022-04-29
SPARQL variable assignment fails instead of producing unbound variable


jindrichmynarz:
Using the SPARQL query engine of rdflib 6.1.1, when an assignment's operand is an error (e.g., running a function on an unbound variable), an error is thrown, for example:

```py
from rdflib import Graph

[r for r in Graph().query('SELECT (concat("", sha1(?x)) AS ?y) WHERE {}')]

# AttributeError: 'NotBoundError' object has no attribute 'datatype'
```

However, the [SPARQL 1.1 specification](https://www.w3.org/TR/sparql11-query/#assignment) expects that the assigned variable would remain unbound in such case:

> If the evaluation of the expression produces an error, the variable remains unbound for that solution but the query evaluation continues.

This is also what Apache Jena 4.4.0 does.

mielvds:
THis might be related: 

```
BIND ('https://data.hetarchief.be/id/organisatie/' AS ?ns)
```
results into `AttributeError: 'SPARQLError' object has no attribute 'datatype'`

```
BIND (URI('https://data.hetarchief.be/id/organisatie/') AS ?ns)
``` 
results into `AttributeError: 'URIRef' object has no attribute 'datatype'`

gjhiggins:
Hmm, perhaps this just needs a guard.

aucampia:
- https://github.com/RDFLib/rdflib/pull/2013 should have fixed this and it was included in RDFLib 6.2.0

Closing this issue but if it still persists please re-open it.

jindrichmynarz:
Thanks! I can no longer reproduce the above issue when using rdflib version 6.2.0.

PR #2013, opened 2022-07-07
Fix for #1873 avoid AttributeError raised ...

gjhiggins:
from attempting to access language and datatype attributes of non-Literal objects

Fix for #1873 ... 
> “Using the SPARQL query engine of rdflib 6.1.1, when an assignment's operand is an error (e.g., running a function on an unbound variable), an error is thrown, for example: # AttributeError: 'NotBoundError' object has no attribute 'datatype'. However, the [SPARQL 1.1 specification](https://www.w3.org/TR/sparql11-query/#assignment) expects that the assigned variable would remain unbound in such case:”

# Summary of changes

Added `isinstance` check to avoid AttributeErrors being raised by attempting to access `language` and `datatype`attribs of non-Literal objects being passed back up the call stack resulting from SPARQL BIND failures/errors.

# Checklist

### DIFF
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -298,10 +298,10 @@ def Builtin_CONCAT(expr, ctx):
 
     # dt/lang passed on only if they all match
 
-    dt = set(x.datatype for x in expr.arg)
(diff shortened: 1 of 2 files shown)


aucampia (approved):

niklasl (approved):
Looks good to me!

Issue #1880, opened 2022-05-03
Addition of `override` argument to `rdflib.store.Store.bind` breaks compatibility


aucampia:
See:
- https://github.com/RDFLib/rdflib/pull/1843#issuecomment-1116074731
  > Um, the change to `Store.bind()` breaks the public API, store plugins (such as [oxrdflib](https://github.com/oxigraph/oxrdflib/blob/67c460897207d2de316d2b94255053ee6f538cc7/oxrdflib/__init__.py#L117), [rdflib-leveldb](https://github.com/RDFLib/rdflib-leveldb/blob/a0f3386c71e6b1cfbd09c257e29400f5fde43ed3/rdflib_leveldb/leveldbstore.py#L466), [rdflib-kyotocabinet](https://github.com/RDFLib/rdflib-kyotocabinet/blob/c65bf64f278078b9fb6d1242b5dadced77d3ef28/rdflib_kyotocabinet/KyotoCabinet.py#L333) and [rdflib-sqlachemy](https://github.com/RDFLib/rdflib-sqlalchemy/blob/d4c057934cd2675083d3df943103bdffb20341d4/rdflib_sqlalchemy/store.py#L637)) are throwing exceptions on "`bind` got an unexpected keyword 'override'".

aucampia:
So the way I see it, the addition of "override" fixed a bug, but to fix the bug we need the interface of bind on stores to accept the argument.

I think the most reasonable solution here is to call store bind via this adaptor method:

https://github.com/RDFLib/rdflib/blob/98292415600ec912ce9816bfe882eeb6fb678947/rdflib/namespace/__init__.py#L621-L626

This way, if the underlying store does not support the override argument we just call Store.bind without it, and then the bug will still be present.

I'm still testing this and will try refine it a bit so that it does not catch unrelated errors and I will improve the log message.

CC: @RDFLib/core-reviewers

PR #2018, opened 2022-07-13
Fallback to old `Store.bind` signature on `TypeError`

aucampia:
# Summary of changes

If `Store.bind` raises a `TypeError`, and the string conversion of this
TypeError contains `override`, then log a warning and call `Store.bind`
without an override.

This is done so that stores that do not accept `override` on
`Store.bind` still work, but at the cost of still having the bug that
was fixed by introducing the `override` parameter.

Also added a private flag which can be used to disable the fix entirely
and never use `override` when calling `Store.bind`.

- Fixes https://github.com/RDFLib/rdflib/issues/1880

# Checklist

  the same change.


gjhiggins (approved):
nice, a great HOW-TO example

PR #2032, opened 2022-07-17
build: Taskfile improvements

aucampia:
# Summary of changes

- Added individual tasks for black, isort and mypy to make it easier to
  run these directly.
- Change flag variables to be JSON based so that they can be disabled
  with "false", this makes it easier to work with tooling that does not
  distinguish between empty and unset environment variables.
- Add a `venv:run` target to make it easier to run things inside the
  venv, e.g. `task venv:run -- pip install lxml`.
- Change pytest invocation to use the pytest script instead of using
  `python -m pytest` as the latter does some things which are not
  desirable.

# Checklist

  the same change.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -73,7 +73,7 @@ jobs:
           task \
             TOX_EXTRA_COMMAND="${{ matrix.TOX_EXTRA_COMMAND }}" \
             OS=${{ matrix.os }} \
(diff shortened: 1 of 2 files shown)


aucampia:
I'm going to update this to also change how pytest is invoked to address this comment: https://github.com/RDFLib/rdflib/issues/1519#issuecomment-1186887630

PR #2033, opened 2022-07-17
Infixowl cleanup iii

gjhiggins:
Maintenance-positive changes recommended by flake8-bugbear, mostly explicitly indicating unused loop variables with a prefixing underscore (it identiified one case where _none_ of the loop variables were being used) and a couple of don't-use-mutable-objects-as-args instances.

# Summary of changes

1. Added leading underscore to unused loop variables, removed mutable objects from args, (replacing with None and then performing an explicit “if None” assignment in the body of the method).

2. Swapped out an excessively weak, coverage-motivated length test of __hash__() (which was causing intermittent test failures) for an equality test of two runs.

# Checklist


aucampia (approved):
LGMT, also updated the CHANGELOG.

PR #2041, opened 2022-07-19
fix: import xml.sax.handler from the right place

aucampia:
# Summary of changes

Change the import of `xml.sax.handler` in the TriX parser so that it
imports from `xml.sax` and not from `xml.sax.saxutils`. Importing from
`xml.sax.saxutils` causes mypy to fail but it is also wrong as there is
no documented `handler` in `xml.sax.saxutils`.

# Checklist

  the same change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,22 @@ and will be removed for release.
 <!-- -->
 <!-- -->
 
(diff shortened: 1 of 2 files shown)


gjhiggins (approved):
lgtm

Issue #2035, opened 2022-07-18
6.2.0: pytest is failing


kloczek:
I'm packaging your module as an rpm package so I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix>
- run pytest with PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>

kloczek:
Here is pytest output 
[python-rdflib-pytest-log.txt](https://github.com/RDFLib/rdflib/files/9128854/python-rdflib-pytest-log.txt)

aucampia:
@kloczek this is a known issue: https://github.com/RDFLib/rdflib/issues/1847

To fix it, don't run tests with lxml installed.

aucampia:
RDFLib has a bug which is tracked in https://github.com/RDFLib/rdflib/issues/1847 - and because of this bug, using lxml with XML based SPARQL results fail.

We should fix the bug, but we should also mark the tests in our test suite that are known to fail under some conditions with xfail under those conditions, so that the test suite still passes.

For example:

https://github.com/RDFLib/rdflib/blob/d2630095bf242f7dd3a678d864b479e97e36de5a/test/test_w3c_spec/test_n3_w3c.py#L117-L119

This test is known to fail on windows, so we mark it with xfail on windows.

Similarly, with lxml installed, the failing tests here should have an xfail marker.

kloczek:
Thank you very much for the explanation 😄

Issue #1847, opened 2022-04-18
SPARQL result parsing does not work with lxml


aucampia:
thanks for @edmondchuc for reporting this, this came up when he was working on integrating pyld in https://github.com/RDFLib/rdflib/pull/1836 - pyld pulls in lxml and with lxml some tests fail that pass otherwise:

https://github.com/RDFLib/rdflib/runs/6061070121?check_suite_focus=true#step:11:674

```
  =========================== short test summary info ============================
  FAILED test/test_sparql/test_result.py::test_select_result_serialize_parse[xml-None-utf-8]
  FAILED test/test_sparql/test_result.py::test_select_result_serialize_parse[xml-TEXT_IO-utf-8]
  FAILED test/test_sparql/test_result.py::test_select_result_serialize_parse[xml-BINARY_IO-utf-8]
  FAILED test/test_sparql/test_result.py::test_select_result_serialize_parse[xml-STR_PATH-utf-8]
  FAILED test/test_sparql/test_result.py::test_select_result_parse_serialized[xml-TEXT_IO-utf-8]
  = 5 failed, 5115 passed, 2 skipped, 115 xfailed, 54 warnings in 129.80s (0:02:09) =
```

Exceptions all occur inside `XMLResultParser.parse`:

```
__________________________________________________________ test_select_result_serialize_parse[xml-BINARY_IO-utf-8] __________________________________________________________
Traceback (most recent call last):
  File "/home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/test_sparql/test_result.py", line 318, in test_select_result_serialize_parse
    check_serialized(format.name, select_result, serialized_data)
  File "/home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/test_sparql/test_result.py", line 110, in check_serialized
    parsed_result = Result.parse(StringIO(data), format=format)
  File "/home/iwana/sw/d/github.com/iafork/rdflib.cleanish/rdflib/query.py", line 215, in parse
    return parser.parse(source, content_type=content_type, **kwargs)
  File "/home/iwana/sw/d/github.com/iafork/rdflib.cleanish/rdflib/plugins/sparql/results/xmlresults.py", line 31, in parse
    return XMLResult(source)
  File "/home/iwana/sw/d/github.com/iafork/rdflib.cleanish/rdflib/plugins/sparql/results/xmlresults.py", line 40, in __init__
    tree = etree.parse(source, parser)
[...]

PR #2044, opened 2022-07-21
fix: SPARQL XML result parsing

aucampia:
# Summary of changes

Fixed the following problems with the SPARQL XML result parsing:

- Both the parse method of both the lxml and `xml` modules does not work
  work well with `TextIO` objects, the `xml` module works with `TextIO`
  objects if the XML encoding is `utf-8`, but not if it is `utf-16`, and
  with `lxml` parse fails for both `utf-8` and `utf-16`. To fix this I
  changed the XML result parser to first convert `TextIO` to `bytes` and
  then feed the `bytes` to `parse()` using `BytesIO`.
- The parser was operating on all elements inside `results` and
  `result` elements, even if those elements were not `result` and `binding`
  elements respectively. This was causing problems with `lxml`, as `lxml`
  also returns comments when iterating over elements. To fix this I added
  a check for the element tags so that only the correct elements are
  considered.

Other changes:

- Added type hints to `rdflib.plugins.sparql.results.xmlresults`.
- Run with `lxml` one some permutations in the test matrix.
- Removed `rdflib.compat.etree`, as this was not very helpful for the
  SPARQL XML Result parser and it was not used elsewhere.
- Added an `lxml` environment to tox which installs `lxml` and
  `lxml-stubs`.
- Expanded SPARQL result testing by adding some additional
  parameters.

Related issues:

- Fixes https://github.com/RDFLib/rdflib/issues/2035
- Fixes https://github.com/RDFLib/rdflib/issues/1847

# Checklist

  the same change.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -33,6 +33,7 @@ jobs:
           - python-version: "3.9"
             os: ubuntu-latest
             TOX_EXTRA_COMMAND: "- black --check --diff ./rdflib"
+            TOXENV_SUFFIX: "-lxml"
           - python-version: "3.10"
             os: ubuntu-latest
             TOX_EXTRA_COMMAND: "flake8 --exit-zero rdflib"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,31 @@ and will be removed for release.
 <!-- -->
 <!-- -->
 
+
+<!-- -->
+<!-- -->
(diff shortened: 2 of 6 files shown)


gjhiggins (approved):
lgtm

aucampia:
Just a heads up, I'm considering this somewhat of a last ditch effort to save lxml integration. Seeing as lxml integration was broken for a long time I don't think there is anyone relying on it, and if it is causing more problems and nobody is interested in maintaining it I will remove it.

PR #2046, opened 2022-07-21
Toplevel n80x

gjhiggins:
Handful of fixes for PEP8 naming convention violations.

# Summary of changes

Changes to argument names not impacting the API, `noqa` pragmas added for convention violations that appear in the published API. Added deprecation to `paths.evalPath()`

# Checklist

### DIFF
--- a/rdflib/compat.py
+++ b/rdflib/compat.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
     import xml.etree.ElementTree as etree
 else:
     try:
-        from lxml import etree  # n0qa: N813
+        from lxml import etree  # noqa: N813
     except ImportError:
         import xml.etree.ElementTree as etree  # noqa: N813 F401
 
--- a/rdflib/compat.py
+++ b/rdflib/compat.py
@@ -12,9 +12,9 @@ if TYPE_CHECKING:
     import xml.etree.ElementTree as etree
 else:
     try:
-        from lxml import etree  # nqqa: N813
+        from lxml import etree  # n0qa: N813
     except ImportError:
-        import xml.etree.ElementTree as etree  # noqa: N813 F401
+        import xml.etree.ElementTree as etree  # noqa: N813
 
 
 def cast_bytes(s, enc="utf-8"):
--- a/rdflib/compat.py
+++ b/rdflib/compat.py
@@ -12,9 +12,9 @@ if TYPE_CHECKING:
     import xml.etree.ElementTree as etree
 else:
     try:
-        from lxml import etree
+        from lxml import etree  # nqqa: N813
     except ImportError:
-        import xml.etree.ElementTree as etree
+        import xml.etree.ElementTree as etree  # noqa: N813
 
 
 def cast_bytes(s, enc="utf-8"):
--- a/rdflib/compat.py
+++ b/rdflib/compat.py
@@ -14,7 +14,7 @@ else:
     try:
         from lxml import etree  # n0qa: N813
     except ImportError:
-        import xml.etree.ElementTree as etree  # noqa: N813
+        import xml.etree.ElementTree as etree  # noqa: N813 F401
 
 
 def cast_bytes(s, enc="utf-8"):
--- a/rdflib/compat.py
+++ b/rdflib/compat.py
@@ -14,7 +14,7 @@ else:
     try:
         from lxml import etree  # nqqa: N813
     except ImportError:
-        import xml.etree.ElementTree as etree  # noqa: N813
+        import xml.etree.ElementTree as etree  # noqa: N813 F401
 
 
 def cast_bytes(s, enc="utf-8"):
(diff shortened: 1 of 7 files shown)


rdflib/compat.py
```diff
@@ -12,9 +12,9 @@
     import xml.etree.ElementTree as etree
 else:
     try:
-        from lxml import etree
+        from lxml import etree  # nqqa: N813
```
aucampia, line 15:
```suggestion
        from lxml import etree  # noqa: N813
```

rdflib/paths.py
```diff
@@ -462,6 +462,17 @@ def path_sequence(self, other):
...
+        DeprecationWarning(
+            "rdflib.path.evalPath() is deprecated, use the (all-lowercase) evalpath(). "
+            "The mixed-case evalPath() function name is incompatible with PEP8 "
+            "recommendations and will be replaced by evalpath() in rdflib 7.0.0."
+        )
+    )
+    return evalpath(graph, t)
+
+
+def evalpath(graph, t):
```
aucampia, line 475: Given the naming of other functions like `mul_path`, `inv_path`, `neg_path` I think it may be better to go with `eval_path` instead.

Happy to change it in your branch if you agree.

rdflib/query.py
```diff
@@ -57,7 +59,7 @@ class update method.
     def __init__(self, graph):
         pass
 
-    def update(self, strOrQuery, initBindings={}, initNs={}):
+    def update(self, str_or_query, initBindings={}, initNs={}):  # noqa: N803
```
aucampia, line 62: I think maybe you intended to have:

```suggestion
    def update(self, strOrQuery, initBindings={}, initNs={}):  # noqa: N803
```

But even if not, I think it is best to keep the old `strOrQuery`.

Alternatively, we can retain it as a keyword only arg so that if someone passes it we can both log a deprecation warning and still do the right thing.

```suggestion
    def update(self, str_or_query = None, initBindings={}, initNs={}, *, strOrQuery = None):  # noqa: N803
```

But I think for now just keeping it as is with noqa is simplest.

gjhiggins: Ah-ha, even with the fix to `str_or_query`, the `# noqa: N803` is still required for the `initBindings` and `initNS` violations, I didn't spot my typos until _far_ to late :frowning_face:

rdflib/compat.py
```diff
@@ -12,9 +12,9 @@
     import xml.etree.ElementTree as etree
 else:
     try:
-        from lxml import etree
+        from lxml import etree  # n0qa: N813
```
aucampia, line 15:
```suggestion
        from lxml import etree  # noqa: N813
```

rdflib/query.py
```diff
@@ -37,7 +37,9 @@ class Processor(object):
     def __init__(self, graph):
         pass
 
-    def query(self, strOrQuery, initBindings={}, initNs={}, DEBUG=False):
+    def query(
+        self, str_or_query, initBindings={}, initNs={}, DEBUG=False  # noqa: N803
```
aucampia, line 41: I still would prefer we don't break our API just for this, or at least split the interface breaking into another PR.

```suggestion
        self, strOrQuery, initBindings={}, initNs={}, DEBUG=False  # noqa: N803
```

gjhiggins: I don't have a position on this but my model is letting me down here. If `strOrQuery` isn't a kwarg, how does changing it break the API?

aucampia: someone can be calling `Graph.query(strOrQuery="SELECT * WHERE { ?s ?p ?o }")` as by default all arguments can be used as keywords or positional arguments, and if they are then changing `strOrQuery` to `str_or_query` will cause their code to fail.

Since python 3.8 there is an option for making arguments positional only [[PEP-0570](https://peps.python.org/pep-0570/)], so for example `def query(self, str_or_query, /, initBindings={}, initNS={}, DEBUG=false)` - and with a positional only arg it is impossible to call it using a keyword, but without this all args can be used as keywords.

gjhiggins: > by default all arguments can be used as keywords or positional arguments
> 
> Since python 3.8 there is an option for making arguments positional only ... and with a positional only arg it is impossible to call it using a keyword, but without this all args can be used as keywords.

Oh, I didn't know that. Thanks for the explanation. Will revert it.

gjhiggins: Reverted.

rdflib/compat.py
gjhiggins: Aargh, thanks for spotting that.

rdflib/query.py
gjhiggins: Reverted change to snakecase.

aucampia (approved):
LGTM

aucampia:
Will merge tomorrow

PR #2047, opened 2022-07-21
Fix for Issue #948, reworked PR #1087 escaping turtle prefix names

gjhiggins:
# Summary of changes

Refined the candidate solution down to an additional on-exception fallback recovery attempt in the turtle serializer. If there's an existing namespace that starts the uri, use that to split it and escape the local part.

# Checklist

### DIFF
--- a/rdflib/plugins/serializers/turtle.py
+++ b/rdflib/plugins/serializers/turtle.py
@@ -284,6 +284,26 @@ class TurtleSerializer(RecursiveSerializer):
 
             if pfx is not None:
                 parts = (pfx, uri, "")
+
+            # Does the uri contain chars that need to be escaped?
+            elif len(self.namespaces) > 0:
+                for prefix, local in [
+                    (_prefix, uri.split(_namespace)[1])
+                    for _prefix, _namespace in self.namespaces.items()
+                    if uri.startswith(_namespace)
+                ]:
+                    return "{prefix}:{escaped_local}".format(
+                        prefix=prefix,
+                        escaped_local="".join(
+                            [
+                                f"\\{char}" if char in "_~.-!$&'()*+,;=/?#@%" else char
+                                for char in local
+                            ]
+                        ),
+                    )
+                # Was worth a try
+                return None
+
             else:
                 # nothing worked
                 return None
--- a/test/test_serializers/test_serializer_turtle.py
+++ b/test/test_serializers/test_serializer_turtle.py
@@ -113,3 +113,24 @@ def test_turtle_namespace():
     assert "GENO:0000385" in output
     assert "SERIAL:0167-6423" in output
     assert r"EX:name_with_\(parenthesis\)" in output
+
+
+def test_issue948_escapeturtleprefixname():
+    foo = Namespace("http://www.example.com/foo/")
+    g = Graph()
+    g.bind("foo", foo)
+    g.add((foo["subj"], foo["pred"], foo["obj"]))
+    g.add((foo["subj"], foo["pred"], foo["obj/bar"]))
+    g.add((foo["subj"], foo["pred"], foo["obj#boo"]))
+    g1 = Graph()
+    g1.bind("foo", foo)
+    g1.parse(data=g.serialize(format="turtle"), format="turtle")
+    expected = r"""@prefix foo: <http://www.example.com/foo/> .
+
+foo:subj foo:pred foo:obj,
+        foo:obj\#boo,
+        foo:obj\/bar .
+
+"""
+    assert g.serialize(format="turtle") == expected
+    assert g1.serialize(format="turtle") == expected


aucampia:
Just to clarify, AFAICT this does not fix any bug, it just makes RDFLib work harder to serialize using PNames instead of falling back to using URIs?

rdflib/plugins/serializers/turtle.py
```diff
@@ -284,6 +284,26 @@ def getQName(self, uri, gen_prefix=True):
 
             if pfx is not None:
                 parts = (pfx, uri, "")
+
```
aucampia, line 287: Ideally compute_qname (which is incorrectly named) should return the PrefixedName in components, and then all that should have to happen here is some standard escaping of the PN_LOCAL.

Not sure why this is not the case, I will debug a bit. Having compute_qname do what it should do is preferable to having similar logic in here.

aucampia: It will also make it easier to fix for other serializers, such as TriG and N3.

test/test_serializers/test_serializer_turtle.py
```diff
@@ -113,3 +113,24 @@ def test_turtle_namespace():
...
+    g1.parse(data=g.serialize(format="turtle"), format="turtle")
+    expected = r"""@prefix foo: <http://www.example.com/foo/> .
+
+foo:subj foo:pred foo:obj,
+        foo:obj\#boo,
+        foo:obj\/bar .
+
+"""
+    assert g.serialize(format="turtle") == expected
+    assert g1.serialize(format="turtle") == expected
```
aucampia, line 136: I was trying to add some more tests, and it seems this actually only works if there is more than 1 triple in the graph.

https://github.com/aucampia/rdflib/blob/4d31b1548340f513579354755e3c0bf5613d7f00/test/test_serializers/test_serializer_turtlelike.py#L18-L58

These tests fail if line 44 is removed.

aucampia: Looking into it further though.

rdflib/plugins/serializers/turtle.py
aucampia: The actual exception here is:

```
2022-07-22T01:23:42.590 DEBUG    root         turtle.py:283:getQName caught ...
  File "/home/iwana/sw/d/github.com/iafork/rdflib.reviews/rdflib/plugins/serializers/turtle.py", line 280, in getQName
    parts = self.store.compute_qname(uri, generate=gen_prefix)
  File "/home/iwana/sw/d/github.com/iafork/rdflib.reviews/rdflib/graph.py", line 1057, in compute_qname
    return self.namespace_manager.compute_qname(uri, generate)
  File "/home/iwana/sw/d/github.com/iafork/rdflib.reviews/rdflib/namespace/__init__.py", line 522, in compute_qname
    raise KeyError(
KeyError: 'No known prefix for http://www.example.com/foo/obj# and generate=False'
```

Which makes me think the real issue is not in this function but elsewhere.

aucampia:
I will try and fix `compute_qname` as I'm about 10% of the way through digging through what is going on there, but it will take some time as there are many functions involved with no tests and no documentation that has very weird specificity. My suspicion is that it would be simpler to write a compute_curie from scratch than to try and fix what is there, but I will see what I can do.

In general though while this enhancement would be nice, it is probably not really that critical.

test/test_serializers/test_serializer_turtle.py
aucampia: I'm fairly sure it is failing because the namespace is only in the list of checked namespaces the second time it is encountered. 

i.e. the namespace would only be added here:

https://github.com/RDFLib/rdflib/blob/1a60704c2ee6a3a1bf8a0824da5174037b87ed4e/rdflib/plugins/serializers/turtle.py#L319

I may be wrong though, but I'm generally not clear on why it is using a list of namespaces other than the list of namespaces provided by the namespace manager.

rdflib/plugins/serializers/turtle.py
```diff
@@ -284,6 +284,26 @@ def getQName(self, uri, gen_prefix=True):
...
+            # Does the uri contain chars that need to be escaped?
+            elif len(self.namespaces) > 0:
+                for prefix, local in [
+                    (_prefix, uri.split(_namespace)[1])
+                    for _prefix, _namespace in self.namespaces.items()
```
aucampia, line 292: Why is this using `self.namespaces` instead of `self.store.namespaces` ? I'm guessing if you do that it will work in all cases.

Even so, while I think that may fix the parameterizes tests I added, I still don't think it is ideal to have logic to compute a CURIE inside the recovery handler for the function that should be computing the CURIE (i.e. `compute_qname`).

I feel it is making this function more complicated than it has to be and already I'm a bit unsure of the complexity. To me for example it seems wrong that this function is doing `self.addNamespace(prefix, namespace)` - and I'm also unclear on why the serializer has `addNamespace` and a list of namespaces.

aucampia: If I change this to use `self.store.namespaces` it works fine with the parameterized tests I added in https://github.com/gjhiggins/rdflib/compare/fix-issue-948...aucampia:rdflib:iwana-20220721T2317-fix_turtle_serialization

gjhiggins:
Happy to close this without merging. I originally took my cue from the discussion about the original still-open PR (with the the aim of addressing the majority of lingering open PRs) and rebased it in order to be able to check it with current master.

Issue #1834, opened 2022-04-16
Make serialize() on a CONSTRUCT result act like normal g.serialize()


aucampia:
### Discussed in https://github.com/RDFLib/rdflib/discussions/1612

<div type='discussions-op-text'>

<sup>Originally posted by **nicholascar** June 11, 2021</sup>
Currently (RDFlib 6.0.0a0 AKA `master`), `g.serialize()` defaults to turtle format and is decoded, however calling `serialize()` on a SPARQL `CONSTRUCT` result (i.e. on a `Result` object) returns the RDFlib 5.0.0 form of an RDF/XML, UTF-8-encoded result.

The old style `serialize()` in `Result` should be updated to match the main `Graph()` `serialize()`.</div>

nicholascar:
I'm encountering issues related to this now. I can easily (two word change) make the `serialize()` return Turtle, not XML, by default, however the result object is still a `rdflib.plugins.sparql.processor.SPARQLResult`, not a `rdflib.graph.Graph` and I think it should be a `Graph`.

gjhiggins:
> I think it should be a `Graph`.

Indeed it should be, according to [the SPARQL 1.1 spec](https://www.w3.org/TR/sparql11-query/#construct) - “The CONSTRUCT query form returns a single RDF graph specified by a graph template.”

Priyansh19077:
We are interested in this issue and are currently working on having a way to get the turtle format in the returned results for CONSTRUCT query. We looked into the xmlresults.py file in the reults directory of sparql. We realised that the serialize() function is being called in that file to actually serialize those results. 

We are planning to have a way to convert the generated xml statements into turtle and then result that. Another way could be to have a separate file like turtleresults.py and then call the serialize function there. 

In case you have any ideas we can go forward with this, please do share them.

aucampia:
There was a PR for this before: https://github.com/RDFLib/rdflib/pull/1418

aucampia:
- While working on https://github.com/RDFLib/rdflib/issues/2067 I had a bit of a think about https://github.com/RDFLib/rdflib/pull/1418, the only breaking change is the default value for encoding, and we can just change this with a very small PR for version 7.0, so I will update that PR slightly to just not change the default encoding.

PR #2065, opened 2022-07-29
fix: issues with string destination handling in `{Graph,Result}.serialize`

aucampia:
# Summary of changes

Change `{Graph,Result}.serialize` to only handle string destinations as URIs if
the schema is `file` and to treat it as operating system paths in all
other cases.

This is for the following reasons:
- `Result.serialize` was treating URI paths as OS paths which only works
  for some cases, for special charachters and percentage encoding it
  does not work.
- Many valid Unix and Windows paths parse using `urlparse` and have no netloc, e.g. `C:\some\path`
  and `some:/path`, however they should be treated as paths and not as URIs.
- `Graph` and `Result` should behave consistently.

Some caveats in this change:
- non-file URIs will now be treated as OS paths which may result in
  slightly confusing error messages, such as
  `FileNotFoundError: [Errno 2] No such file or directory: 'http://example.com/'`
  if http://example.com/ is passed.
- some valid file URIs (e.g. `file:/path/to/file` from https://datatracker.ietf.org/doc/html/rfc8089)
  are also valid Posix paths but will be treated as file-URIs instead of
  Posix paths. For `Graph.serialize` this ambiguity can be avoided by
  using `pathlib.Path`, but for `Result.serialize` there is currently no
  way to avoid it, though I will work on https://github.com/RDFLib/rdflib/issues/1834
  soon and in that provide a way to avoid the ambiguity there also.

- Fixes https://github.com/RDFLib/rdflib/issues/2067

# Checklist

  the same change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,6 +159,28 @@ and will be removed for release.
 <!-- -->
 <!-- -->
 
+
+<!-- -->
+<!-- -->
+<!-- CHANGE BARRIER: START #2068 -->
+<!-- -->
+<!-- -->
+
+- Improve file-URI and path handling in `Graph.serialize` and `Result.serialize` to
+  address problems with windows path handling in `Result.serialize` and to make
+  the behavior between `Graph.serialize` and `Result.serialie` more consistent.
+  Closed [issue #2067](https://github.com/RDFLib/rdflib/issues/2067).
+  [PR #2068](https://github.com/RDFLib/rdflib/pull/2068).
+  - String values for the `destination` argument will now only be treated as
+    file URIs if `urllib.parse.urlparse` returns their schema as `file`.
+  - Simplified file writing to avoid a temporary file.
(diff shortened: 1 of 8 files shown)


aucampia:
Not sure what is up with coverage, I'm guessing it is somehow related to the latest build on master failing (https://github.com/RDFLib/rdflib/actions?query=branch%3Amaster), I will rebase once https://github.com/RDFLib/rdflib/pull/2068 is merged.

aucampia:
> Not sure what is up with coverage, I'm guessing it is somehow related to the latest build on master failing (https://github.com/RDFLib/rdflib/actions?query=branch%3Amaster), I will rebase once #2068 is merged.

Nevermind, it is because I fiddled with tox.ini when I was trying to debug this.

gjhiggins (approved):
Huh, windows tax - paid, gj :+1:

PR #2073, opened 2022-08-03
Fix passing ConjunctiveGraph as namespace_manager

mielvds:
# Summary of changes

- passes the `NamespaceManager` of the `ConjunctiveGraph` as namespace_manager of the `Graph` created by the `get_context()` method instead of the `ConjunctiveGraph` object itself.

# Checklist

  the same change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ and will be removed for release.
 
  - Fixes passing `NamespaceManager` in `ConjunctiveGraph`'s method `get_context()`. 
    The `get_context()` method will now pass the `NamespaceManager` of `ConjunctiveGraph` to the `namespace_manager` attribute of the newly created context graph, instead of the `ConjunctiveGraph` object itself. This cleans up an old FIXME commment.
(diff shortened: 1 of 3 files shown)


aucampia (approved):
@mielvds thanks for the PR. While I think it is pretty safe and uncontroversial it would still be good to have tests for this.

Also, we are trying to reduce the effort it takes to make a release, and one part of this is to keep the changelog up to date always:

https://github.com/RDFLib/rdflib/blob/cfa418074b27b12aac905ba266b002a237c5ff4c/CHANGELOG.md?plain=1#L201-L217

So it would be good if you could add tests and update the changelog, but I will try find some time this weekend to add tests and update the change on your PR branch if it is not done yet, as this is rather important to fix.

aucampia:
@mielvds please share how you managed to trigger this if you can.

mielvds:
Hi @aucampia 
I agree. I will add some tests and update the changelog. You can then check if they suffice. 

I triggered this while developing a documentation tool for shacl. I manage different shacl graphs as named graphs and got errors when printing the namespaces.

aucampia (approved):
Thank you very much for the test and changelog update @mielvds, the change looks good.

PR #2076, opened 2022-08-07
fix: handling of Literal datatype

aucampia:
# Summary of changes

Check datatype against `None` instead of checking it's truthiness (i.e.
`if datatype is not None:` instead of `if datatype:`).

Checking truthiness instead of `is not None` causes a blank string to
be treated the same as None. The consequence of this was that
`Literal.datatype` could be a `str`, a `URIRef` or `None`, instead of
just a `URIRef` or `None` as was seemingly intended.

Other changes:
- Changed the type of `Literal.datatype` to be `Optional[URIRef]`
  instead of `Optional[str]` now that `str` will always be converted to
  `URIRef` even if it is a blank string.
- Changed `rdflib.util._coalesce` to make it easier and safer to use
  with a non-`None` default value.
- Added `rdflib.util._convert_optional` that makes it easy to convert
  optional values while retaining their optional nature.
- Changed `rdflib.util` to avoid issues with circular imports.

# Checklist

  the same change.

### DIFF
--- a/rdflib/util.py
+++ b/rdflib/util.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 """
 Some utility functions.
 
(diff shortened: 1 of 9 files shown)


rdflib/util.py
```diff
@@ -456,27 +466,68 @@ def get_tree(
...
 
 
+_InputT = TypeVar("_InputT")
+_OutputT = TypeVar("_OutputT")
+
+
+def _convert_optional(
+    converter: Callable[[_InputT], _OutputT],
+    input: Optional[_InputT],
+) -> Optional[_OutputT]:
```
aucampia, line 476: Actually I thought a bit about this, it is not worth the cost of a function call for this I think, so I will eliminate it, and it is easy enough to do do with `z = None if x is None else convert(x)`

aucampia: Done now.

gjhiggins (approved):
:+1:

PR #2085, opened 2022-08-19
fix: issue with trig reference counting across graphs

aucampia:
# Summary of changes

The TriG serializer was only considering BNode references inside a single
graph and not counting the BNodes subjects as references when considering if a
BNode should be serialized as unlabeled blank nodes (i.e. `[ ]`), and as a
result it was serializing BNodes as unlabeled if they were in fact referencing
BNodes in other graphs.

One caveat of this change is that some RDF Datasets may be serialized
less sussinctly in that it would not use unlabeled blank nodes where it is
technically possible. This can be trivially fixed, but a trivial fix
increases the compuational complexity of serialization significantly.

Other changes:
- Removed the roundtrip xfail that this change fixed.
- Added another roundtrip test which has various combinations of BNode
  references across graphs in a dataset, this test fails for JSON-LD
  however, so while this change removes one xfail it also now adds
  another.
- Set the default indent_size and style in .editorconfig as to avoid
  relying on undefined system defaults.

# Checklist

  the same change.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,6 +6,8 @@ root = true
 
 # Unix-style newlines with a newline ending every file
 [*]
+indent_size = 4
+indent_style = space
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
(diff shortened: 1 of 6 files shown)


gjhiggins (approved):
a-ha, well spotted

PR #2091, opened 2022-08-20
ci: publish test reports

aucampia:
# Summary of changes

Use the dorny/test-reporter GitHub Action to publish test reports for
both mypy and pytest.

This should make it easier to see what fails without having to dig
through the rather noisy console logs.

# Checklist

  the same change.

### DIFF
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -1,0 +1,27 @@
+name: "Test Report"
+on:
+  workflow_run:
+    workflows: ["Validate"]
+    types:
+      - completed
+jobs:
+  report:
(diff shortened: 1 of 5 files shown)


aucampia:
The reports will only start working once this is merged to master but for an example see: \
- PR: https://github.com/aucampia/rdflib/pull/40
- Test failure: https://github.com/aucampia/rdflib/runs/7932777048?check_suite_focus=true
- Test pass: 

One caveat is that xfails get reported as skips, but otherwise the output is useful for seeing what specific tests failed.

gjhiggins (approved):
nice touch

PR #2096, opened 2022-08-22
fix: Don't register berkelydb as a store if it is not available on the system

### LINKED ISSUE #1816: Don't register berkelydb as a store if it is not available on the system
Currently we are always registering BerkelyDB regardless of whether it is available or not:

https://github.com/RDFLib/rdflib/blob/cdaee27415bd669082d90c9b6bbff3191a147761/rdflib/plugin.py#L202-L207

This causes problems if we want to operate on all stores, for example to test some functionality, as a special exception is needed for BerkeleyDB:

https://github.com/RDFLib/rdflib/blob/13ec1df32bdb6a09de491ba2a931008c9dd9adea/test/test_graph/test_graph.py#L68-L84

I think it would be best to only register the plugin if the store is available.

Jtmcflea:
# Summary of changes

Fixes #1816
- Added berkeleydb check in plugin.py
- Removed exception in test_graph.py
- Removed `import rdflib.plugins.stores.berkeleydb.has_bsddb`

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -618,5 +618,3 @@ register(
     "rdflib.plugins.sparql.results.tsvresults",
     "TSVResultParser",
 )
-
-# if plugins whatever doesn't have bddb, unregister it
--- a/rdflib/plugin.py
+++ b/rdflib/plugin.py
@@ -180,6 +180,15 @@ def plugins(
 
 
 # Register Stores
+
+if rdflib.plugins.stores.berkeleydb.has_bsddb:
+    # Checks for BerkeleyDB before registering it
+    register(
+        "BerkeleyDB",
+        Store,
+        "rdflib.plugins.stores.berkeleydb",
+        "BerkeleyDB",
+    )
 register(
     "default",
     Store,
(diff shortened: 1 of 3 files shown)


Jtmcflea:
Whoops! Sorry about the unprofessional language and weird descriptions in the commit list! I'm still learning how github works and thought it would update differently.

aucampia:
> Whoops! Sorry about the unprofessional language and weird descriptions in the commit list! I'm still learning how github works and thought it would update differently.

No worries :) thanks for the PR, from a quick glance it looks good, will check a bit closer once CI has completed

rdflib/plugin.py
```diff
@@ -613,3 +618,5 @@ def plugins(
     "rdflib.plugins.sparql.results.tsvresults",
     "TSVResultParser",
 )
+
+# if plugins whatever doesn't have bddb, unregister it
```
aucampia, line 622:
```suggestion
```

Don't think this comment is needed.

aucampia (approved):
Change looks good to me, but maybe the comment at the end of `rdflib/plugin.py` is not needed.

Thanks again for the PR, I will include some note in the `CHANGELOG.md` and merge sometime in the next couple of days, probably before friday.

rdflib/plugin.py
Jtmcflea: oh whoops, forgot that was in there lol

aucampia: Removed it now.

aucampia:
I added a CHANGELOG entry to your branch, feel free to change it if you want, I will check it again before I merge it.

PR #2101, opened 2022-08-26
chore: rename default branch to `main`

aucampia:
# Summary of changes

We are still using a somewhat outmoded name for our default branch, this
PR captures the decision to change it as a decision record and prepares
the rest of the repo for this change.

Once merged, the main branch of RDFLib will be renamed using the
procedure documented [here](https://github.com/github/renaming).

Other changes:

- Add area for decision records (this may be merged in another PR before
  this one).
- Remove duplicated sections from `docs/upgrade4to5.rst` as they cause
  problems when using `sphinx.ext.autosectionlabel` and serve no
  purpose.

I am not particularly invested in renaming `master`, so happy to abandon if there are objections, but I think on balance it is the right thing to do.

# Checklist

  the same change.


white-gecko (approved):
These are many points to rename the branch, thank you for doing this. In generall I like this idea.

This pr also adds the decision restored mechanism, which I like. Also the it includes cleanup in 4to5 doc which is ok for me with the reason given.

For me this looks good.

gjhiggins (approved):
thorough approach, lgtm

Issue #2011, opened 2022-07-07
SPARQL Object Lists and Predicate-Object Lists produce different results


mielvds:
Hi all,

We notice some incorrect parsing wrt. object lists and predicate object lists.

The following query produces no results:

```
PREFIX source: <https://data.hetarchief.be/ns/source#>

SELECT ?val1 ?val2
WHERE {
    ?o source:custom_fields [
            source:value        ?val1 ;
         ], [
           source:value        ?val2 ;
         ] .
}
```

But this one does:

```
PREFIX source: <https://data.hetarchief.be/ns/source#>

SELECT ?val1 ?val2
WHERE {
    ?o source:custom_fields [
           source:value        ?val1 ;
         ];
         source:custom_fields [
           source:value        ?val2 ;
         ] .
}
```

According to the spec, or at least how I interpret it, these two queries should be semantically identical. Pyoxigraph returns results for both.

PR #2107, opened 2022-09-02
Add test for issue #2011

mielvds:
This PR adds an xfail test to reproduce issue #2011

### DIFF
--- a/test/test_sparql/test_sparql.py
+++ b/test/test_sparql/test_sparql.py
@@ -1,5 +1,6 @@
 import logging
 from test.utils import eq_
+from test.utils.result import assert_bindings_collections_equal
 from typing import Any, Callable, Mapping, Sequence, Type
 
 import pytest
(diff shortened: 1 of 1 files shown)


aucampia:
pre-commit.ci autofix

aucampia:
thanks for the PR @mielvds, I will double check the spec tomorrow and confirm with jena and then merge.

test/test_sparql/test_sparql.py
```diff
@@ -90,6 +90,36 @@ def test_sparql_bnodelist():
...
+
+    """
+
+    g = Graph()
+    g.parse(
+        data="""
+    @prefix : <urn:ns1:> .
+    :s :p [ :v 1 ], [ :v 2].
+    """,
+        format="n3",
```
aucampia, line 107:
```suggestion
        format="n3",
```

If possible we should avoid using n3 in tests unless absolutely needed, in part because it is not standardized, and further it is also not exactly RDF, N3 allows encoding things that cannot be represented in any other RDF formats, and that is not in line with RDF abstract syntax.

aucampia: Fixed this with some other minor things.

aucampia:
The test looks good and seems valid to me also, I made some minor changes to your branch in fc1c859 and will merge it now.

PR #2112, opened 2022-09-09
add charset encoding to SPARQLConnector.update() request.

### LINKED ISSUE #2095: SPARQLConnector.update should set ContentType charset
I use rdflib to connect to a Blazegraph triplestore using the SPARQLUpdateStore. 

When I add data containing non-ASCII characters using `addN()` it results in garbled data in the triplestore (UTF-8 interpreted as ISO-8859).

The problem seems to be that SPARQLConnector.update() does not set "charset=UTF-8" in the Content-Type header:

https://github.com/RDFLib/rdflib/blob/a70a9c82649f9345c3691d69df6f1108e9e05871/rdflib/plugins/stores/sparqlconnector.py#L170

but it does use UTF-8 in `data=query.encode()` in line 183.

The problem goes away if I patch the Content-Type to be "application/sparql-update; charset=UTF-8".

I can provide a PR with this one-line change if you like.

robcast:
# Summary of changes

Add encoding "charset=UTF-8" to Content-Type header in SPARQLConnector.update() request.

Fixes #2095 

I added a separate test which is basically a copy of the existing test_graph_update(). The assertion could be integrated in the existing test instead.

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugins/stores/sparqlconnector.py
+++ b/rdflib/plugins/stores/sparqlconnector.py
@@ -167,7 +167,7 @@ class SPARQLConnector(object):
 
         headers = {
             "Accept": _response_mime_types[self.returnFormat],
-            "Content-Type": "application/sparql-update",
+            "Content-Type": "application/sparql-update; charset=UTF-8",
         }
 
         args = dict(self.kwargs)  # other QSAs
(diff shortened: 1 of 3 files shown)


aucampia:
Change looks good, I will double check why test is failing when I have time.

aucampia:
pre-commit.ci autofix

aucampia (approved):
Thanks for the pull request @robcast - the change and the test looks good.

PR #2134, opened 2022-10-10
Pass `service_query` to `_buildQueryStringForServiceCall` instead of a `Match`

veyndan:
In order to use the function, we must know that we need to pass a `Match` with a capture group at position 2. This is non-obvious.

### DIFF
--- a/rdflib/plugins/sparql/evaluate.py
+++ b/rdflib/plugins/sparql/evaluate.py
@@ -325,7 +325,7 @@ def evalServiceQuery(ctx: QueryContext, part):
 
     if match:
         service_url = match.group(1)
-        service_query = _buildQueryStringForServiceCall(ctx, match)
+        service_query = _buildQueryStringForServiceCall(ctx, match.group(2))
 
         query_settings = {"query": service_query, "output": "json"}
         headers = {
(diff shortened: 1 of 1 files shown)


aucampia:
@veyndan thanks for the pull request, we try and avoid merging PRs that affect runtime behaviour if they don't have tests, so please add tests to this if you can, otherwise it will have to wait for someone else to get time to look at it and add tests.

veyndan:
@aucampia I'm not too sure what sort of additional test would make sense here. I would assume that such a refactoring is covered by the tests in `test_service.py`. To make sure of this, I passed in random arguments to `_buildQueryStringForServiceCall` and they failed as expected.

aucampia:
> @aucampia I'm not too sure what sort of additional test would make sense here. I would assume that such a refactoring is covered by the tests in `test_service.py`

If this is only a refactor then yes, and actually it does seem to be only a refactor, apologies for not noticing. I will double check tomorrow night.

aucampia (approved):
Looks pretty good to me and it is covered by existing tests. It also makes a lot more sense to pass a string than the match, thanks @veyndan

PR #2142, opened 2022-10-24
Replace ParamList with Group

veyndan:
# Summary of changes

No functional changes. I've wanted to better understand the parsing of SPARQL, and have found that `ParamList` makes the parsing unnecessarily more complex. Replacing `ParamList` with the built-in `Group` operator simplifies the ability to understand how a SPARQL query is being parsed.

In order to replace `ParamList` with `Group`, I just did a find and replace of `ParamList(foo)` with `Group(Param(foo))`. Perhaps replacing instances of `ZeroOrMore(ParamList(foo))` should be replaced with `Group(ZeroOrMore(Param(foo))` instead of `ZeroOrMore(Group(Param(foo))`, but I'm not too sure of the benefits of one over the other, so I instead opted for what semantically matches what we had prior to this change.

# Checklist

- [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so maintainers can fix minor issues and keep your PR up to date.

### DIFF
--- a/rdflib/plugins/sparql/parserutils.py
+++ b/rdflib/plugins/sparql/parserutils.py
@@ -224,10 +224,11 @@ class Comp(TokenConverter):
         for t in tokenList:
             if isinstance(t, ParseResults):
                 for i in t:
-                    if isinstance(i, ParamValue):
-                        if i.name not in res:
-                            res[i.name] = []
-                        res[i.name].append(i.tokenList)
+                    if not isinstance(i, ParamValue):
+                        raise TypeError("ParseResult must be a ParamValue")
+                    if i.name not in res:
+                        res[i.name] = []
+                    res[i.name].append(i.tokenList)
             elif isinstance(t, ParamValue):
                 res[t.name] = t.tokenList
         return res
--- a/rdflib/plugins/sparql/parserutils.py
+++ b/rdflib/plugins/sparql/parserutils.py
@@ -94,8 +94,7 @@ class ParamValue(object):
     All cleverness is in the CompValue
     """
 
-    def __init__(self, name, tokenList, isList):
-        self.isList = isList
+    def __init__(self, name, tokenList):
         self.name = name
         if isinstance(tokenList, (list, ParseResults)) and len(tokenList) == 1:
             tokenList = tokenList[0]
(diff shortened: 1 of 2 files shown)


rdflib/plugins/sparql/parserutils.py
```diff
@@ -241,16 +228,14 @@ def postParse(self, instring, loc, tokenList):
...
-            # if isinstance(t,CompValue):
-            #    res.update(t)
+            if isinstance(t, ParseResults):
+                for i in t:
+                    if isinstance(i, ParamValue):
+                        if i.name not in res:
+                            res[i.name] = plist()
+                        res[i.name].append(i.tokenList)
+            elif isinstance(t, ParamValue):
+                res[t.name] = t.tokenList
```
aucampia, line 238: @veyndan could you explain the significance of this? It may be right but I can't say I entirely understand why, I could look into it more though and it may be right but would be good to understand why you changed it.

aucampia: Okay nevermind, I see now it is needed to accomodate the change from `ParamList` to `Group(Param(...))`

rdflib/plugins/sparql/parserutils.py
```diff
@@ -241,16 +228,14 @@ def postParse(self, instring, loc, tokenList):
...
-                        res[t.name] = plist()
-                    res[t.name].append(t.tokenList)
-                else:
-                    res[t.name] = t.tokenList
-                # res.append(t.tokenList)
-            # if isinstance(t,CompValue):
-            #    res.update(t)
+            if isinstance(t, ParseResults):
+                for i in t:
+                    if isinstance(i, ParamValue):
```
aucampia, line 233: Could it make sense to raise an exception if this condition does not match?

rdflib/plugins/sparql/parserutils.py
aucampia: Resolving.

aucampia (commented):
@veyndan thanks for the PR and apologies for the late review. It looks correct to me but I'm going to try get someone with some knowledge of `pyparsing` to also weigh in and do another review after thinking about it a bit, but I think this approach is significantly better than using a custom class.

rdflib/plugins/sparql/parserutils.py
veyndan: Yeah I think that's a lot clearer, as otherwise it looks like we're filtering our for `ParamValue`s, but they should all be `ParamValue`s. Fixed in latest commit.

ptmcg: Raising a TypeError here conflicts with this line in the docstring: "Any sub-tokens that are not Params will be ignored." So raising TypeError here may start breaking some existing code.

veyndan: I completely missed that! I'm not sure how true that line in the docstring is, but I think it's worth an investigation in a separate PR.

veyndan:
@aucampia Is this okay to merge?

ptmcg:
I have a few more comments to make on the use of Group, which I'll try to write up this evening.

ptmcg:
It has been a hairy few weeks, I'm sorry this took so long.

I've started a new page in the pyparsing wiki for pyparsing design topics, with [the first topic about using `Group`](https://github.com/pyparsing/pyparsing/wiki/Design-Notes#what-is-group-for). Please look it over (comments and suggestions welcome!), and if you like, we can discuss in more detail how this relates to ParamList in your parser.

aucampia:
Thanks @ptmcg , I will have a look.

@veyndan sorry I have been a bit busy, will try look again this week and process the inputs from @ptmcg

PR #2143, opened 2022-10-24
Remove redundant class

veyndan:
Despite the description of the class, we never type check for `plist`, nor is there a clear reason why we need to. I'm removing it for clarity.


aucampia (approved):
Sorry for the delay in review.

I'm okay with this, no good reason for this class to exist.

Issue #2098, opened 2022-08-25
AttributeError: module 'posixpath' has no attribute 'sep'


zero0nee:
When running `Graph.pars` with `format="json-ld"` in python 3.10.4 I get. the following error. 

`AttributeError: module 'posixpath' has no attribute 'sep'`

The problem seems to stem from the [following line](https://github.com/RDFLib/rdflib/blob/6f2c11cd2c549d6410f9a1c948ab3a8dbf77ca00/rdflib/plugins/shared/jsonld/util.py#L16) due to the fact that `sep` is no longer a part of `posixpath`. 

Replacing
```
from posixpath import normpath, sep
```
with
```
from posixpath import normpath 
from os import sep
```

should solve the problem. 

**How to reproduce:**
Run the following code in python 3.10.4 or higher:

```
from rdflib import Graph
g = Graph()
g.parse(data='{"@type": "Actor", "@id": "http://test.io/olcaId/0044417a-cd8e-447c-8205-d158af171f81", "@context": "http://test.github.io/olca-schema/context.jsonld"}', format="json-ld")
```

aucampia:
> Replacing
> 
> ```
> from posixpath import normpath, sep
> ```
> 
> with
> 
> ```
> from posixpath import normpath 
> from os import sep
> ```
> 
> should solve the problem.

Thanks for the report, however this will use the wrong separator on windows I think as os.sep is OS dependent.

Will look at it when we can though and PRs are always welcome. We should also figure out why this was not picked up by our continuous integration which do run the test suite on 3.10.4

aucampia:
Related:
- https://github.com/stub42/pytz/issues/72
- https://github.com/pallets/jinja/issues/1697

zero0nee:
Just made a [PR](https://github.com/RDFLib/rdflib/pull/2156). @aucampia

aucampia:
> The problem seems to stem from the [following line](https://github.com/RDFLib/rdflib/blob/6f2c11cd2c549d6410f9a1c948ab3a8dbf77ca00/rdflib/plugins/shared/jsonld/util.py#L16) due to the fact that `sep` is no longer a part of `posixpath`.

Is it not? I really am not sure why you get this error when we don't get it on our CI systems. It is also still documented in python 3.11

```console
$ pydoc3.11 posixpath | sed -n '/^DATA/,/^$/p'
DATA
    __all__ = ['normcase', 'isabs', 'join', 'splitdrive', 'split', 'splite...
[...]

PR #2156, opened 2022-11-16
Issue #2098

zero0nee:
# Summary of changes

Import sep from os instead of posixpath to avoid the following [issue](https://github.com/RDFLib/rdflib/issues/2098).

### DIFF
--- a/rdflib/plugins/shared/jsonld/util.py
+++ b/rdflib/plugins/shared/jsonld/util.py
@@ -13,8 +13,8 @@ else:
         import simplejson as json
 
 from io import TextIOBase, TextIOWrapper
-from posixpath import normpath
 from os import sep
+from posixpath import normpath
 from urllib.parse import urljoin, urlsplit, urlunsplit
 
 from rdflib.parser import (
--- a/rdflib/plugins/shared/jsonld/util.py
+++ b/rdflib/plugins/shared/jsonld/util.py
@@ -13,7 +13,8 @@ else:
         import simplejson as json
 
 from io import TextIOBase, TextIOWrapper
-from posixpath import normpath, sep
+from posixpath import normpath
+from os import sep
 from urllib.parse import urljoin, urlsplit, urlunsplit
 
 from rdflib.parser import (


rdflib/plugins/shared/jsonld/util.py
```diff
@@ -13,7 +13,8 @@
         import simplejson as json
 
 from io import TextIOBase, TextIOWrapper
-from posixpath import normpath, sep
+from os import sep
+from posixpath import normpath
```
aucampia, line 17: The problem with this is, os.sep is platform dependent, while posixpath is not.

On Linux:

```console
$ uname -a
Linux teekai.zoic.eu.org 6.0.8-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Nov 11 15:03:58 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
$ python -m pydoc os | grep -E '^\s+sep = '
    sep = '/'
$ python -m pydoc posixpath | grep -E '^\s+sep = '
    sep = '/'
```

On Windows:

```console
$ uname -a
MINGW64_NT-10.0-19044 DESKTOP-KLQG489 3.1.7-340.x86_64 2021-10-12 16:29 UTC x86_64 Msys
$ python -m pydoc os | grep -E '^\s+sep = '
    sep = r'\'
$ python -m pydoc posixpath | grep -E '^\s+sep = '
    sep = '/'
```

If this is causing problems, which it seems it is, we should rather just use a private constant for it, like `_posixpath_sep = '/'`

However I still don't understand how you get this issue. python 3.7 to 3.11 has sep as part of posixpath as far as I can tell. It would be helpful if you could give an example of how your system was setup and where python came from and what version it is.

aucampia:
We really need to better understand why the problem you reported in https://github.com/RDFLib/rdflib/issues/2098 happens.

rdflib/plugins/shared/jsonld/util.py
aucampia: I would change it but you did not give maintainers permission to modify your branch, if you do I will update your branch.

aucampia:
I'm closing this as there has been no feedback, I'm happy to make a change to solve your issue, but the change in this PR won't quite work as it is platform-dependent. I would also really like to understand more about your environment for reproducing this.

Issue #2164, opened 2022-11-21
JSON-LD parser doesn't restore root directory


hsolbrig:
It appears that the working directory isn't refreshed when reading when reading relative context files.

In the directory structure:

```
test.jsonld
contexts/person.jsonld
contexts/place.jsonld
```

Where test.jsonld contains:
```
{
  "@context": [
    "contexts/person.jsonld",
    "contexts/place.jsonld"
  ]
}
```

`rdfpipe -i json-ld -o turtle ldbug.jsonld`  returns the following:
```
Traceback (most recent call last):
  File "/Users/solbrig/Development/json_ld_tests/venv/bin/rdfpipe", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/tools/rdfpipe.py", line 198, in main
    parse_and_serialize(
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/tools/rdfpipe.py", line 52, in parse_and_serialize
    graph.parse(fpath, format=use_format, **kws)
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/graph.py", line 1977, in parse
    context.parse(source, publicID=publicID, format=format, **args)
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/graph.py", line 1330, in parse
    parser.parse(source, self, **args)
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/plugins/parsers/jsonld.py", line 115, in parse
    to_rdf(data, conj_sink, base, context_data, version, generalized_rdf)
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/plugins/parsers/jsonld.py", line 134, in to_rdf
    return parser.parse(data, context, dataset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/plugins/parsers/jsonld.py", line 154, in parse
    context.load(local_context, context.base)
  File "/Users/solbrig/Development/json_ld_tests/venv/lib/python3.11/site-packages/rdflib/plugins/shared/jsonld/context.py", line 375, in load
[...]

Issue #415, opened 2014-08-05
Parsing error for Unicode in freebase


rosshorne:
Parsing fails for certain literals. For example, the following curl returns apparently fine ntriples.
curl https://www.googleapis.com/freebase/v1/rdf/m/012s3l

However, the following code using rdflib results in a parse error.
g.parse("https://www.googleapis.com/freebase/v1/rdf/m/012s3l", format="n3")

I'm reproducing the relevant error below, which indicates that there is a parse problem with "^\xab" that appears in a literal.

rdflib.plugins.parsers.notation3.BadSyntax: at line 14 of <>:
Bad syntax (bad escape) at ^ in:
"...438\u043d \u0437 \u0442\u0432\u043e\u0440\u0446\u0456\u0432 ^\xab\u0448\u043a\u0456\u043b\u044c\u043d\u043e\u0457 \u0456\..."

Since this URI is for freebase, we can't change the data. Would it be possible to fix the parser? I notice there are sever comments elsewhere about problems parsing Unicode, so this problem probably applies to a class of escape characters.

joernhees:
closing this for now, please re-open if needed

PR #2165, opened 2022-11-22
Issue 2164

hsolbrig:
# Summary of changes

Added a test case for issue #2164 and proposed a fix.  The problem arose in line #415 in context.py, where the base URI passed to _prep_sources was overwritten in anticipation of processing inner nestings.  Changed the assignment to `base` to `new_base`


aucampia:
pre-commit.ci autofix

aucampia:
> This appears to be happening because of [test_localsuite.py](https://github.com/RDFLib/rdflib/blob/f9d7132d042be61690954d8a0678863498ae1f9c/test/jsonld/test_localsuite.py#L11) expects the data to be located in https://rdflib.github.io/rdflib-jsonld/local-testsuite/ instead of the local testing directory.

The reason this URI is used is so that there is a system neutral way of referring to tests and test content, the best reference I can find for this now is [this](https://github.com/w3c/rdf-tests/blob/22cfdfda79153b6dba91844e69043c019c89fd84/turtle/README#L36-L41)

> ```
> The home of the test suite is <http://www.w3.org/2013/TurtleTests/>.
> Per RFC 3986 section 5.1.3, the base IRI for parsing each file is the
> retrieval IRI for that file. For example, the tests turtle-subm-01 and
> turtle-subm-27 require relative IRI resolution against a base of
> <http://www.w3.org/2013/TurtleTests/turtle-subm-01.ttl> and
> <http://www.w3.org/2013/TurtleTests/turtle-subm-27.ttl> respectively.
> ```

There really are no good or simple solutions to this. We had a similar problem for the SPARQL test suite and there I resorted to monkey patching:

https://github.com/RDFLib/rdflib/blob/0e61a7fc82bd35ff0cec800c446647fe7d9c39d5/test/utils/sparql_checker.py#L349-L374

In this case I think what you did is fine as there is no need for system neutral references in the local-suite JSON-LD tests.

The change and tests looks fine to me otherwise. I will just do a double check tomorrow but I think it is good to merge.

aucampia:
Okay I made a small change to instead use a file URI for `TC_BASE` and also added a comment to explain why, this should keep things closer to the way it was.

aucampia (approved):
Change looks good to me @hsolbrig - I will merge by this weekend if there is no further feedback.

PR #2169, opened 2022-11-23
compute_qname handle case where name could be unbound

tgbugs:
In NamespaceManager.compute_qname it is possible for the variable name to be unbound when a namespace is used as a predicate e.g. ns2: and the namespace itself cannot be split any further.

Fixing this usually just kicks the can down the road because the only reason to hit this condition is that you are trying to serialize a predicate that has no valid NCName form into an xml format that requires one, making it impossible to complete the conversion. That said, this fix should make it easier to understand what is going on by unmasking the real issue.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -394,6 +394,7 @@ class NamespaceManager(object):
     """
 
     def __init__(self, graph: "Graph", bind_namespaces: "_NamespaceSetString" = "core"):
+        logging.debug("self = %r", self)
         self.graph = graph
         self.__cache: Dict[str, Tuple[str, URIRef, str]] = {}
         self.__cache_strict: Dict[str, Tuple[str, URIRef, str]] = {}
@@ -468,6 +469,7 @@ class NamespaceManager(object):
         registered prefix) or (unlike compute_qname) the Notation 3
         form for URIs: <...URI...>
         """
+        logging.debug("normalizeUri: rdfTerm = %s", rdfTerm)
         try:
             namespace, name = split_uri(rdfTerm)
             if namespace not in self.__strie:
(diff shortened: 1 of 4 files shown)


aucampia:
pre-commit.ci autofix

rdflib/namespace/__init__.py
```diff
@@ -504,6 +504,7 @@ def compute_qname(self, uri: str, generate: bool = True) -> Tuple[str, URIRef, s
             except ValueError as e:
                 namespace = URIRef(uri)
                 prefix = self.store.prefix(namespace)
+                name = ""  # empty prefix case, safe since not prefix is error
```
aucampia, line 507: @tgbugs It looks reasonable, but could you add a test for this? It will make it easier to understand the failure mode, if you are unsure I can have a look later.

tgbugs: Will do.

aucampia: Thanks, I had a look now, and added some more tests here: https://github.com/aucampia/rdflib/commits/fix-nm-name-error

https://github.com/aucampia/rdflib/blob/2145bdbe75e7c085f2050677e84f8bb6928197d4/test/test_namespacemanager.py#L241-L296


However things don't quite go wrong as I expect them to, and I would like to be able to make this go wrong in an isolated situation. I will look at it further when I have more time.

tgbugs: I assume you ran your new tests without the `name = ''` fix.

I think the test you have as written won't cause the issue because the blank nodes (or maybe just one blank node?) are required to trigger the issue. There are a number of other existing tests where the xml roundtrip is marked as xfail which I think match the new ones you have, but I could be wrong.

aucampia: @tgbugs I may be missing something, but my view is, if something is wrong with `compute_qname` that is fixed with this bug, then we should be able to define a unit test case that reproduces it.

`xfail` is not ideal for reproducing and testing this as xfails designate known issues, I think the xfail you added is a valid xfail, but I would like to have a non-xfail test for the functionality that fails without the fix and passes with it.

I will debug more and look specifically if I can figure out what happens with blank nodes that does not happen in my unit tests.

tgbugs: > I will debug more and look specifically if I can figure out what happens with blank nodes that does not happen in my unit tests.

Got it.

aucampia: Okay I managed to reproduce this independently with this case:

https://github.com/RDFLib/rdflib/blob/458e34baed5cda3c3130fc0f65e0860a778c62b6/test/test_namespacemanager.py#L213-L221

The key was that the prefix should only be bound in the store and nowhere else.

Without your fix that case fails, with your fix it passes.

aucampia: Resolving as I'm happy given the new limited tests I added to your branch.

aucampia (approved):
@tgbugs thanks for the fix, looks good.

And apologies for the dodgy commit names to your branch, I would fix them but I don't want to force push to your branch so I'm just leaving them as is. I will sqash merge this PR later.

aucampia:
I will merge tomorrow if there is no further feedback.

PR #2174, opened 2022-12-07
Add namespace_manager argument for n3 method on Paths

mgberg:
# Summary of changes

The `n3` method on the `Term` class optionally takes a `NamespaceManager` as an argument which, if provided, is used to create a prefixed name instead of fully qualified URIs for `URIRef`s and typed `Literal`s where applicable if an appropriate prefix has been bound for the relevant namespace. `Path` objects also have a `n3` method to create N3/SPARQL compatible serializations for those objects which call the `n3` method on the underlying `URIRef`s. However, `Path.n3` did not have the argument for the `NamespaceManager`.

This PR simply adds the same optional `NamespaceManager` argument to the `n3` method for `Path`s to be consistent with the `n3` method on `Term`s and passes that `NamespaceManager` to any nested calls to `n3` so it behaves as expected.

# Checklist

  the same change.

### DIFF
--- a/rdflib/paths.py
+++ b/rdflib/paths.py
@@ -188,6 +188,7 @@ from rdflib.term import Node, URIRef
 
 if TYPE_CHECKING:
     from rdflib.graph import Graph, _ObjectType, _SubjectType
+    from rdflib.namespace import NamespaceManager
 
 
 # property paths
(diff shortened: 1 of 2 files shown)


aucampia (commented):
@mgberg the PR looks pretty good, I made some minor changes to get validation to pass and to add some additional type hints:

- https://github.com/corning-incorporated/rdflib/pull/1

If you make PRs, it is better to make it from your personal account, that way we can push to your branch instead of needing to make PR to your branch.

aucampia (approved):
Change looks good to me, it makes sense to have the same signature for `def n3()` across all classes. I will merge before this weekend if there is no further feedback.

aucampia:
> I believe that my organizational policy disallows that, but I'll double check on that. Thanks!

It really is also up to you, I don't really mind either way, makes things a bit easier if we can push to your branch but it is not a requirement for PRs and if the PR does something valuable I don't mind making a PR to the PR branch to add fixes.

PR #2180, opened 2022-12-13
Fix missing query string params in sparqlconnector when using POST method 

davidandreoletti:
# Summary of changes

The  SPARQLConnector instance created by a SPARQLStore instance providing endpoint url + **query string params** using POST method did not append the query string to endpoint url.

This PR fixes this.

Backward compatible: yes.

  the same change.


aucampia (approved):
Thank you for the PR @davidandreoletti. I updated your code slightly to cover the edge case where `default-graph-uri` is set, and I also changed your test to a parameterized test which checks for this case.

I also made some other slight cosmetic changes to the test.

aucampia:
I will merge tomorrow if there is no further feedback.

davidandreoletti:
> I will merge tomorrow if there is no further feedback.

Sounds good. Thanks.

davidandreoletti:
@aucampia Last release (v6.2.0) was published around July 2022. What's the most probable timeline for the next release ?

PR #2183, opened 2022-12-17
Issue2179 incorrect headers

gitmpje:
# Summary of changes

# Checklist

  the same change.


aucampia (approved):
Thanks for the PR @gitmpje, copy was not enough, it needed a deepcopy, but the test was great and you found the exact source of the problem which is very helpful.

I changed copy to deepcopy and also changed the test to run everything twice so that any issues with ordering is picked up both ways.

aucampia:
I will merge by the weekend if there are no further changes.

gitmpje:
@aucampia Thank you for fixing this issue more fundamentally. Using `copy` seemed to solve the issue when testing locally, but `deepcopy` should definitely fix it.

PR #2187, opened 2022-12-29
move to poetry for dependency management; consolidate more settings into pyproject.toml

jclerman:
## Summary

Goals:
1. move from `setup.py` & setuptools to `poetry`, for dependency and project management.
2. move other settings, from `setup.cfg`, into `pyproject.toml`.
3. Consolidate all dependencies into one file, `pyproject.toml`, and remove the need for all `requirements*txt` files.

Includes moving the package version into `pyproject.toml`, and in `__init__.py`, reading the version from `pyproject.toml` via use of `importlib.metadata`.

## TODO
* ~Fixes to `tox.ini` are still needed.~ DONE
* ~Updates to `pyproject.toml` are needed, to add more dependency groups, which should then be referenced in `tox.ini` and other build/linting configs that have steps requiring development-dependencies.~ DONE
* Remove references to `requirements*txt` files from 
  * ~`Taskfile.yaml`~ DONE
  * ~`workflows/validate.yaml`~ DONE
* ~Remove `requirements*txt` files once they are no longer referenced anywhere.~ DONE

## Other details

* With this change, versions of package dependencies are now explicit.
* This change eliminates the use of static `requirements*txt` files from `Dockerfile.devcontainer`

# Checklist

  the same change.
  - [x] Considered updating our changelog (`CHANGELOG.md`). (Elected not to update `CHANGELOG.md` since this PR does not have user-facing changes)
- [x] ~Considered~ granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),


aucampia:
@jclerman thanks for this, would be very happy to get to poetry as that makes things a lot easier.

pyproject.toml
```diff
@@ -82,3 +148,47 @@ log_format = "%(asctime)s.%(msecs)03d %(levelname)-8s %(name)-12s %(filename)s:%
...
+warn_unreachable = true
+warn_unused_ignores = true
+no_implicit_optional = false
+
+[[tool.mypy.overrides]]
+module = "pyparsing.*"
+# This is here because of an upstream issue with pyparsing:
+# https://github.com/pyparsing/pyparsing/issues/385
+# Once the issue is fixed this should be removed.
+follow_imports = "skip"
```
aucampia, line 194: You also need a `build-system` section:

```suggestion
follow_imports = "skip"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

pyproject.toml
```diff
@@ -1,3 +1,69 @@
...
+flake8 = [
+    {version = "^5.0.0", python = ">=3.7,<3.8"},
+    {version = "^6.0.0", python = ">=3.8.1"},
+]
+isort = "^5.10.0"
+mypy = "0.991"
+
+[tool.poetry.extras]
+berkeleydb = ["berkeleydb"]
+networkx = ["networkx"]
```
aucampia, line 65:
```suggestion
html = ["html5lib"]
```

pyproject.toml
```diff
@@ -1,3 +1,69 @@
...
+#setuptools  # not needed anymore?
+importlib_metadata = {version = "^4.2.0", python = ">=3.7,<3.8"}
+berkeleydb = {version = "^18.1.5", optional = true}
+networkx = [
+    {version = "^2.6.2", python = ">3.7,<3.8", optional = true},
+    {version = "^2.8.8", python = "^3.8", optional = true},
+]
+
+[tool.poetry.group.dev.dependencies]
+html5lib = "^1.1"
```
aucampia, line 47: So that it can be used in `[tool.poetry.extras]`

```suggestion
networkx = [
    {version = "^2.6.2", python = ">3.7,<3.8", optional = true},
    {version = "^2.8.8", python = "^3.8", optional = true},
]
html5lib = "^1.1"

[tool.poetry.group.dev.dependencies]
```

aucampia:
@jclerman let me know if you are okay for me to commit to this branch.

jclerman:
> @jclerman let me know if you are okay with me commiting to your branch.

@aucampia that should be fine. I wanted to make a few additional changes to demonstrate how we can get tox to use `pyproject.toml`-specified dependencies, without the use of `extras` which don't quite make sense since they are intended only for development use, not for end-users. I have that working now (I think) - though I am seeing issues with both the `flake8` and `docs` builds, at least locally. Not clear to me why either of the failures are occurring, or whether they are even new.

pyproject.toml
aucampia: Actually this is somewhat wrong, but the essence of it is that we should try and retain the same extras as before, and we used to have `html` installing `html5lib` so would be best to keep it like that.

pyproject.toml
jclerman: Added now.

pyproject.toml
jclerman: Got it. I didn't understand this request at first but now I do - I've made the update in the most recent commit.  Note that in order to make it so that html5lib is only installed as an "extra", we need to specify it with `optional = true`.  I've done that in my commit, and also added the necessary line to the `extras` section.

Looks like I've omitted some other previously-configured `extra` groups too - I'll need to address those, unless we decide they shouldn't be extras after all.

aucampia:
@jclerman I won't make changes to your branch, I will make a separate branch  with any changes that I think may be relevant and link to it here, and I will merge your branch into it when there is changes.

pyproject.toml
aucampia: I would say the only extras that we should definitely keep are the one for normal users:

```
{
"html": ["html5lib"],
"berkeleydb": ["berkeleydb"],
"networkx": ["networkx"],
}
```

The others, `tests`, `docs`, `dev` are less important to keep as extras think.

Currently readthedocs uses the `docs` extras:

https://github.com/RDFLib/rdflib/blob/8213b7465783e636a394c86fd2387a44523d9c2f/.readthedocs.yaml#L19-L24


But I think we should find a way to get readthedocs to use the poetry.lock file so the builds are more reproducible, I will look at that.

There are also this issue:

- https://github.com/RDFLib/rdflib/issues/1519#issuecomment-999191547

I'm not entirely sure of the workflow here but it seems there is some expectation of having `tests` extras maybe, however the right solution here could just be to fix the tests that use optional dependencies to only run when those are installed.

I would be fine if we don't keep extras for `docs`, `dev` and `tests`. We can always add it back if it causes serious problems.

aucampia: CC: @kloczek you may have some input that is relevant here also.

jclerman:
@aucampia, for the time being, I have updated `pyproject.toml` to preserve all of the "extras" groups as they were previously configured.  I think we agree that it'd be nice to remove the "extras" that aren't intended for end-users, replacing them with `poetry` dependency-groups (as was my initial plan).  However, there are a lot of moving parts here, and it probably make sense to stay focused on the move to `poetry`, and to tackle the migration away from extra "extras" as a follow-up PR.

Unfortunately, tox (and probably other tools) does not (yet?) support built-in access to poetry dependency-groups.  The way I have accessed them in other projects is to add a clause like this to a given tox-environment's config clause:

```
allowlist_externals = poetry
commands_pre =
    poetry export --only XYZ --output {temp_dir}/{envname}-requirements-XYZ-only.txt
    pip install -r {temp_dir}/{envname}-requirements-XYZ-only.txt
```

replacing "XYZ" with the name of the poetry dep-group I want to use for that tox-env.

pyproject.toml
jclerman: I think this suggestion is resolved with my latest changes?

tox.ini
```diff
@@ -36,9 +37,7 @@ commands =
     {envpython} -m coverage report
 
 [testenv:flake8]
-deps =
-    -rrequirements.flake8.txt
-skip_install = true
```
jclerman, line 41: The `skip_install` directive seemed to be causing issues for me, so I removed it. The flake8 requirements are now moved to an "extra" group defined in `pyproject.toml`; the `requirements.flake8.txt` file should no longer be needed.

jclerman:
I think just one issue remains to address before we can deprecate `setup.py`: The following clause in that file provides logic for conditional installation of the `examples` package.  That kind of conditional installation is not supported through `poetry`, as far as I can tell.  I'm also not sure if the condition is ever met, since I couldn't find the `READTHEDOCS` environment-variable set in any of the build-configs - but maybe it's set elsewhere?

This is the clause:
```
if os.environ.get("READTHEDOCS", None):
    # if building docs for RTD
    # install examples, to get docstrings
    packages.append("examples")
```

I think the simplest thing here would be to move `examples` down into the `rdflib` distribution and make it a sub-package (which is always included in the build), and to update `docs/apidocs/examples.rst` accordingly. If there is a simpler solution though, that'd be great. @aucampia

aucampia:
@jclerman thanks for all the work here, I will look at what we can do about the examples, I also think it may not be needed.

Dockerfile.devcontainer
```diff
@@ -41,17 +41,22 @@ RUN \
...
+RUN poetry export \
+    -E dev \
+    -E tests \
+    -E docs \
+    --output /var/tmp/requirements-via-poetry-including-all-dev.txt
+RUN \
+    python -m pip install \
+        -r /var/tmp/requirements-via-poetry-including-all-dev.txt \
         && \
     true
```
aucampia, line 62: For the devconainer there is really no need to install the dependencies globally, they should always be used from the poetry venv.

```suggestion
```

pyproject.toml
jclerman: @aucampia, looks like this snippet provides a useful example of how to use poetry with readthedocs.yml: https://github.com/readthedocs/readthedocs.org/issues/4912#issuecomment-1143587902

aucampia:
I think we should just drop flakeheaven, it is sad, but they are too far behind on flake8.

aucampia:
We have to use sphinx>=5, and if I set that constraint flake8 4.x becomes an issue. I will see what I can do to run flake8 only on the diffs, but will do that later, for now I would rather go without flake8. I will look at it further tomorrow. My branch is https://github.com/aucampia/rdflib/tree/switch-to-poetry

jclerman:
> We have to use sphinx>=5, and if I set that constraint flake8 4.x becomes an issue. I will see what I can do to run flake8 only on the diffs, but will do that later, for now I would rather go without flake8. I will look at it further tomorrow. My branch is https://github.com/aucampia/rdflib/tree/switch-to-poetry

Would it be helpful if I were to start manually merging your changes into this (https://github.com/jclerman/rdflib/tree/switch-to-poetry) branch?

aucampia:
> Are there tests (or can we make one) that show what breaks when we use sphinx 4.x?
> 

Sphinx 4 creates warnings with python 3.7 relating to type hints and these are set to result in errors. We can maybe disable that but then we will miss other real problems, but maybe just disabling it on python 3.7 makes sense.

> 
> Would it be helpful if I were to start manually merging your changes into this (https://github.com/jclerman/rdflib/tree/switch-to-poetry) branch?

I wouldn't yet. Let me figure out things first, I will clean it up as much as I can and make sure it only contains things that really should be in the same PR. I should be done tomorrow.

aucampia:
Some thoughts ...

The main benefit we want from poetry is reproducible dev environments, which we can get from poetry's lockfile. However, for this to provide maximal value we really need to take this into account for testing also. I'm looking at options now.

Further, we should be as liberal with version restrictions that end up in the wheel, we should not really place upper limits unless we know later versions break, and the lower limits should be as low as possible.

As for the importlib-metadata and flakeheaven situation, I'm trying to resolve it without removing flakeheaven.

aucampia:
> The main benefit we want from poetry is reproducible dev environments, which we can get from poetry's lockfile. However, for this to provide maximal value we really need to take this into account for testing also. I'm looking at options now.

Somewhat burying the lede here, to just clarify, we really need to take this into account for tox. So tox should use the versions from poetry.lock. At least I think so. Maybe we can look at it later though.

aucampia:
> > The main benefit we want from poetry is reproducible dev environments, which we can get from poetry's lockfile. However, for this to provide maximal value we really need to take this into account for testing also. I'm looking at options now.
> 
> Somewhat burying the lede here, to just clarify, we really need to take this into account for tox. So tox should use the versions from poetry.lock. At least I think so. Maybe we can look at it later though.

yeah I will figure it out in a later PR.

jclerman:
> > > The main benefit we want from poetry is reproducible dev environments, which we can get from poetry's lockfile. However, for this to provide maximal value we really need to take this into account for testing also. I'm looking at options now.
> > 
> > 
> > Somewhat burying the lede here, to just clarify, we really need to take this into account for tox. So tox should use the versions from poetry.lock. At least I think so. Maybe we can look at it later though.
> 
> yeah I will figure it out in a later PR.

It's not clear that using the values from poetry.lock is desirable, since rdflib is a library and may be used with other (later) versions of dependencies - i.e., whatever is advertised as compatible, per pyproject.toml.  

However, it's possible to get tox to use the poetry.lock values, and that does have the advantage of reproducible tests (we can always test with later versions of deps by rerunning `poetry lock`).  To do that, we could:
* move development dependencies back into poetry dependency groups (instead of "extras")
* Within `tox.ini`, include pre-commands that (1) use poetry to export the dependencies (that will effectively use the values in poetry.lock) to a temporary, tox-venv-specific requirements.txt file, and then (2) use pip to install those (additional) packages from the temp requirements.txt into the tox venv.  That is what I've done in other projects and it works well.

jclerman:
> Some thoughts ...
> 
> The main benefit we want from poetry is reproducible dev environments, which we can get from poetry's lockfile. However, for this to provide maximal value we really need to take this into account for testing also. I'm looking at options now.
> 
> Further, we should be as liberal with version restrictions that end up in the wheel, we should not really place upper limits unless we know later versions break, and the lower limits should be as low as possible.

My usual practice is to assume that (other) libraries use semantic versioning, such that we should assume that major version-changes include breaking changes and minor/patchlevel version-changes don't.  Thus my use of "^X.Y.Z" in most of the version-specs in my proposed pyproject.toml.  We haven't tested version "X+1" so we shouldn't advertise rdflib as being compatible with it.

That imposes more of a maintenance-burden on the rdflib side but also should make it more robust.

aucampia:
@jclerman please update the branch with main when you can (merge or rebase, up to you), or let me know if you are fine if I do it, I want to be sure we are not dropping any coverage.

aucampia:
> I think just one issue remains to address before we can deprecate `setup.py`: The following clause in that file provides logic for conditional installation of the `examples` package. That kind of conditional installation is not supported through `poetry`, as far as I can tell. I'm also not sure if the condition is ever met, since I couldn't find the `READTHEDOCS` environment-variable set in any of the build-configs - but maybe it's set elsewhere?
> 
> This is the clause:
> 
> ```
> if os.environ.get("READTHEDOCS", None):
>     # if building docs for RTD
>     # install examples, to get docstrings
>     packages.append("examples")
> ```
> 
> I think the simplest thing here would be to move `examples` down into the `rdflib` distribution and make it a sub-package (which is always included in the build), and to update `docs/apidocs/examples.rst` accordingly. If there is a simpler solution though, that'd be great. @aucampia

I looked at this now, it is being set by the readthedocs builder [[ref](https://docs.readthedocs.io/en/stable/environment-variables.html#envvar-READTHEDOCS)], however, I'm pretty sure we don't need to do this, as the readthedocs builder does not use the installed code to generate APIdocs and instead uses the source directory.

The readthedocs builder works fine as far as I can tell: 

Screenshot_2023-01-03_20-40-42

The build logs are here: https://readthedocs.org/projects/rdflib/builds/19065922/

And the generated docs look mostly fine, there are some odd things but nothing related to the examples as far as I can tell: https://rdflib--2187.org.readthedocs.build/en/2187/

The source links from the [examples](https://rdflib--2187.org.readthedocs.build/en/2187/apidocs/examples.html) page even work.

So as far as I can tell this is not a problem, if someone reports an issue we can look at it.

aucampia:
I'm mostly done with <https://github.com/aucampia/rdflib/tree/switch-to-poetry> (<https://github.com/RDFLib/rdflib/pull/2192>) - I'm just trying to figure out why tox is not installing networkx

aucampia:
I will clean up my branch's commit log and explain the changes then make a PR into your branch.

pyproject.toml
aucampia: resolving this as I'm fairly happy with what you did, will though make a further suggestion in a PR to your branch.

aucampia:
> I'm mostly done with https://github.com/aucampia/rdflib/tree/switch-to-poetry (#2192) - I'm just trying to figure out why tox is not installing networkx.
> 
> ### Update:
> Actually I figured it out I think:
> 
> ```diff
> $ git diff
> diff --git a/pyproject.toml b/pyproject.toml
> index 65320735..f0ae154f 100644
> --- a/pyproject.toml
> +++ b/pyproject.toml
> @@ -38,7 +38,7 @@ pyparsing = "^3.0.9"
>  importlib_metadata = {version = "^4.2.0", python = ">=3.7,<3.8"}
>  berkeleydb = {version = "^18.1.5", optional = true}
>  networkx = [
> -    {version = "^2.6.2", python = ">3.7,<3.8", optional = true},
> +    {version = "^2.6.2", python = ">=3.7,<3.8", optional = true},
>      {version = "^2.8.8", python = "^3.8", optional = true},
>  ]
>  html5lib = {version = "^1.1", optional = true}
> ```

Not quite working yet:

```console
20230103T221820 iwana@teekai.zoic.eu.org:~/sw/d/github.com/iafork/rdflib.reviews
$ TOX_PYTEST_ARGS=./test/test_extras/test_extras_external_graph_libs.py tox -e py37-extensive -v --recreate
py37-extensive: remove tox env folder /home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/py37-extensive
.pkg: remove tox env folder /home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/.pkg
py37-extensive: find interpreter for spec PythonSpec(major=3, minor=7)
py37-extensive: proposed PythonInfo(spec=CPython3.11.1.final.0-64, system=/usr/bin/python3.11, exe=/home/iwana/.local/pipx/venvs/tox/bin/python, platform=linux, version='3.11.1 (main, Dec  7 2022, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]', encoding_fs_io=utf-8-utf-8)
py37-extensive: proposed PathPythonInfo(spec=CPython3.10.9.final.0-64, system=/home/linuxbrew/.linuxbrew/opt/python@3.10/bin/python3.10, exe=/home/linuxbrew/.linuxbrew/bin/python3, platform=linux, version='3.10.9 (main, Dec  6 2022, 18:44:57) [GCC 11.3.0]', encoding_fs_io=utf-8-utf-8)
py37-extensive: proposed PathPythonInfo(spec=CPython3.7.16.final.0-64, exe=/usr/bin/python3.7, platform=linux, version='3.7.16 (default, Dec  7 2022, 12:54:52) \n[GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]', encoding_fs_io=utf-8-utf-8)
py37-extensive: create virtual environment via CPython3Posix(dest=/home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/py37-extensive, clear=False, no_vcs_ignore=False, global=False)
py37-extensive: add seed packages via FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/iwana/.local/share/virtualenv)
py37-extensive: add activators for Bash, CShell, Fish, Nushell, PowerShell, Python
py37-extensive: install_deps> python -I -m pip install berkeleydb-stubs
py37-extensive: exit 0 (0.85 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python -I -m pip install berkeleydb-stubs pid=424180
.pkg: find interpreter for spec PythonSpec(path=/home/iwana/.local/pipx/venvs/tox/bin/python)
.pkg: proposed PythonInfo(spec=CPython3.11.1.final.0-64, system=/usr/bin/python3.11, exe=/home/iwana/.local/pipx/venvs/tox/bin/python, platform=linux, version='3.11.1 (main, Dec  7 2022, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]', encoding_fs_io=utf-8-utf-8)
.pkg: create virtual environment via CPython3Posix(dest=/home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/.pkg, clear=False, no_vcs_ignore=False, global=False)
.pkg: add seed packages via FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/iwana/.local/share/virtualenv)
.pkg: add activators for Bash, CShell, Fish, Nushell, PowerShell, Python
.pkg: install_requires> python -I -m pip install 'poetry-core>=1.4.0'
.pkg: exit 0 (1.11 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python -I -m pip install 'poetry-core>=1.4.0' pid=424229
.pkg: _optional_hooks> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api
.pkg: exit None (0.05 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api pid=424271
.pkg: get_requires_for_build_sdist> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api
.pkg: exit None (0.00 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api pid=424271
.pkg: prepare_metadata_for_build_wheel> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api
.pkg: exit None (0.13 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api pid=424271
.pkg: build_sdist> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api
.pkg: exit None (0.41 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api pid=424271
py37-extensive: install_package_deps> python -I -m pip install 'berkeleydb<19.0.0,>=18.1.5' black==22.12.0 'coverage[toml]<8.0.0,>=7.0.1' 'flake8>=4.0.1' 'flakeheaven<4.0.0,>=3.2.1' 'html5lib<2.0,>=1.1' 'importlib_metadata<5.0.0,>=4.2.0; python_version >= "3.7" and python_version < "3.8"' 'isodate<0.7.0,>=0.6.1' 'isort<6.0.0,>=5.10.0' mypy==0.991 'networkx<3.0.0,>=2.6.2; (python_version >= "3.7" and python_version < "3.8") and extra == "networkx"' 'networkx<3.0.0,>=2.8.8; (python_version >= "3.8" and python_version < "4.0") and extra == "networkx"' 'pep8-naming<0.14.0,>=0.13.2' 'pyparsing<4.0.0,>=3.0.9' 'pytest-cov<5.0.0,>=4.0.0' 'pytest<8.0.0,>=7.1.3'
py37-extensive: exit 0 (5.31 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python -I -m pip install 'berkeleydb<19.0.0,>=18.1.5' black==22.12.0 'coverage[toml]<8.0.0,>=7.0.1' 'flake8>=4.0.1' 'flakeheaven<4.0.0,>=3.2.1' 'html5lib<2.0,>=1.1' 'importlib_metadata<5.0.0,>=4.2.0; python_version >= "3.7" and python_version < "3.8"' 'isodate<0.7.0,>=0.6.1' 'isort<6.0.0,>=5.10.0' mypy==0.991 'networkx<3.0.0,>=2.6.2; (python_version >= "3.7" and python_version < "3.8") and extra == "networkx"' 'networkx<3.0.0,>=2.8.8; (python_version >= "3.8" and python_version < "4.0") and extra == "networkx"' 'pep8-naming<0.14.0,>=0.13.2' 'pyparsing<4.0.0,>=3.0.9' 'pytest-cov<5.0.0,>=4.0.0' 'pytest<8.0.0,>=7.1.3' pid=424278
py37-extensive: install_package> python -I -m pip install --force-reinstall --no-deps /home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/.tmp/package/2/rdflib-6.3.0a0.tar.gz
py37-extensive: exit 0 (2.35 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python -I -m pip install --force-reinstall --no-deps /home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/.tmp/package/2/rdflib-6.3.0a0.tar.gz pid=424344
py37-extensive: commands[0]> .tox/py37-extensive/bin/python -m mypy --show-error-context --show-error-codes --junit-xml=test_reports/mypy-junit.xml
Success: no issues found in 348 source files
py37-extensive: exit 0 (13.34 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> .tox/py37-extensive/bin/python -m mypy --show-error-context --show-error-codes --junit-xml=test_reports/mypy-junit.xml pid=424440
py37-extensive: commands[1]> .tox/py37-extensive/bin/python -m pytest -ra --tb=native ./test/test_extras/test_extras_external_graph_libs.py
platform linux -- Python 3.7.16, pytest-7.2.0, pluggy-1.0.0
cachedir: .tox/py37-extensive/.pytest_cache
rootdir: /home/iwana/sw/d/github.com/iafork/rdflib.reviews, configfile: pyproject.toml
plugins: cov-4.0.0
collected 2 items                                                                                                                                                           

test/test_extras/test_extras_external_graph_libs.py ss                                                                                                                [100%]

========================================================================== short test summary info ==========================================================================
SKIPPED [1] test/test_extras/test_extras_external_graph_libs.py:10: couldn't find networkx
SKIPPED [1] test/test_extras/test_extras_external_graph_libs.py:61: couldn't find graph_tool
============================================================================ 2 skipped in 0.01s =============================================================================
py37-extensive: exit 0 (0.27 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> .tox/py37-extensive/bin/python -m pytest -ra --tb=native ./test/test_extras/test_extras_external_graph_libs.py pid=424805
.pkg: _exit> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api
.pkg: exit None (0.00 seconds) /home/iwana/sw/d/github.com/iafork/rdflib.reviews> python /home/iwana/.local/pipx/venvs/tox/lib64/python3.11/site-packages/pyproject_api/_backend.py True poetry.core.masonry.api pid=424271
  py37-extensive: OK (24.19=setup[10.59]+cmd[13.34,0.27] seconds)
  congratulations :) (24.23 seconds)
```

aucampia:
Okay, I think the problem is with how tox is calling pip (see below). For berkeleydb it just puts berkeleydb, but for networkx it adds `extra == "networkx"` and nothing is setting that so it does not match.

Will enquire with tox somehow.

```console
20230103T222036 iwana@teekai.zoic.eu.org:~/sw/d/github.com/iafork/rdflib.reviews
$ .tox/py37-extensive/bin/python -I -m pip install 'berkeleydb<19.0.0,>=18.1.5' black==22.12.0 'coverage[toml]<8.0.0,>=7.0.1' 'flake8>=4.0.1' 'flakeheaven<4.0.0,>=3.2.1' 'html5lib<2.0,>=1.1' 'importlib_metadata<5.0.0,>=4.2.0; python_version >= "3.7" and python_version < "3.8"' 'isodate<0.7.0,>=0.6.1' 'isort<6.0.0,>=5.10.0' mypy==0.991 'networkx<3.0.0,>=2.6.2; (python_version >= "3.7" and python_version < "3.8") and extra == "networkx"' 'networkx<3.0.0,>=2.8.8; (python_version >= "3.8" and python_version < "4.0") and extra == "networkx"' 'pep8-naming<0.14.0,>=0.13.2' 'pyparsing<4.0.0,>=3.0.9' 'pytest-cov<5.0.0,>=4.0.0' 'pytest<8.0.0,>=7.1.3'
Ignoring networkx: markers '(python_version >= "3.7" and python_version < "3.8") and extra == "networkx"' don't match your environment
Ignoring networkx: markers '(python_version >= "3.8" and python_version < "4.0") and extra == "networkx"' don't match your environment
Requirement already satisfied: berkeleydb<19.0.0,>=18.1.5 in ./.tox/py37-extensive/lib64/python3.7/site-packages (18.1.5)
Requirement already satisfied: black==22.12.0 in ./.tox/py37-extensive/lib64/python3.7/site-packages (22.12.0)
Requirement already satisfied: coverage[toml]<8.0.0,>=7.0.1 in ./.tox/py37-extensive/lib64/python3.7/site-packages (7.0.2)
Requirement already satisfied: flake8>=4.0.1 in ./.tox/py37-extensive/lib/python3.7/site-packages (4.0.1)
Requirement already satisfied: flakeheaven<4.0.0,>=3.2.1 in ./.tox/py37-extensive/lib/python3.7/site-packages (3.2.1)
Requirement already satisfied: html5lib<2.0,>=1.1 in ./.tox/py37-extensive/lib/python3.7/site-packages (1.1)
Requirement already satisfied: importlib_metadata<5.0.0,>=4.2.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (4.2.0)
Requirement already satisfied: isodate<0.7.0,>=0.6.1 in ./.tox/py37-extensive/lib/python3.7/site-packages (0.6.1)
Requirement already satisfied: isort<6.0.0,>=5.10.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (5.11.4)
Requirement already satisfied: mypy==0.991 in ./.tox/py37-extensive/lib64/python3.7/site-packages (0.991)
Requirement already satisfied: pep8-naming<0.14.0,>=0.13.2 in ./.tox/py37-extensive/lib/python3.7/site-packages (0.13.2)
Requirement already satisfied: pyparsing<4.0.0,>=3.0.9 in ./.tox/py37-extensive/lib/python3.7/site-packages (3.0.9)
Requirement already satisfied: pytest-cov<5.0.0,>=4.0.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (4.0.0)
Requirement already satisfied: pytest<8.0.0,>=7.1.3 in ./.tox/py37-extensive/lib/python3.7/site-packages (7.2.0)
Requirement already satisfied: pathspec>=0.9.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from black==22.12.0) (0.10.3)
Requirement already satisfied: mypy-extensions>=0.4.3 in ./.tox/py37-extensive/lib/python3.7/site-packages (from black==22.12.0) (0.4.3)
Requirement already satisfied: tomli>=1.1.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from black==22.12.0) (2.0.1)
Requirement already satisfied: platformdirs>=2 in ./.tox/py37-extensive/lib/python3.7/site-packages (from black==22.12.0) (2.6.2)
Requirement already satisfied: typing-extensions>=3.10.0.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from black==22.12.0) (4.4.0)
Requirement already satisfied: typed-ast>=1.4.2 in ./.tox/py37-extensive/lib64/python3.7/site-packages (from black==22.12.0) (1.5.4)
Requirement already satisfied: click>=8.0.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from black==22.12.0) (8.1.3)
Requirement already satisfied: pyflakes<2.5.0,>=2.4.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from flake8>=4.0.1) (2.4.0)
Requirement already satisfied: mccabe<0.7.0,>=0.6.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from flake8>=4.0.1) (0.6.1)
Requirement already satisfied: pycodestyle<2.9.0,>=2.8.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from flake8>=4.0.1) (2.8.0)
Requirement already satisfied: urllib3 in ./.tox/py37-extensive/lib/python3.7/site-packages (from flakeheaven<4.0.0,>=3.2.1) (1.26.13)
Requirement already satisfied: colorama in ./.tox/py37-extensive/lib/python3.7/site-packages (from flakeheaven<4.0.0,>=3.2.1) (0.4.6)
Requirement already satisfied: toml in ./.tox/py37-extensive/lib/python3.7/site-packages (from flakeheaven<4.0.0,>=3.2.1) (0.10.2)
Requirement already satisfied: entrypoints in ./.tox/py37-extensive/lib/python3.7/site-packages (from flakeheaven<4.0.0,>=3.2.1) (0.4)
Requirement already satisfied: pygments in ./.tox/py37-extensive/lib/python3.7/site-packages (from flakeheaven<4.0.0,>=3.2.1) (2.14.0)
Requirement already satisfied: six>=1.9 in ./.tox/py37-extensive/lib/python3.7/site-packages (from html5lib<2.0,>=1.1) (1.16.0)
Requirement already satisfied: webencodings in ./.tox/py37-extensive/lib/python3.7/site-packages (from html5lib<2.0,>=1.1) (0.5.1)
Requirement already satisfied: zipp>=0.5 in ./.tox/py37-extensive/lib/python3.7/site-packages (from importlib_metadata<5.0.0,>=4.2.0) (3.11.0)
Requirement already satisfied: pluggy<2.0,>=0.12 in ./.tox/py37-extensive/lib/python3.7/site-packages (from pytest<8.0.0,>=7.1.3) (1.0.0)
Requirement already satisfied: exceptiongroup>=1.0.0rc8 in ./.tox/py37-extensive/lib/python3.7/site-packages (from pytest<8.0.0,>=7.1.3) (1.1.0)
Requirement already satisfied: packaging in ./.tox/py37-extensive/lib/python3.7/site-packages (from pytest<8.0.0,>=7.1.3) (22.0)
Requirement already satisfied: iniconfig in ./.tox/py37-extensive/lib/python3.7/site-packages (from pytest<8.0.0,>=7.1.3) (1.1.1)
Requirement already satisfied: attrs>=19.2.0 in ./.tox/py37-extensive/lib/python3.7/site-packages (from pytest<8.0.0,>=7.1.3) (22.2.0)

[notice] A new release of pip available: 22.2.2 -> 22.3.1
[notice] To update, run: /home/iwana/sw/d/github.com/iafork/rdflib.reviews/.tox/py37-extensive/bin/python -m pip install --upgrade pip

```

aucampia:
@RDFLib/core-reviewers please have a look when you can, let me know if you have any objections.

Taggin some others for feedback @rchateauneu @edmondchuc

tox.ini
aucampia: I'm okay with this

Dockerfile.devcontainer
aucampia: resolved in PR to @jclerman's branch.

aucampia:
> > Some thoughts ...
> > The main benefit we want from poetry is reproducible dev environments, which we can get from poetry's lockfile. However, for this to provide maximal value we really need to take this into account for testing also. I'm looking at options now.
> > Further, we should be as liberal with version restrictions that end up in the wheel, we should not really place upper limits unless we know later versions break, and the lower limits should be as low as possible.
> 
> My usual practice is to assume that (other) libraries use semantic versioning, such that we should assume that major version-changes include breaking changes and minor/patchlevel version-changes don't. Thus my use of "^X.Y.Z" in most of the version-specs in my proposed pyproject.toml. We haven't tested version "X+1" so we shouldn't advertise rdflib as being compatible with it.
> 
> That imposes more of a maintenance-burden on the rdflib side but also should make it more robust.
> 
> Regarding lower limits: Agreed. I don't know what's been tested though so my default approach was to specify recent versions of dependencies. If you know that earlier versions are compatible, it makes sense to allow for those earlier versions in pyproject.toml.

Fair enough, major version upper bounds are probably best, I will tune the lower limits tomorrow, I think we work with pyparsing 2.x but I will test first.

aucampia:
In trying to move the Docker image building to poetry I came across this bug:

- https://github.com/python-poetry/poetry-plugin-export/issues/34

It won't be fixed anytime soon. There are basically two options here, we can either remove the dev stuff from extras, or we have to use something like [`TomWright/dasel`](https://github.com/TomWright/dasel) or [`kislyuk/yq`](https://github.com/kislyuk/yq) to first remove the dev dependency group and then relock and then export.

I think of these the best would be to just use poetry dependency groups for dev related extras. This however means we will have to change how tox invokes things and how readthedocs builds things.

jclerman:
> In trying to move the Docker image building to poetry I came across this bug:
> 
> * [poetry export - should it include optional extras by default? python-poetry/poetry-plugin-export#34](https://github.com/python-poetry/poetry-plugin-export/issues/34)
> 
> It won't be fixed anytime soon. There are basically two options here, we can either remove the dev stuff from extras, or we have to use something like [`TomWright/dasel`](https://github.com/TomWright/dasel) or [`kislyuk/yq`](https://github.com/kislyuk/yq) to first remove the dev dependency group and then relock and then export.
> 
> I think of these the best would be to just use poetry dependency groups for dev related extras. This however means we will have to change how tox invokes things and how readthedocs builds things.

I do like the idea of using poetry dep-groups for dev-related dependencies; that's the pattern that poetry is really designed for, so things should be easier (and, as you note, better supported) if we do that.  It'll require other refactoring (again, as you note) which I'd hoped to segregate into a separate PR - but maybe it's best to just make all these related changes in a single PR.

aucampia:
@jclerman I think most problems are now solved in:

- https://github.com/jclerman/rdflib/pull/2

  (from https://github.com/aucampia/rdflib/tree/switch-to-poetry and https://github.com/RDFLib/rdflib/pull/2192)

I will go through everything again at least once before merging.

edmondchuc:
Not sure if the devcontainer setup has been tested yet but I can't seem to run `poetry install` without root. The mounted volume is owned by root. Maybe I've missed a step to set up the devcontainer correctly?

```bash
@edmondchuc ➜ /srv/workspace/ (switch-to-poetry) $ poetry install
The virtual environment found in /srv/workspace/.venv seems to be broken.
Recreating virtualenv rdflib in /srv/workspace/.venv
usage: virtualenv [--version] [--with-traceback] [-v | -q] [--read-only-app-data] [--app-data APP_DATA] [--reset-app-data] [--upgrade-embed-wheels] [--discovery {builtin}] [-p py] [--try-first-with py_exe]
                  [--creator {builtin,cpython3-posix,venv}] [--seeder {app-data,pip}] [--no-seed] [--activators comma_sep_list] [--clear] [--no-vcs-ignore] [--system-site-packages] [--symlinks | --copies] [--no-download | --download]
                  [--extra-search-dir d [d ...]] [--pip version] [--setuptools version] [--wheel version] [--no-pip] [--no-setuptools] [--no-wheel] [--no-periodic-update] [--symlink-app-data] [--prompt prompt] [-h]
                  dest
virtualenv: error: argument dest: the destination . is not write-able at /srv/workspace/.venv
```

```bash
@edmondchuc ➜ /srv/workspace/ (switch-to-poetry) $ ls -la /srv/workspace/.venv/
total 8
drwxr-xr-x   2 root   root 4096 Jan 11 00:57 .
drwxrwxrwx+ 16 vscode root 4096 Jan 11 00:57 ..
```

aucampia:
> Not sure if the devcontainer setup has been tested yet but I can't seem to run `poetry install` without root. The mounted volume is owned by root. Maybe I've missed a step to set up the devcontainer correctly?

No you have not, I think for me it works because I have rootless docker. The main reason for using a volume for .venv is so that the devcontainer venv does not conflict with your normal venv. I will see about changing how we deal with this. Now that I have learnt we can control the venv with just setting the VIRTUAL_ENV environment that may be an option.

edmondchuc:
Oh dear.. sorry, I thought I pushed to my fork

edmondchuc:
PR updating the `Dockerfile.devcontainer` is here https://github.com/jclerman/rdflib/pull/3

aucampia:
> PR updating the `Dockerfile.devcontainer` is here [jclerman#3](https://github.com/jclerman/rdflib/pull/3)

On this, i think the right solution is to remove these lines:

https://github.com/RDFLib/rdflib/blob/9b778b39a7759a9c94d5437e1d5f02196bdc9c51/docker-compose.yml#L12-L15

https://github.com/RDFLib/rdflib/blob/9b778b39a7759a9c94d5437e1d5f02196bdc9c51/docker-compose.yml#L24-L27

https://github.com/jclerman/rdflib/pull/3#discussion_r1067477252

> @edmondchuc I think I'm just trying to be too clever with what I did here.
> 
> The reasoning for adding these volumes were so that usage of the devcontainer does not contaminate your host environment while still keeping state outside of the container. This works with rootless docker on Linux, but I'm not sure how it works on MacOS and Windows - so I think best is maybe just to not try and be too clever. if it works for you with the following lines removed then we should maybe just remove them instead:
> 
> https://github.com/RDFLib/rdflib/blob/9b778b39a7759a9c94d5437e1d5f02196bdc9c51/docker-compose.yml#L12-L15
> 
> ```yaml
>       - dot-venv:/srv/workspace/.venv
>       - dot-tox:/srv/workspace/.tox
>       - dot-mypy-cache:/srv/workspace/.mypy_cache
>       - dot-pytest-cache:/srv/workspace/.pytest_cache
> ```
>

aucampia:
I made a PR to add a tox environment with minumum versions and lowered the lower bounds for most dependencies to match:

- https://github.com/jclerman/rdflib/pull/4

Dependabot should keep poetry.lock and pyproject.toml up to date, and that is what will run under normal circumstances, so if new versions are released we will get a dependabot PR which should pass or fail depending on if the  new version is okay with RDFLib.

This constraints should ensure that we don't inadvertently drop support for older versions of libraries and ensures and ensures that the minumum versions we set are actually valid.

aucampia:
I think after merging these two you can mark this as ready for review:
- https://github.com/jclerman/rdflib/pull/4
- https://github.com/jclerman/rdflib/pull/3

I will do another check and then we should be merged before Monday.

aucampia:
I will do a final review tonight and maybe wait a day or so for more reviews.

Thanks for all the effort and patience.

CC: tagging in case anyone has time to review @rchateauneu @edmondchuc @RDFLib/core-reviewers @mwatts15

aucampia:
Will fix the build error tonight

mwatts15:
Can't likely do a full review, but I can try out basic project setup stuff
and verify, if that helps.

aucampia:
> Can't likely do a full review, but I can try out basic project setup stuff and verify, if that helps.

@mwatts15 That will be appreciated if you have time.

cclauss (approved):
A few possible upgrades

aucampia:
Thanks for the review @cclauss

.github/workflows/docker-images.yaml
```diff
@@ -44,6 +44,10 @@ jobs:
         uses: arduino/setup-task@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Python Poetry Action
+        uses: abatilo/actions-poetry@v2.1.6
```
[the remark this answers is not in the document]
aucampia: bumped

.github/workflows/docker-images.yaml
```diff
@@ -44,6 +44,10 @@ jobs:
...
+        with:
+          poetry-version: 1.3.1
```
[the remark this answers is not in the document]
aucampia: bumped

aucampia:
@jclerman I bumped the versions now in:

- https://github.com/RDFLib/rdflib/pull/2187/commits/0a5db59b205d84ef46e39ce7363a93d96632dbd1

In a subsequent PR we can create a `devtools/requirements-gha.txt` with tox poetry, and then have dependabot manage it and use it to install tox and poetry in the github actions pipeline.

aucampia (approved):
Thanks for the effort and patience, I think poetry makes rdflib a lot easier to develop for newcomers, and I do prefer it to manually managing my venv.

There are some more things we can do in subsequent PRs but I think this is good to merge.

jclerman:
Thanks so much @aucampia for your huge effort refining my initial PR and helping get this over the finish line.  I agree that poetry will make development/contributions much easier for newcomers (:wave:), and looking forward to making contributions to the library itself once this is in place.

pyproject.toml
```diff
@@ -82,3 +160,70 @@ log_format = "%(asctime)s.%(msecs)03d %(levelname)-8s %(name)-12s %(filename)s:%
...
+    '.mypy_cache',
+    '.pytest_cache',
+    '.tox',
+    '.venv',
+    '.var',
+    '.github',
+    '_build',
+    'htmlcov',
+    'benchmarks',
+    'examples',       # No need to Black examples
```
mwatts15, line 183: minor: I think this comment may be misleading, since this is config for isort rather than Black.

mwatts15:
Possibly a Poetry noob issue, but after I do a `poetry install`, I can't run poetry anymore because `poetry install` uninstalls html5lib (I take it that's because poetry uninstalls extras you don't specify and `html` is an extra...). This is surprising to me -- I'd think I could `poetry install` again and basically poetry would maybe check what's installed is up to date (compare against poetry.lock and pyproject.toml, etc.).

mwatts15:
I made a pretty shallow review. I really like how so much of the project configuration is laid out in one file so I don't have to go hunting. Good stuff!

jclerman:
> Possibly a Poetry noob issue, but after I do a `poetry install`, I can't run poetry anymore because `poetry install` uninstalls html5lib (I take it that's because poetry uninstalls extras you don't specify and `html` is an extra...). This is surprising to me -- I'd think I could `poetry install` again and basically poetry would maybe check what's installed is up to date (compare against poetry.lock and pyproject.toml, etc.).

Thanks @mwatts15!  Yes, `poetry install` with no flags for extras does seem to remove any extras previously installed.  Not sure what you mean though by not being able to run poetry anymore though?

pyproject.toml
jclerman: Sounds reasonable to me - though we inherited that wording from the config that was in `setup.cfg`; we simply moved it to `pyproject.toml` here. I'll let @aucampia comment on whether we should update the wording here to say "No need to **lint** example", or do that in another PR.

mwatts15:
sorry, should have posted the stack trace. poetry seems to have an html5lib
dependency. it tries to import the library for all the commands I tried.

mwatts15:
> sorry, should have posted the stack trace. poetry seems to have an html5lib dependency. it tries to import the library for all the commands I tried.
Here's the error poetry gives me. It doesn't seem like html5lib should be needed really and especially doesn't seem like poetry should let you uninstall its own dependencies. I just did `pip install poetry` in a virtualenv (https://python-poetry.org/docs/#installing-manually), so I don't think I did anything weird.
```
% poetry --verbose                                                                                                                                                                                                                                                       1

  ModuleNotFoundError

  No module named 'html5lib'

  at env/lib/python3.10/site-packages/poetry/repositories/link_sources/html.py:22 in <module>
       18│ 
       19│ 
       20│ with warnings.catch_warnings():
       21│     warnings.simplefilter("ignore")
    →  22│     import html5lib
       23│ 
       24│ 
       25│ class HTMLPage(LinkSource):
       26│     def __init__(self, url: str, content: str) -> None:

The following error occurred when trying to handle this error:


  ModuleNotFoundError

  No module named 'html5lib'

  at env/lib/python3.10/site-packages/poetry/repositories/link_sources/html.py:22 in <module>
       18│ 
       19│ 
       20│ with warnings.catch_warnings():
       21│     warnings.simplefilter("ignore")
    →  22│     import html5lib
       23│ 
       24│ 
       25│ class HTMLPage(LinkSource):
       26│     def __init__(self, url: str, content: str) -> None:
```

aucampia:
> I just did `pip install poetry` in a virtualenv (https://python-poetry.org/docs/#installing-manually), so I don't think I did anything weird.

If you do `poetry install` in a venv (i.e. with `VIRTUAL_ENV` environment variable set) then poetry will use that virtualenv instead of it's own, and if `VIRTUAL_ENV` is set to the poetry virtual env then it will mess with poetry's dependencies.

I usually use pipx for poetry, as that then manages my venv for me, and then `poetry` works quite well. If you manually made a poetry venv just be sure to not have it active when running `poetry install`, rather use the full path to `${VIRTUAL_ENV}/bin/poetry` or add `${VIRTUAL_ENV}/bin` - just as long as VIRTUAL_ENV is not your poetry venv when `poetry install` runs.

pyproject.toml
aucampia: I changed it now on the PR branch, though actually we should probably be blacking and linting examples, but that can be sorted out in future PRs.

aucampia:
Planning to merge a bit later tonight (CET).

PR #2188, opened 2022-12-29
Fixing bug applying VALUES outside of a GROUP BY

robons:
# Summary of changes

Altering order in which aggregate variable aliases are renamed to user-defined variable names to ensure that when defining a `VALUES` pattern outside of a `GROUP BY`, the variables in the query are correctly joined to those defined in the `VALUES` pattern. 

The change fixes the following bug...

## Example of bug

```python
from pprint import pprint

from rdflib import Graph, URIRef, Literal, RDFS

graph = Graph()

graph.add((URIRef("http://example.com/something"), RDFS.label, Literal("Match on this.")))
graph.add((URIRef("http://example.com/something-else"), RDFS.label, Literal("Don't match on this.")))

results = graph.query("""
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT ?obj ?label
WHERE {
    ?obj rdfs:label ?label
}
GROUP BY ?obj ?label
VALUES ( ?obj ) {
    ( <http://example.com/something> )
}
""")

pprint([row.asdict() for row in results])
``` 

results in the following erroneous output:

```
[{'label': rdflib.term.Literal('Match on this.'),
  'obj': rdflib.term.URIRef('http://example.com/something')},
 {'label': rdflib.term.Literal("Don't match on this."),
  'obj': rdflib.term.URIRef('http://example.com/something-else')}]
```

I would expect the the `VALUES` pattern to have filtered the results down to only include the result where `?obj` is `http://example.com/something`, i.e.:

```
[{'label': rdflib.term.Literal('Match on this.'),
  'obj': rdflib.term.URIRef('http://example.com/something')}]
```

this is indeed the outcome when passing the data and query through the Apache JENA command line tools.

# Checklist

  the same change.
  - [x] (NA) Updated relevant documentation to avoid inaccuracies.
  - [x] (NA) Considered adding additional documentation.
  - [x] (NA) Considered adding an example in `./examples` for new features.
  - [x] (NA) Considered updating our changelog (`CHANGELOG.md`).


aucampia (approved):
Thanks for the PR @robons and apologies for the late review. The change and test looks good to me, so I'm going to merge it before Monday if there is no further feedback.

Issue #1722, opened 2022-02-23
SPARQL Sub-query inside sub-query fails


robons:
When I write a query which has a sub-query inside a sub-query (as per the following example) I don't get the results I would expect back. 

```python
import json
from typing import List
from rdflib import Graph, Literal, URIRef, RDFS
from rdflib.query import ResultRow

graph = Graph()
graph.add((URIRef("http://example.com/something"), RDFS.label, Literal("Some label")))

results: List[ResultRow] = list(
    graph.query(
        """
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT *
WHERE {   
    {
        SELECT *
        WHERE {
                {
                    SELECT ?label
                    WHERE {
                        [] rdfs:label ?label.
                    }
                }
        }
    }
}
        """
    )
)

print(json.dumps(results, indent=4))
```

This yields the following output:

```
[]
```

However, if I remove one level of sub-query from the SPARQL query I end up with results that I agree with:

```sparql
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT *
WHERE {   
    {
        SELECT ?label
        WHERE {
            [] rdfs:label ?label.
        }
    }
}
```

I get the result I would have expected from the first query:

```
[
    [
        "Some label"
    ]
]
```

I've tested both queries in Apache JENA and get the results that I'd expect so I suspect that this is a symptomatic of a bug in rdflib's SPARQL querying functionality somewhere.

version info:

rdflib 6.1.1
python 3.9.0

gargayush18:
I have found a method and solution to solve this particular problem. I have noticed that the when the multiple SELECT* is introduced, the nested Select* works but the results are not passed to the above SparQl query

gargayush18:
To solve this there are two possible paths: 
a) Preprocessing the Sparqual Query before evaluation 
b) Changing the way of the evaluation of the query i.e once the query is parsed then passing the results to the above query , in the case of the Select*. 

Following the path b may lead to change in the way in some other queries are parsed and evaluated. 
So it will be a safe approach to change the query in the preprocessing in such a way that it gives the desired output.

gargayush18:
Now if we change the query in the preprocessing , lets take the example of the above query itself, 

SELECT *
WHERE {   
    {
        SELECT *
        WHERE {
                {
                    SELECT ?label
                    WHERE {
                        [] rdfs:label ?label.
                    }
                }
        }
    }
}
        """
[...]

PR #2190, opened 2022-12-30
Fix for `SELECT *` inside `SELECT *` bug

robons:
# Summary of changes

Fixing use of `SELECT *` in sub-select within `SELECT *` parent query as discovered in https://github.com/RDFLib/rdflib/issues/1722.

Now when an instance of `SELECT *` is encountered, the query tree/plan builder now correctly considers the projected variables of any sub-select statements when deciding which variables should be projected out. 

- fixes https://github.com/RDFLib/rdflib/issues/1722

# Checklist

  the same change.
  - [x] (NA) Updated relevant documentation to avoid inaccuracies.
  - [x] (NA) Considered adding additional documentation.
  - [x] (NA) Considered adding an example in `./examples` for new features.
  - [x] (NA) Considered updating our changelog (`CHANGELOG.md`).


aucampia (approved):
Thanks for the PR @robons and apologies for the late review. The change and test looks good to me, so I'm going to merge it before Monday if there is no further feedback.

PR #2192, opened 2023-01-02
[DO NOT MERGE, DO NOT REVIEW] Switch to poetry

aucampia:
Just testing things so I don't create spam on #2187

- Changes to #2187: https://github.com/jclerman/rdflib/compare/switch-to-poetry...aucampia:rdflib:switch-to-poetry
- Changes to main: https://github.com/RDFLIB/rdflib/compare/main...aucampia:rdflib:switch-to-poetry


tox.ini
```diff
@@ -16,7 +16,7 @@ setenv =
     extensive: POETRY_ARGS_extensive = --extras=html --extras=berkeleydb --extras=networkx
     lxml: POETRY_ARGS_lxml = --extras=lxml
 commands_pre =
-    py3{7,8,9,10,11}: poetry --version
+    py3{7,8,9,10,11}: poetry lock --check
     py3{7,8,9,10,11}: poetry install --no-root --only=main --only=dev --only=tests --extras=html  {env:POETRY_ARGS_docs:} {env:POETRY_ARGS_extensive:} {env:POETRY_ARGS_lxml:} --sync
```
jclerman, line 20: There may be an issue with using "poetry install" here instead of "poetry export" followed by "pip install" using the exported requirements.  The problem is that "poetry install" will install into a poetry-controlled venv - but tox creates/uses its own virtual environments.  Even if the former method works to some extent, it'll start to be problematic with `tox -p`, since the various `poetry install` commands will all be installing into the same venv. That causes much pain, as I've learned.

jclerman: By the way, instead of the multiple specifications of "--only", which could be confusing, you could also do `--with=dev --with=tests` - `main` will be included by default.

aucampia: poetry seems to be installing into the environment variable indicated by `VIRTUAL_ENV` - so for `tox -e py37` it is using `.tox/py37`: https://gist.github.com/aucampia/acc2fb08b0f577a7592042e1021f8910#file-output-txt-L69-L83

I will commit shortly with more debugging output and quite from there, but as far as I can tell it is fine.

It was surprising to me that it worked when I noticed it but I think it makes some sense.  See also:

- https://github.com/python-poetry/poetry/issues/2056
- https://github.com/python-poetry/poetry/issues/281

This seems to be a consequence of:

- https://github.com/python-poetry/poetry/blob/3fe8838659415f7495d415503ff64eea1aa0bf57/src/poetry/utils/env.py#L691-L752

This is also how poetry recommends using it: https://python-poetry.org/docs/faq/#usecase-2

aucampia: Build log can be seen here: https://github.com/RDFLib/rdflib/actions/runs/3887504878/jobs/6633837906#step:8:224

For this commit: https://github.com/aucampia/rdflib/tree/2292b877281f0f29d20701c445064677c6551a31

aucampia:
Closing this as it has served its purpose.

PR #2209, opened 2023-01-17
build: rename minimum constraints file to evade dependabot

aucampia:
# Summary of changes

Dependabot is updating `devtools/constraints-min.txt`, but this file should be the minimum supported versions and should not be automatically updated.

Dependabot seems to operate on all .txt and .in files, so I'm changing it to be `devtools/requirements.min` to evade dependabot.

# Checklist

  the same change.


aucampia:
Will merge when CI passes.

CC: @jclerman

PR #2220, opened 2023-02-01
namespace.get_longest_namespace avoid concurrent modification issues

tgbugs:
When using async code that operates on namespaces it is possible for the namespacies trie dict to be modified in one coro while it is being iterated through in another, causing a RuntimeError. To avoid this we switch to always iterate over trie.keys(). This is safe because any concurrent modification will either have no impact or will duplicate the operation leaving the trie in the same state.

# Summary of changes

# Checklist

  the same change.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -842,7 +842,7 @@ def insert_strie(strie: Dict[str, Any], trie: Dict[str, Any], value: str) -> Non
 
 
 def get_longest_namespace(trie: Dict[str, Any], value: str) -> Optional[str]:
-    for key in trie:
+    for key in trie.keys():
         if value.startswith(key):
             out = get_longest_namespace(trie[key], value)
             if out is None:


aucampia (commented):
Hi @tgbugs - thanks for the PR, but I'm not entirely sure I understand the reasoning, though there may be some nuance that I'm missing.

From [Built-in Types — Mapping Types — dict](https://docs.python.org/3.7/library/stdtypes.html#dict)
> **iter(d)**
> Return an iterator over the keys of the dictionary. This is a shortcut for `iter(d.keys())`.

From [Built-in Types — Dictionary view objects](https://docs.python.org/3.7/library/stdtypes.html#dictionary-view-objects)

> The objects returned by [dict.keys()](https://docs.python.org/3.7/library/stdtypes.html#dict.keys), [dict.values()](https://docs.python.org/3.7/library/stdtypes.html#dict.values) and [dict.items()](https://docs.python.org/3.7/library/stdtypes.html#dict.items) are view objects. They provide a dynamic view on the dictionary’s entries, which means that when the dictionary changes, the view reflects these changes.

There may be nuance somewhere here that I'm missing though.

The change is probably fine in that under my interpretation it should have no impact, but I would rather not merge it unless I'm clearer on why it will help.

I think ultimately what is happening with data strucutres being modified in operations that really should only be reading data should be eliminated.

tgbugs:
In another place in the code near here I caste to a tuple to make doubly sure, but I can do a separate test to see what the behavior of all the variants is with regard to concurrent changes.

aucampia:
@tgbugs not sure if there is any update on this from your side, as I indicated, the behaviour here should be unchanged with the changes in your pull request. I think it may be best to close this PR unless there is something I missed.

aucampia:
Closing this PR as the change should have no impact on behaviour, happy to re-consider it with follow-up.

PR #2221, opened 2023-02-02
Add SPARQL DESCRIBE query implementation

### LINKED ISSUE #479: DESCRIBE query not working
Apparently DESCRIBE queries are not fully implemented yet? This is what I get with two different queries:
###### 

1) g.query('DESCRIBE http://www.example.org/a')
...
/usr/lib/python2.7/site-packages/rdflib/plugins/sparql/algebra.pyc in translate(q)
    530 
    531     # all query types have a where part
--> 532     M = translateGroupGraphPattern(q.where)
    533 
    534     aggregate = False

/usr/lib/python2.7/site-packages/rdflib/plugins/sparql/algebra.pyc in translateGroupGraphPattern(graphPattern)
    261     """
    262 
--> 263     if graphPattern.name == 'SubSelect':
    264         return ToMultiSet(translate(graphPattern)[0])
    265 

AttributeError: 'NoneType' object has no attribute 'name'

2) g.query('DESCRIBE ?s WHERE {?s ?p ?o FILTER (?s = http://www.example.org/a)}')

...
/usr/lib/python2.7/site-packages/rdflib/plugins/sparql/evaluate.pyc in evalPart(ctx, part)
    260 
    261     elif part.name == 'DescribeQuery':
--> 262         raise Exception('DESCRIBE not implemented')
    263 
    264     else:

Exception: DESCRIBE not implemented
###### 

The graph is created via this sequence:

import rdflib
from rdflib import RDF

ex = rdflib.Namespace('http://www.example.org/')
g = rdflib.Graph()
g.add((ex['a'], RDF['type'], ex['Cl']))

mgberg:
# Summary of changes
This adds an implementation for SPARQL DESCRIBE queries, using the built-in `cbd` method. I see there are several issues and PRs for DESCRIBE implementation. I believe this should close #479 and should resolve #1913, or at least pick up where it left off. It should also resolve #1096.

This implementation should support the full SPARQL specification for DESCRIBE; either explicit IRIs can be provided (with no WHERE clause), or variables projected from a WHERE clause can be provided, or variables projected from a WHERE clause AND explicit IRIs can be provided. If a WHERE clause is provided, it should be evaluated the same way as it would for a SELECT DISTINCT query (including dataset specifications).

The expected results for the test cases provided match the behavior seen when running the same queries against the same data using ARQ.

A possible future extension would be to add a global option (similar to `rdflib.plugins.sparql.SPARQL_LOAD_GRAPHS`) to change the method used to describe resources instead of always using CBD.

- Resolves <https://github.com/RDFLib/rdflib/issues/479>

# Checklist

  the same change.
  - [X] Updated relevant documentation to avoid inaccuracies.
  - [X] Considered adding additional documentation.
  - [X] Considered adding an example in `./examples` for new features.
  - [X] Considered updating our changelog (`CHANGELOG.md`).

### DIFF
--- a/rdflib/plugins/sparql/parser.py
+++ b/rdflib/plugins/sparql/parser.py
@@ -1479,7 +1479,7 @@ DescribeQuery = Comp(
     "DescribeQuery",
     Keyword("DESCRIBE")
     + (OneOrMore(ParamList("var", VarOrIri)) | "*")
-    + Param("datasetClause", ZeroOrMore(DatasetClause))
+    + ZeroOrMore(ParamList("datasetClause", DatasetClause))
     + Optional(WhereClause)
     + SolutionModifier
     + ValuesClause,
--- a/rdflib/plugins/sparql/algebra.py
+++ b/rdflib/plugins/sparql/algebra.py
@@ -335,7 +335,11 @@ def translateGroupGraphPattern(graphPattern: CompValue) -> CompValue:
     """
 
     if graphPattern.name == "SubSelect":
-        return ToMultiSet(translate(graphPattern)[0])
+        # The first output from translate cannot be None for a subselect query
+        # as it can only be None for certain DESCRIBE queries.
+        # type error: Argument 1 to "ToMultiSet" has incompatible type "Optional[CompValue]";
+        #   expected "Union[List[Dict[Variable, str]], CompValue]"
+        return ToMultiSet(translate(graphPattern)[0])  # type: ignore[arg-type]
 
     if not graphPattern.part:
         graphPattern.part = []  # empty { }
--- a/rdflib/plugins/sparql/algebra.py
+++ b/rdflib/plugins/sparql/algebra.py
@@ -623,7 +623,7 @@ def translateValues(
     return Values(res)
 
 
-def translate(q: CompValue) -> Tuple[CompValue, List[Variable]]:
+def translate(q: CompValue) -> Tuple[Optional[CompValue], List[Variable]]:
     """
     http://www.w3.org/TR/sparql11-query/#convertSolMod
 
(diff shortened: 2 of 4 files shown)


aucampia:
Just for reference, two previous abandoned attempts at doing this: 

- https://github.com/RDFLib/rdflib/pull/1913
- https://github.com/gjhiggins/rdflib/commit/5fb6f43e6b1d2264ac5015c260b32fc4e7dbf240#

rdflib/plugins/sparql/parser.py
```diff
@@ -1479,7 +1479,7 @@ def expandCollection(terms):
     "DescribeQuery",
     Keyword("DESCRIBE")
     + (OneOrMore(ParamList("var", VarOrIri)) | "*")
-    + Param("datasetClause", ZeroOrMore(DatasetClause))
+    + ZeroOrMore(ParamList("datasetClause", DatasetClause))
```
aucampia, line 1482: Does this make a meaningful difference? In most other cases it is as it was before, and the code that processes this remains unchanged as far as I can tell, so I would rather keep it the same unless there is a good reason to change it, which there may be, but I can't quite tell. If there is a good reason to change it then I think a comment would maybe be warranted so that others that come later can easily figure out why DESCRIBE is different from say SELECT in this regard.

aucampia (approved):
@mgberg thanks for the PR, it looks great as always, I just have a question regarding the change in `rdflib/plugins/sparql/parser.py` which I'm not entirely clear on.

The change should also be pretty safe as nothing should impact non-describe queries.

aucampia:
pre-commit.ci run

gjhiggins (approved):
lgtm

rdflib/plugins/sparql/parser.py
mgberg: Two points about this change:
- This does make a meaningful difference; if left the old way, the `datasetClause` attribute would be a single object if one `FROM` or `FROM NAMED` clause is present in the query, which would cause `evalQuery` to fail [here](https://github.com/RDFLib/rdflib/blob/63755b8330cd6ce7650130732cb13a4712131944/rdflib/plugins/sparql/evaluate.py#L642). The updated way would be a list of one element in that case which would allow `evalQuery` to work as intended.
- If you look [here](https://github.com/RDFLib/rdflib/blob/63755b8330cd6ce7650130732cb13a4712131944/rdflib/plugins/sparql/parser.py#L1431), you'll see that I simply copied what was already present for `SELECT` queries to use for `DESCRIBE`; `DESCRIBE` used to be different, but now it actually matches `SELECT`.

PR #2239, opened 2023-02-27
Implement bind_namespaces strategy for Prefix.cc

cthoyt:
# Summary of changes

This PR implements a web-dependent loader for prefix-namespace definitions from Prefix.cc by parsing its context document (https://prefix.cc/context.jsonld)

# Checklist

  the same change.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -6,6 +6,7 @@ from functools import lru_cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Set, Tuple, Union
 from unicodedata import category
+from urllib.request import urlopen
 from urllib.parse import urldefrag, urljoin
 
 from rdflib.term import URIRef, Variable, _is_valid_uri
@@ -418,11 +419,16 @@ class NamespaceManager(object):
             for prefix, ns in _NAMESPACE_PREFIXES_CORE.items():
                 self.bind(prefix, ns)
         elif bind_namespaces == "cc":
+            for prefix, ns in _NAMESPACE_PREFIXES_RDFLIB.items():
+                self.bind(prefix, ns)
+            for prefix, ns in _NAMESPACE_PREFIXES_CORE.items():
+                self.bind(prefix, ns)
             # bind any prefix that can be found with lookups to prefix.cc
-            # first bind core and rdflib ones
-            # work out remainder - namespaces without prefixes
-            # only look those ones up
-            raise NotImplementedError("Haven't got to this option yet")
+            response = urlopen("https://prefix.cc/context.jsonld")
+            context = json.loads(response.read())
+            for prefix, ns in context["@context"].items():
+                # note that prefixes are lowercase-only in prefix.cc
+                self.bind(prefix, ns)
         elif bind_namespaces == "core":
             # bind a few core RDF namespaces - default
             for prefix, ns in _NAMESPACE_PREFIXES_CORE.items():


rdflib/namespace/__init__.py
```diff
@@ -722,6 +726,13 @@ def absolutize(self, uri: str, defrag: int = 1) -> URIRef:
         return URIRef(result)
 
 
+def _get_prefix_cc():
+    """Get the context from Prefix.cc."""
+    response = urlopen("https://prefix.cc/context.jsonld")
+    context = json.loads(response.read())
+    return context["@context"]
```
aucampia, line 733: One concern here is that there are some valid "@context" values [[ref](https://www.w3.org/TR/json-ld/#context-definitions)] that will not be handled correctly here, it may be that they don't appear in prefix.cc, but it may be good to confirm that somehow.

aucampia (commented):
You will need to add `cc` to https://github.com/RDFLib/rdflib/blob/4aaaa6fa6a57af8a59ecc28fdcdbd1661941b5fa/rdflib/_type_checking.py#L29

And also update https://github.com/RDFLib/rdflib/blob/4aaaa6fa6a57af8a59ecc28fdcdbd1661941b5fa/rdflib/namespace/__init__.py#L373-L375

rdflib/namespace/__init__.py
cthoyt: yes, it's the case that there are no special values in this context dictionary based on the way that the context is constructed (https://github.com/cygri/prefix.cc/blob/cbc85c00e59e00cf4fee697374109fdd9027231a/templates/format/jsonld.php) and the strict requirements on prefixes (though right now I am having a hard time finding where it's documented that these have to be lowercase strings of alphanumeric characters length <= 10)

aucampia:
Will try to get this into the next release, I will make any changes needed directly to your branch.

cthoyt:
@aucampia thank you for the feedback! Feel free to make any edits you think are appropriate or request I make some updates.

aucampia:
@cthoyt I will likely make another patch release soon, so will hold back a bit on merging this.

aucampia (approved):
This looks good to me, though I may add a warning that indicates that the "cc" selector is offered as a best-effort feature and may stop working if `prefix.cc` stops working or if they change their interface.

rdflib/namespace/__init__.py
```diff
@@ -418,11 +418,14 @@ def __init__(self, graph: "Graph", bind_namespaces: "_NamespaceSetString" = "cor
             for prefix, ns in _NAMESPACE_PREFIXES_CORE.items():
                 self.bind(prefix, ns)
         elif bind_namespaces == "cc":
```
niklasl, line 420: Why use the TLD name for `prefix.cc`? `cc` is also the conventional prefix for [creativecommons](http://creativecommons.org/ns#), so this might be confusing.

(I would also prefer having this as a utility (e.g. `graph.bind_namespaces(util.get_prefix_cc()`) rather than a flag to Graph. That would be more explicit and lets the user control network access and caching.)

cthoyt: @niklasl I am happy to do whatever the RDFLib team decides on, but this interface and nomenclature was already predefined, so I just filled it in with an implementation as suggested.

I agree that since prefix.cc relies on a network connection that this is a bit of a tricky situation.

niklasl: Actually,  you can achieve the same effect using RDFLib as is, using just:
```python
from rdflib import Graph

graph = Graph()
graph.parse('https://prefix.cc/context.jsonld')

for pfx, ns in graph.namespaces():
    print(pfx, ns)
```
This yields one very interesting difference though: the `go` prefix won't work, as it ends in a `_`, which is not [treated as a namespace prefix in JSON-LD 1.1](https://www.w3.org/TR/json-ld/#compact-iris), since it does not en with a [URI gen-delim](https://www.rfc-editor.org/rfc/rfc3986#section-2.2) (it has to be explicitly declared using `"@prefix": true` in the context to be treated as a prefix anyway).

aucampia:
Since we had this in the docstring, I'm going to merge it (possibly with some minor changes), though possibly it would have been better in `rdflib._contrib`, in future I will channel functionality there instead.

aucampia:
> Since we had this in the docstring, I'm going to merge it (possibly with some minor changes), though possibly it would have been better in `rdflib._contrib`, in future I will channel functionality like this there instead.

Actually, I think it would be better to just move this into `rdflib._contrib`, some ways in which we can implement this is:

- Change type of bind_namespaces to also accept a mapping of prefix to string (similar to `rdflib.namespace._NAMESPACE_PREFIXES_RDFLIB`), add a function to `rdflib._contrib` that returns such a mapping for prefix.cc, this would then be used as follow:
  ```python
  from rdflib._contrib import get_prefixcc_mapping
  from rdflib import Graph
  
  g = Graph(bind_namespaces=rdflib._contrib.get_prefixcc_mapping())
  ```

- Change type of bind_namespaces to also accept a function that returns a mapping of prefix to string
  ```python
  from rdflib._contrib.namespace_providers import prefix_cc
  from rdflib import Graph
  
  g = Graph(bind_namespaces=prefix_cc)
  ```
  `rdflib._contrib.namespace_providers.prefix_cc` would then return the mapping when called.

Maybe something better can also be done, but I would rather have it in `rdflib._contrib` than explain in the doc string that we cannot cover this with normal guarantees for our public interface stability.

aucampia:

aucampia:
> 

cthoyt:
thanks for the comments until now, but I have run out of motivation to try and get this PR in

I would suggest changing the following line to link back to this PR, and maybe state that this is a low-priority (or remove entirely)

 https://github.com/RDFLib/rdflib/blob/a0be37fce44a8cf96da87a7181080f6c235dcf19/rdflib/namespace/__init__.py#L452

PR #2255, opened 2023-03-08
fix: small InputSource related issues

aucampia:
# Summary of changes

I have added a bunch of tests for `InputSource` handling, checking every kind of input source with most of the parsers. During this, I detected the following issues that I fixed:

- `rdflib.util._iri2uri()` should not URL quote the `netloc` parameter, the `idna` encoding already takes care of special characters. I removed the URL quoting of `netloc`
- HexTuple parsing was handling the input source in a way that would only work for some input sources, and not raising errors for other input sources. I changed the input source handling to be more generic.
- `rdflib.parser.create_input_source()` incorrectly uses `file.buffer` instead of `source.buffer` when dealing with IO stream sources.

Other changes with no runtime impact include:

- Changed the HTTP mocking stuff in test slightly to accommodate serving arbitrary files, as I used this in the `InputSource` tests.
- Don't use google in tests as we keep getting
  `urllib.error.HTTPError: HTTP Error 429: Too Many Requests`
  from it.

# Checklist

  the same change.


aucampia:
I did this while looking at <https://github.com/RDFLib/rdflib/issues/1844> - but this does not quite address anything there. I plan to make a separate PR with some security documentation and warnings, and maybe an example of how to use python auditing and `urllib.request.install_opener` to mitigate the issue. I'm going to consider that as closing the matter.

aucampia:
Reduced the size of changes a bit, will merge with no review as this is pretty well tested code and the non-test changes are very minor.

PR #2257, opened 2023-03-10
GH#2256 Introduce a query optimizer concept

JervenBolleman:
First implementation is to make it more likely that VALUES is used on the left side of a join.

Added a test, but the optimizer is not invoked by default at this time

# Summary of changes

Introduced a way to add new generic query optimizers that work on the query algebra.

# Checklist

  the same change.


rdflib/plugins/sparql/processor.py
```diff
@@ -63,8 +64,9 @@ def update(self, strOrQuery, initBindings={}, initNs={}):
 
 
 class SPARQLProcessor(Processor):
-    def __init__(self, graph):
+    def __init__(self, graph, optimizers: List[SPARQLOptimizer] = None):
```
aucampia, line 67: For simplicity sake I think it would be best to just accept any callable that maps a query to a query:

So somewhere before this:

```python
_QueryTranslatorType = Callable[[Query],Query]
```

Then:

```suggestion
    def __init__(self, graph, query_translators: Optional[List[_QueryTranslatorType]] = None):
```

rdflib/plugins/sparql/optimizer.py
```diff
@@ -0,0 +1,46 @@
...
+in an faster to evaluate version.
+"""
+
+
+class SPARQLOptimizer:
+    def optimize(self, query: Query) -> Query:
+        return query
+
+
+class ValuesToTheLeftOfTheJoin(SPARQLOptimizer):
```
aucampia, line 25: While this is valuable, I think it may be better to keep it in `rdflib._contrib`, as we don't necessarily want to offer the same level of compatibility guarantees as we do for other code.

rdflib/plugins/sparql/optimizer.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def optimize(self, query: Query) -> Query:
```
aucampia, line 27:
```suggestion
    @classmethod
    def optimize(cls, query: Query) -> Query:
```

As these methods don't use the class state and are side effect free it is best to make them class methods, that way it is clearer to users that they don't have to be concerned with concurrency issues.

rdflib/plugins/sparql/optimizer.py
```diff
@@ -0,0 +1,46 @@
...
+        main = query.algebra
+        query.algebra = self._optimize_node(main)
+        return query
+
+    def _optimize_node(self, cv: Any) -> Any:
```
aucampia, line 32:
```suggestion
    @classmethod
    def _optimize_node(cls, cv: Any) -> Any:
```

rdflib/plugins/sparql/processor.py
aucampia: Your tests should still work fine with that.

aucampia:
@JervenBolleman I made a PR against your branch with some changes: https://github.com/JervenBolleman/rdflib/pull/5

One problem with taking translators as an argument to `SPARQLProcessor.__init__` is that it is a bit awkward to call it, it may be better to take translators as an argument to `SPARQLProcessor.query()`, that way someone can just pass translators when calling `Graph.query()`.

There may be reasons to not do this also, and keep it in the constructor, but I'm not sure that there is.

CC: @niklasl @westurner @RDFLib/core-reviewers

JervenBolleman:
@aucampia thank you for helping with this. 

The reason why I did it via the constructor, is that stores normally know how to optimize queries the best. And if you just want to modify a query before sending it for evaluation, there is no need for the store to have any knowledge of the modification, or do it for the user. My natural habitat being RDF4j I think that the approach of query optimizer pipelines is natural, and they are store constructor based.

westurner:
If this is a config file parameter, constructor args may be easier than
passing those through for each .query()

I don't know whether it makes a difference for testing whether constructor
args create a default SPARQLProcessor/Optimizer on a public attr or
the/*every* store just has extra private methods and args/kwarg?

aucampia:
I will have to double-check how exactly this will be called then by passing it to the store, using just graph with the default store, however, the current approach requires this as far as I can tell:

https://github.com/JervenBolleman/rdflib/blob/4dfaaec2b599748360613f58bd6734da225ec7c5/test/test_sparql/test_translators.py#L65-L94

```python
def test_graph_query(rdfs_graph: Graph):
    requested_query_string = """
    PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
    SELECT ?x {
        ?x rdfs:label "subClassOf".
    }
    """

    translated_query_string = """
    PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
    SELECT ?x {
        ?x rdfs:label "subPropertyOf".
    }
    """

    requested_query = _prepare_query(requested_query_string)
    translated_query = _prepare_query(translated_query_string)

    translate = Mock(return_value=translated_query)

    processor = SPARQLProcessor(rdfs_graph, translators=[translate])
    result = rdfs_graph.query(requested_query, processor)
    rows = []
    for row in result:
        assert isinstance(row, ResultRow)
        rows.append(row.asdict())

    assert rows == [{"x": RDFS.subPropertyOf}]

    translate.assert_called_once_with(requested_query)
```

It may be worth adding an example of how users should be using this, and also adding a tests for the idiomatic end-to-end usage.

@JervenBolleman when you have a moment do check https://github.com/JervenBolleman/rdflib/pull/5 - happy to update it if there is something you want different.

westurner:
To not have to pass the processor as a positional arg on every query,:

```python
result = rdfs_graph.query(requested_query, processor)
```

(If it isn't already) couldn't it be:

```python
Store.__init__(*, processor:SPARQLProcessor=None)

store = Store(, processor=processor)
# ...
result = rdfs_graph.query(requested_query)
```

aucampia:
> To not have to pass the processor as a positional arg on every query,:
> 
> ```python
> result = rdfs_graph.query(requested_query, processor)
> ```
> 
> (If it isn't already) couldn't it be:
> 
> ```python
> Store.__init__(*, processor:SPARQLProcessor=None)
> 
> store = Store(, processor=processor)
> # ...
> result = rdfs_graph.query(requested_query)
> ```

Will check this weekend if this will work or can be made to work.

aucampia:

aucampia:
> 

PR #2261, opened 2023-03-11
feat: add type hints to `rdflib.path`

aucampia:
# Summary of changes

There are some deprecation warnings coming from here as `evalPath` is used internally. I want to fix them, but I want type checking to ensure I don't mess up, so this just adds type checking, will fix the deprecation warnings in a separate PR.

This PR does not include any runtime changes, which can be confirmed from looking at the compact diff shared [here](https://gist.github.com/aucampia/4bc97199a0cfcc308676e675a20b0f3a).

# Checklist

  the same change.

### DIFF
--- a/rdflib/_type_checking.py
+++ b/rdflib/_type_checking.py
@@ -18,6 +18,7 @@ import sys
 
 __all__ = [
     "_NamespaceSetString",
+    "_MulPathMod",
 ]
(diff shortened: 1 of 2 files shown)


aucampia:
planning to merge without review as this has no runtime impact.

PR #2265, opened 2023-03-12
more type-hinting for SPARQL plugin

jclerman:
# Summary of changes

Here, adding type-hints to some of the SPARQL parser plugin code.

Includes a couple of small consequent changes:

1. Minor refactor of `prettify_parsetree()`, separating the public-facing callable from the internal code that does not need to be public-facing.  That allows the public-facing callable to have more informative and restrictive type-hints for its arguments.
2. Added some test-coverage for `expandUnicodeEscapes()` - initially for my own understanding, but seems useful to leave it in place since I didn't see test-coverage for that function.

There should be no backwards-incompatible changes in this PR - at least, not intentionally.

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugins/sparql/parser.py
+++ b/rdflib/plugins/sparql/parser.py
@@ -8,7 +8,7 @@ import re
 import sys
 from typing import Any
 from typing import Optional as OptionalType
-from typing import TextIO, Tuple, Union
+from typing import IO, Tuple, Union
 
 from pyparsing import CaselessKeyword as Keyword  # watch out :)
 from pyparsing import (
@@ -1531,7 +1531,7 @@ def expandUnicodeEscapes(q: str) -> str:
     return expandUnicodeEscapes_re.sub(expand, q)
 
 
-def parseQuery(q: Union[str, bytes, TextIO]) -> ParseResults:
+def parseQuery(q: Union[str, bytes, IO]) -> ParseResults:
     if hasattr(q, "read"):
         q = q.read()
     if isinstance(q, bytes):
@@ -1541,7 +1541,7 @@ def parseQuery(q: Union[str, bytes, TextIO]) -> ParseResults:
     return Query.parseString(q, parseAll=True)
 
 
-def parseUpdate(q: Union[str, bytes, TextIO]):
+def parseUpdate(q: Union[str, bytes, IO]):
     if hasattr(q, "read"):
         q = q.read()
 
--- a/rdflib/plugins/sparql/parser.py
+++ b/rdflib/plugins/sparql/parser.py
@@ -6,6 +6,9 @@ based on pyparsing
 
 import re
 import sys
+from typing import Any
+from typing import Optional as OptionalType
+from typing import TextIO, Tuple, Union
 
 from pyparsing import CaselessKeyword as Keyword  # watch out :)
 from pyparsing import (
@@ -37,15 +40,15 @@ DEBUG = False
 # ---------------- ACTIONS
 
 
-def neg(literal):
+def neg(literal) -> rdflib.Literal:
     return rdflib.Literal(-literal, datatype=literal.datatype)
 
 
-def setLanguage(terms):
+def setLanguage(terms: Tuple[Any, OptionalType[str]]) -> rdflib.Literal:
     return rdflib.Literal(terms[0], lang=terms[1])
 
 
-def setDataType(terms):
+def setDataType(terms: Tuple[Any, OptionalType[str]]) -> rdflib.Literal:
     return rdflib.Literal(terms[0], datatype=terms[1])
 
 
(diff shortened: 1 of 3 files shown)


aucampia:
Thanks for the PR @jclerman, looks good from a very brief glance - I'm planning to release today, but will see if I can include this, I may make some changes directly to your branch though if needed, but i will try to keep it minimal.

jclerman:
Sure thing @aucampia. Total coincidence that I submitted this today; hadn't realized you were about to make a release - but it'd be great if this could be included!  Minor changes are fine, certainly. Not sure if the flake8 violation should be fixed or allowed - I named the test after a function, and because the function had capital letters in its name, the test does too. I don't mind breaking that symmetry if we'd prefer not to add more flake8 violations.

aucampia:
> Not sure if the flake8 violation should be fixed or allowed - I named the test after a function, and because the function had capital letters in its name, the test does too.

For those I will just add `noqa` comments, they were being ignored by flakeheaven baseline, but that resets once the code is edited, and then the generall approach is to just ignore them explicitly if they can't be fixed, or if fixing them is out of scope of the PR.

rdflib/plugins/sparql/parser.py
```diff
@@ -1508,25 +1511,27 @@ def expandCollection(terms):
...
-        except:  # noqa: E722
-            raise Exception("Invalid unicode code point: " + m)
+        except (ValueError, OverflowError) as e:
+            raise ValueError("Invalid unicode code point: " + m.group(1)) from e
 
     return expandUnicodeEscapes_re.sub(expand, q)
 
 
-def parseQuery(q):
+def parseQuery(q: Union[str, bytes, TextIO]) -> ParseResults:
```
aucampia, line 1534:
```suggestion
def parseQuery(q: Union[str, bytes, IO]) -> ParseResults:
```

This will work with both IO and TextIO, and as the functions are designed to work with bytes or str I think it is best to acceot TextIO and BinaryIO (i.e. `typing.IO`).

rdflib/plugins/sparql/parser.py
```diff
@@ -1536,7 +1541,7 @@ def parseQuery(q):
     return Query.parseString(q, parseAll=True)
 
 
-def parseUpdate(q):
+def parseUpdate(q: Union[str, bytes, TextIO]):
```
aucampia, line 1544:
```suggestion
def parseUpdate(q: Union[str, bytes, IO]):
```

rdflib/plugins/sparql/parser.py
```diff
@@ -6,6 +6,9 @@
 
 import re
 import sys
+from typing import Any
+from typing import Optional as OptionalType
+from typing import TextIO, Tuple, Union
```
aucampia, line 11:
```suggestion
from typing import IO, Tuple, Union
```

See below

rdflib/plugins/sparql/parser.py
```diff
@@ -1508,25 +1511,27 @@ def expandCollection(terms):
...
     Unicode code points may also be expressed using an \ uXXXX (U+0 to U+FFFF) or \ UXXXXXXXX syntax (for U+10000 onwards) where X is a hexadecimal digit [0-9A-F]
     """
 
-    def expand(m):
+    def expand(m: re.Match) -> str:
         try:
             return chr(int(m.group(1), 16))
-        except:  # noqa: E722
-            raise Exception("Invalid unicode code point: " + m)
+        except (ValueError, OverflowError) as e:
```
aucampia, line 1528: I'm not sure if `OverflowError` could be raised from here, not that it is that much of a problem if it can't but would like to know just for interest’s sake.

test/test_sparql/test_sparql.py
```diff
@@ -957,3 +957,25 @@ def test_sparql_describe(
...
+        ("1234", "1234", True),
+        (r"1234\u0050", "1234P", True),
+        (r"1234\u00e3", "1234\u00e3", True),
+        (r"1234\u00e3\u00e5", "1234ãå", True),
+        (r"1234\u900000e5", "", False),
+        (r"1234\u010000e5", "", False),
+        (r"1234\u001000e5", "1234\U001000e5", True),
+    ],
+)
+def test_expandUnicodeEscapes(arg: str, expected_result: str, expected_valid: bool):
```
aucampia, line 975: Not the most ideal to name it differently from the function, but I guess not clearly worse than a `# noqa`:

```suggestion
def test_expand_unicode_escapes(arg: str, expected_result: str, expected_valid: bool):
```

aucampia (approved):
I made a very small change, adding `BinaryIO` in addition to `TextIO`, as that should also work. Everything looks great otherwise, thanks again for this.

aucampia:
I plan to merge with only one review unless there is more feedback soon, as I want to finish the release today and there are more changes I want to get in for SPARQL typing, as the current typing in `Graph.{query,update}` is too strict.

rdflib/plugins/sparql/parser.py
jclerman: In my local testing I found that `OverflowError` could be raised for large 8-digit hexadecimal codepoint numbers - apparently because they can be larger than a 32-bit int.  I'm not sure why the max isn't 64-bit, but that might be a limitation caused by the use of `int()`.

Valid codepoints can't be larger than `\U10000000`, so it's kind of a moot point, except that due to the way the regex is written, the function will try to handle `\U99999999` and will trigger `OverflowError`.

PR #2267, opened 2023-03-12
Eliminate file intermediary in translate algebra

jclerman:
# Summary of changes

Previously, `rdflib.plugins.sparql.algebra.translateAlgebra()` maintained state via a file, with a fixed filename `query.txt`.  With this change, use of that file is eliminated; state is now maintained in memory (so that multiple concurrent `translateAlgebra()` calls, for example, should no longer interfere with each other).

The change is accomplished with no change to the client interface.  Basically, the actual functionality has been moved into a class, which is instantiated and used as needed (once per call to `algrebra.translateAlgebra()`).  The class is available for direct use, though that's not anticipated to be useful in normal operation.

# Checklist

  the same change.
  - [x] Considered adding additional documentation (added PyDoc for new class)


aucampia:
Thanks for the PR, quite important I would say, but will have to wait for the next release which hopefully will be not too distant in the future. I will do a thorough review next week.

aucampia (approved):
Thanks again for the PR, the change looks good, I only renamed `AlgebraTranslator` to `_AlgebraTranslator` so it is not part of our public API, and I made some small changes to the docstrings, mainly so that references to functions are picked up by Sphinx.

aucampia:
I will merge this after releasing 6.3.2

PR #2282, opened 2023-03-17
Include test in sdist

bollwyvl:
# Summary of changes

Congratulations on 6.3.0!

A perhaps minor regression from earlier versions is that the `sdist` does not include the `test` folder, which makes it harder for downstreams to use a single source of truth to build and test a reliable package. This restores the `test` folder for sdists.

While the tests are of course _large_, most users' agents (modern `pip`) will prefer fetching the `.whl`, so this should create minimal disturbance for most users.

# Checklist

  the same change.


aucampia (approved):
LGTM and thanks for the PR, I will make 6.3.1 with this tonight still.

aucampia:
> LGTM and thanks for the PR, I will make 6.3.1 with this tonight still.

That is, if kroki is going to play along :facepalm:

aucampia:
This is done now.

bollwyvl:
Thanks! We'll see if the conda-forge bot picks it up for me, but otherwise
will take a look some time later this weekend.

bollwyvl:
Yep, the bot [picked up the release](https://github.com/conda-forge/rdflib-feedstock/pull/26) just fine, but needs some help finding the new tests.

Of note: `definednamespace_creator` tests use a file from `examples`, but otherwise looking good!

mgorny:
> Of note: `definednamespace_creator` tests use a file from `examples`, but otherwise looking good!

Yeah, just wanted to report that it's missing. Could you add it too, please?

aucampia:
I will add examples to sdist, but I will also have a look at how you run the tests and see if we can't maybe run it similarly in our CI pipeline to ensure that it will work correctly for you.

mgorny:
For the record, I'm from Gentoo and not conda-forge, so it's probably different but our code right now is kinda ugly: https://github.com/gentoo/gentoo/blob/master/dev-python/rdflib/rdflib-6.3.0.ebuild#L42

Long story short:
- we strip `doctest-modules` option to disable tests using Internet
- we strip some `pytest-cov` related option because we run tests without coverage testing
- we explicitly deselect a bunch of tests (`EPYTEST_DESELECT`) and ignore one whole file (`EPYTEST_IGNORE`) that use Internet
- we pass `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1` because some plugin (it wasn't obvious which one and I haven't had time to check it more precisely) has caused the test suite to fail

That said, I haven't had time to recheck much of this, so perhaps there is a better way to solve some of these problems.

One thing that would really be helpful is having some marker on tests using Internet, so we could easily deselect them all.

aucampia:
@mgorny I have made a new issue <https://github.com/RDFLib/rdflib/issues/2293> and will look at it when I have time.

mgorny:
Thanks a lot!

PR #2284, opened 2023-03-17
docs: don't use kroki

aucampia:
# Summary of changes

The kroki server is currently experiencing some issues which breaks our
build, this change eliminates the use of kroki in favour of directly
using the generated SVG images which is checked into git alongside the
plantuml sources.

I also added a task to the Taskfile to re-generate the SVG images from
the plantuml sources by calling docker.

# Checklist

  the same change.


aucampia:
Merging without review as this has no runtime changes.

PR #2289, opened 2023-03-19
fix: include docs and examples in the sdist tarball

aucampia:
# Summary of changes

The sdists generated by setuptools included the `docs` and `examples` directories, and they are needed for building docs and running tests using the sdist.

This change includes these directories in the sdist tarball.

A `test:sdist` task is also added to `Taskfile.yml` which uses the sdists to run pytest and build docs.

# Checklist

  the same change.


aucampia:
Merging without review, as this has no runtime changes and only includes additional content in sdist.

Issue #2293, opened 2023-03-19
Provide an easy way to disable tests that use the internet


aucampia:
This is a request from Gentoo:

- https://github.com/RDFLib/rdflib/pull/2282#issuecomment-1475321347

> One thing that would really be helpful is having some marker on tests using Internet, so we could easily deselect them all.

Not sure what the best way is to implement this, but it would be useful.

aucampia:
We should probably add a custom marker (<https://docs.pytest.org/en/7.1.x/example/markers.html>), that way someone can do:

```
pytest -v -m "not uses_internet"
```

aucampia:
I guess the biggest question that remains is, what should we call the marker? The pytest docs use `webtest`, which I guess works, but not sure if that is best, I think `uses_internet` may be better, but if `webtest` is widely used then we should use that rather.

aucampia:
I will go for webtest:
- https://github.com/search?q=%22%40pytest.mark.webtest%22&type=code

niklasl:
Makes sense. In theory `uses_network` might be more proper, but in practise, RDFLib should probably only use http(s) for network access anyway.

mgorny:
Thank you!

aucampia:
> Thank you!

Hope it helps. If there are other things we can do to make it easier to work with our releases, please let me know.

PR #2295, opened 2023-03-19
test: add `webtest` marker to tests that use the internet

aucampia:
# Summary of changes

This is being done so that it is easier for downstream packagers to run the test
suite without requiring internet access.

To run only tests that does not use the internet, run `pytest -m "not webtest"`.

The validation workflow validates that test run without internet access by
running the tests inside `firejail --net=none`.

- Closes <https://github.com/RDFLib/rdflib/issues/2293>.

# Checklist

  the same change.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -52,6 +52,10 @@ jobs:
             os: ubuntu-latest
             TOX_EXTRA_COMMAND: "flake8 --exit-zero rdflib"
             TOXENV_SUFFIX: "-docs"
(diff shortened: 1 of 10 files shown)


aucampia:
Hopefully the coverage remains the same, if it does, then I guess we can go with this.

aucampia:
Okay, coverage remains the same, so I think this is good to go.

I did this now, mainly so that it is less annoying for downstream to package our releases.

gjhiggins (approved):
:+1:

PR #2301, opened 2023-03-20
Streamline typing and implementation of SPARQL processor

cthoyt:
# Summary of changes

This PR first removes situations where a mutable object (i.e., a dictionary) was used as the default value for functions in the `rdflib.plugins.sparql.processor` module and related code. It replaces these situations with the an "optional" that defaults to None, and is then handled within the function. Luckily, some of the code that the SPARQL Processor relied on already had this style, meaning not a lot of changes had to be made.

Second, this PR makes a small update to the logic in the SPARQL Processor's query function to simplify the if/else statement. This better mirrors the implementation in the UpdateProcessor

These changes should not affect any user-level code and should already be covered by existing tests. The relaxation of the typing should not affect any downstream type checking.

# Checklist

  the same change.


aucampia:
@cthoyt thanks for the PR, I'm only going to be merging bugfixes until next week as I want to release 6.3.2 with some regression fixes, but I will review this when I have time.

cthoyt:
Thanks @aucampia. I hope some of my recent PRs are more helpful than stressful for you. Please take your time, and thank you for all of your efforts. Cheers

rdflib/query.py
```diff
@@ -57,8 +57,8 @@ def __init__(self, graph: "Graph"):
     def query(  # type: ignore[empty-body]
         self,
         strOrQuery: Union[str, "Query"],  # noqa: N803
-        initBindings: Mapping["str", "Identifier"] = {},  # noqa: N803
-        initNs: Mapping[str, Any] = {},  # noqa: N803
+        initBindings: Optional[Mapping["str", "Identifier"]] = None,  # noqa: N803
+        initNs: Optional[Mapping[str, Any]] = None,  # noqa: N803
```
aucampia, line 61: While the current function signature is problematic due to its use of mutable data structures in argument defaults, One problem with changing this is, plugins that implement this as it stands could break if they now receive valid values according to the new function signature, as they don't expect `None` as a value.

We need to figure out a way to upgrade our plugin interfaces, I think the right approach is to create ProcessorV2, but we need to also consider what other changes we want to introduce before that.

I think it is best to change this back to what it was, though the rest of the PR seems good and quite helpful.

```suggestion
        initBindings: Mapping["str", "Identifier"] = {},  # noqa: N803
        initNs: Mapping[str, Any] = {},  # noqa: N803
```

aucampia (commented):
The PR looks good, but out of an abundance of caution, we should probably keep the signature of `rdflib.query.Processor.query` the same as it was.

rdflib/query.py
cthoyt: I see how this could be an issue. I will switch the defaults to retain the empty dictionary

cthoyt: Addressed in 259c10db and 0ad0cffe

aucampia: Done now, please have a look and let me know if things look okay, but essentially I just reverted https://github.com/RDFLib/rdflib/commit/259c10db10e5ead9f48da007cf30d28af1c9b9a9 and then changed `rdflib/query.py` back to what is in the main branch.

aucampia (approved):
Looks good to me now, only change is essentially restoring `rdflib/query.py` to the state from the main branch.

Issue #2120, opened 2022-09-21
Regression in 6.2.0 when URL source to be parsed contains a query string


nutjob4life:
`Graph.parse` in 6.2.0 is producing inconsistent statement counts when reading over HTTP versus 6.1.1 when the source URL contains a query string.

To reproduce:
```console
$ date -u
Wed Sep 21 16:57:54 UTC 2022
$ cd /tmp
$ python3.10 --version
Python 3.10.5
$ python3.10 -m venv venv
$ cd venv
$ bin/pip install --quiet --upgrade setuptools pip wheel rdflib==6.2.0
$ bin/python
Python 3.10.5 (main, Jun 23 2022, 17:15:25) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import rdflib
>>> a, b = rdflib.Graph(), rdflib.Graph()
>>> a.parse('https://bmdb.jpl.nasa.gov/rdf/example')
>>> b.parse('https://bmdb.jpl.nasa.gov/rdf/example?all=right')
>>> len(a), len(b), len(a) != len(b)
(2, 2, False)
```
This should be `(2, 4, True)`.

Using 6.1.1 produces the correct results:
```console
$ bin/pip install --quiet rdflib==6.1.1
$ bin/python
Python 3.10.5 (main, Jun 23 2022, 17:15:25) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import rdflib
>>> a, b = rdflib.Graph(), rdflib.Graph()
>>> a.parse('https://bmdb.jpl.nasa.gov/rdf/example')
>>> b.parse('https://bmdb.jpl.nasa.gov/rdf/example?all=right')
>>> len(a), len(b), len(a) != len(b)
(2, 4, True)
```

Apparently in 6.2.2 the query string part of the URL `?all=right` gets stripped out. It is necessary with certain web services APIs in order to select different sets of RDF statements.

aucampia:
Regression came in with <https://github.com/RDFLib/rdflib/pull/1902>, the problem is that we are using quote(), which is designed for quoting paths, for non-path elements.

The exact fix is not that simple, but not that complex either,  should have this fixed for 6.3.2

aucampia:
[...]

PR #2304, opened 2023-03-21
fix: IRI to URI conversion

aucampia:
# Summary of changes

The URI to IRI conversion was percentage-quoting characters that should not have been quoted, like equals in the query string. It was also not quoting things that should have been quoted, like the username and password components of a URI.

This change improves the conversion by only quoting characters that are not allowed in specific parts of the URI and quoting previously unquoted components. The safe characters for each segment are taken from [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986).

The new behavior is heavily inspired by
[`werkzeug.urls.iri_to_uri`](https://github.com/pallets/werkzeug/blob/92c6380248c7272ee668e1f8bbd80447027ccce2/src/werkzeug/urls.py#L926-L931) though there are some differences.

- Closes <https://github.com/RDFLib/rdflib/issues/2120>.

# Checklist

  the same change.

### DIFF
--- a/rdflib/util.py
+++ b/rdflib/util.py
@@ -522,32 +522,92 @@ def _coalesce(
     return default
 
 
+_RFC3986_SUBDELIMS = "!$&'()*+,;="
+"""
+``sub-delims`` production from `RFC 3986, section 2.2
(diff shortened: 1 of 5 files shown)


gjhiggins (approved):
lgtm :+1:

PR #2305, opened 2023-03-22
Update CONTRIBUTORS from commit history

gjhiggins:
Update CONTRIBUTORS from commit history, add .mailmap to fix early mis-attributed contributions

# Checklist


.mailmap
aucampia, line 1: @gjhiggins while this is useful for keeping track of committers, it may also be linking identifiers that users don't want linked, and exposing details people may not want exposed.

All of this info is in the commit log anyway, so probably it should be fine, but GitHub takes some care to add some hurdles to getting to the email addresses, and this may somewhat counteract that.

I will ask around a bit about this and see what people think.

gjhiggins: I don't think there are many current addresses in this list and the first one can alwas be set to `<devnull@localhost>`

aucampia (approved):
LGTM, must have been some effort, but I think it is very valuable to have an accurate list and mailmap.

aucampia:
I will merge tomorrow.

white-gecko (approved):

Issue #2303, opened 2023-03-21
rdflib/plugins/shared/jsonld/context.py UnboundLocalError: local variable 'new_base' referenced before assignment


crisely09:
We migrated from **rdflib 6.2.0 to 6.3.1** and now we are getting a completely new error initializing a `Context` instance.

See error below:

![Screenshot 2023-03-21 at 14 04 46](https://user-images.githubusercontent.com/3536415/226615151-bd88c1d6-fb07-466b-b9ad-1152b0a00330.png)

I believe the variable `new_base` was named just `base` in the previous version.

aucampia:
Thanks for the report @crisely09 - I will try have it fixed in 6.3.2.

aucampia:
@crisely09 could you share some more details of the code that triggered this?

crisely09:
> @crisely09 could you share some more details of the code that triggered this?

We inherit the class, we use it mainly to resolve documents. See https://github.com/BlueBrain/nexus-forge/blob/master/kgforge/core/commons/context.py

crisely09:
Great, I will try again on Monday after the release.
Thanks!

PR #2306, opened 2023-03-22
fix: JSON-LD context construction from a `dict`

aucampia:
# Summary of changes

A variable was only being initialized for string-valued inputs, but if a `dict` input was passed the variable would still be accessed, resulting in a `UnboundLocalError`.

This change initializes the variable always, instead of only when string-valued input is used to construct a JSON-LD context.

- Closes <https://github.com/RDFLib/rdflib/issues/2303>.

# Checklist

  the same change.


aucampia:
@hsolbrig please have a look if you have time, I think you know this code better than most.

Issue #2072, opened 2022-08-03
function call fails as there is a reference to args.target_namespace which I believe shoud be target_namespace


konstantin-mergenthaler:
https://github.com/RDFLib/rdflib/blob/cfa418074b27b12aac905ba266b002a237c5ff4c/rdflib/tools/defined_namespace_creator.py#L98

aucampia:
Thanks for the bug report @konstantin-mergenthaler, this should be fixed in the next release in mid october.

PR #2311, opened 2023-03-24
fix: reference to global inside `get_target_namespace_elements`

aucampia:
# Summary of changes

`get_target_namespace_elements` references the `args` global, which is not defined if the function is called from outside the module. This commit fixes that instead referencing the argument passed to the function.

- Closes <https://github.com/RDFLib/rdflib/issues/2072>.

# Checklist

  the same change.


gjhiggins (approved):
:+1:

PR #2312, opened 2023-03-25
fix: change the prefix for `https://schema.org/` back to `schema`

aucampia:
# Summary of changes

The default prefix for `https://schema.org/` registered with `rdflib.namespace.NamespaceManager` was inadvertently changed to `sdo` in 6.2.0, this however constitutes a breaking change, as code that was using the `schema` prefix would no longer have the same behavior. This change changes the prefix back to `schema`.

# Checklist

  the same change.

### DIFF
--- a/rdflib/namespace/__init__.py
+++ b/rdflib/namespace/__init__.py
@@ -364,8 +364,8 @@ class NamespaceManager(object):
     * rdflib:
         * binds all the namespaces shipped with RDFLib as DefinedNamespace instances
         * all the core namespaces and all the following: brick, csvw, dc, dcat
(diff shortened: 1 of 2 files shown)


gjhiggins (approved):
:+1:

tobiasschweizer:
I really liked that you changed the default prefix back to sdo (no TLS) previous to RDFlib 6.3.2 ;-)

Is there an easy way to override the default? I got this [hint](https://github.com/RDFLib/pySHACL/issues/118#issuecomment-1026252226) once but never tried it (lazy me).

Background: We are still using https://schema.org in our SHACL shapes and Ontospy to document them. everything is finde in the RDF sense but schema1 simply looks odd in the docs.

tobiasschweizer:
@aucampia Great, thanks for the clarification!

PR #2322, opened 2023-03-27
Add optional graph argument to Graph.cbd, use for DESCRIBE queries

mgberg:
# Summary of changes

I often find myself wanting to create subgraphs including descriptions of multiple resources, so I end up doing some pattern like this:
```python
src_graph: Graph  # A large graph for which I want a subgraph
out_graph = Graph()
for uri in resources:
    out_graph += src_graph.cbd(uri)
```
While this works fine, there is overhead every time `cbd` is called:
1. A new `Graph` object is created each time.
2. Once triples are added to that new graph, they must be copied over to `out_graph`, i.e. added to a graph a second time.

This PR adds an optional `target_graph` argument to the `cbd` method. If a `Graph` is provided for that argument, the computed CBD is added directly to that provided graph instead of creating a new graph. Consequently, the above snippet could be written like this to avoid that extra overhead:
```python
src_graph: Graph  # A large graph for which I want a subgraph
out_graph = Graph()
for uri in resources:
    src_graph.cbd(uri, out_graph)
```

If the `target_graph` argument is not provided, then it behaves the same as it currently does- it should not break any existing code.

One place where this situation occurs is in the SPARQL DESCRIBE implementation. This PR tweaks that implementation to use this new argument to improve performance.

# Checklist

  the same change.

### DIFF
--- a/docker/latest/requirements.in
+++ b/docker/latest/requirements.in
@@ -1,4 +1,4 @@
 # This file is used for building a docker image of hte latest rdflib release. It
 # will be updated by dependabot when new releases are made.
-rdflib==6.3.1
+rdflib==6.3.2
 html5lib
--- a/docker/latest/requirements.txt
+++ b/docker/latest/requirements.txt
@@ -10,7 +10,7 @@ isodate==0.6.1
     # via rdflib
 pyparsing==3.0.9
     # via rdflib
-rdflib==6.3.1
+rdflib==6.3.2
     # via -r docker/latest/requirements.in
 six==1.16.0
     # via
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "rdflib"
-version = "6.3.2"
+version = "6.4.0a0"
 description = """RDFLib is a Python library for working with RDF, \
 a simple yet powerful language for representing information."""
 authors = ["Daniel 'eikeon' Krech <eikeon@eikeon.com>"]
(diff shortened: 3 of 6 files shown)


aucampia:
@mgberg this seems pretty useful. Another way to approach this may be to extract something which computes CBDs into a separate function freestanding, function, and then have that operate on a source and target graph. The benefit there would be that the graph interface remains smaller.

I will think about it a bit.

@westurner @niklasl any feedback from you would be appreciated here.

westurner:
Maybe `def cbd(self, *, target_graph=None)`?

aucampia:
@mgberg it is slightly better to make pull requests from your personal account if you are making them across organizations, that way we can update your PR branch directly with suggestions.

The pull requests mostly looks fine, but I think making `target_graph` a keyword only argument is probably best.

I will make a PR against your branch doing that and adding some unit tests.

aucampia:
> @mgberg it is slightly better to make pull requests from your personal account if you are making them across organizations, that way we can update your PR branch directly with suggestions.
> 
> The pull requests mostly looks fine, but I think making `target_graph` a keyword only argument is probably best.
> 
> I will make a PR against your branch doing that and adding some unit tests.

PR here: https://github.com/corning-incorporated/rdflib/pull/2

mgberg:
> @mgberg it is slightly better to make pull requests from your personal account if you are making them across organizations, that way we can update your PR branch directly with suggestions.

I'll do that in the future!

aucampia (approved):
Change looks good to me, thanks @mgberg

PR #2336, opened 2023-04-09
refactor: eliminate unneeded `rdflib.compat` imports

aucampia:
# Summary of changes

Compatibility handling for `collections.abc.Mapping` and `collections.abc.MutableMapping` is not needed as RDFLib currently only support Python 3.7 and newer, and those classes are available from `collections.abc` in Python 3.7.

# Checklist

  the same change.


aucampia:
Merging this without review as this has no runtime impact.

PR #2337, opened 2023-04-09
refactor: eliminate unused imports

aucampia:
# Summary of changes

`XMLNS` was seen as unused in `rdflib/__init__.py` but this change adds it to `__all__` in that module so that it is not unused.

This change also removes actual unused imports in
`rdflib/namespace/__init__.py`.

# Checklist

  the same change.


aucampia:

PR #2338, opened 2023-04-09
refactor: narrow imports

aucampia:
# Summary of changes

This change narrows import so that things are imported from the Python module where they are defined instead of importing them from a module that re-exports them, e.g. change import of `Graph` to import from the `rdflib.graph` module instead of from the `rdflib` module. This helps avoid problems with circular imports.

# Checklist

  the same change.


aucampia:

PR #2339, opened 2023-04-09
refactor: eliminate inheritance from object

aucampia:
This change removes the redundant inheritance from `object` (i.e. `class Foo(object): pass`) that is no longer needed in Python 3 and is a relic from Python 2.

# Checklist

  the same change.


aucampia:

PR #2350, opened 2023-04-11
fix: eliminate bare `except:`

aucampia:
# Summary of changes

Replace bare `except:` with `except Exception`, there are some cases where it can be narrowed further, but this is already an improvement over the current situation.

This is somewhat pursuant to eliminating
[flakeheaven](https://github.com/flakeheaven/flakeheaven), as it no longer supports the latest version of flake8
[[ref](https://github.com/flakeheaven/flakeheaven/issues/132)]. But it also is just the right thing to do as bare exceptions can cause problems.

# Checklist

  the same change.


gjhiggins (approved):
:+1:

Issue #2348, opened 2023-04-10
NamespaceManager.expand_curie() does not handle the default prefix


jclerman:
I have found that rdflib's handling of cURIEs for entities within an ontology's own namespace is somewhat difficult.

One of the issues is already documented (https://github.com/RDFLib/rdflib/issues/2077) - namely, that NamesapceManager does not handle mapping multiple prefixes to a single URI.

I also see that a cURIE like `:my_class` is rejected by `expand_curie()` as invalid.  It seems like it should be valid though, based on https://www.w3.org/TR/curie/#s_syntax, in particular
> A host language MAY declare a default prefix value, or MAY provide a mechanism for defining a default prefix value. In such a host language, when the prefix is omitted from a CURIE, the default prefix value MUST be used.

rdflib does record the default prefix value, mapping it to the empty string, as can be seen by examining NamespaceManager.namespaces() - but that default value is never used in `expand_curie()`.  The problematic code appears to be https://github.com/RDFLib/rdflib/blob/5c3c78af33530298a7db2350304ce253370e91c1/rdflib/namespace/__init__.py#L618, which requires that the prefix in CURIE to be expanded has non-zero length - but that's in conflict with the (correct) storage of the default prefix in `namespaces()`.

The two issues make it difficult (impossible?) to expand a CURIE for an entity in the default namespace of an ontology:

 - The default namespace prefix is inaccessible per the above
 - A substitute prefix, mapping to the same namespace, is dropped by NamespaceManager, per https://github.com/RDFLib/rdflib/issues/2077.

Is there a workaround via which I could write a CURIE that maps to an item in a parsed ontology (e.g. one stored in RDF/XML), such that rdflib could expand it to the correct URI?

Issue #2077, opened 2022-08-11
rdflib 6.2.0 SPARQL parsing regression: multiple prefixes for same IRI not supported?


jclerman:
I have a standard block of SPARQL prefixes that I use in some of my code, and I started getting test failures when I tried to update from rdflib 6.1.1 to rdflib 6.2.0.  I noticed that the problem seems to be that somehow, support for multiple prefixes that resolve to the same URI seems to have been dropped.

Example block at the beginning of a SPARQL query:

```
PREFIX foo: <http://example.com#>
PREFIX bar: <http://example.com#>
```

The above should, as far as I know, be supported - but in rdflib 6.2.0, the `foo:` prefix will not be recognized when used in the query - and in fact, during parsing of the query, the `foo:` prefix is deleted.

The issue seems to be somewhere in `rdfllib.namespace.__init__.py`, in `NamespaceManager.bind()` - but I can't quite pin down what change is causing this new issue.

I do see this comment in the code, which was present in 6.1.1 but maybe it wasn't working as described?

```
        # Check if the bound_namespace contains a URI
        # and if so convert it into a URIRef for comparison
        # This is to prevent duplicate namespaces with the
        # same URI
```

however, if I correctly understand the intent behind that comment, it seems problematic - there is nothing wrong (as far as I know?) with multiple prefixes for the same URI in SPARQL.

aucampia:
> however, if I correctly understand the intent behind that comment, it seems problematic - there is nothing wrong (as far as I know?) with multiple prefixes for the same URI in SPARQL.

I agree with you that I think it is problematic, I don't really see anything wrong with multiple prefixes either, maybe one concern is that we could end up with ambiguity in what the output of serializers that support namespaces will be, but I don't really think that is reason enough to not support multiple prefixes mapping to the same URI.

aucampia:
Sorry for the accidental closure.

jclerman:
I've looked into this further, motivated by the issues I described in #2348.

The root of the problem is at least reflected in the fact that the `Store` class (in `rdflib/store.py`) defines the `prefix` method with the expectation that it'll return an `Optional[str]`.  I think instead it should be returning a `Set[str]` (which might be empty).

All of the implementations of `Store` would need to be updated with that in mind, including updates of their implementations of `bind()`.  For example, in the `Memory` store, we have

[...]

PR #2365, opened 2023-04-18
fix handling of URIs in the default namespace

jclerman:
**Fixes:** https://github.com/RDFLib/rdflib/issues/2348
**Adds failing test-case for:** https://github.com/RDFLib/rdflib/issues/2077 (marked xfail for now)

# Summary of changes

Addressing two issues with NamespaceManager:

1. Filed as https://github.com/RDFLib/rdflib/issues/2348: `NamespaceManager.expand_curie()` did not support expansion of CURIEs with a leading colon, even though those are valid and should resolve using the default namespace for a graph.
2.  Also addressing the converse case: generation of a CURIE (with leading colon, so that it can be handled by `expand_curie()` from a URI in the default namespace.  The new support is added with a new method, `curie()` on NamespaceManager, so as not to change (break) existing behavior of `NamespaceManager.qname()`.

Tests are included for both methods, including expected failure-cases for expand_curie().

 - this should fix https://github.com/RDFLib/rdflib/issues/2348
 - also addresses the converse case from #2348, mapping a URI in the default namespace to a CURIE, using a new method to avoid a breaking change to `qname()`.

# Checklist

  the same change.


test/data/owl/test_ontology.owl
aucampia, line 1: I'm a bit hesitant to include large RDF files data that does not have clear provenance because it is hard to evolve it, in part because it is hard to gauge whether changes to it are warranted or not.

My preference is to rather use data maintained by others and versioned, and then download them with `test/data/fetcher.py`, e.g.

https://github.com/RDFLib/rdflib/blob/e103078be6c75b2f9a1dc515abffcdf6a674407d/test/data/fetcher.py#L274-L277


In cases where there is a specific case that should be illustrated, the variants test harness is possibly best. I will try to find a suitable well known RDF file that can be used for the tests.

aucampia: I think [The Organization Ontology](https://www.w3.org/TR/vocab-org/) (<https://www.w3.org/ns/org.ttl> and <https://www.w3.org/ns/org.rdf>) would work, going to use them

aucampia: Actually, [PROV-O](https://www.w3.org/TR/prov-o/)  may be best, as that has a blank namespace for both XML and Turtle, while the ORG ontology only has it for Turtle.

aucampia: I think it is best to just set it up manually.

test/test_namespace/test_namespacemanager.py
```diff
@@ -484,3 +484,69 @@ def check() -> None:
...
+        # next case only works with fix for https://github.com/RDFLib/rdflib/issues/2348
+        (":chair", "http://www.example.org/ontologies/mini-ont#chair", None, ""),
+        pytest.param(
+            # failure case that should succeed once https://github.com/RDFLib/rdflib/issues/2077 is fixed
+            "mini-ont:chair",
+            "http://www.example.org/ontologies/mini-ont#chair",
+            None,
+            "",
+            marks=pytest.mark.xfail,
+        ),
```
aucampia, line 511: This is failing for a reason unrelated to that bug though, your owl file defines: `xmlns:mini-ont="http://example.org/ontologies/mini-ont#"` - and the check here is for `http://www.example.org/` instead of `http://example.org/`.

aucampia:
@jclerman please see if you are okay with <https://github.com/RDFLib/rdflib/pull/2365/commits/598f9679b27e18755d4fbc54772f23ab0af007a7>

> 
> * Changed the tests to use an in-module fixture instead of the
>   `test_ontology.owl` graph and removed the `test_ontology.owl` file. The
>   problem with arbitrary test data is that it is hard to evolve. If possible
>   test data should be kept as terse as possible. In this case I just created a
>   namespace manager with three namespaces added, of which only two will be
>   effective because of <https://github.com/RDFLib/rdflib/issues/2077>.
> * Added a warning to `NamespaceManager.curie` to warn users that it is not a
>   pure function and explain the specific side effects.

test/test_namespace/test_namespacemanager.py
aucampia: Fixed in https://github.com/RDFLib/rdflib/commit/598f9679b27e18755d4fbc54772f23ab0af007a7

test/data/owl/test_ontology.owl
aucampia: I changed to an in-module fixture in https://github.com/RDFLib/rdflib/commit/598f9679b27e18755d4fbc54772f23ab0af007a7

jclerman: The use of an in-module fixture is concise, more than the RDF file that I used.  The possible disadvantage, as I see it, is that it's less obvious (to someone who is reading the code) what we expect the code to do with an actual RDF file - one has to learn the test-fixture, and infer how it corresponds to the namespace-manager one would get by parsing an RDF file.

I don't have really strong feelings about it though - the main thing here is to provide the new functionality and show (in some way) that it works.

test/test_namespace/test_namespacemanager.py
jclerman: Ouch. Thanks for catching that!

aucampia:
> I think I'd like to update the PR to allow an optional `generate` parameter on `curie()`, and pass its value to `compute_qname()`'s parameter of the same name. My own preference would be to make the default value `False` - I think that would still make this technically an impure function, but at least the default behavior would be to leave the NamespaceManager unchanged.

A parameter like that would be nice, but I would not say essential. If the namespace manager was a blank slate, I would prefer to default to `generate=False`, but the existing behaviour of other functions makes it a bit controversial to do that, as they generate by default (i.e. have `generate=True` as a default). I think having `NamespaceManager.curie` behave the same as most other methods on `NamespaceManager` is probably best, even though it does not feel entirely right to default to `generate=True`.

So it is up to you, happy if you change this PR to add it, but it is also fine to add it later if you want.

test/data/owl/test_ontology.owl
aucampia: > The possible disadvantage, as I see it, is that it's less obvious (to someone who is reading the code) what we expect the code to do with an actual RDF file - one has to learn the test-fixture, and infer how it corresponds to the namespace-manager one would get by parsing an RDF file.

This is true. In this case, it was also a bit challenging understanding what `NamespaceManager.namespaces` would return for the RDF/XML file, another option would be to just have a very minimal RDF document in a python multiline string instead of the array I have now, but it does somewhat complicate the testing a bit as there are multiple things that have to go right.

gjhiggins (commented):
lgtm :+1:

jclerman:
All your comments make sense.  I think the generate option would be very useful, so I'm adding it in a new commit here, with the default=True for the reason you mentioned (and I agree with everything you said).

I updated the test function to test out the 3 possible settings of the parameter - not passed, True, and False.

aucampia (approved):
LGTM, thanks for fixing this bug and the new functionality.

PR #2366, opened 2023-04-19
Add a cbd_subject() method

Natureshadow:
# Summary of changes

This PR adds a new `Graph.cbd_subject()` method. It is the counter-part to `Graph.cbd()`: It checks whether the graph is a CBD for some resource and, if it is, returns the subject of that resource.

I need this for implementing federation between graphs, where incoming pushes shall be ckecked to be CBDs because CBDs are nice to test against ACLs (cf. https://socialhub.activitypub.rocks/t/formalizing-inbox-posts-to-be-cbds/3123)

It also adds a `Graph.roots()` method returning the roots of the graph, used as a utility in `cbd_subject`, but also useful otherwise.

# Checklist

  the same change.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -951,6 +951,12 @@ class Graph(Node):
         ):
             yield s, p, o
 
+    def roots(self) -> Generator["_SubjectType", None, None]:
+        """A generator of subjects that are roots of the graph"""
+        for s in self.subjects(unique=True):
+            if (None, None, s) not in self:
+                yield s
+
     @overload
     def value(
         self,
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1879,6 +1879,23 @@ class Graph(Node):
 
         return subgraph
 
+    def cbd_subject(self) -> Optional[_SubjectType]:
+        """Determine the subject for which the graph is a Concise Bounded Description
+
+        :return: The subject of the CBD or None
+        """
+        roots = set(self.roots())
+        if len(roots) != 1:
+            # A CBD has exactly one root
+            return None
+        root = roots.pop()
+
+        real_cbd = self.cbd(root)
+        if not real_cbd.isomorphic(self):
+            return None
+
+        return root
+
 
 _ContextType = Graph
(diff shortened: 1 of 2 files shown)


aucampia:
Thanks for the PR @Natureshadow, as this is an enhancement I will want some feedback from other maintainers also, and I will still think about it a bit and maybe extend the docstrings a bit just to make sure I understand the intent.

gjhiggins (approved):
Interesting use case, can see your requirement and its potential generality. Also nice to see RDFLib being used to contribute to the development of the fediverse and impressed that Melvin is both taking an interest and endorsing your stance.

aucampia:

aucampia:
> 

aucampia:
`Graph.cbd_subject()` is a very special case I think, it may be useful in some select circumstances, but I think it will be the first thing we drop if we ever rework things as the surface area of RDFLib is already way too big for the maintenance resources. I will be open to integrating something like this in `rdflib._contrib` where we don't have any API guarantees.

I will leave this open but as of right now my main priority is improving quality.

ashleysommer:
I've reviewed this, and I'm not particularly sure of its utility.

I don't think "roots" of a graph is a real linked-data or RDF concept. It looks like it returns all nodes that are in the subject position in any triple in the graph, but never in the object position of any triple in the graph.

That really only works where properties are top-down, eg a DCAT Catalog, where all children defined on the catalog with `dcterms:hasPart`. It doesn't work where children reference their parents, eg, a Dataset where all records use `void:inDataset`. In this case `graph.roots()` doesn't return the Dataset.

The assertion that a CBD graph will contain exactly 1 root is also incorrect. `graph.cbd(uri)` will also include in the graph triples where ?uri is the object (that is part of the definition of concise bounded description, or DESCRIBE in SPARQL). So for _most_ CBD graphs, the graph.roots() call will return zero items.

Finally, the use of `cbd_subjects()` seems like a very special case like @aucampia said. I feel like you could do the same with some careful `graph.isomorphic()` calls in your application.

Issue #2382, opened 2023-05-12
HTTP 308 handling does not correctly change Host header in request.


aucampia:
RDFLib handles HTTP 308 responses with this code:

https://github.com/RDFLib/rdflib/blob/cbd61510ec581aee262bc2bb8ad95d94f7784842/rdflib/parser.py#L270-L283

This is because Python's urllib did not support 308 handling [[ref](https://github.com/python/cpython/issues/84501)] before Python 3.11 [[ref](https://docs.python.org/3/whatsnew/changelog.html)]:

> ### Python 3.11.0 alpha 2
> ...
> [bpo-40321](https://bugs.python.org/issue?@action=redirect&bpo=40321): Adds support for HTTP 308 redirects to [urllib](https://docs.python.org/3/library/urllib.html#module-urllib). See [RFC 7538](https://datatracker.ietf.org/doc/html/rfc7538.html) for details. Patch by Jochem Schulenklopper.

However, the way RDFLib handles 308 does not accommodate changes in HTTP host.

The following code does the same as RDFLib for HTTP 308 errors:

```python
from typing import Any
from urllib.error import HTTPError
from urllib.request import HTTPHandler, HTTPSHandler, Request, build_opener
from urllib.response import addinfourl

opener = build_opener(HTTPHandler(debuglevel=1), HTTPSHandler(debuglevel=1))


def _opener(req: Request) -> Any:
    try:
        return opener.open(req)
    except HTTPError as ex:
        if ex.code == 308:
            req.full_url = ex.headers.get("Location")
            return _opener(req)
        else:
            raise


response: addinfourl = _opener(Request("http://www.w3.org/ns/adms.ttl"))
```

Running it with Python 3.10 gives this output:

```console
$ VIRTUAL_ENV=.venv/py310 poetry run python var/check.py
[... elided ...]
send: b'GET /w3c/ns/adms.ttl HTTP/1.1\r\nAccept-Encoding: identity\r\nHost: uri.semic.eu\r\nUser-Agent: Python-urllib/3.10\r\nConnection: close\r\n\r\n'
reply: 'HTTP/1.1 308 Permanent Redirect\r\n'
header: Connection: close
header: Location: https://uri.semic.eu/w3c/ns/adms.ttl
header: Server: Caddy
header: Date: Fri, 12 May 2023 09:12:53 GMT
header: Content-Length: 0
[...]

PR #2389, opened 2023-05-17
fix: HTTP 308 Permanent Redirect status code handling

aucampia:
# Summary of changes

Change the handling of HTTP status code 308 to behave more like `urllib.request.HTTPRedirectHandler`, most critically, the new 308 handling will create a new `urllib.request.Request` object with the new URL, which will prevent state from being carried over from the original request.

One case where this is important is when the domain name changes, for example, when the original URL is `http://www.w3.org/ns/adms.ttl` and the redirect URL is `https://uri.semic.eu/w3c/ns/adms.ttl`. With the previous behaviour, the redirect would contain a `Host` header with the value `www.w3.org` instead of `uri.semic.eu`, because the `Host` header is placed in `Request.unredirected_hdrs` and takes precedence over the `Host` header in `Request.headers`.

Other changes:
- Only handle HTTP status code 308 on Python versions before 3.11 as Python 3.11 will handle 308 by default [[ref](https://docs.python.org/3.11/whatsnew/changelog.html#id128)].
- Move code which uses `http://www.w3.org/ns/adms.ttl` and `http://www.w3.org/ns/adms.rdf` out of `test_guess_format_for_parse` into a separate parameterized test which instead uses the embedded http server.

  This allows the test to fully control the `Content-Type` header in the response instead of relying on the value that the server is sending.

  This is needed because the server is sending `Content-Type: text/plain` for the `adms.ttl` file, which is not a valid RDF format, and the test is expecting `Content-Type: text/turtle`.

Fixes:
- <https://github.com/RDFLib/rdflib/issues/2382>.

# Checklist

  the same change.

### DIFF
--- a/rdflib/_networking.py
+++ b/rdflib/_networking.py
@@ -1,0 +1,117 @@
+from __future__ import annotations
+
+import string
+import sys
+from typing import Dict
+from urllib.error import HTTPError
+from urllib.parse import quote as urlquote
+from urllib.parse import urljoin, urlsplit
+from urllib.request import HTTPRedirectHandler, Request, urlopen
+from urllib.response import addinfourl
+
+
+def _make_redirect_request(request: Request, http_error: HTTPError) -> Request:
+    """
+    Create a new request object for a redirected request.
+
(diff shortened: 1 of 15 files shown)


aucampia:
Merging this without reviews, as it is quite well-tested code, and the problem is breaking the CI.

Happy for post-merge feedback if any is forthcoming.

PR #2390, opened 2023-05-19
fix: bugs with `rdflib.extras.infixowl`

aucampia:
# Summary of changes

Fix the following issues in `rdflib.extras.infixowl`:
- getting and setting of max cardinality only considered identifiers and not other RDF terms.
- The return value of `manchesterSyntax` was wrong for some cases.
- The way that `BooleanClass` was generating its string representation (i.e. `BooleanClass.__repr__`) was wrong for some cases.

Other changes:
- Added an example for using infixowl to create an ontology.
- Updated infixowl tests.
- Updated infixowl documentation.

# Checklist

  the same change.


aucampia:
@gjhiggins please see if you are okay with the changes I made/ All changes can be seen here:

- <https://github.com/gjhiggins/rdflib/compare/infixowl-fix...aucampia:rdflib:aucampia/20230519T1021-infix-fixes>

Of particular import may be:

- <https://github.com/RDFLib/rdflib/pull/2390/commits/5130e365442ef142e8a9b9f0b5666e73303c663c>

  > Add back Individual.serialize
  > 
  > Removing this would be a breaking change, it should be marked as
  > deprecated instead and if it is being replaced by another function then
  > it should call the other function.
  >

rdflib/extras/infixowl.py
```diff
@@ -1560,7 +1581,10 @@ def _getExtent():  # noqa: N802
 
 class Callable:
     def __init__(self, anycallable):
-        self.__call__ = anycallable
+        self.__callfn__ = anycallable
+
+    def __call__(self, *args, **kwargs):
+        return self.__callfn__(*args, **kwargs)
```
aucampia, line 1587: @gjhiggins I'm not entirely clear what the purpose of this change is, is it really needed? Does it make any real difference?

I also think it would be best to use `self._callfn` instead of `self.__callfn__`, as we should not use dunder methods for our own internal purposes as they later could be used for something else by python. It is probably not that serious, but unless there is a good reason to use `__callfn__` I would rather change it to `_callfn`.

gjhiggins (approved):
:+1:

rdflib/extras/infixowl.py
aucampia: I will change `__callfn__` to `_callfn` and merge.

aucampia: Done now.

PR #2391, opened 2023-05-19
docs: recommend making an issue before making an enchancement

aucampia:
# Summary of changes

Suggest that contributors first make an issue to get in principle agreement for pull requests before making the pull request.

Enhanements can be controvarsial and we may reject the enhancement sometimes even if the code is good, as it may just not be deemed important enough to increase the maintenance burder of RDFLib.

# Checklist

  the same change.


gjhiggins (approved):
well worth making it clear and explicit :+1:

Issue #2404, opened 2023-05-22
Triples from the default graph in RDF documents do not go into the default_context of Dataset or ConjunctiveGraph


aucampia:
I expect data from the default graph in RDF documents to go into the default context/graph of Dataset and ConjunctiveGraph, but they don't.

Given my expectation, this test should pass:

```python
import logging
from typing import Type, Union

import pytest
from pytest_check import check

from rdflib.graph import DATASET_DEFAULT_GRAPH_ID, ConjunctiveGraph, Dataset, Graph
from rdflib.term import BNode, URIRef

RDF_DOCUMENT = """
@prefix egdo: <http://example.org/>.

{
    egdo:gds0 egdo:note "This is a triple inside the default graph." .
}

egdo:gds0 egdo:note "This is an additional note on a triple which is in the default graph." .

egdo:gds1 egdo:note "This is a triple inside the default graph." .

egdo:g0 {
    egdo:g0s0 egdo:note "This is a triple inside a graph named <http://example.org/g0>." .
}

egdo:g1 {
    egdo:g1s0 egdo:note "This is a triple inside a graph named <http://example.org/g1>." .
}
"""


@pytest.mark.parametrize(
    ["container_type"],
    [
        (Dataset,),
        (ConjunctiveGraph,),
    ],
)
def test_issue(container_type: Union[Type[Dataset], Type[ConjunctiveGraph]]) -> None:
    logging.debug("container_type = %s", container_type)
    container = container_type()

    if container_type is Dataset:
        # An empty dataset has 1 default graph and no named graphs, so 1 graph in
        # total.
        with check:
            assert 1 == sum(1 for _ in container.contexts())
        with check:
            assert DATASET_DEFAULT_GRAPH_ID == next((context.identifier for context in container.contexts()), None)
        with check:
            assert container.default_context == next(container.contexts(), None)
    else:
        with check:
            assert isinstance(container.default_context.identifier, BNode)

    # Load an RDF document with triples in three graphs into the dataset.
    container.parse(data=RDF_DOCUMENT, format="trig")

    context_identifiers = set(context.identifier for context in container.contexts())

    logging.info("context_identifiers = %s", context_identifiers)
    logging.info(
        "container.default_context.triples(...) = %s",
        set(container.default_context.triples((None, None, None))),
    )

    all_contexts = set(container.contexts())
    logging.info("all_contexts = %s", (context.identifier for context in all_contexts))

[...]

Issue #2375, opened 2023-04-30
Dataset adds graph labels where there were none


sdasda7777:
If you run the following code

```Python
from rdflib import Dataset

input_data = "<http://s> <http://v> <http://o>."
ds = Dataset()
ds.parse(data=input_data, format="nquads")
print(ds.serialize(format="nquads"))
```

you will get something like `<http://s> <http://v> <http://o> _:Nf7773353eb67460ab92041894df76998 .`

I believe that is wrong, since there was no graph label in the input data, and RDF standard clearly distinguishes between the default graph and named graphs.

Is this a bug, or is there a way to make it not do that? If it is necessary to have some graph in there internally, I think I saw some tools use specific IRIs like `http://www.w3.org/1999/02/22-rdf-syntax-ns#defaultGraph`, but in that case there wouldn't be a reason not to leave those out when serializing.

aucampia:
This is essentially the same as this issue:
- https://github.com/RDFLib/rdflib/issues/1804

aucampia:
Fix is more or less ready, please have a look:
- https://github.com/RDFLib/rdflib/pull/2406

Issue #1804, opened 2022-04-11
rdfpipe adds a graph name to triples in the default graph, consequently breaking round tripping


aucampia:
I would expect to be able to round-trip quads with rdfpipe, but this does not work because rdfpipe labels quads in the default graph with a graph name and then serializes them with the injected graph name.

```console
~/sw/d/github.com/iafork/rdflib.cleanish
$ cat ./test/w3c/trig/trig-turtle-03.trig 
# Turtle is TriG
prefix : <http://example/> 

[ :p 123 ; :q 456 ] :r 1 .
```

```console
~/sw/d/github.com/iafork/rdflib.cleanish
$ pipx run --spec git+https://github.com/RDFLib/rdflib.git@master#egg=rdflib rdfpipe -i trig -o nquads  ./test/w3c/trig/trig-turtle-03.trig
_:n33872095eabf41d58f85dc3a7c672883b1 <http://example/p> "123"^^<http://www.w3.org/2001/XMLSchema#integer> <file:///home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/w3c/trig/trig-turtle-03.trig> .
_:n33872095eabf41d58f85dc3a7c672883b1 <http://example/q> "456"^^<http://www.w3.org/2001/XMLSchema#integer> <file:///home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/w3c/trig/trig-turtle-03.trig> .
_:n33872095eabf41d58f85dc3a7c672883b1 <http://example/r> "1"^^<http://www.w3.org/2001/XMLSchema#integer> <file:///home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/w3c/trig/trig-turtle-03.trig> .
```

```console
~/sw/d/github.com/iafork/rdflib.cleanish
$ pipx run --spec git+https://github.com/RDFLib/rdflib.git@master#egg=rdflib rdfpipe -i trig -o nquads  ./test/w3c/trig/trig-turtle-03.trig | pipx run --spec git+https://github.com/RDFLib/rdflib.git@master#egg=rdflib rdfpipe -i nquads -o trig -

@prefix ns1: <file:///home/iwana/sw/d/github.com/iafork/rdflib.cleanish/test/w3c/trig/> .
@prefix ns2: <http://example/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ns1:trig-turtle-03.trig {
    [] ns2:p 123 ;
[...]

PR #2406, opened 2023-05-23
BREAKING CHANGE: Don't use `publicID` as the name for the default graph.

aucampia:
# Summary of changes

When parsing data into a `ConjunctiveGraph` or `Dataset`, the triples in the
default graphs in the sources were loaded into a graph named `publicID`.

This behaviour has been changed, and now the triples from the default graph in
source RDF documents will be loaded into `ConjunctiveGraph.default_context` or
`Dataset.default_context`.

The `publicID` parameter to `ConjunctiveGraph.parse` and `Dataset.parse`
constructors will now only be used as the base URI for relative URI resolution.

- Fixes https://github.com/RDFLib/rdflib/issues/2404
- Fixes https://github.com/RDFLib/rdflib/issues/2375
- Fixes https://github.com/RDFLib/rdflib/issues/436
- Fixes https://github.com/RDFLib/rdflib/issues/1804

# Checklist

  the same change.
  
### DIFF
--- a/docs/upgrade6to7.rst
+++ b/docs/upgrade6to7.rst
@@ -28,7 +28,7 @@ into a named graph, use the following code:
     from rdflib import ConjunctiveGraph
 
     cg = ConjunctiveGraph()
-    cg.get_context("example:graph_name").parse("http://example.com/source.trig", format="trig")
+    cg.get_context("example:graph_name").parse("http://example.com/source.ttl", format="turtle")
 
 If you want to move triples from the default graph into a named graph, use the
 following code:
--- a/docs/upgrade6to7.rst
+++ b/docs/upgrade6to7.rst
@@ -1,0 +1,45 @@
+.. _upgrade4to5: Upgrading from RDFLib version 6 to 7
+
+============================================
+Upgrading 6 to 7
+============================================
+
+New behaviour for ``publicID`` in ``parse`` methods.
+----------------------------------------------------
+
+Before version 7, the ``publicID`` argument to the
+:meth:`~rdflib.graph.ConjunctiveGraph.parse` and
+:meth:`~rdflib.graph.Dataset.parse` methods was used as the name for the default
+graph, and triples from the default graph in a source were loaded into the graph
+named ``publicID``.
+
+In version 7, the ``publicID`` argument is only used as the base URI for relative
+URI resolution as defined in `IETF RFC 3986
+<https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.4>`_.
+
+To accommodate this change, ensure that use of the ``publicID`` argument is
+consistent with the new behaviour.
+
+If you want to load triples from a format that does not support named graphs
+into a named graph, use the following code:
+
+.. code-block:: python
+    
+    from rdflib import ConjunctiveGraph
+
+    cg = ConjunctiveGraph()
+    cg.get_context("example:graph_name").parse("http://example.com/source.trig", format="trig")
+
+If you want to move triples from the default graph into a named graph, use the
+following code:
+
+.. code-block:: python
+
+    from rdflib import ConjunctiveGraph
+
+    cg = ConjunctiveGraph()
+    cg.parse("http://example.com/source.trig", format="trig")
+    destination_graph = cg.get_context("example:graph_name")
+    for triple in cg.default_context.triples((None, None, None)):
+        destination_graph.add(triple)
+        cg.default_context.remove(triple)
--- a/test/test_conjunctivegraph/test_conjunctive_graph.py
+++ b/test/test_conjunctivegraph/test_conjunctive_graph.py
@@ -22,7 +22,7 @@ def test_bnode_publicid():
     b = BNode()
     data = "<d:d> <e:e> <f:f> ."
     print("Parsing %r into %r" % (data, b))
-    g.parse(data=data, format="turtle", publicID=b)
+    g.get_context(b).parse(data=data, format="turtle", publicID=b)
 
     triples = list(g.get_context(b).triples((None, None, None)))
     if not triples:
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Help with maintenance of all of the RDFLib family of packages is always welcome 
 
 ## Versions & Releases
 
-* `6.4.0a0` current `main` branch
+* `7.0.0a0` current `main` branch
 * `6.x.y` current release and support Python 3.7+ only. Many improvements over 5.0.0
     * see [Releases](https://github.com/RDFLib/rdflib/releases)
 * `5.x.y` supports Python 2.7 and 3.4+ and is [mostly backwards compatible with 4.2.2](https://rdflib.readthedocs.io/en/stable/upgrade4to5.html).
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,6 +55,7 @@ apidoc_output_dir = "apidocs"
 
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
 autodoc_default_options = {"special-members": True}
+autodoc_inherit_docstrings = True
 
 # https://github.com/tox-dev/sphinx-autodoc-typehints
 always_document_param_types = True
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -66,6 +66,7 @@ If you are familiar with RDF and are looking for details on how RDFLib handles i
    namespaces_and_bindings
    persistence
    merging
+   upgrade6to7
    upgrade5to6
    upgrade4to5
    security_considerations
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "rdflib"
-version = "6.4.0a0"
+version = "7.0.0a0"
 description = """RDFLib is a Python library for working with RDF, \
 a simple yet powerful language for representing information."""
 authors = ["Daniel 'eikeon' Krech <eikeon@eikeon.com>"]
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1400,26 +1400,26 @@ class Graph(Node):
            :doc:`Security Considerations </security_considerations>`
            documentation.
 
-        :Parameters:
-
-          - ``source``: An InputSource, file-like object, or string. In the case
-            of a string the string is the location of the source.
-          - ``location``: A string indicating the relative or absolute URL of
-            the source. Graph's absolutize method is used if a relative location
+        :param source: An `InputSource`, file-like object, `Path` like object,
+            or string. In the case of a string the string is the location of the
+            source.
+        :param location: A string indicating the relative or absolute URL of the
+            source. `Graph`'s absolutize method is used if a relative location
             is specified.
-          - ``file``: A file-like object.
-          - ``data``: A string containing the data to be parsed.
-          - ``format``: Used if format can not be determined from source, e.g.
+        :param file: A file-like object.
+        :param data: A string containing the data to be parsed.
+        :param format: Used if format can not be determined from source, e.g.
             file extension or Media Type. Defaults to text/turtle. Format
             support can be extended with plugins, but "xml", "n3" (use for
             turtle), "nt" & "trix" are built in.
-          - ``publicID``: the logical URI to use as the document base. If None
+        :param publicID: the logical URI to use as the document base. If None
             specified the document location is used (at least in the case where
-            there is a document location).
-
-        :Returns:
-
-          - self, the graph instance.
+            there is a document location). This is used as the base URI when
+            resolving relative URIs in the source document, as defined in `IETF
+            RFC 3986
+            <https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.4>`_,
+            given the source document does not define a base URI.
+        :return: ``self``, i.e. the :class:`~rdflib.graph.Graph` instance.
 
         Examples:
 
@@ -2206,15 +2206,18 @@ class ConjunctiveGraph(Graph):
         **args: Any,
     ) -> "Graph":
         """
-        Parse source adding the resulting triples to its own context
-        (sub graph of this graph).
+        Parse source adding the resulting triples to its own context (sub graph
+        of this graph).
 
         See :meth:`rdflib.graph.Graph.parse` for documentation on arguments.
 
+        If the source is in a format that does not support named graphs it's triples
+        will be added to the default graph (i.e. `Dataset.default_context`).
+
         :Returns:
 
-        The graph into which the source was parsed. In the case of n3
-        it returns the root context.
+        The graph into which the source was parsed. In the case of n3 it returns
+        the root context.
 
         .. caution::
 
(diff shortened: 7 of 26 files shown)


aucampia:
I marked this ready for review as I want to get feedback on the essential changes.

I will probably clean up the docstrings a bit still.

test/test_conjunctivegraph/test_conjunctive_graph.py
gjhiggins, line 1: I believe this is a supported use of publicID. The test succeeds with current main but fails with the change to `Graph.parse`:
```python
def test_parse_graph_with_publicid_as_new_cg_subgraph():
    from test.data import tarek, hates, cheese
    cg = ConjunctiveGraph()
    assert len(list(cg.contexts())) == 0  # Empty contexts exclded from listing
    # Add a statement to the default graph/context
    cg.add((tarek, hates, cheese))
    assert len(cg) == 1
    assert len(list(cg.contexts())) == 1
    assert list(cg.contexts())[0] == cg.default_context
    # Parse ttl graph into subgraph identified with publicID
    cg.parse(
        data="<urn:example:tarek> <urn:example:likes> <urn:example:pizza> .",
        publicID="urn:x-rdflib:context-a",  # Now ignored
        format="ttl",
    )
    # New subgraph should be created
    assert len(list(cg.contexts())) == 2
    # And the newly-created subgraph wth the publicID in the above turtle should exist
    assert URIRef("urn:x-rdflib:context-a") in [c.identifier for c in list(cg.contexts())]
```

aucampia: You are right, the test in question will work on main and fail on this PR branch, and this is because on main `publicID` is used as the identifier for the default graph in addition to the base for relative URI resolution in source documents, while in this PR branch, it is only used as the base for relative URI resolution.

With the issues that this PR tries to address, what we need to answer is:

1. How should parse work? If it should work the way it works in main, we close the issues and move on, if not, we have to change how it works, which means the behaviour will be different after changing it.
2. What should we do about the issues this tries to address?

I think the way it is working in main, i.e. using publicID as the name of the graph that triples inside the default graph is loaded into, is wrong. This is also not what the documentation of publicID suggests will happen.

https://github.com/RDFLib/rdflib/blob/ad56044cf3033f8d2a3c625d322f3d091d7c9322/rdflib/graph.py#L1416-L1418

Given that description, what I would expect is that the `publicID` is what relative URIs will be resolved against. And that is also what it is used for, and the only thing it should be used for, I think. And I don't see why the base for URI relative URI resolution should be the same as the graph name that default triples are being loaded into, which should not be named to begin with because as per the spec, has no name:

https://www.w3.org/TR/rdf11-concepts/#section-dataset

> Exactly one default graph, being an [RDF graph](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-graph). The default graph does not have a name and may be empty.

Another way to think about this is, what is worse:

1. Not having the content of default graph in sources loaded into the default graph of Dataset/ConjunctiveGraph
2. Having to explicitly load something into a sub-graph if you don't want it in the default graph

I think 1 is worse.

aucampia: All that being said, the long and short of it is: How should parse work and how do we address the issues at hand, not how does parse work in main.

aucampia: Also, if you think the code should work the way it does in main for loading TTL into a ConjunctiveGraph, how should it work when you change `ttl` to `trig` in your code?

I think if it is difficult to find consensus on what `{Dataset,ConjunctiveGraph}.parse(...)` should do if the source does not support labelled graphs, the best option is to just refuse to parse sources that don't support labelled graphs into Datasets or ConjuctiveGraphs. So then the right behaviour for your example would be to raise an exception, as the operation cannot be defined in an unambiguous way.

aucampia: It is also important to ask if something gets us into a better state or a worse state, or just moves sideways. I think given the amount of issues that require API breakages we won't be in a perfect state ever, and it is not realistic to aim for perfection anyway, as it is unattainable.

aucampia: I added some documentation and updated the description to make it clearer.

edmondchuc: I agree with the new changes here. `publicID` is used to resolve relative IRIs (I actually never knew it was used as the name of the graph to load the source's default graph statements for `ConjunctiveGraph/Dataset`).

So what is the behaviour now with this change when loading source data in `ttl` format into a `Dataset`?

ashleysommer:
Moving a topic from the code review thread into the main Issue thread:
@aucampia 
> I think if it is difficult to find consensus on what {Dataset,ConjunctiveGraph}.parse(...) should do if the source does not support named graphs, the best option is to just refuse to parse sources that don't support named graphs into Datasets or ConjuctiveGraphs. So then the right behaviour for your example would be to raise an exception, as the operation cannot be defined in an unambiguous way.

This is an issue I have been facing in my PySHACL application. In the past I used to use `Graph.parse()` for parsing input files, but in order to support `trig and nquads` files, I switched to using `Dataset.parse()`  for all input documents.

This gave arise to an issue when parsing ttl and ntriples files, where the Dataset has a default context graph (named `DATASET_DEFAULT_GRAPH_ID`) but that graph is empty, and the contents are placed into a _new_ graph in the Dataset. This makes targeting the correct triples difficult because you need to determine which graph in the Dataset they are in.

For me personally the ideal behaviour would be:
1. If calling `Dataset.parse()` on a non-named-graph file (eg, turtle, ntriples) _and_ 
2. If the Dataset has a default context graph named `DATASET_DEFAULT_GRAPH_ID` _and_
3. If the default context graph is empty, has no triples in it
Then: remove the empty default graph, parse the contents into a new graph, and set that new graph as the default context on the Dataset.

I know that would introduce inconsistent behaviour when any of those conditions differ slightly.

test/test_conjunctivegraph/test_conjunctive_graph.py
gjhiggins: > I actually never knew it was used as the name of the graph to load the source's default graph statements for `ConjunctiveGraph/Dataset`.

Me neither until I started to investigate the ramifications of the change - but as aucampia has noted, it *is* documented.

It's probably worth bearing in mind that the implementation of `ConjunctiveGraph` predates the W3's definition of Dataset. In RDFLib, a `Graph` object *always* has an identifier so I guess it was both unavoidable and natural for the `Graph` object that was bound to `ConjunctiveGraph.default_context` to be able to have a user-definable identifier (somewhere in the code/issues someone notes the difficulty of distinguishing a BNode-identified default graph from a BNode-identified named graph).

And this is probably why `ConjunctiveGraph` can behave a little unintuitively when passed an identifier:

```python
cg = ConjunctiveGraph()
assert cg.default_context.identifier == cg.identifier  # False
# AssertionError: assert BNode('Nd463bea483a649c6918c222b20ad8832') == BNode('N1feef746236f44cea47c83ed58f3e752')

cg = ConjunctiveGraph(identifier=URIRef('urn:example:context-0'))
assert cg.default_context.identifier == cg.identifier  # True
```

`Dataset` is a different beast and benefits from [a concrete assertion](https://www.w3.org/TR/2014/REC-rdf11-concepts-20140225/#section-dataset):
> “Exactly one default graph, being an RDF graph. The default graph does not have a name.”

Unlike `ConjunctiveGraph`, `Dataset.__init__()` [does not take an `identifier` param](https://github.com/RDFLib/rdflib/blob/ad56044cf3033f8d2a3c625d322f3d091d7c9322/rdflib/graph.py#L2389), a BNode is bound to the `Dataset` object’s `identifier` property but the (ahem, *unnamed*, sigh) default graph has a fixed, unchageable `identifier` property bound to `<urn:x-rdflib:default>`:

```python
ds = Dataset()
assert isinstance(ds.identifier, BNode)
assert ds.default_context.identifier == URIRef('urn:x-rdflib:default')
```
(I s'pose it's *technically* user-modifiable by rebinding [DATASET_DEFAULT_ID](https://github.com/RDFLib/rdflib/blob/ad56044cf3033f8d2a3c625d322f3d091d7c9322/rdflib/graph.py#L2263))

> So what is the behaviour now with this change when loading source data in ttl format into a Dataset?

Current state-of-play is unchanged, source ttl is loaded into the `Graph` bound to `default_context`: 
```python
ds = Dataset()
ds.parse(location=TEST_DATA_DIR / "variants" / "diverse_triples.ttl", format="ttl")
assert len(ds) == 5
assert list(ds.contexts()) == [ds.default_context]
assert len(ds.default_context) == 5
```

gjhiggins:
> .. an issue when parsing ttl and ntriples files, where the Dataset has a default context graph (named `DATASET_DEFAULT_GRAPH_ID`) but that graph is empty, and the contents are placed into a _new_ graph in the Dataset.

Yeah, that's wrong, the contents of statements without a context (i.e. triples) *shouldn't* be placed in an arbitrary new graph, it's a behaviour inherited from a longstanding [bug in ConjunctiveGraph](https://github.com/RDFLib/rdflib/issues/436) which aucampia has addressed with this PR. 

`ConjunctiveGraph` docstring clearly sets expectations for users:

> All methods that add triples work against this default graph.

and `parse` is a method that adds triples.

> For me personally the ideal behaviour would be:
> 
>     1. If calling `Dataset.parse()` on a non-named-graph file (eg, turtle, ntriples) _and_
> 
>     2. If the Dataset has a default context graph named `DATASET_DEFAULT_GRAPH_ID` _and_
> 
>     3. If the default context graph is empty, has no triples in it
>        Then: remove the empty default graph, parse the contents into a new graph, and set that new graph as the default context on the Dataset.
> 
> 
> I know that would introduce inconsistent behaviour when any of those conditions differ slightly.

What we have atm is that `parse` **adds** to the default graph (identified by DATASET_DEFAULT_GRAPH_ID):

```python
ds = Dataset()
ds.parse(location=TEST_DATA_DIR / "variants" / "diverse_triples.ttl", format="ttl")
assert len(ds.default_context) == 5
ds.parse(location=TEST_DATA_DIR / "variants" / "simple_triple.ttl", format="ttl")
assert len(ds.default_context) == 6
```

test/test_conjunctivegraph/test_conjunctive_graph.py
gjhiggins: > Also, if you think the code should work the way it does in main

I wouldn't aspire to hold an opinion. My intention was merely to bring the issue to the attention of reviewers in as clear and straightforward a manner as I could so that any potential impact on existing downstream code could be discussed and advice prepared.

docs/upgrade6to7.rst
```diff
@@ -0,0 +1,45 @@
...
+
+If you want to load triples from a format that does not support named graphs
+into a named graph, use the following code:
+
+.. code-block:: python
+    
+    from rdflib import ConjunctiveGraph
+
+    cg = ConjunctiveGraph()
+    cg.get_context("example:graph_name").parse("http://example.com/source.trig", format="trig")
```
aucampia, line 31:
```suggestion
    cg.get_context("example:graph_name").parse("http://example.com/source.ttl", format="turtle")
```

aucampia:
> This gave arise to an issue when parsing ttl and ntriples files, where the Dataset has a default context graph (named `DATASET_DEFAULT_GRAPH_ID`) but that graph is empty, and the contents are placed into a _new_ graph in the Dataset.


What I suggested in the "Upgrading 6 to 7" document I added here is to do:

```python
    dataset = Dataset()
    dataset.get_context("example:graph_name").parse("http://example.com/source.ttl", format="turtle")
```

I guess if you don't know whether the source supports named graphs or not, this is not that helpful.

> For me personally the ideal behaviour would be:
> 
> 1. If calling `Dataset.parse()` on a non-named-graph file (eg, turtle, ntriples) _and_
> 2. If the Dataset has a default context graph named `DATASET_DEFAULT_GRAPH_ID` _and_
> 3. If the default context graph is empty, has no triples in it
>    Then: remove the empty default graph, parse the contents into a new graph, and set that new graph as the default context on the Dataset.

If the default context is empty, would it not be fine to just parse it directly into the empty default graph instead of parsing it into a new graph, and setting the new graph as the default graph?

Regarding the second point, I think we should try and make that a given - or at least we should not support cases where the default context has another name and try and prevent it, as per the spec the default graph is not named. I guess for now the special reserved name is okay, but I would say we should not allow changing the way the default graph is named as it is not clear in what sense that is still the default graph.

Just to understand this better, what would happen if the default context graph is not empy? Would the triples from the source then be added to it?

aucampia:
I will add some trest in a separate PR for:

- Relative URI resolution
- Calling parse on a graph with triples in

aucampia:
> > For me personally the ideal behaviour would be:
> > 
> > 1. If calling `Dataset.parse()` on a non-named-graph file (eg, turtle, ntriples) _and_
> > 2. If the Dataset has a default context graph named `DATASET_DEFAULT_GRAPH_ID` _and_
> > 3. If the default context graph is empty, has no triples in it
> >    Then: remove the empty default graph, parse the contents into a new graph, and set that new graph as the default context on the Dataset.
> 
> If the default context is empty, would it not be fine to just parse it directly into the empty default graph instead of parsing it into a new graph, and setting the new graph as the default graph?

Just as a final warning, with this change `parse()` does not clear any Graph in the dataset. I think in previous versions it cleared some graphs in the dataset, but not the default graph. parse should behave consistently, and in most cases it does not clear graphs, so I think the most reasonable thing to do is to not clear graphs.

However, that being said, if all conditions you give in your steps hold (i.e. "If the default context graph is empty, has no triples in it"), the behaviour with this PR will have the same outcome.

test/test_conjunctivegraph/test_conjunctive_graph.py
aucampia: @edmondchuc I never answered you explicitly on this question

> So what is the behaviour now with this change when loading source data in ttl format into a Dataset?

The new behaviour is that it goes into the default graph.

https://github.com/RDFLib/rdflib/blob/528465c51b48cb96dbbcfa0dd3351039ce3785b7/rdflib/graph.py#L2214-L2215

aucampia:
I'm going to merge this by this weekend if there is no further feedback, I may include the additional tests for unrelated functionality first, but I am uncertain if I will have time and there are other changes that are urgent.

aucampia:
> I'm going to merge this by this weekend if there is no further feedback, I may include the additional tests for unrelated functionality first, but I am uncertain if I will have time and there are other changes that are urgent.

Adding tests for relative URIs will be easier with this, as I want to use the variants test infrastructure for it, and this has some changes there that eliminate the hacks related to the default graph not being the default graph.

If there are problems discovered with the tests that I will add for relative URIs and parsing multiple documents or one document multiple times I will fix it.

gjhiggins:
Just FTR, the use of publicID as destination for uncontextualised statements was a fix for issue #535, made in PR https://github.com/RDFLib/rdflib/pull/536, sorry it took me so long to track it down.

aucampia:
> Just FTR, the use of publicID as destination for uncontextualised statements was a fix for issue #535, made in PR #536, sorry it took me so long to track it down.

Thanks for sharing @gjhiggins - @pchampin as you raised the original issue it may be worth getting your input here.

pchampin:
> > Just FTR, the use of publicID as destination for uncontextualised statements was a fix for issue #535, made in PR #536, sorry it took me so long to track it down.
> 
> Thanks for sharing @gjhiggins - @pchampin as you raised the original issue it may be worth getting your input here.

Sorry for the late response. I am satisfied with this change. Reusing `publicID` as the name of the default graph was indeed a workaround (either `default_context` didn't exist at the time, or I missed it...); the new behaviour  is cleaner.

PR #2443, opened 2023-06-12
Fix docs comment doesn't describe behavior

Kibubu:
Comment refers to a person that knows bob and code would return a name, but this would only work if the triple 
`person foaf:name bob .`
is part of the dataset. As this is a very uncommon way to model a `foaf:knows` the code was adjusted to match the description.


aucampia (approved):
Thanks for the fix @Kibubu, it looks good. I also fixed the flake8/flakeheaven config to not error out.

PR #2444, opened 2023-06-13
Update to https for dbpedia

miselico:
# Summary of changes

The URL for the service keyword had the http address for the dbpedia endpoint, which does no longer work. Changing it to https does the trick.

I only modified this one example in the documentation.

# Checklist

  the same change.


aucampia (approved):
Thanks for the fix @miselico

Issue #2229, opened 2023-02-14
Serializing SPARQL Query Results with Aggregates over Variables from Optional Graph Pattern


prohde:
I ran into an issue when serializing the results of SPARQL queries with aggregates from optional graph patterns, i.e., they might potentially be unbound. I am using ``rdflib==6.2.0``.

The query in question is:
```sparql
SELECT DISTINCT ?x (COUNT(DISTINCT ?inst) AS ?cnt) WHERE {
  ?x a <http://swat.cse.lehigh.edu/onto/univ-bench.owl#GraduateStudent> 
  OPTIONAL {
    VALUES ?inst { <http://www.University0.edu> <http://www.University1.edu> }. 
    ?x <http://swat.cse.lehigh.edu/onto/univ-bench.owl#undergraduateDegreeFrom> ?inst .
  }
} GROUP BY ?x
```

For each graduate student, I want to know how many undergraduate degrees he/she has from the list of universities provided using the VALUES clause. I am using OPTIONAL here since I am also interested in getting `0` if the student doesn't have a degree from one of the specified universities.

The query runs fine in my SPARQL endpoint but when I try to use rdflib as an in-memory RDF graph, I get the following exception:
```
Traceback (most recent call last):
  File "/home/my_path/venv/lib/python3.9/site-packages/rdflib/plugins/sparql/evalutils.py", line 68, in _eval
    return ctx[expr]
  File "/home/my_path/venv/lib/python3.9/site-packages/rdflib/plugins/sparql/sparql.py", line 175, in __getitem__
    return self.ctx.initBindings[key]  # type: ignore[index]
KeyError: rdflib.term.Variable('inst')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/my_path/graph_test.py", line 33, in run_query
    res_json = res.serialize(format='json')
  File "/home/my_path/venv/lib/python3.9/site-packages/rdflib/query.py", line 252, in serialize
    serializer.serialize(stream2, encoding=encoding, **args)  # type: ignore
  File "/home/my_path/venv/lib/python3.9/site-packages/rdflib/plugins/sparql/results/jsonresults.py", line 43, in serialize
    self._bindingToJSON(x) for x in self.result.bindings
  File "/home/my_path/venv/lib/python3.9/site-packages/rdflib/query.py", line 184, in bindings
    self._bindings += list(self._genbindings)
[...]

PR #2448, opened 2023-06-14
Fix: query over empty optional aggregate; added test and fix for issue #2229

WhiteGobo:
testing counting of optional nodes. Zero optional nodes may throw a NotBoundError
Added fix for NotBoundError, for this test.

# Summary of changes

As stated in [issue 2229](https://github.com/RDFLib/rdflib/issues/2229#issue-1583853300) a sparql query fails, if you count over an aggregate of optional values and zero values where found.
This patch catches the resulting NotBoundError and enables with that the count over the empty value-list. Also added the reason for the catch as comment.
Also added a test, that makes such a query. 

- Fixes <https://github.com/RDFLib/rdflib/issues/2229>

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugins/sparql/aggregates.py
+++ b/rdflib/plugins/sparql/aggregates.py
@@ -89,7 +89,11 @@ class Counter(Accumulator):
         return row
 
     def use_row(self, row: FrozenBindings) -> bool:
-        return self.eval_row(row) not in self.seen
+        try:
+            return self.eval_row(row) not in self.seen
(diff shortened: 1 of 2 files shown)


aucampia:
pre-commit.ci autofix

aucampia (approved):
Thanks for the fix @WhiteGobo, this looks good to me. I changed the test to check the whole result though, not just one row, and I confirmed this with Jena's SPARQL CLI.

PR #2460, opened 2023-06-20
BREAKING CHANGE: Make Literal more standards compliant

### LINKED ISSUE #670: Literal.datatype doesn't get initialized for Turtle string
Say we've this RDF statement in Turtle:

> ex:obs3 ex:hasStatus "online" . 

At rdflib 4.2.1 rdflib.term.Literal does not init 'datatype' attr for string type if not explicitly written as "value"^^xsd:string. As per W3C specification of Turtle, writing just "value" implicitly means it's xsd:string.

This assignment is needed for the program logic to be consistent when expecting to check for literal type using such attribute. 

The attribute get updated for other types like decimal, integer without explicitly written in Turtle.

### LINKED ISSUE #1326: Literal should always have a datatype

### LINKED ISSUE #2123: Literal equality fails for datatypes `None` and `XSD.string`
Example:
```python
>>> from rdflib import Literal
>>> from rdflib.namespace import XSD
>>> Literal('foo') == Literal('foo', datatype=XSD.string)
False
```

According to [https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal](https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal), every literal has a datatype (and a simple literal without a datatype should be considered syntactic sugar for literals of datatype `http://www.w3.org/2001/XMLSchema#string`).

The implementation in rdflib seems to follow the original RDF (not RDF 1.1), which allowed missing datatypes.

Related issues:
* #1326 proposes to always set a datatype (which would work, but might break things).

PS:
A similar problem would presumably occur with `XSD.langString`, except that setting both `lang` and `datatype` results in an error (which might also be an issue).

aucampia:
- Literal's datatype attribute is now no longer optional and will never
  be `None`.
- Constructing a literal with only a language tag will set the datatype
  to `rdf:langString`.

Related issues:

- Fixes <https://github.com/RDFLib/rdflib/issues/2123>
- Fixes <https://github.com/RDFLib/rdflib/issues/1326>
- Fixes <https://github.com/RDFLib/rdflib/issues/670>

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -20,6 +20,8 @@ underlying Graph:
 * Numerical Ranges
 
 """
+from __future__ import annotations
+
 import re
 from fractions import Fraction
 
@@ -51,11 +53,13 @@ from typing import (
     Callable,
     Dict,
     List,
+    NoReturn,
     Optional,
     Tuple,
     Type,
     TypeVar,
     Union,
+    overload,
 )
 from urllib.parse import urldefrag, urljoin, urlparse
 
@@ -69,6 +73,8 @@ from isodate import (
 )
 
 import rdflib
+
+# import rdflib.namespace as ns
 import rdflib.util
 from rdflib.compat import long_type
 
@@ -97,8 +103,27 @@ def _is_valid_uri(uri: str) -> bool:
 _lang_tag_regex = compile("^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$")
 
 
-def _is_valid_langtag(tag: str) -> bool:
-    return bool(_lang_tag_regex.match(tag))
+@overload
+def _validate_language_tag(tag: None) -> NoReturn:
+    ...
+
+
+@overload
+def _validate_language_tag(tag: str) -> None:
+    ...
+
+
+def _validate_language_tag(tag: Optional[str]) -> None:
+    if tag is None:
+        raise ValueError("language can not be None")
+    if not tag:
+        raise ValueError("language can not be empty string")
+    if not _lang_tag_regex.match(tag):
+        raise ValueError(
+            "language tag must match the pattern "
+            "^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$ "
+            f"but was {tag}"
+        )
 
 
 def _is_valid_unicode(value: Union[str, bytes]) -> bool:
@@ -370,6 +395,15 @@ class URIRef(IdentifiedNode):
             raise Exception("<%s> is not a skolem URI" % self)
 
 
+# Cannot import Namespace/XSD because of circular dependencies
+_XSD_PFX = "http://www.w3.org/2001/XMLSchema#"
+_RDF_PFX = "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+
+_RDF_LANGSTRING = URIRef(_RDF_PFX + "langString")
+
+_XSD_STRING = URIRef(_XSD_PFX + "string")
+
+
 class Genid(URIRef):
     __slots__ = ()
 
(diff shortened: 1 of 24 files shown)


rdflib/term.py
```diff
@@ -599,75 +657,114 @@ class Literal(Identifier):
...
+            if not _lang_tag_regex.match(lang):
+                raise ValueError(
+                    "language tag must match the pattern "
+                    "^[a-zA-Z]+(?:-[a-zA-Z0-9]+)*$ "
+                    f"but was {lang}"
+                )
+
+            if datatype is None:
+                datatype = _RDF_LANGSTRING
+
```
WhiteGobo, line 728: Let `q=Literal("ex", datatype=something)`. This will also overwrite any datatype given to a new Literal `Literal(q)`.
```
#prevous:
Literal(q)
# >> Literal("ex", datatype=something)

# now:
Literal(q)
# >> Literal("ex", datatype=RDF.langString)
```

Before the datatype was inherited. But i havent seen this behaviour documented, so there was now expected behaviour before. But we could mention this change in behaviour nonetheless.
Also this will handle issue #2603

rdflib/term.py
```diff
@@ -370,6 +395,15 @@ def de_skolemize(self) -> "BNode":
             raise Exception("<%s> is not a skolem URI" % self)
 
 
+# Cannot import Namespace/XSD because of circular dependencies
+_XSD_PFX = "http://www.w3.org/2001/XMLSchema#"
+_RDF_PFX = "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+
+_RDF_LANGSTRING = URIRef(_RDF_PFX + "langString")
```
WhiteGobo, line 402: It might be more useful to make this a class attribute for Literal. Then someone can change the default datatype by using a childclass instead of changing it programwide.
But im not sure

WhiteGobo, line 402: It might be more useful to make `_RDF_LANGSTRING` a class attribute for Literal. Then someone can change the default datatype for langstring by using a childclass instead of changing it programwide.
But im not sure

nicholascar:
@WhiteGobo @aucampia are either of you likely to be able to update / improve this PR?

ashleysommer:
I'd like to see this merged after the release of v7.1.0, along with this related change https://github.com/RDFLib/rdflib/issues/2866 as part of the "lots of breaking changes" release that RDFLib-8 will be.

PR #2467, opened 2023-07-03
Modify TriG serializer to not generate new prefixes for graph URIs

mgberg:
# Summary of changes

I personally have found it very annoying that the TriG serializer will generate new prefixes for named graph URIs if there is no appropriate prefix for that graph. This is a common occurrence when compiling TriG files where the named graphs each contain an ontology, and the names of the graphs are the URIs of the ontology. Many ontology URIs are the preferred namespace _without_ the trailing `#` or `/`.

For example, the [W3C Data Catalog Vocabulary](https://www.w3.org/ns/dcat.ttl) URI is `http://www.w3.org/ns/dcat` while the preferred prefix/namespace is `PREFIX dcat: <http://www.w3.org/ns/dcat#>`. If you run the following python snippet that simulates loading the DCAT ontology into the named graph `http://www.w3.org/ns/dcat` (or at least one triple in it):

```python
from rdflib import Dataset, URIRef
from rdflib.namespace import RDF, OWL

DCAT_URI = URIRef("http://www.w3.org/ns/dcat")

ds = Dataset()
g = ds.graph(DCAT_URI)
g.add((DCAT_URI, RDF.type, OWL.Ontology))

print(ds.serialize(format="trig"))
```

you should see the following output:
```trig
@prefix ns1: <http://www.w3.org/ns/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .

ns1:dcat {
    ns1:dcat a owl:Ontology .
}
```

That `ns1` prefix is generally unhelpful/unexpected, especially since you're probably expecting to see the URI written out in full [like in the source file](https://www.w3.org/ns/dcat.ttl). Since a new namespace is generated per graph (when the conditions are right), you can end up with a bunch of unhelpful prefixes if you load a bunch of graphs into the dataset. An example would be merging all the [QUDT](https://qudt.org/) Turtle files into one TriG file.

This PR includes modifications to two lines in the TriG parser to prevent this from happening. Consequently, running the above snippet with this PR will result in the following, which I think most people would expect.
```trig
@prefix owl: <http://www.w3.org/2002/07/owl#> .

<http://www.w3.org/ns/dcat> {
    <http://www.w3.org/ns/dcat> a owl:Ontology .
}
```

Of course, if an appropriate namespace does already exist or one gets bound, the graph URI will still be shortened appropriately. Adding this snippet at the end:
```python
ds.bind("MANUAL_NS", "http://www.w3.org/ns/")
print(ds.serialize(format="trig"))
```
still results in the expected output:
```trig
@prefix MANUAL_NS: <http://www.w3.org/ns/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .

MANUAL_NS:dcat {
    MANUAL_NS:dcat a owl:Ontology .
}
```

I'm not sure if there are any concerns around backwards compatibility or not- the content of the file doesn't _really_ change, it's just cosmetic.

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugins/serializers/trig.py
+++ b/rdflib/plugins/serializers/trig.py
@@ -40,7 +40,8 @@ class TrigSerializer(TurtleSerializer):
             if len(context) == 0:
                 continue
             self.store = context
-            self.getQName(context.identifier)
+            # Don't generate a new prefix for a graph URI if one already exists
+            self.getQName(context.identifier, False)
             self._subjects = {}
 
             for triple in context:
@@ -97,7 +98,8 @@ class TrigSerializer(TurtleSerializer):
                 if isinstance(store.identifier, BNode):
                     iri = store.identifier.n3()
                 else:
-                    iri = self.getQName(store.identifier)
+                    # Show the full graph URI if a prefix for it doesn't already exist
+                    iri = self.getQName(store.identifier, False)
                     if iri is None:
                         # type error: "IdentifiedNode" has no attribute "n3"
                         iri = store.identifier.n3()  # type: ignore[attr-defined]
--- a/test/test_trig.py
+++ b/test/test_trig.py
@@ -60,15 +60,15 @@ def test_remember_namespace():
     # prefix for the graph but later serialize() calls would work.
     first_out = g.serialize(format="trig", encoding="latin-1")
     second_out = g.serialize(format="trig", encoding="latin-1")
-    assert b"@prefix ns1: <http://example.com/> ." in second_out
-    assert b"@prefix ns1: <http://example.com/> ." in first_out
+    assert b"@prefix ns1: <http://example.com/> ." not in second_out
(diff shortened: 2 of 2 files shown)


aucampia:
@mgberg thanks for the PR, I don't have any backwards compatibility concerns with this because as you said it is just a cosmetic difference, but it is somewhat a matter of preference, and that is not that easy to gauge. I somewhat prefer the behaviour you are changing it to with this PR, but it may be good to reach out on Twitter to see if there are any concerns.

I will wait a while and see if there is any objection before merging.

I think long term we need to make serializers more easily customizable and then provide flags to control this behaviour, but right now adding custom flags for this may not be ideal.

mgberg:
> I think long term we need to make serializers more easily customizable and then provide flags to control this behaviour, but right now adding custom flags for this may not be ideal.

I also think making the serializers more customizable would be a good feature.

PR #2468, opened 2023-07-04
trig Parser now raises SyntaxError, when finding nested graphs

WhiteGobo:
# Summary of changes

When loading [trig-graph-bad-07.trig](https://www.w3.org/2013/TrigTests/trig-graph-bad-07.trig) an exception should be raised.
I implemented a normal SyntaxError for this occasion.

The xfail for the corresponding test was removed.

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/plugins/parsers/trig.py
+++ b/rdflib/plugins/parsers/trig.py
@@ -105,8 +105,7 @@ class TrigSinkParser(SinkParser):
         # type error: Incompatible types in assignment (expression has type "Graph", variable has type "Optional[Formula]")
         self._context = self._store.newGraph(graph)  # type: ignore[assignment]
         if self._context is not None and self._parentContext is not None:
-            raise SyntaxError("GRAPH may not include a GRAPH "
-                              "(#trig-graph-bad-07)")
+            raise SyntaxError("GRAPH may not include a GRAPH " "(#trig-graph-bad-07)")
 
         while 1:
             i = self.skipSpace(argstr, j)
--- a/rdflib/plugins/parsers/trig.py
+++ b/rdflib/plugins/parsers/trig.py
@@ -104,6 +104,9 @@ class TrigSinkParser(SinkParser):
         self._reason2 = becauseSubGraph
         # type error: Incompatible types in assignment (expression has type "Graph", variable has type "Optional[Formula]")
         self._context = self._store.newGraph(graph)  # type: ignore[assignment]
+        if self._context is not None and self._parentContext is not None:
+            raise SyntaxError("GRAPH may not include a GRAPH "
+                              "(#trig-graph-bad-07)")
 
         while 1:
             i = self.skipSpace(argstr, j)
--- a/rdflib/plugins/parsers/trig.py
+++ b/rdflib/plugins/parsers/trig.py
@@ -98,14 +98,15 @@ class TrigSinkParser(SinkParser):
 
         j = i + 1
 
+        if self._context is not None:
+            self.BadSyntax(argstr, i, "Nested graphs are not allowed")
+
         oldParentContext = self._parentContext
         self._parentContext = self._context
         reason2 = self._reason2
         self._reason2 = becauseSubGraph
         # type error: Incompatible types in assignment (expression has type "Graph", variable has type "Optional[Formula]")
         self._context = self._store.newGraph(graph)  # type: ignore[assignment]
-        if self._context is not None and self._parentContext is not None:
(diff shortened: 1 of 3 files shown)


WhiteGobo:
I havent checked if the Syntaxerror should be subclassed.

aucampia:
pre-commit.ci autofix

rdflib/plugins/parsers/trig.py
```diff
@@ -104,6 +104,8 @@ def graph(self, argstr: str, i: int) -> int:
         self._reason2 = becauseSubGraph
         # type error: Incompatible types in assignment (expression has type "Graph", variable has type "Optional[Formula]")
         self._context = self._store.newGraph(graph)  # type: ignore[assignment]
+        if self._context is not None and self._parentContext is not None:
+            raise SyntaxError("GRAPH may not include a GRAPH " "(#trig-graph-bad-07)")
```
aucampia, line 108:
```suggestion
            raise SyntaxError("GRAPH may not include a GRAPH")
```

aucampia:
Thanks for the PR @WhiteGobo - I will check it in a bit more detail a bit later tonight but it looks good at a glance.

One concern may be that people prefer more lax parsers, but that can go too far [[ref](https://www.ietf.org/archive/id/draft-iab-protocol-maintenance-05.html)], and in this case, I definitely would much rather have the default be compliant with the spec and if someone wants it more lax we can consider adding an option to control parser behaviour to make it more lax; however, it is hard to see who would want it in this case.

aucampia:
For reference, this is the test file in question:

https://github.com/RDFLib/rdflib/blob/6f53455c788b123826ad7321c7cea5f381ae7a77/test/data/suites/w3c/trig/trig-graph-bad-07.trig#L1-L9

rdflib/plugins/parsers/trig.py
```diff
@@ -104,6 +104,8 @@ def graph(self, argstr: str, i: int) -> int:
         self._reason2 = becauseSubGraph
         # type error: Incompatible types in assignment (expression has type "Graph", variable has type "Optional[Formula]")
         self._context = self._store.newGraph(graph)  # type: ignore[assignment]
+        if self._context is not None and self._parentContext is not None:
```
aucampia, line 107: `self._store.newGraph(graph)` always returns a value so `self._context is not None` should always be true and can be omitted as far as I can tell.

This can also move up a bit, I think it would be best at line 100 here, and then it can also only use `self._context` it won't be moved to `self._parentContext` yet.

https://github.com/RDFLib/rdflib/blob/6f53455c788b123826ad7321c7cea5f381ae7a77/rdflib/plugins/parsers/trig.py#L99-L101

rdflib/plugins/parsers/trig.py
aucampia: This will trigger even for cases where GRAPH is not used, and that is desirable, but in those cases the error message will be a bit misleading.

Also, while I don't exactly condone it, the way that errors are handled for the N3 related parsers is with `self.BadSyntax(...)`, and the `argstr: str, i: int` parameters are passed to it. The benefit of this is that it reports the line number, or at least tries to.

aucampia (commented):
I'm going to fix the two issues I noted in comments on your branch.

rdflib/plugins/parsers/trig.py
aucampia: Fixed now.

rdflib/plugins/parsers/trig.py
aucampia: Fixed now.

aucampia:
> I'm going to fix the two issues I noted in comments on your branch.

done now.

aucampia (approved):
Thanks for the fix, I'm not sure what was happening with the triples when nested graphs were encountered, but it could not have been good.

I will merge later this week if there is no more feedback.

WhiteGobo:
Thx for the review. Ive felt really lost, when working on the Trig- and Notation3-Parser. I worked more with trial and error, than thinking:) Im much more confident with the parsers, that are written with pyparsing.

aucampia:
> Thx for the review. Ive felt really lost, when working on the Trig- and Notation3-Parser. I worked more with trial and error, than thinking:) Im much more confident with the parsers, that are written with pyparsing.

We really should get rid of the hand-written parsers, I don't like working on it either. I think using https://github.com/lark-parser/lark would be best, as that allows generating parsers that don't need any runtime dependencies, but PyParsing is already a lot easier.

PR #2469, opened 2023-07-04
trig GRAPH needs graphname (as in w3c-test:trig-graph-bad-01)

WhiteGobo:
# Summary of changes

trig parser will now expect a graphname if Keyword GRAPH is used before {...}.
removed expected failure for testcase https://www.w3.org/2013/TrigTests/#trig-graph-bad-01

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/plugins/parsers/trig.py
+++ b/rdflib/plugins/parsers/trig.py
@@ -82,7 +82,7 @@ class TrigSinkParser(SinkParser):
             graph = r[0]
             i = j
         elif need_graphid:
-            return -1
+            self.BadSyntax(argstr, i, "GRAPH keyword must be followed by graph name")
         else:
             graph = self._store.graph.identifier  # hack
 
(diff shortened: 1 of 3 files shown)


aucampia:
Test file in question:

https://github.com/RDFLib/rdflib/blob/6f53455c788b123826ad7321c7cea5f381ae7a77/test/data/suites/w3c/trig/trig-graph-bad-01.trig#L1-L5

rdflib/plugins/parsers/trig.py
```diff
@@ -69,16 +69,20 @@ def graph(self, argstr: str, i: int) -> int:
...
             i = j
+            need_graphid = True
 
         r: MutableSequence[Any] = []
         j = self.labelOrSubject(argstr, i, r)
         if j >= 0:
             graph = r[0]
             i = j
+        elif need_graphid:
+            return -1
```
aucampia, line 85:
```suggestion
            self.BadSyntax(argstr, i, "GRAPH keyword must be followed by graph name")
```

I think it may be best to also use `self.BadSyntax` here, as it makes the problem a bit clearer to users.

aucampia (commented):
Thanks for the fix, it looks good. I just made one suggestion regarding how the error is raised.

aucampia:
I will merge later this week.

PR #2470, opened 2023-07-04
sparql parser identifying decimal correctly

WhiteGobo:
# Summary of changes

sparql10-test manifest#term-6 failed because decimal wasnt identified.
Changed parser so '456.' can be identified as decimal.
removed xfail for w3c-testcase

# Checklist

  the same change.
  

aucampia (commented):
@WhiteGobo thanks for the PR, however, I think it is best to keep it as is in this case.

The W3C specs don't exactly follow SEMVER, thus splitting the baby with minor versions that are not backwards compatible.

SPARQL 1.1's DECIMAL production is [[ref](https://www.w3.org/TR/sparql11-query/#rDECIMAL)]:


```bnf
[147]  	DECIMAL	  ::=  	[0-9]* '.' [0-9]+
```

The SPARQL 1.0 DECIMAL production, instead, is as follows [[ref](https://www.w3.org/TR/rdf-sparql-query/#rDECIMAL)]:

```bnf
[78]  	DECIMAL	  ::=  	[0-9]+ '.' [0-9]* | '.' [0-9]+
```

The xfail marker is somewhat questionable here, though. It should be used to indicate that the test should pass, but does not due to a defect. In this case, the test should not pass, and it is not a defect that prevents it from passing, it is just not relevant.

I think the right option is to instead mark it with `pytest.mark.skip` and change the reason to something that indicates that it depends on SPARQL 1.0 gramar.

WhiteGobo:
mh i see. sounds good to me.

For context, the query file 

https://github.com/RDFLib/rdflib/blob/6f53455c788b123826ad7321c7cea5f381ae7a77/test/data/suites/w3c/dawg-data-r2/basic/term-6.rq#L5

and the data file:

aucampia:
Closing this as <https://github.com/RDFLib/rdflib/pull/2471> was merged.

Issue #2475, opened 2023-07-08
Casting literal with content to rdf:HTML datatype leads incorrectly to empty literal


floresbakker:
Casting a literal to rdf:HTML datatype leads incorrectly to an empty literal.

**Example code**

```
prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

select * where {
  bind(strdt("<body>", rdf:HTML) as ?tag1) # incorrectly disappearing literal

  bind("<body>" as ?tag2)                  # correctly appearing literal
}

```
**Expected results for ?tag1 and ?tag2**

```
"<body>"^^rdf:HTML
"<body>"

```
**Instead we get:**

```
""^^rdf:HTML
"<body>"


```

WhiteGobo:
Mh it seems this is interpreted as a builtin type and then translated directly into an xml-thingy. The `<html>` is stored in `self.value` but i have no idea where it is supposed to be read:

```
class Literal(Identifier, Node):
    ....
    def __repr__(self):
        args = [super().__repr__()]
        ...
```
But neither `Identifier` nor `Node` have a `__repr__`-method

You can get a hold on the '<html>' as a value. But it is in the form of a `xml.dom.minidom.DocumentFragment`.

```
q = Literal("<body>", datatype="http://www.w3.org/1999/02/22-rdf-syntax-ns#HTML")
print(q.value)
>>> <xml.dom.minidom.DocumentFragment object at 0x7f61f112b750>
```

floresbakker:
Ah, so it has not completely disappeared as it is still present somehow in the underlying data structure of RDFLib.  Yet, it is unavailable within a SPARQL query, right? Do you think this is a rather fundamental and difficult issue to solve in RDFlib? It doesn't sound like an easily solvable bug unfortunately.

WhiteGobo:
Mh it should be possible to find the bug within short time.

WhiteGobo:
Mh i think i will produce a hotfix just by overriding some of the init parts. And then i want to know, why `Literals` dont use the literal str that is given to them but instead convert via some obscure methods.

The bug is that a wrong object is created at this position:

https://github.com/RDFLib/rdflib/blob/0ea6ca579442219d67ffb1fc7313f05fd16d8d49/rdflib/term.py#L1641C1-L1648C22

Or something is wrong when that object is transformed back to a string at this position:

[...]

PR #2477, opened 2023-07-08
Hotfix so that: literal html uses initial value instead

WhiteGobo:
# Summary of changes

Created a hotfix for Problem missing [literal-html string](https://github.com/RDFLib/rdflib/issues/2475).

If Literal is XSD:HTML the literal value wont be overwritten by the value produced in Literal.__init__

In its current form `Literal` produces an empty string if `datatype=xsd:HTML`.

If `Literal(..,datatype=xsd.HTML)` prodcues an empty `value`, this hotfix doesnt fix that. (I havent checked this.)

# Longer description

For more information see the discussion in Issue above. In short:
`Literal.__init__` doesnt respect [Literal-term-equality](https://www.w3.org/TR/rdf11-concepts/#dfn-literal-term-equality). Instead it produces for builtin-datatypes a special `value`. Because Literal is a subclass to str, it needs a `str` back from the produced `value`. During this process the value of the html string is lost and so only an empty `str` remains.
This hotfix circumvents this by checking if `datatype==xsd.HTML` and just uses the original string given to `__init__`.

The value that is produced within `Literal.__init__` may be corrupted because it just consists of an empty `""`
It is created at this position:

https://github.com/RDFLib/rdflib/blob/0ea6ca579442219d67ffb1fc7313f05fd16d8d49/rdflib/term.py#L1641C1-L1648C22

At this position it is transformed back to a string:

https://github.com/RDFLib/rdflib/blob/0ea6ca579442219d67ffb1fc7313f05fd16d8d49/rdflib/term.py#L1656C1-L1663C31

The type of the value is `xml.dom.minidom.DocumentFragment`. Checking what is inside that object may be a little bit harder.

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -654,7 +654,11 @@ class Literal(Identifier):
             if value is not None and normalize:
                 _value, _datatype = _castPythonToLiteral(value, datatype)
                 if _value is not None and _is_valid_unicode(_value):
-                    lexical_or_value = _value
+                    if _datatype != _RDF_HTMLLITERAL:
+                        # Hotfix. See here for more information:
+                        # https://github.com/RDFLib/rdflib/issues/2475
+                        # html produce an empty _value
+                        lexical_or_value = _value
 
         else:
             # passed some python object


aucampia:
@WhiteGobo thanks for the PR, I think a better interim option is to just set `rdlib.NORMALIZE_LITERALS` to `False`. I'm not entirely clear what is going wrong though, I suspect the HTML parsing is a bit broken as the document fragment seems emtpy in this case.

WhiteGobo:
Oh i've missed that option.

Yeah sadly the xml.dom-things werent easy for me to understand, when i tried to find the bug. Maybe ill come back to it some time later. I mean, i think i know where to look for the problem.

PR #2483, opened 2023-07-11
Stopping normalization of invalid html literals

### LINKED ISSUE #2475: Casting literal with content to rdf:HTML datatype leads incorrectly to empty literal

WhiteGobo:
# Summary of changes

literal stops normalization process if html5lib cant parse literal
Tests now check, that invalid html literals wont be normalized

Fixes #2475

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -1641,16 +1641,17 @@ def _parseXML(xmlstring: str) -> xml.dom.minidom.Document:  # noqa: N802
 def _parseHTML(htmltext: str) -> xml.dom.minidom.DocumentFragment:  # noqa: N802
     try:
         import html5lib
-
-        parser = html5lib.HTMLParser(tree=html5lib.treebuilders.getTreeBuilder("dom"))
-        retval = parser.parseFragment(htmltext)
-        retval.normalize()
-        return retval
     except ImportError:
         raise ImportError(
             "HTML5 parser not available. Try installing"
             + " html5lib <http://code.google.com/p/html5lib>"
         )
+    parser = html5lib.HTMLParser(tree=html5lib.treebuilders.getTreeBuilder("dom"))
+    retval = parser.parseFragment(htmltext)
+    if parser.errors:
+        raise ValueError("html5lib failed to parse.", htmltext)
+    retval.normalize()
+    return retval
 
 
 def _writeXML(  # noqa: N802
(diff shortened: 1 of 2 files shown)


rdflib/term.py
```diff
@@ -1641,16 +1641,17 @@ def _parseXML(xmlstring: str) -> xml.dom.minidom.Document:  # noqa: N802
...
         raise ImportError(
             "HTML5 parser not available. Try installing"
             + " html5lib <http://code.google.com/p/html5lib>"
         )
+    parser = html5lib.HTMLParser(tree=html5lib.treebuilders.getTreeBuilder("dom"))
+    retval = parser.parseFragment(htmltext)
+    if parser.errors:
+        raise ValueError("html5lib failed to parse.", htmltext)
+    retval.normalize()
+    return retval
```
aucampia, line 1654: I looked at the html5lib code and I think it may be slightly better to run the parser in strict mode, in which case it will throw by itself.

```suggestion
    parser = html5lib.HTMLParser(tree=html5lib.treebuilders.getTreeBuilder("dom"), strict=True)
    retval = parser.parseFragment(htmltext)
    retval.normalize()
    return retval
```

For more info see [[ref](https://html5lib.readthedocs.io/en/latest/html5lib.html#html5lib.html5parser.HTMLParser)]

aucampia (commented):
Thanks for looking into this, it is indeed the right thing to do, however running the HTML5Lib parser in strict mode may be slightly better as that results in an exception with some more information about the error. We could technically recover the same information and add it in a custom exception, but errors is an array of errors and could indicate multiple errors.

WhiteGobo:
Thank you for reviewing the code. Ive missed that option. Sounds good.

aucampia:
pre-commit.ci autofix

PR #2490, opened 2023-07-16
fix: handling of `rdf:HTML` literals

aucampia:
# Summary of changes

Previously, if without `html5lib` installed, literals with`rdf:HTML` datatypes were treated as
[ill-typed](https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal), even if they were not ill-typed.

With this change, if `html5lib` is not installed, literals with the `rdf:HTML` datatype will not be treated as ill-typed, and will have `Null` as their `ill_typed` attribute value, which means that it is unknown whether they are ill-typed or not.

This change also fixes the mapping from `rdf:HTML` literal values to lexical forms.

Other changes:

- Add tests for `rdflib.NORMALIZE_LITERALS` to ensure it behaves correctly.

Related issues:

- Fixes <https://github.com/RDFLib/rdflib/issues/2475>

# Checklist

  the same change.

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -76,6 +76,15 @@ if TYPE_CHECKING:
     from .namespace import NamespaceManager
     from .paths import AlternativePath, InvPath, NegatedPath, Path, SequencePath
 
+_HAS_HTML5LIB = False
+
+try:
+    import html5lib
+
+    _HAS_HTML5LIB = True
+except ImportError:
+    html5lib = None
+
 _SKOLEM_DEFAULT_AUTHORITY = "https://rdflib.github.io"
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


aucampia:
@WhiteGobo if you have a moment for a review please have a look.

test/test_literal/test_literal_html5lib.py
```diff
@@ -0,0 +1,79 @@
...
+        # Well-typed literals, these have lexical forms that parse
+        # without errors with html5lib.
+        (
+            lambda: Literal("<table></table>", datatype=RDF.HTML),
+            LiteralChecker(..., None, RDF.HTML, False, "<table></table>"),
+        ),
+        (
+            lambda: Literal("  <table>  </table>  ", datatype=RDF.HTML, normalize=True),
+            LiteralChecker(..., None, RDF.HTML, False, "  <table>  </table>  "),
+        ),
```
WhiteGobo, line 63: Just from a brief look, i would have expected to compare the normalized  value with `<table></table>`.

WhiteGobo:
Looks good to me. I find the idea to load the normalizer only if html5lib exists good.
I was only confused by one of the tests but as far as ive seen this is in line with, how other normalization tests of literals are done.

rdflib/term.py
```diff
@@ -1967,14 +1990,21 @@ def _castPythonToLiteral(  # noqa: N802
...
-    # bit random
-    (xml.dom.minidom.DocumentFragment, (_writeXML, _RDF_HTMLLITERAL)),
     (Fraction, (None, _OWL_RATIONAL)),
 ]
 
+if html5lib is not None:
+    # This is a bit dirty, by accident the html5lib parser produces
+    # DocumentFragments, and the xml parser Documents, letting this
+    # decide what datatype to use makes roundtripping easier, but it a
+    # bit random.
```
WhiteGobo, line 2000: And this first part of the comment seems to suggest, that DocumentFragments isnt the expected type, although it is. But i have to little knowledge about the roundtripping tests to understand what the author wanted to say otherwise. Maybe we could just drop it.

test/test_literal/test_literal_html5lib.py
aucampia: Yeah, me too, I actually have no idea what normalize is doing, I tried to find cases where it changes the lexical value but I could not find any.

rdflib/term.py
aucampia: The comment is a bit weird, but what it tries to point out is that it is more or less just luck that gives it a different type from `rdf:XML`, it could be that someone constructs a literal with a `DocumentFragment` which is not representative of HTML, in which case it should not become `rdf:HTML`. This points to some drawbacks in how python values are used to construct literals.

WhiteGobo: mh ok makes sense. You dont know by chance why rdf:xml generates a [document](https://rdflib.readthedocs.io/en/stable/rdf_terms.html#python-conversions) instead of a [documentfragment](https://www.w3.org/TR/rdf12-concepts/#section-XMLLiteral)?

WhiteGobo (approved):

PR #2502, opened 2023-07-26
Add CITATION.cff according to zenodo record

### LINKED ISSUE #2059: Create a CITATION.cff file
now that we are integrated with Zenodo we should do somewhat better in curating metadata, this can be done with: https://citation-file-format.github.io/

white-gecko:
Add CITATION.cff according to https://zenodo.org/record/7771749.

Close #2059.

# Summary of changes

Add the CITATION.cff file according to the spec https://citation-file-format.github.io/ and with the data of https://zenodo.org/record/7771749 and orcids added manually, left out in doubt.

# Checklist

  the same change.
  
  - [x] Created an issue to discuss the change and get in-principle agreement. -> exists already: #2059
  - [x] ~Considered adding an example in `./examples`.~
- ~If the change has a potential impact on users of this project:~
  

CITATION.cff
```diff
@@ -0,0 +1,75 @@
...
+- family-names: "Cogrel"
+  given-names: "Benjamin"
+  orcid: "https://orcid.org/0000-0002-7566-4077"
+- family-names: "Stuart"
+  given-names: "Veyndan"
+title: "RDFLib"
+version: 6.3.2
+date-released: 2023-03-26
+url: "https://github.com/RDFLib/rdflib"
+doi: 10.5281/zenodo.7771749
```
aucampia, line 75:
```suggestion
doi: 10.5281/zenodo.6845245
```

The DOI is a bit of a problem.

image

The above is from https://zenodo.org/account/settings/github/, how I interpret it, and how it mostly has worked, is that after creating a release using GitHub releases, Zenodo fetches the archive of the release (which for 6.3.2 is https://github.com/RDFLib/rdflib/tree/0593c75753b0b80256d448d56328dbc3967cb055) and then assigns a DOI and publishes the archive under the DOI.

The `CITATION.cff` file should be taken from the release archive, but before Zenodo publishes it we don't know the version-specific DOI.

However, we do have a DOI which is not version-specific which Zenodo refers to as the [concept DOI](https://help.zenodo.org/faq/#hversioning-3). For RDFLib it is https://doi.org/10.5281/zenodo.6845245. I think if we put any value in here, the best is to put https://doi.org/10.5281/zenodo.6845245 in here. The CITATION.cff documentation have an example of both using the versioned and concept DOI in here [[ref](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#doi)], so I think it is acceptable.

aucampia (approved):
Looks good, I will add a note in the release that if someone that contributed and is not listed in here wants their details included they can make a PR, and it will be in the next release.

PR #2504, opened 2023-07-30
Corrected n3 method of path objects (fixes #2503)

### LINKED ISSUE #2503: N3 method for path objects is bugged
The n3 method for path objects returns invalid/unexpected paths for non-trivial paths.

Example:
```
import rdflib
import rdflib.paths

path = rdflib.RDF.type / rdflib.paths.AlternativePath(rdflib.SKOS.note, rdflib.RDFS.comment)
```
Calling `path.n3()` here returns `<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>/<http://www.w3.org/2004/02/skos/core#note>|<http://www.w3.org/2000/01/rdf-schema#comment>` (essentially `rdf:type/skos:note|rdfs:comment`).
However, it the method should rather return `<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>/(<http://www.w3.org/2004/02/skos/core#note>|<http://www.w3.org/2000/01/rdf-schema#comment>)` (thus `rdf:type/(skos:note|rdfs:comment)`).

The n3 method ignores all groups as it never introduces any brackets.

Related to https://github.com/RDFLib/rdflib/issues/421

lukasmu:
# Summary of changes

This is an attempt to fix #2503. Previously, the n3 method of path objects returned unexpected/invalid strings when the path objects were non-trivial (i.e. nested, consisting of property path groups). Examples can be found in the related issue and the tests.

This change fixes a bug. It might impact on users that previously relied on the invalid implementation of the n3 method. Not sure if/where this should be added to the documentation.

- Fixes https://github.com/RDFLib/rdflib/issues/2503

# Checklist

  the same change.
  
  - [x] ~~Updated relevant documentation to avoid inaccuracies.~~
  - [x] ~~Considered adding additional documentation.~~

### DIFF
--- a/rdflib/paths.py
+++ b/rdflib/paths.py
@@ -213,6 +213,15 @@ OneOrMore = "+"
 ZeroOrOne = "?"
 
 
+def _n3(
+    arg: Union["URIRef", "Path"], namespace_manager: Optional["NamespaceManager"] = None
+):
+    # type error: Item "Path" of "Union[Path, URIRef]" has no attribute "n3"  [union-attr]
+    if isinstance(arg, (SequencePath, AlternativePath)) and len(arg.args) > 1:
+        return "(%s)" % arg.n3(namespace_manager)
+    return arg.n3(namespace_manager)  # type: ignore[union-attr]
+
+
 @total_ordering
 class Path:
     __or__: Callable[["Path", Union["URIRef", "Path"]], "AlternativePath"]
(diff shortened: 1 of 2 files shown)


aucampia (approved):
Thanks for the pull request @lukasmu - it looks good to me. I added some additional tests that all passed, and a return type annotation.

PR #2520, opened 2023-08-06
fix: make rdflib.term.Node abstract (fixes #2518)

### LINKED ISSUE #2518: rdflib.term.IdentifiedNode should be abstract
I just saw issue #2509 (and the corresponding PR #2516).

I would argue that `rdflib.term.IdentifiedNode` should also be abstract and have an abstract `n3` method because:
- The docstring of `IdentifiedNode` already states `An abstract class, primarily [...]`
- When annotating code with type hints we are now often forced to refer somewhat awkwardly to `typing.Union[URIRef, BNode]` instead of  simply `IdentifiedNode` (when our code relies on the presence of the `n3` method).

Any objections/thoughts?

lukasmu:
# Summary of changes

Attempt to make `rdflib.term.Node` as well as `rdflib.term.Identifier` and `rdflib.term.IdentifiedNode` abstract base classes.
Due to this change a few `type: ignore` comments can be removed from the main library. However, three `type: ignore` comments also need to be added to a test file (though this seems reasonable in the specific case).

# Checklist

  the same change.

### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1609,10 +1609,9 @@ class Graph(Node):
 
         return processor.update(update_object, initBindings, initNs, **kwargs)
 
-    def n3(self) -> str:
+    def n3(self, namespace_manager: Optional["NamespaceManager"] = None) -> str:
         """Return an n3 identifier for the Graph"""
-        # type error: "IdentifiedNode" has no attribute "n3"
-        return "[%s]" % self.identifier.n3()  # type: ignore[attr-defined]
+        return "[%s]" % self.identifier.n3(namespace_manager=namespace_manager)
 
     def __reduce__(self) -> Tuple[Type[Graph], Tuple[Store, _ContextIdentifierType]]:
         return (
@@ -2591,14 +2590,12 @@ class QuotedGraph(Graph):
         )
         return self
 
-    def n3(self) -> str:
+    def n3(self, namespace_manager: Optional["NamespaceManager"] = None) -> str:
         """Return an n3 identifier for the Graph"""
-        # type error: "IdentifiedNode" has no attribute "n3"
-        return "{%s}" % self.identifier.n3()  # type: ignore[attr-defined]
+        return "{%s}" % self.identifier.n3(namespace_manager=namespace_manager)
 
     def __str__(self) -> str:
-        # type error: "IdentifiedNode" has no attribute "n3"
-        identifier = self.identifier.n3()  # type: ignore[attr-defined]
+        identifier = self.identifier.n3()
         label = self.store.__class__.__name__
         pattern = (
             "{this rdflib.identifier %s;rdflib:storage "
@@ -2900,7 +2897,7 @@ class ReadOnlyGraphAggregate(ConjunctiveGraph):
     ) -> NoReturn:  # noqa: N803
         raise ModificationException()
 
-    def n3(self) -> NoReturn:
+    def n3(self, namespace_manager: Optional["NamespaceManager"] = None) -> NoReturn:
         raise UnSupportedAggregateOperation()
 
     def __reduce__(self) -> NoReturn:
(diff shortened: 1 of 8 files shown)


aucampia (approved):
Thanks for the PR @lukasmu - it looks very good.

I don't exactly like the n3 method though, at some point we have to try get rid of it, N3 is not ratified by W3C and is not really using the RDF data model, so having a method named after it is a bit odd, and it also does very different things for Graph and for Identifier.

That being said, this PR essentially just documents the facts as they are in the type system, so I think it's the best option at the moment.

aucampia:
@ajnelson-nist if you have a moment, please have a look. The only slightly controversial thing here is to have `n3` as high up in the class hierarchy as `rdflib.term.Node`, but I think it's justified.

rdflib/graph.py
```diff
@@ -1609,10 +1609,9 @@ def update(
 
         return processor.update(update_object, initBindings, initNs, **kwargs)
 
-    def n3(self) -> str:
+    def n3(self, namespace_manager: Optional["NamespaceManager"] = None) -> str:
```
ajnelson-nist, line 1612: I'm a little confused.  The body of this method is updated to use `namespace_manager` as a keyword argument on an `n3()` method; but, this syntax looks like it's specifying `namespace_manager` as a positional argument.

Could the signature become instead:

```python
def n3(
    self,
    *args: Any,
    namespace_manager: Optional["NamespaceManager"]=None,
    **kwargs: Any,
) -> str: ...
```

rdflib/graph.py
```diff
@@ -1609,10 +1609,9 @@ def update(
...
         """Return an n3 identifier for the Graph"""
-        # type error: "IdentifiedNode" has no attribute "n3"
-        return "[%s]" % self.identifier.n3()  # type: ignore[attr-defined]
```
ajnelson-nist, line 1615: I'm glad to see these no-attribute issues being addressed.

ajnelson-nist (changes requested):
Overall, I agree with the effects of this PR.  `n3()` method calls have been in place for some time, so this looks like overall the typing becomes more coherent.

My one request for a change is, I think, minor w.r.t. the rest of the PR: to decide whether `namespace_manager` is a positional or a keyword argument.  The function signatures seem to be saying positional, the function calls seem to be saying keyword.

lukasmu:
Thanks for the feedback! 

When coding in Python and calling a function I always tend to provide the name of the arguments even if they are positional arguments. Just because "Explicit is better than implicit" (related https://www.youtube.com/watch?v=wf-BqAjZb8M&t=43m15s).
In Python we almost never see positional-only arguments (though they exist as of Python 3.8): 
```
def f(positional_parameter, /, positional_or_keyword_parameter, *, keyword_parameter):
    pass
```
Arguments like `namespace_manager` are thus technically positional-or-keyword parameters.

ajnelson-nist:
> Thanks for the feedback!
> 
> When coding in Python and calling a function I always tend to provide the name of the arguments even if they are positional arguments. Just because "Explicit is better than implicit" (related https://www.youtube.com/watch?v=wf-BqAjZb8M&t=43m15s). In Python we almost never see positional-only arguments (though they exist as of Python 3.8):
> 
> ```
> def f(positional_parameter, /, positional_or_keyword_parameter, *, keyword_parameter):
>     pass
> ```
> 
> Arguments like `namespace_manager` are thus technically positional-or-keyword parameters.

I suppose my complaint amounts to a difference of opinion in coding style.  I've often been confused over the years with function definition styles in Python slightly obscuring whether an argument was positional or a keyword, particularly when default values were used too.  I became a happier programmer when I learned about the asterisk delimiter.  I've avoided the `/` because I needed to support Python 3.7 applications, but now that that version's sunset, I'll consider it if I ever come across a good reason.

I tried to get examples in your explicit-name style to fail type review with confusion of a keyword vs positional argument, but your call style seemed to work fine.  So, I'll leave it to the RDFLib developers to settle on style.

ajnelson-nist (approved):
I have no blocking feedback on this PR.  @aucampia , I don't think my Request Changes review prevents a merge, but if it does, and this Approval action doesn't undo it, please feel free to override it.

aucampia:
Thank you very much for having a look @ajnelson-nist - and thanks again for the fix @lukasmu

PR #2523, opened 2023-08-10
refactor: don't use the same variable name for different types

aucampia:
# Summary of changes

Mypy does not like it when the same variable name is used for different types.

# Checklist

  the same change.


aucampia:
I'm fixing this because dmypy gets super confused here with <https://github.com/RDFLib/rdflib/pull/2520>.

Issue #2538, opened 2023-08-22
With Datasets/ConjunctiveGraphs, `Graph.{query,update}` should use the default graph as the value for the `queryGraph` when calling `Store.{query,update}` 


aucampia:
When using a `rdflib.store.Store` that implements a `parse()` method, `rdflib.graph.Graph.query()` calls it instead of using the builtin SPARQL processor.

`rdflib.store.Store.query` has a parameter `queryGraph`, which presumably specifies the initial [active graph](https://www.w3.org/TR/sparql11-query/#sparqlDataset) (this should be clarified in docstrings).

When SPARQL is evaluated against a Dataset, the initial active graph should be the default graph [[ref](https://www.w3.org/TR/sparql11-query/#sparqlAlgebraEval)], however, when calling `rdflib.store.Store.query`, `rdflib.graph.Graph.query` passes `rdflib.graph.Graph.identifier` as the value of `queryGraph`, and this will not be the identifier of the default graph if the graph is a `rdflib.graph.Dataset`, and will instead be an identifier that has really no meaning whatsoever  (see <https://github.com/RDFLib/rdflib/issues/2537>).

https://github.com/RDFLib/rdflib/blob/70a3049e67d084fd086cd3afed48a8c09b5676a3/rdflib/graph.py#L1549-L1555

One consequence of this issue is:

- https://github.com/oxigraph/oxrdflib/issues/28

Selecting the value of `queryGraph` as follows seems to fix the issue:

```python
queryGraph = self.default_context.identifier if isinstance(self, ConjunctiveGraph) else self.identifier
```

aucampia:
The exact same problem exists in `Graph.update`, will submit a PR to fix both shortly.

Issue #2537, opened 2023-08-22
What does `Dataset.identifier` identify?


aucampia:
It's not clear that it identifies anything sensible, porbably it just an artifact that the inheritance tree is not optiomal and that Dataset should not inherit from ConjunctiveGraph

namedgraph:
Randomly found some ancient but maybe relevant discussion :)
https://code.alcidesfonseca.com/docs/rdflib/graphs_bnodes.html

gjhiggins:
> It's not clear that it identifies anything sensible, porbably it just an artifact that the inheritance tree is not optiomal and that Dataset should not inherit from ConjunctiveGraph

It will have significance for graph-aware stores as a means of distinguishing different instantiations of Dataset.

Most, if not all, examples of Dataset usage in the RDFLib documentation inadvertently elide some subtle issues pertaining to persistence by instantiating Datasets without specifying a store, relying on the "default" (Memory) store ...

```
ds1 = Dataset()
d2 = Dataset()
```

Each of the above Datasets is instantiated *with its own separate `Store` object* and there is a necessary but not typically-advertised caveat because when Datasets share a Store, there are some not-immediately-obvious-but-quite-predictable issues which pertain to the fact that every Dataset has a default graph with the identifier: `URIRef("urn:x-rdflib:default")` which needs to be stored using a Dataset-specific modifier (not implemented in the `Memory` Store), otherwise they will all share the same default graph, e.g.:

```python
def test_shared_memory():
    from rdflib.plugins.stores.memory import Memory
    memorystore00 = Memory(identifier=rdflib.URIRef("urn:example:memorystore00"))

    ds1 = rdflib.Dataset(store=memorystore00)
    ds1.default_graph.add((tarek, likes, pizza))

    ds2 = rdflib.Dataset(store=memorystore00)
    assert list(ds2.default_graph) == [(tarek, likes, pizza)]  # Unsurprising even if atypical
```

Removing the Dataset identifier will create the restriction of a maximum of one Dataset per instanced persistence. This may be an existing unspoken convention (or one I haven't yet found) because Fuseki implies/works under this restriction by separating Datasets via the API URLs.

Users of Datasets probably also need to be aware that this convention extends to the fact that each named graph source file effectively defines a complete Dataset and parsing two distinct named graph sources into a single Dataset might well lead to unexpected collisions in the default graph:

```python
def test_dataset_default_graph_collision() -> None:
    from rdflib.namespace import DCTERMS as DCT
[...]

PR #2546, opened 2023-08-26
fix: `queryGraph` selection for `query` and `update`

aucampia:
# Summary of changes

`Store.query` and `Store.update` have a parameter `queryGraph`, which
presumably specifies the initial [active
graph](https://www.w3.org/TR/sparql11-query/#sparqlDataset).

When SPARQL is evaluated against a `Dataset`, the initial active graph
should be the default graph
[[ref](https://www.w3.org/TR/sparql11-query/#sparqlAlgebraEval)],
however, when calling `rdflib.graph.Graph.query` or
`rdflib.graph.Graph.update`, they were passing
`rdflib.graph.Dataset.identifier` as the value of `queryGraph` to
`Store.query` and `Store.update`. The passed value would not have been
the identifier of the default graph of the Dataset, and would instead
have been an identifier that has really no meaning whatsoever (see
<https://github.com/RDFLib/rdflib/issues/2537>).

This change fixes the selection of the initial active graph by passing
`ConjunctiveGraph.default_context.identifier` if the `Graph.query` or
`Graph.update` method is called on a `ConjunctiveGraph` and
`ConjunctiveGraph.default_union` is `False`.

Related issues:
- Fixes <https://github.com/RDFLib/rdflib/issues/2538>.

# Checklist

  the same change.

### DIFF
--- a/test/test_graph/test_graph_store.py
+++ b/test/test_graph/test_graph_store.py
@@ -4,27 +4,32 @@ Tests for usage of the Store interface from Graph/NamespaceManager.
 
 import itertools
 import logging
+from test.data import SIMPLE_TRIPLE_GRAPH
 from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
     Dict,
     Iterable,
+    Mapping,
     Optional,
     Sequence,
     Tuple,
     Type,
     Union,
 )
+from unittest.mock import patch
 
 import pytest
 
 import rdflib.namespace
-from rdflib.graph import Graph
+from rdflib.graph import ConjunctiveGraph, Dataset, Graph
 from rdflib.namespace import Namespace
+from rdflib.plugins.sparql.sparql import Query
 from rdflib.plugins.stores.memory import Memory
+from rdflib.query import Result
 from rdflib.store import Store
-from rdflib.term import URIRef
+from rdflib.term import Identifier, URIRef, Variable
 
 if TYPE_CHECKING:
     from _pytest.mark.structures import ParameterSet
@@ -69,7 +74,7 @@ EGNS_V1 = EGNS["v1"]
 EGNS_V2 = EGNS["v2"]
 
 
-def make_test_graph_store_bind_cases(
+def make_graph_store_bind_cases(
     store_type: Type[Store] = Memory,
     graph_type: Type[Graph] = Graph,
 ) -> Iterable[Union[Tuple[Any, ...], "ParameterSet"]]:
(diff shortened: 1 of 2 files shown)


aucampia:
After this <https://github.com/oxigraph/oxrdflib/issues/28> should hopefully not be a thing.

test/test_graph/test_graph_store.py
```diff
@@ -205,9 +210,121 @@ def test_graph_store_bind(
...
+    result = Result("SELECT")
+    result.vars = [Variable("s"), Variable("p"), Variable("o")]
+    result.bindings = [
+        {
+            Variable("s"): URIRef("http://example.org/subject"),
+            Variable("p"): URIRef("http://example.org/predicate"),
+            Variable("o"): URIRef("http://example.org/object"),
+        },
+    ]
+
```
aucampia, line 305: Note to self: This is not needed as update has no result.

```suggestion
```

test/test_graph/test_graph_store.py
```diff
@@ -205,9 +210,121 @@ def test_graph_store_bind(
...
+    ) -> Result:
+        assert query_string == query
+        assert dict(store.namespaces()) == initNs
+        assert {} == initBindings
+        assert query_graph == queryGraph
+        assert {} == kwargs
+        return result
+
+    with patch.object(store, "query", wraps=mock_query) as wrapped_query:
+        result = graph.query(query_string)
```
aucampia, line 272: Note to self: rename variable

test/test_graph/test_graph_store.py
```diff
@@ -205,9 +210,121 @@ def test_graph_store_bind(
...
+        assert update_string == query
+        assert dict(store.namespaces()) == initNs
+        assert {} == initBindings
+        assert query_graph == queryGraph
+        assert {} == kwargs
+
+    with patch.object(store, "update", wraps=mock_update) as wrapped_update:
+        graph.update(update_string)
+        assert result.type == "SELECT"
+        assert list(result) == list(SIMPLE_TRIPLE_GRAPH.triples((None, None, None)))
```
aucampia, line 329: Note to self: not needed as this is an update, not a select.
```suggestion
```

aucampia: Fixed.

test/test_graph/test_graph_store.py
aucampia: Fixed.

test/test_graph/test_graph_store.py
aucampia: Fixed.

aucampia:
The other option is to make `Dataset.identifier` the same as `Dataset.default_context.identifier` - that may actually be somewhat sensible :thinking:

aucampia:
> The other option is to make `Dataset.identifier` the same as `Dataset.default_context.identifier` - that may actually be somewhat sensible 🤔

Maybe starting with this PR is best, we can change to having  `Dataset.identifier` the same as `Dataset.default_context.identifier` later and then just adjust the behaviour here if needs be.

Tpt (approved):
Thank you! It fixes oxrdflib

aucampia:
> Thank you! It fixes oxrdflib

Thank you for having a look

PR #2554, opened 2023-08-28
fix: SPARQL `LOAD ... INTO GRAPH` handling

aucampia:
# Summary of changes

`LOAD ... INTO GRAPH` stopped working correctly after the change to handling of the `publicID` `Graph.parse` parameter in RDFLib 7.0.0 (<https://github.com/RDFLib/rdflib/pull/2406>).

This is because `LOAD` evaluation relied on `publicID` to select the graph name. So after <https://github.com/RDFLib/rdflib/pull/2406> data would be loaded into the default graph even if a named graph is specified.

This change adds tests for `LOAD ... INTO GRAPH` and fixes the load evaluation.

# Checklist

  the same change.

### DIFF
--- a/test/utils/__init__.py
+++ b/test/utils/__init__.py
@@ -19,6 +19,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Sequence,
     Set,
     Tuple,
     Type,
@@ -35,7 +36,7 @@ import rdflib.plugin
 from rdflib import BNode, ConjunctiveGraph, Graph
 from rdflib.graph import Dataset
 from rdflib.plugin import Plugin
-from rdflib.term import Identifier, Literal, Node, URIRef
+from rdflib.term import IdentifiedNode, Identifier, Literal, Node, URIRef
 
 PluginT = TypeVar("PluginT")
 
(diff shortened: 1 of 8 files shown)


aucampia:
@rdfguy I think this should fix your problem https://twitter.com/oralassila/status/1695913300423225687

aucampia:
A consequence of this change is also that relative IRI lookup for graphs
loaded with `LOAD ... INTO GRAPH` is now relative to the source document
URI instead of the base URI of the graph being loaded into, which is
more correct I think.

I will add a test for this also tomorrow to confirm.

aucampia:
Also seems like windows does not like file URIs, I will look into that tomorrow also.

aucampia:
> Also seems like windows does not like file URIs, I will look into that tomorrow also.

Okay nevermind, tests pass on windows now, I just forgot to use URIs.

rdfguy:
Reading the commits the changes make sense, and in line with what I observed yesterday when my `LOAD`s ended up in wrong named graphs. Haven't tried the code yet, though.

test/utils/__init__.py
```diff
@@ -257,6 +258,23 @@ def assert_quad_sets_equals(
         else:
             assert lhs_set != rhs_set
 
+    @classmethod
+    def assert_collection_graphs_equal(
+        cls, lhs: ConjunctiveGraph, rhs: ConjunctiveGraph
+    ) -> None:
+        """
+        Assert that all graphs in provides collections are equal,
```
aucampia, line 266:
```suggestion
        Assert that all graphs in the provided collections are equal,
```

aucampia:
> A consequence of this change is also that relative IRI lookup for graphs loaded with `LOAD ... INTO GRAPH` is now relative to the source document URI instead of the base URI of the graph being loaded into, which is more correct I think.
> 
> I will add a test for this also tomorrow to confirm.

Done now.

test/utils/__init__.py
aucampia: fixed.

aucampia:
I will merge this tomorrow if there is no futher feedback.

PR #2580, opened 2023-09-10
feat: enable `check_untyped_defs` for Mypy

aucampia:
# Summary of changes

This change enables `check_untyped_defs` so that Mypy will check all code, even if it is not annotated with type hints.

This change also fixes all errors that were raised after `check_untyped_defs` was enabled.

# Checklist

  the same change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -255,6 +255,11 @@ no_implicit_optional = false
 implicit_reexport = false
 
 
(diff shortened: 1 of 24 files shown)


aucampia:
Compact diff can be seen here: https://gist.github.com/aucampia/e8089b17012ecd6814eb7e06954a8d4a

PR #2592, opened 2023-09-12
Draft: Dataset add named graph

Murloc6:
This is a draft PR to add to the discussion from #2591


aucampia (commented):
Thanks for the draft PR @Murloc6 - I'm not sure that the setter for identifier will interface properly with the backing store.

I'm also not entirely sure about named graph, I think we need to ensure the semantics is clear for this in terms of pre and post conditions, but if it will ensure all triples in g will go into the graph named `named_graph_identifier` then it is good.

rdflib/graph.py
```diff
@@ -2522,6 +2530,13 @@ def add_graph(
         """alias of graph for consistency"""
         return self.graph(g)
 
+    def add_named_graph(
+            self, named_graph_identifier: IdentifiedNode, g: Optional[Union[_ContextIdentifierType, _ContextType, str]]
```
aucampia, line 2534:
```suggestion
            self, named_graph_identifier: IdentifiedNode, g: Optional[Union[_ContextIdentifierType, _ContextType, str]]
```
```suggestion
            self, identifier: IdentifiedNode, graph: Optional[Union[_ContextIdentifierType, _ContextType, str]]
```

namedgraph:
Can we add `graphs: Dict[IdentifiedNod, Graph]` to `Dataset` instead of using `Graph.identifier`?

aucampia:
> Can we add `graphs: Dict[IdentifiedNod, Graph]` to `Dataset` instead of using `Graph.identifier`?

We should get rid of Graph.identifier but `Dataset.graphs` that would have to probably be a property and I would start with making it read only, it would have to interface with the store somehow.

nicholascar:
Maintainers will try to incorporate all these changes into a new Dataset API scheduled for RDFLib 8.x due for release in late 2024

PR #2699, opened 2024-01-31
Add SHACL path to RDFLib Path utility and corresponding tests

mgberg:
# Summary of changes

This adds a utility to parse valid SHACL Paths (e.g. the objects of triples with predicate `sh:path`) into an RDFLib-friendly object. The resulting object could either be:
 - A `URIRef`, when the path is simply a predicate
 - A `Path`, when the path is complex (e.g. `[sh:inversePath skos:broader]`)

This enables easy evaluation and manipulation of SHACL paths using RDFLib. For example:
- The output can be passed to any of several methods (`Graph.triples`, `Graph.value`, etc.) as a predicate to get value(s) at that path
- The `n3` method can be used to get a SPARQL property path representation of the path

Note that `pyshacl` does not have a utility to do this- rather, it only evaluates SHACL paths.
# Checklist

  the same change.

### DIFF
--- a/rdflib/extras/shacl.py
+++ b/rdflib/extras/shacl.py
@@ -1,0 +1,92 @@
+"""
+Utilities for interacting with SHACL Shapes Graphs more easily.
+"""
+from __future__ import annotations
+
+from typing import Optional, Union
+
+from rdflib import Graph, Literal, URIRef, paths
+from rdflib.namespace import RDF, SH
+from rdflib.paths import Path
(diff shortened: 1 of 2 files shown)


nicholascar (approved):
Would be good to see more than just one utility function for a "Utilities" SHACL file!

mgberg:
> Would be good to see more than just one utility function for a "Utilities" SHACL file!

Agreed!

Issue #2734, opened 2024-03-16
Graph.parse: Guess format when remote source content type is "text/plain"


lu-pl:
Parsing remote sources with `Graph.parse` relies on the [source.content_type](https://github.com/RDFLib/rdflib/blob/98346657aaea6bad23fe574b7e575d8fbc002da7/rdflib/graph.py#L1466) for determining the format/Parser.

This however fails if the response header returns "text/plain" as the content type, because "text/plain" is not registered in plugin.py and there is no further resolution strategy for that case in `Graph.parse`.

I would suggest the following change: If [`source.content_type`](https://github.com/RDFLib/rdflib/blob/98346657aaea6bad23fe574b7e575d8fbc002da7/rdflib/graph.py#L1466)  in `Graph.parse` is "text/plain" then  `rdflib.util.guess_format` should be invoked on `source`.

Suggestion for [line 1465](https://github.com/RDFLib/rdflib/blob/98346657aaea6bad23fe574b7e575d8fbc002da7/rdflib/graph.py#L1465)f.:
```python
from rdflib.util import guess_format

if format is None:
    if source.content_type == "text/plain":
        format = guess_format(source)
    else:
        format = source.content_type
```

I am currently using the following kludge for this:

```python
from rdflib.util import guess_format

class ParseGraph(Graph):
    def parse(self, *args, **kwargs):
        try:
            graph = super().parse(*args, **kwargs)
        except PluginException:
            guessed_format = guess_format(kwargs["source"])

            logger.info("Encountered PluginException.")
            logger.info(f"Retrying to parse with guessed format '{guessed_format}'.")

            kwargs.update({"format": guessed_format})
            graph = super().parse(*args, format=guessed_format, **kwargs)

        return graph
```

PR #2735, opened 2024-03-16
fix: use guess_format when plugin for format not found

edmondchuc:
# Summary of changes

Fixes https://github.com/RDFLib/rdflib/issues/2734
- add fallback to use `rdflib.util.guess_format` when `PluginException` occurs
- ensure `PluginException` still occurs when `guess_format` fails

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1475,7 +1475,19 @@ class Graph(Node):
             if format is None:
                 format = "turtle"
                 could_not_guess_format = True
-        parser = plugin.get(format, Parser)()
+        try:
+            parser = plugin.get(format, Parser)()
+        except plugin.PluginException:
+            # Handle the case when a URLInputSource returns RDF but with the headers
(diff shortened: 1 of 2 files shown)


rdflib/graph.py
```diff
@@ -1475,7 +1475,18 @@ def parse(
...
+        except plugin.PluginException:
+            # Handle the case when a URLInputSource returns RDF but with the headers
+            # as a format that does not exist in the plugin system.
+            # Use guess_format to guess the format based on the input's file suffix.
+            format = rdflib.util.guess_format(
+                source if not isinstance(source, InputSource) else str(source)
+            )
+            if format is None:
+                raise
+            parser = plugin.get(format, Parser)()
```
edmondchuc, line 1489: We may be able to factor this logic out into a function. That way, we can test it with a mock URL object rather than a real web request over the network.

nicholascar (approved):

PR #2761, opened 2024-04-12
Sparql connector default graph

guizsantos:
# Summary of changes

This PR adds the functionality to pass the `default_graph` parameter from within `kwargs`. This will be particularly useful to the LangChain `RdfGraph` object usage. I am pushing a PR there that will leverage the funcionality from this PR.

Furthermore, the `default_graph` parameter which is passed directly to the `query` method has priority over the `default_graph` parameter set from `kwargs`, which should be aligned with users expected behavior and offer backwards compatibilty.

Finally, I am not sure an Issue should be created and linked or new tests should be introduced, this is actually my first ever PR on a public repository, so any guidance is appreciated.

# Checklist

  the same change.
  
- If the change has a potential impact on users of this project: the changes should have no impact


WhiteGobo (changes requested):
(I am not of the maintainers but) this looks like a good functionality.

I would catch default_graph in `__init__(..., default_graph=None, ...)` and save it as `self.default_graph=default_graph` and later default to `self.default_graph` instead of `self.kwargs.get("default_graph")`. Because currently `self.kwargs` is only used as params for request.

I would consider adding short docstring for `SPARQLConnector.query` and`.update` that `default_graph` defaults to `self.kwargs["default_graph"]`.
A test for this functionality might be a good idea as well.

guizsantos:
Thanks for the feedback @WhiteGobo , I have little experience witht the usage of `rdflib` so I used the most straightforward way to pass the parameter over. However, I agree with your thoughts there... I will update the PR accordinly.

In the matter of testing, though, I am not sure how to proceed there. Let me take a look at the current tests and try to figure it out on my own. I will report back here.

nicholascar:
@ashleysommer @edmondchuc could one or both of you please review? I'm keen to see straightforward PRs be handled qiickly

nicholascar:
@guizsantos I'm a maintainer.

Just echoing @WhiteGobo: can you please add a short docstring? If you don't it won't be obvious at the function signature level that `query()` and `update()` can work with `kwargs` in addition to the function parameter option.

rdflib/plugins/stores/sparqlconnector.py
```diff
@@ -162,7 +163,8 @@ def update(
             raise SPARQLConnectorException("Query endpoint not set!")
 
         params = {}
-
+        
+        default_graph = default_graph or self.kwargs.get("default_graph")
```
ashleysommer, line 167:
```suggestion
        default_graph = default_graph or self.kwargs.get("default_graph")
```

ashleysommer: Ruff is complaining about an extra blank line here, that is what the remaining failing test in the CI is.

ashleysommer (changes requested):

nicholascar:
@guizsantos would you be able to provide any updates for this PR now that the GitHub testing mechanism is working properly again?

nicholascar (changes requested):
As per previous comment about DocString

ashleysommer (approved):

ashleysommer:
I just noticed one potentially concerning issue with this PR.

The Docstring for Query and Update in the SPARQL Connector states:
> Any additional keyword arguments will be passed to to the request, and can be used to setup timeouts etc.

So I was checking whether there might be a problem caused if the `default_graph` parameter also gets forwarded to the `Request` instance.  However it seems the way the Request is constructed must've changed since the docstring was written because in the current code only `kwargs["headers"]` and `kwargs["params"]` are used, so the comment about "all other kwargs" and "can be used to set up timeouts, etc" are both incorrect.

@guizsantos Are you able to fix up the docstring for both `query` and `update` methods reflect the actual purpose of `kwargs`, including the new `default_graph` parameter.

nicholascar:
I've assigned myself to add the doc string that @guizsantos hasn't added before this gets merged.

PR #2785, opened 2024-05-12
Assign consistent bnode identifiers in N3 formulas

eyusupov:
# Summary of changes

Blank nodes are assigned inconsistent identifiers when they are within different quoted graphs.
I could not find a normative section in the spec that would define the scope of blank nodes, but [section 3.6.1](https://w3c.github.io/N3/spec/#bnodeprplist) says that we "Then, we use this identifier within our [N3 graph](https://w3c.github.io/N3/spec/#dfn-n3-graph)", where "N3 document represents an [N3 graph](https://w3c.github.io/N3/spec/#dfn-n3-graph)".

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/plugins/parsers/notation3.py
+++ b/rdflib/plugins/parsers/notation3.py
@@ -905,7 +905,7 @@ class SinkParser:
         term = self._anonymousNodes.get(ln, None)
         if term is not None:
             return term
-        term = self._store.newBlankNode(self._context, why=self._reason2)
+        term = self._store.newBlankNode(self._context, uri='#' + ln, why=self._reason2)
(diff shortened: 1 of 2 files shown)


eyusupov:
Note: interestingly, QuotedGraph documentation mentions that it's intended to be completely disjoint from other graphs https://github.com/RDFLib/rdflib/blob/main/rdflib/graph.py#L33-L49, which contradicts my understanding of the situation

eyusupov:
Looks like this is expected behavior at the moment: https://w3c.github.io/N3/spec/semantics.html#example3

Issue #2670, opened 2023-12-19
Bug when using FROM statements


jcbiddle:
Using a FROM statement to query a Dataset induces a number of problems. Consider the following example:
```Python
from rdflib import Dataset
from rdflib.plugins import sparql

sparql.SPARQL_LOAD_GRAPHS = False  # Needed to ensure remote lookups aren't performed
sparql.SPARQL_DEFAULT_GRAPH_UNION = False  # Needed otherwise queries don't work at all!

data = """\
@prefix ex: <http://example.com#> .
ex:Graph1 {
ex:Alice a ex:Person .
}

ex:Graph2 {
ex:Charlie a ex:Person .
}
"""

query = """
PREFIX ex: <http://example.com#>
SELECT ?person
FROM ex:Graph1
WHERE {
?person a ex:Person
}
"""

ds = Dataset()
ds.parse(data=data, format="trig")

for row in ds.query(query):
    print(*row)
# Correctly outputs "http://example.com#Alice"
```

This snippet loads two named graphs, each with a single triple, then queries `ex:Graph1`.

After this query, the graph now contains a duplicate of `ex:Graph1` in the default graph:
```
for quad in ds.quads():
    print([term.fragment for term in quad[0:3]], quad[3])
# Outputs:
# ['Charlie', 'type', 'Person'] http://example.com#Graph2
# ['Alice', 'type', 'Person'] http://example.com#Graph1
# ['Alice', 'type', 'Person'] urn:x-rdflib:default
```

Aside from inadvertently increasing the size of the Dataset, this also induces a bug when querying other graphs in the Dataset. For example, if we now query `ex:Graph2`, we find that we get an erroneous result:

```
query = """
PREFIX ex: <http://example.com#>
SELECT ?person
FROM ex:Graph2
WHERE {
?person a ex:Person
}
"""
res = ds.query(query)
print("Second query results")
for row in res:
    print(*row)
# Outputs
# http://example.com#Alice <---- SHOULDN'T BE HERE
# http://example.com#Charlie
```
I have no clue as to the cause of this behaviour, but clearly something is wrong with the handling of FROM statements. Furthermore, none of the queries show here work if `sparql.SPARQL_DEFAULT_GRAPH_UNION = True`, which appears to be a separate but related problem.

WhiteGobo:
related discussion #2591

jcbiddle:
[...]

Issue #811, opened 2018-02-22
Using FROM and FROM NAMED on ConjunctiveGraph behaves not standard conform


white-gecko:
I want to use `FROM` and `FROM NAMED` in a SPARQL query to select the default graph resp. named graphs to execute the query on. But the RDFlib implementation does not act as it is described in the SPARQL 1.1 specification especially the section "13.2 Specifying RDF Datasets" (https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#specifyingDataset)

To demonstrate this behavior, I've created an MWE:

```
#!/usr/bin/env python3

import rdflib.plugins.sparql
from rdflib import ConjunctiveGraph
 
data = """
<urn:a> <urn:a> <urn:a> <urn:a> .
<urn:b> <urn:b> <urn:b> <urn:b> .
<urn:c> <urn:c> <urn:c> <urn:c> .
"""

if __name__ == "__main__":
    rdflib.plugins.sparql.SPARQL_DEFAULT_GRAPH_UNION = False  # Line A
    rdflib.plugins.sparql.SPARQL_LOAD_GRAPHS = False
    graph = ConjunctiveGraph()
    graph.parse(data=data, format='nquads')
    result = graph.query("SELECT * {?s ?p ?o}")               # Line B
    for resrow in result:
        print(resrow)
```

Running this code behaves as expected, while I will show derived examples in the following, by altering Line A and Line B:

**1. Replacing the Default Graph**

    rdflib.plugins.sparql.SPARQL_DEFAULT_GRAPH_UNION = True       # Line A
    result = graph.query("SELECT * FROM <urn:b> {?s ?p ?o}")      # Line B

- What I see: returns all three statements.
- What I expect: Only the statement `<urn:b> <urn:b> <urn:b>` as result
- Why is the actual result not correct?:

> A SPARQL query may specify the dataset to be used for matching by using the FROM clause and the FROM NAMED clause to describe the RDF dataset. If a query provides such a dataset description, then it is used in place of any dataset that the query service would use if no dataset description is provided in a query.
> (https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#specifyingDataset)

**2. Specifying a Named Graph but Querying the Default Gaph**

    rdflib.plugins.sparql.SPARQL_DEFAULT_GRAPH_UNION = True         # Line A
    result = graph.query("SELECT * FROM NAMED <urn:b> { ?s ?p ?o}") # Line B

- What I see: returns all three statements.
[...]

Issue #2147, opened 2022-11-02
Unusually high memory consumption in SPARQL query


karlb:
I am having problems with queries being killed due to exceeding the available memory on my machine. I'm using rdflib-hdt to work on a 70MB HDT file and don't see how querying that file could reasonably use up more than 10GB of RAM memory. Is there a way to deal with such a situation other than just writing simpler queries? Any way as a user to find out where the problem comes from (in an SQL query plan I would look for things like large sort buffers)?

The query I'm using is
```
    SELECT ?lexentry ?other_form
    WHERE {
        ?lexentry a ontolex:LexicalEntry ;
                  ontolex:otherForm ?other_form .

        ?other_form ontolex:writtenRep ?other_written .
        OPTIONAL { ?lexentry lexinfo:partOfSpeech ?pos }

        OPTIONAL { ?other_form olia:hasMood ?mood }
        OPTIONAL { ?other_form olia:hasNumber ?number }
        OPTIONAL { ?other_form olia:hasPerson ?person }
        OPTIONAL { ?other_form olia:hasTense ?tense }
        OPTIONAL { ?other_form olia:hasVoice ?voice }

        OPTIONAL { ?other_form olia:hasCase ?case }
        OPTIONAL { ?other_form olia:hasInflectionType ?inflection }
        OPTIONAL { ?other_form olia:hasDefiniteness ?definiteness }
        OPTIONAL { ?other_form olia:hasGender ?gender }
    }
```

I can provide exact code and data to reproduce this, if it is of any help. I have executed the same query successfully on the same dataset in virtuoso.

Versions used:
```
Python 3.9.2
rdflib        6.2.0
rdflib-hdt    3.0
```

According to https://github.com/RDFLib/rdflib-hdt/issues/17#issuecomment-1296772401, the HDT-Store is unlikely to be the culprit.

karlb:
Any feedback on this would be welcome, even it if is something like "what you do is a bad idea, because ...", "queries with many optionals are expected to use much memory in our current implementation" or "looks like a bug, but I don't have the time to look into it".

aucampia:
@karlb we likely won't get time to look into performance any time soon, but if you have time to look at it, please do. We are always happy for pull requests.

It may be that other stores than the default one work better.

olk-m:
@karlb  Is this still a problem? Can you show `pip freeze` output and show a small reproducible example that I can copy and run? Otherwise, please close the issue.

karlb:
[...]

Issue #1775, opened 2022-03-25
Sparql extremely slow with this query and content


rchateauneu:
This query takes dozen of minutes, but less than one second with Protégé.

```
        prefix cim:  <http://www.primhillcomputers.com/ontology/survol#>
        prefix rdfs:    <http://www.w3.org/2000/01/rdf-schema#>
        select ?my_dir_name3
        where {
        ?my_dir1 rdf:type cim:Win32_Directory .
        ?my_dir1 cim:Name "C:" .
        ?my_assoc_dir1 rdf:type cim:Win32_SubDirectory .
        ?my_assoc_dir1 cim:GroupComponent ?my_dir1 .
        ?my_assoc_dir1 cim:PartComponent ?my_dir2 .
        ?my_dir2 rdf:type cim:Win32_Directory .
        ?my_assoc_dir2 rdf:type cim:Win32_SubDirectory .
        ?my_assoc_dir2 cim:GroupComponent ?my_dir2 .
        ?my_assoc_dir2 cim:PartComponent ?my_dir3 .
        ?my_dir3 rdf:type cim:Win32_Directory .
        ?my_dir3 cim:Name ?my_dir_name3 .
        }
```

[content_snippet_Testing_Win32_Directory_Win32_SubDirectory_Win32_SubDirectory.txt](https://github.com/RDFLib/rdflib/files/8343131/content_snippet_Testing_Win32_Directory_Win32_SubDirectory_Win32_SubDirectory.txt)

PR #2794, opened 2024-05-29
Fix explicit dataset (`FROM` and `FROM NAMED` clauses)

apicouSP:
# Summary of changes

### `FROM` or `FROM NAMED` clause: redefine entirely the query's RDF dataset
When a query (`SELECT`,  `CONSTRUCT`,  `ASK`,  `DESCRIBE`) is using a `FROM` or `FROM NAMED` clause, we redefine entirely the query's RDF dataset
Include only the graphs in `FROM` clauses in the query's default graph
Include only the graphs in the `FROM NAMED` clauses in the query's named graphs

As a consequence, when a user defines a `FROM`  clause in his query but does not define `FROM NAMED`, then the named graphs are considered empty set.
And vice versa: if a user defines a `FROM NAMED` clause but does not define a `FROM` clause, then the default graph is considered empty.
I added tests specifically for that
That's my interpretation of [this](https://www.w3.org/TR/sparql11-query/#specifyingDataset:~:text=If%20a%20query%20provides%20such%20a%20dataset%20description%2C%20then%20it%20is%20used%20in%20place%20of%20any%20dataset%20that%20the%20query%20service%20would%20use%20if%20no%20dataset%20description%20is%20provided%20in%20a%20query.)

Also since the RDF Dataset is entirely redefined when using `FROM` or `FROM NAMED` clauses then for those queries the parameter `SPARQL_DEFAULT_GRAPH_UNION` will be ignored

### Load external graphs only if they don't already exist

Try to load external graphs only if they don't already exist in the given `ConjunctiveGraph`

As a consequence, you don't have to set `SPARQL_DEFAULT_GRAPH_UNION`  to `False` if all the graphs mentioned in `FROM` and `FROM NAMED` clauses already exist in the given `ConjunctiveGraph`

## WHAT it is not
- I didn't change the behavior of triples duplicated in different named graphs when those named graphs are merged into the query's default graph. Would need further discussion
- I didn't change the behavior or the updates (`DELETE` and `INSERT`). Since code is independent, I will probably make another PR for that

## WHY:
So queries behave closer to [W3C spec](https://www.w3.org/TR/sparql11-query/#specifyingDataset)

## Issues:
This PR fixes [this](https://github.com/RDFLib/rdflib/issues/2670) issue (confirmed)
And also [that issue](https://github.com/RDFLib/rdflib/issues/811) as far as I understand
This also solves partially [this discussion](https://github.com/RDFLib/rdflib/discussions/1541)

# Checklist

  the same change.
- [X] Checked that all tests passes .
- [] type checking: black and flake8 do not agree on how to format some lines, I chose to take black's suggestions
  
  - [X] Added or updated tests that fail without the change.
  - [ ] Updated relevant documentation to avoid inaccuracies. Just following W3C spec, which link is already in the docs.


apicouSP:
I now use Dataset instead of ConjunctiveGraph  as this one is now deprecated.
I also fixed style and type errors in new tests. Now black, ruff and mypy should pass.
Let me know if you need something to merge this!
Sorry I made too much commits.

ashleysommer:
Thank you @apicouSP for this Pull Request.
This looks like a good change to bring RDFLib SPARQL executor more in line with the SPARQL 1.1 specs.
Note, I've fixed all the issues that were preventing our CI pipelines from working properly, so I can see now your tests are passing.

I'm not super familiar with the RDFLib SPARQL executor (I don't know if any current maintainers are) so this will take a bit longer than normal to review this, but we'll try to get it merged before our upcoming RDFLib v7.1 release.

rdflib/plugins/sparql/sparql.py
```diff
@@ -264,11 +265,27 @@ def __init__(
         self.graph: Optional[Graph]
         self._dataset: Optional[ConjunctiveGraph]
         if isinstance(graph, ConjunctiveGraph):
-            self._dataset = graph
-            if rdflib.plugins.sparql.SPARQL_DEFAULT_GRAPH_UNION:
-                self.graph = self.dataset
+            if datasetClause:
+                self._dataset = ConjunctiveGraph()
```
ashleysommer, line 269: Should this be `Dataset()`?

nicholascar:
@recalcitrantsupplant can you please review this?

rdflib/plugins/sparql/sparql.py
nicholascar: Yes, `ConjunctiveGraph` is deprecated in 7.x for removal in 8.x, so shift all references to `Dataset`

ashleysommer:
I'm happy to merge this as-is now, and remove references to ConjunctiveGraph across the whole codebase in a separate PR.

rdflib/plugins/sparql/sparql.py
```diff
@@ -264,11 +265,27 @@ def __init__(
...
+                self.graph = Graph()
+                for d in datasetClause:
+                    if d.default:
+                        self.graph += graph.get_context(d.default)
+                        if not graph.get_context(d.default):
+                            self.load(d.default, default=True)
```
ashleysommer, line 275: Why call `graph.get_context(d.default)` twice?
```suggestion
                       from_graph = graph.get_context(d.default)
                       self.graph += from_graph
                        if not from_graph:
                            self.load(d.default, default=True)
```

rdflib/plugins/sparql/sparql.py
```diff
@@ -264,11 +265,27 @@ def __init__(
...
         if isinstance(graph, ConjunctiveGraph):
-            self._dataset = graph
-            if rdflib.plugins.sparql.SPARQL_DEFAULT_GRAPH_UNION:
-                self.graph = self.dataset
+            if datasetClause:
+                self._dataset = ConjunctiveGraph()
+                self.graph = Graph()
+                for d in datasetClause:
+                    if d.default:
+                        self.graph += graph.get_context(d.default)
```
ashleysommer, line 273: Seems like its pretty costly to copy _every_ triple from each `FROM` graph into a new Graph. Could it be done by adding the existing graph to a new Dataset (or ConjunctiveGraph), and setting default_union on that store?

rdflib/plugins/sparql/sparql.py
apicouSP: Thanks for your suggestion @ashleysommer, I also did the same for the named graphs

rdflib/plugins/sparql/sparql.py
apicouSP: I changed dataset init with `Dataset` instead of `ConjunctiveGraph`. I didn't change `if isinstance(graph, ConjunctiveGraph):` so we can still query `ConjunctiveGraph` for the moment. I guess this should be changed when `ConjunctiveGraph` is totally removed

rdflib/plugins/sparql/sparql.py
apicouSP: If I understand well your suggestion, this was previously done in the load function:
`self.graph += self.dataset.get_context(source)  # type: ignore[operator]`
Exactly the same way.
So that's just something that I didn't change.
I will see if I can optimize that

ashleysommer: Thanks for the explanation. I see now this of course isn't something you added, but it was just to this codepath.
There is no need to include an optimization for that in this PR, because it could introduce an unintended change in behaviour to SPARQL processing.

I am planning on doing a large sprint of work on optimising the SPARQL processor after RDFLib v7.1.0 is released.
I suspect efficiency problems (exactly like this one) are behind the odd extremely slow execution users are seeing in issues like:
https://github.com/RDFLib/rdflib/issues/2147
and
https://github.com/RDFLib/rdflib/issues/1775
and 
https://github.com/RDFLib/rdflib/issues/787

rdflib/plugins/sparql/sparql.py
ashleysommer: Thanks!

ashleysommer:
@apicouSP 
Currently tests failing after the latest update:
> py39-docs: commands[3]> poetry run python -m mypy --show-error-context --show-error-codes --junit-xml=test_reports/3.9-ubuntu-latest-mypy-junit.xml
  rdflib/plugins/sparql/sparql.py: note: In member `"__init__"` of class `"QueryContext"`:
  rdflib/plugins/sparql/sparql.py:286: error: Incompatible types in assignment (expression has type `"ConjunctiveGraph"`, variable has type "Optional[Dataset]")  [assignment]
  Found 1 error in 1 file (checked 398 source files)

rdflib/plugins/sparql/sparql.py
```diff
@@ -255,20 +255,39 @@ def __init__(
...
     ):
         self.initBindings = initBindings
         self.bindings = Bindings(d=bindings or [])
         if initBindings:
             self.bindings.update(initBindings)
 
         self.graph: Optional[Graph]
-        self._dataset: Optional[ConjunctiveGraph]
+        self._dataset: Optional[Dataset]
         if isinstance(graph, ConjunctiveGraph):
```
ashleysommer, line 267: The `isinstance(graph, ConjunctiveGraph)` statement on line below makes `mypy` think `graph` is a specifically a `ConjunctiveGraph` so it doesn't like the assignment to `self._dataset` that is typed `Optional[Dataset]`
So for now, until we strip out all of ConjunctiveGraph, lets make it `Union[Dataset, ConjunctiveGraph]`

```suggestion
        self._dataset: Optional[Union[Dataset, ConjunctiveGraph]]
        if isinstance(graph, (Dataset, ConjunctiveGraph)):
```

ashleysommer (changes requested):

ashleysommer (approved):

Issue #2692, opened 2024-01-28
Turning rdflib jsonld into a "full processor" (a.o. for schema.org compliance)


ghost:
Cloned from https://github.com/RDFLib/rdflib-jsonld/issues/62 :

The JSON-LD 1.1 draft spec mentions different levels of processing for JSON-LD: https://w3c.github.io/json-ld-syntax/#processor-levels

A pure processor can only parse JSON-LD expressed in JSON directly, but a full processor can also parse JSON-LD embedded in HTML.

It would be great if rdflib-jsonld would support this. It would make rdflib-jsonld a library that could be used for HTML documents following the schema.org guidelines for embedding (meta)data in HTML pages as described in their getting started guide https://schema.org/docs/gs.html.

Together with the RDFa & microdata parsers this can then work as a fully RDF based version of the Structured Data Testing tool from Google: https://search.google.com/structured-data/testing-tool.

wallberg-umd:
I'm looking into implementing this using https://github.com/RDFLib/rdflib-jsonld/pull/63 as a starting point.  In that code HTML parsing is attempted after JSON parsing fails, but I'm looking at choosing the parse using the content type of the source.

I see that the the JSON-LD test suite in "test/jsonld/1.1" already provides a number of tests which just need to be enabled.

wallberg-umd:
FYI: I'm occasionally working on this at https://github.com/wallberg-umd/rdflib/tree/issue-2692-embedded-jsonld-draft . I've added the basic functionality and enabled the existing tests.  I'm now working through making the tests pass.

wallberg:
I've completed an initial implementation for this issue, see https://github.com/wallberg/rdflib/tree/issue-2692-embedded-jsonld .

It contains one breaking change: when `rdflib.plugins.shared.jsonld.util.source_to_json` extracts JSON-LD from HTML, it needs to return the value of the HTML base element in addition to the JSON.  I took the simplest path and returned a tuple containing the JSON and the base.  

I can think of other ways to return the base without breaking the current return value:
1) Return `json` when processing a json document and `(json, base)` when processing an html document.
2) Add an optional parameter to return `(json, base)` instead of `json`.
3) Continue returning only `json`, but add an optional parameter which will receive the value of `base`.

I'd like to get some feedback on the preferred approach before submitting the PR.

A note on the current status of validation:
1) `task test` passes
2) `task lint` passes
3) `task mypy` fails, I believe to existing problems in main
[...]

PR #2804, opened 2024-06-20
Add JSON-LD extraction from HTML

wallberg:
Implementation of issue #2692.

See also https://w3c.github.io/json-ld-syntax/#embedding-json-ld-in-html-documents and https://www.w3.org/TR/json-ld11-api/#html-content-algorithms .

# Summary of changes

If `source.content_type` is "text/html" or "application/xhtml+xml" then parse the document as HTML and extract script elements of type="application/ld+json" as JSON-LD.

The default behavior is to extract only the first matching script element.  These overrides are available:
* To extract all script elements: supply an optional `extract_all_scripts=True` parameter to `JsonLDParser.parse()`
* To extract one script element with a specific id attribute value: add the id value as a fragment identifier in the IRI available from `source.getSystemId()`

# Detailed changes

`rdflib.plugins.parsers.jsonld.JsonLDParser.parse`

* add docstring
* change parameter list from **kwargs to explicit list
* add optional extract_all_scripts parameter
* get the fragment identifier from source.getSystemId()
* add fragment_id and extract_all_scripts parameters to the call to source_to_json

`rdflib.plugins.shared.jsonld.util.source_to_json`

* add docstring
* add optional fragment_id and extract_all_scripts parameters
* change the return value to a tuple with the extracted JSON document and value of the HTML base element
* if source.content_type is "text/html" or "application/xhtml+xml" then parse source as HTML and extract the appropriate script element(s) and the HTML base element

`test/jsonld/test_onedotone.py`

* enable all existing html tests (except html/f004-in). (*Note*: for more information on the failing html/f004-in test, see https://lists.w3.org/Archives/Public/public-json-ld-wg/2024May/0000.html)
* if inputpath ends with ".html" (with optional fragment identifier) then invoke runner.do_test_html

`test/jsonld/runner.py`

* add new `do_test_html` function (*Note*: the html test cases from the JSON-LD Test Suite combine testing
for JSON-LD extraction from the HTML with testing for other algorithms (e.g. compact/flatten),
which rdflib does not currently support. In order to test extraction only and ignore
the compact/flatten algorithms, do_test_html performs a graph comparison using
rdflib.compare.isomorphic, without serializing back to JSON)

# Breaking Changes

When `rdflib.plugins.shared.jsonld.util.source_to_json` extracts JSON-LD from HTML, it needs to return the value of the HTML base element in addition to the JSON. I took the simplest path and returned a tuple containing the JSON and the base.

I can think of other ways to return the base without breaking the current return value:

* Return json when processing a json document and tuple (json, base) when processing an html document.
* Add an optional parameter to return tuple (json, base) instead of json.
* Continue returning only json, but add an optional parameter which will receive the value of base.

# Checklist

  the same change.
  - [x] Created an issue to discuss the change and get in-principle agreement. #2692


nicholascar:
@wallberg you have prefixed this PR with "Draft" but it's not actually a draft PR. Do you consider it ready for review?

Since @ashleysommer fixed the GitHub vlaidation pipeline, it appears to be passing all tests.

wallberg:
@nicholascar yes, ready for review.

ashleysommer:
I'm happy to see this is using the built-in `html.parser` library, because we will soon be removing the old `html5lib` dependency from our dependencies.

nicholascar (approved):

Issue #2748, opened 2024-03-20
FAILED test/test_misc/test_parse_file_guess_format.py::TestFileParserGuessFormat::test_warning


ncopa:
The test/test_misc/test_parse_file_guess_format.py::TestFileParserGuessFormat::test_warning fails on Alpine Linux edge x86_64:

Python 3.11.8

```
=================================== FAILURES ===================================
____________________ TestFileParserGuessFormat.test_warning ____________________

self = <rdflib.plugins.parsers.notation3.SinkParser object at 0x7fbecb88c910>
argstr = '<?xml version="1.0"?>\n\n<!--\n  Copyright World Wide Web Consortium, (Massachusetts Institute of\n  Technology, Inst...rdf:datatype="http://www.w3.org/2001/XMLSchema#integer" xml:lang="fr">10</eg:baz>\n </rdf:Description>\n\n</rdf:RDF>\n'
i = 330, res = []

    def uri_ref2(self, argstr: str, i: int, res: MutableSequence[Any]) -> int:
        """Generate uri from n3 representation.
    
        Note that the RDF convention of directly concatenating
        NS and local name is now used though I prefer inserting a '#'
        to make the namesapces look more like what XML folks expect.
        """
        qn: typing.List[Any] = []
        j = self.qname(argstr, i, qn)
        if j >= 0:
            pfx, ln = qn[0]
            if pfx is None:
                assert 0, "not used?"
                ns = self._baseURI + ADDED_HASH  # type: ignore[unreachable]
            else:
                try:
>                   ns = self._bindings[pfx]
E                   KeyError: 'Description'

rdflib/plugins/parsers/notation3.py:1232: KeyError

During handling of the above exception, another exception occurred:

self = <Graph identifier=Nb4b72901e98b4f9f86eddbb8ac3005d9 (<class 'rdflib.graph.Graph'>)>
source = <_io.BufferedReader name='/tmp/tmpmcthgvqs/no_file_ext'>
publicID = None, format = 'turtle', location = None, file = None, data = None
args = {}, could_not_guess_format = True
parser = <rdflib.plugins.parsers.notation3.TurtleParser object at 0x7fbec6ac2510>

    def parse(
        self,
        source: Optional[
            Union[IO[bytes], TextIO, InputSource, str, bytes, pathlib.PurePath]
        ] = None,
        publicID: Optional[str] = None,  # noqa: N803
        format: Optional[str] = None,
        location: Optional[str] = None,
[...]

PR #2817, opened 2024-07-02
Fix testing for warnings with pytest 8 (#2748)

nphilipp:
New versions of pytest don’t catch log.warning() as a UserWarning. Use the caplog fixture instead.

Related: #2727
Fixes: #2748

# Summary of changes

Use the `caplog` fixture instead of `pytest.warns()` to check for warnings emitted with `log.warning()`, as outlined in https://github.com/RDFLib/rdflib/pull/2727#issuecomment-1994503300

# Checklist

  the same change.


nicholascar:
Thanks @nphilipp!

PR #2796, opened 2024-06-10
SPARQL result parsing

white-gecko:
# Summary of changes

- Remove plugin registration for "application/sparql-results+xml; charset=UTF-8"
  - The ResultParser plugin with the name "application/sparql-results+xml; charset=UTF-8" is never called in internal code. The only place where ResultParser plugins are requested is in ["rdflib/query.py:276"](https://github.com/RDFLib/rdflib/blob/53f72d87bbc9c0f193bdc703f030587b359bfb4e/rdflib/query.py#L276). In case of a content-type the charset is split off already.
  - cf. #941
- Resolve TODO: "Pull in these from the result implementation plugins?" for _response_mime_types in sparqlconnector
  - The `_response_mime_types` dict is replaced by a method `response_mime_types` that generates an `Accept` field containing the supported mime types. Per default these are all mime types for which ResultParser plugins are registered.
- Dynamically register the `rdflib.plugins.sparql.results.graph.GraphResultParser` plugin for all names of the Parser plugins since it calls the Parsers internally.

Eventually, these changes could be superseded by a more flexible plugin system, that would distinguish between plugin name, supported format, and supported mime-type.

The changes should maintain backward compatibility.

But this API is now more flexible. The `returnFormat` for `SPARQLConnector` and `SPARQLStore` is now optional and `None` per default. Also, the default Accept header lists all mime-types supported by parsers. Maybe this should be mentioned in the change notes.

~Currently, this PR includes #3240, these commits should be removed, if #3240 is merged.~

# Checklist

- [x] Checked that there aren't other open pull requests for
  the same change.
- [x] Checked that all tests and type checking passes.
- [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

PR #2818, opened 2024-07-05
Prevent Collection from adding 'rdf:nil rdf:rest rdf:nil.' triples

progval:
# Summary of changes

Creating 'Collection(graph, RDF.nil)' used to add a 'rdf:nil rdf:rest rdf:nil.' triple to the graph, which turned the empty list into an infinite list.

This was a side-effect of unconditionally appending 'seq' to the collection, which is not possible when the collection is 'nil' without ill side-effects.

This commit makes the append conditional; and also checks the collection is non-empty in case the user explicitly tries to add item to the empty list

# Checklist

  the same change.


rdflib/collection.py
```diff
@@ -208,6 +212,8 @@ def __iter__(self) -> Iterator[Node]:
         return self.graph.items(self.uri)
 
     def _end(self) -> Node:
+        if self.uri == RDF.nil:
```
ashleysommer, line 215: Can we move this check from _end down to after line 252 in `__iadd__`, I think it makes more sense to catch it there.
That way we can still use `_end()` when `nil` is deliberately added to the end of a list.

ashleysommer (changes requested):

rdflib/collection.py
```diff
@@ -243,6 +250,9 @@ def append(self, item: Node) -> Collection:
         return self
 
     def __iadd__(self, other: Iterable[Node]):
+        if self.uri == RDF.nil:
+            raise ValueError("Cannot append to empty list")
+
         end = self._end()
```
ashleysommer, line 256: I was thinking like this:
```suggestion
        end = self._end()
        if end == RDF.nil:
            raise ValueError("Cannot append to empty list")
```

I think `_end()` will still return `nil` when `self.uri` is `nil`, this is what we want.

Note, I think you can append `nil` to the `rdf:rest` on the end of a populated list, that will cause it to be a finalized list, that cannot have more items added.

So if `_end()` returns `rdf:nil` then the list is either empty or finalized, either way it cannot have another item appended.

ashleysommer (changes requested):

rdflib/collection.py
```diff
@@ -231,6 +235,9 @@ def append(self, item: Node) -> Collection:
 
         """
 
+        if self.uri == RDF.nil:
+            raise ValueError("Cannot append to empty list")
+
         end = self._end()
```
ashleysommer, line 241: Can we make the same change here to `append()` too please?
```suggestion
        end = self._end()
        if end == RDF.nil:
            raise ValueError("Cannot append to empty list")

```

ashleysommer (changes requested):

rdflib/collection.py
progval: Sorry, I missed it

ashleysommer:
There's an odd failing test now, weirdly its the same SPARQL test that is failing in #2796 that is completely unrelated to this PR.

ashleysommer:
That sporadic test is no longer failing, this can merge.

ashleysommer (approved):

PR #2838, opened 2024-07-24
Update dependabot.yml

ashleysommer:
Use the new "increase-if-necessary" strategy for Dependabot to reduce the occurrence of content-hash conflicts.

# Summary of changes

# Checklist

  the same change.
  

ashleysommer:
Another update. I'll set it to "widen", because the semantics for "widen" are already the same as "if-necessary", it is actually like "widen-if-necessary".

PR #2842, opened 2024-07-24
Ruff fixes

ashleysommer:
Last week saw Dependabot bump Ruff from 0.4.1 to 0.5.2, thats 12 releases, then yesterday bumped it from 0.5.2 to 0.5.4, another two releases.
In that jump of 14 releases, there were a bunch of fixes to Ruff, that caused some of RDFLib's files to need to be fixed.

The biggest change is now Ruff doesn't think "test" is a system module (which is correct) so now the isort feature sorts test in with the rest of the user's local modules.

There was also an update to the settings names in the pyproject.toml to stop Ruff from complaining about using the wrong settings names.


ashleysommer:
Finally passed.

PR #2853, opened 2024-07-26
Fix and extend implementation of `BytesIOWrapper`

ashleysommer:
Fix BytesIOWrapper to make it more complete, more useful, and compliant with Python's `BufferedIOBase` base class and `BinaryIO` typing, also compatible with Typeshed's `_WrappedBuffer` protocol.

This originally started out as a small [fix](https://github.com/RDFLib/rdflib/blob/7bc8c5f5f30a3262caeadb50d4b203e7853be4e3/rdflib/parser.py#L575) for an incorrect handling of an edge case in `FileInputSource`, but found it wasn't so simple because `BytesIOWrapper` only wrapped `str` instances, not `TextIO` streams. (I expected it would, because the counterpart `TextIOWrapper` _does_ wrap `BinaryIO` streams). While diving in to fix that too, I also found that many methods on `BytesIOWrapper` were either throwing `NotImplementedError()` or simply missing. So `BytesIOWrapper` was not aligned with its superclass `BufferedIOBase`, and not compliant with the python `BinaryIO` interface.

I implemented all of the missing and unimplemented methods, and reworked it to support wrapping both string instances and TextIO files and streams.

After these changes, BytesIOWrapper can now also wrap a TextIO Unicode file, or a StringIO, as well as the original wrapped str.

I implemented `@property name()` and others like `readable()` and `writable()` on `BytesIOWrapper` because `TextIOWrapper` expects it.  So you could now use `BytesIOWrapper` as a backing buffer to `TextIOWrapper`  but really don't do that.

I've also added a bunch more typing, and some tests for BytesIOWrapper that were missing, hopefully will increase our Coverage number.

### DIFF
--- a/rdflib/parser.py
+++ b/rdflib/parser.py
@@ -27,6 +27,7 @@ from typing import (
     TextIO,
     Tuple,
     Union,
+    cast,
 )
 from urllib.parse import urljoin
 from urllib.request import Request, url2pathname
@@ -40,9 +41,13 @@ from rdflib.term import URIRef
 
 if TYPE_CHECKING:
     from email.message import Message
+    from io import BufferedReader
     from urllib.response import addinfourl
 
+    from typing_extensions import Buffer
+
     from rdflib.graph import Graph
+
 
 __all__ = [
     "Parser",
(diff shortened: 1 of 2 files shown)


ashleysommer:
Fixed failing test introduced by my second commit.

There will be a follow up take here for down the road. The `BytesIOWrapper` class is now very long, and should be moved out of the rdflib `parser.py` file. It is a helper for `InputSource` but is now taking up more than half of the `parser.py` file.

nicholascar (approved):

PR #2863, opened 2024-07-30
Add initial implementation of RDF Patch parser.

recalcitrantsupplant:
# Summary of changes

Introduces support for parsing RDF Patch files, aligned to the specification here: https://afs.github.io/rdf-delta/rdf-patch.html

Supports:
- Adding and Deleting triples and quads from a Dataset.
- Adding and Deleting prefixes from a Dataset.
- blank node ID preservation from input documents. That is, blank nodes are NOT document scoped when parsing patches.
blank nodes can be specified either using the format:
  - `_:identifier`
  - `<_:identifier>`
both will be parsed as blank nodes and their identifier from the document they are parsed from preserved.

Known limitations:
- No support yet for Patch logs (a series of Patch files)

# Checklist

  the same change.
  - [x] ~Created an issue to discuss the change and get in-principle agreement.~ Discussed with a number of maintainers
TODO:

### DIFF
--- a/examples/parse_patch.py
+++ b/examples/parse_patch.py
@@ -19,11 +19,11 @@ ds = Dataset()
 # Apply add patch
 ds.parse(data=add_patch, format="patch")
 print("After add patch:")
-for triple in ds.de_skolemize():
+for triple in ds:
     print(triple)
 
 # Apply delete patch
 ds.parse(data=delete_patch, format="patch")
 print("After delete patch:")
-for triple in ds.de_skolemize():
+for triple in ds:
     print(triple)
--- a/examples/parse_patch.py
+++ b/examples/parse_patch.py
@@ -1,0 +1,29 @@
+from rdflib import Dataset
+
+# RDF patch data
+add_patch = """
+TX .
+A _:bn1 <http://example.org/predicate1> "object1" .
+A _:bn1 <http://example.org/predicate2> "object2" .
+TC .
+"""
+
+delete_patch = """
+TX .
+D _:bn1 <http://example.org/predicate1> "object1" .
+TC .
+"""
+
+ds = Dataset()
+
+# Apply add patch
+ds.parse(data=add_patch, format="patch")
+print("After add patch:")
+for triple in ds.de_skolemize():
+    print(triple)
+
+# Apply delete patch
+ds.parse(data=delete_patch, format="patch")
+print("After delete patch:")
+for triple in ds.de_skolemize():
+    print(triple)
--- a/examples/parse_patch.py
+++ b/examples/parse_patch.py
@@ -1,29 +1,35 @@
 from rdflib import Dataset
 
-# RDF patch data
-add_patch = """
-TX .
-A _:bn1 <http://example.org/predicate1> "object1" .
-A _:bn1 <http://example.org/predicate2> "object2" .
-TC .
-"""
 
-delete_patch = """
-TX .
-D _:bn1 <http://example.org/predicate1> "object1" .
-TC .
-"""
+def main():
+    # RDF patch data
+    add_patch = """
+    TX .
+    A _:bn1 <http://example.org/predicate1> "object1" .
+    A _:bn1 <http://example.org/predicate2> "object2" .
+    TC .
+    """
 
-ds = Dataset()
+    delete_patch = """
+    TX .
+    D _:bn1 <http://example.org/predicate1> "object1" .
+    TC .
+    """
 
-# Apply add patch
-ds.parse(data=add_patch, format="patch")
-print("After add patch:")
-for triple in ds:
-    print(triple)
+    ds = Dataset()
 
-# Apply delete patch
-ds.parse(data=delete_patch, format="patch")
-print("After delete patch:")
-for triple in ds:
-    print(triple)
+    # Apply add patch
+    ds.parse(data=add_patch, format="patch")
+    print("After add patch:")
+    for triple in ds:
+        print(triple)
+
+    # Apply delete patch
+    ds.parse(data=delete_patch, format="patch")
+    print("After delete patch:")
+    for triple in ds:
+        print(triple)
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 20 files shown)


rdflib/plugins/parsers/patch.py
```diff
@@ -0,0 +1,150 @@
...
+            self.add_triple_or_quad()
+        elif operation == Operation.DeleteTripleOrQuad:
+            self.delete_triple_or_quad()
+        elif operation == Operation.AddPrefix:
+            self.add_prefix()
+        elif operation == Operation.DeletePrefix:
+            self.delete_prefix()
+
+    def add_triple_or_quad(self):
+        self.sink.parse(data=self.line, format="nquads", skolemize=True)
```
ashleysommer, line 121: Is this a performance issue? `sink.parse()` will create a whole new instance of a new NQuads parser, each with its own new `Dataset` to parse into, for every line that is added in the rdp file.
Probably not a big deal for small test file, but with patches of hundreds of MB I can see this will become a bottleneck.

rdflib/plugins/parsers/patch.py
```diff
@@ -0,0 +1,150 @@
...
+        removal_ds = Dataset()
+        removal_ds.parse(data=self.line, format="nquads", skolemize=True)
+        triple_or_quad = next(iter(removal_ds))
+        self.sink.remove(triple_or_quad)
+
+    def add_prefix(self):
+        # Extract prefix and URI from the line
+        prefix, ns, _ = self.line.replace('"', "").replace("'", "").split(" ")
+        ns_stripped = ns.strip("<>")
+        self.sink.bind(prefix, ns_stripped)
```
ashleysommer, line 133: This adds the new prefix to the new local Dataset instance created on line 70, not the original dataset passed in. Is that an issue? Do the new prefixes appear on the original dataset's namespace manager because they share a store?

rdflib/plugins/parsers/patch.py
recalcitrantsupplant: I originally used parse line and can had some issue, can't recall what - will have another crack.

edmondchuc: Since this is reusing the existing parse method from the superclass, I'm guessing it doesn't support the `<_:label>` syntax defined in https://afs.github.io/rdf-delta/rdf-patch.html. Probably need to make a note of it somewhere in the docs.

edmondchuc: I don't believe the RDF Patch stores need to be concerned with skolemizing and de-skolemizing blank nodes and skolem IRIs. Would it be possible to override the `nodeid` method in the superclass to return the blank node verbatim from the input rather than generating a new bnode id?

edmondchuc (changes requested):

rdflib/plugins/parsers/patch.py
recalcitrantsupplant: > Since this is reusing the existing parse method from the superclass, I'm guessing it doesn't support the `<_:label>` syntax defined in https://afs.github.io/rdf-delta/rdf-patch.html. Probably need to make a note of it somewhere in the docs.

Sort of - I tested and RDFLib just parses them as URIRefs, so they'll work similar to skolem IRIs in that they won't be document scoped. It will be problematic for any internal operations which they should be blank nodes for though, and any serialisation would need to convert them back to blank nodes.

With your second comment there on skolemisation - I'll give that a go, and also parse any <_:x> as a blank node with x as the node ID.

rdflib/plugins/parsers/patch.py
recalcitrantsupplant: > Do the new prefixes appear on the original dataset's namespace manager because they share a store?
Yes - I've tried binding the prefixes to the original instance and the behaviour is the same.
When the Dataset is instantiated and parse is called, the method creates a ConjunctiveGraph instance, which in turn creates a Graph instance which parse is called on. This is probably why the nquads parser (which I've followed) creates another Dataset instance. Would it be best to remove the creation of a ConjunctiveGraph instance when Dataset.parse() is called in a separate PR, which should remove the need to create a local Dataset instance in the parser? (+ probably the nquads parser too).

rdflib/plugins/parsers/patch.py
recalcitrantsupplant: > I don't believe the RDF Patch stores need to be concerned with skolemizing and de-skolemizing blank nodes and skolem IRIs. Would it be possible to override the `nodeid` method in the superclass to return the blank node verbatim from the input rather than generating a new bnode id?

I've made this change + the support for the <:_label> syntax, could you please review again.

rdflib/plugins/parsers/patch.py
recalcitrantsupplant: Discussed w/ Nick & Ashley not making further changes to this PR, but future tasks to:
- discuss Dataset / Graph model / interfaces (add, parse etc.)
- simplify/standardise parse/serialize interfaces generally

ashleysommer:
@edmondchuc Merging is blocked on your pending review. Can you please re-review and approve to so we can merge it.

edmondchuc (approved):

examples/parse_patch.py
```diff
@@ -0,0 +1,29 @@
...
+ds.parse(data=add_patch, format="patch")
+print("After add patch:")
+for triple in ds.de_skolemize():
+    print(triple)
+
+# Apply delete patch
+ds.parse(data=delete_patch, format="patch")
+print("After delete patch:")
+for triple in ds.de_skolemize():
+    print(triple)
```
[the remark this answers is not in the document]
recalcitrantsupplant: Removed in ba3c04108c24b13cc925e0bc95eb1b995eb0fd40

rdflib/plugins/parsers/patch.py
```diff
@@ -0,0 +1,183 @@
...
+        :type bnode_context: `dict`, optional
+        :param bnode_context: a dict mapping blank node identifiers to `~rdflib.term.BNode` instances.
+                              See `.W3CNTriplesParser.parse`
+        """
+        assert sink.store.context_aware, (
+            "RDFPatchParser must be given" " a context aware store."
+        )
+        # type error: Incompatible types in assignment (expression has type "ConjunctiveGraph", base class "W3CNTriplesParser" defined the type as "Union[DummySink, NTGraphSink]")
+        self.sink: Dataset = Dataset(store=sink.store)
+        self.skolemize = skolemize
```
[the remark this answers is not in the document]
recalcitrantsupplant: Yeah I just mirrored the other parsers here. Could there be use cases where someone wants to skolemize when parsing?

Issue #2445, opened 2023-06-13
The default graph should have no name.


aucampia:
According to <https://www.w3.org/TR/rdf11-concepts/#section-dataset>

> An RDF dataset is a collection of [RDF graphs](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-graph), and comprises:
> 
> - Exactly one default graph, being an [RDF graph](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-graph). The default graph does not have a name and may be empty.
> - Zero or more named graphs. Each named graph is a pair consisting of an [IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-iri) or a blank node (the graph name), and an [RDF graph](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-graph). Graph names are unique within an RDF dataset.

In RDFLib however:

- The default context in ConjunctiveGraph has a BNode as name.
- The default context in Dataset has a IRI as a name.

Related issues:

-

gjhiggins:
> According to https://www.w3.org/TR/rdf11-concepts/#section-dataset
> > The default graph does not have a name and may be empty.

I encountered this in my explorations of a solution ...

> In RDFLib however:
> 
>     * The default context in ConjunctiveGraph has a BNode as name.

If the entire class is to be deprecated, issue can be ignored

>     * The default context in Dataset has a IRI as a name.

Doesn't seem to present an issue for the API: exclude the URI from `list(Dataset.contexts())` but include the `Graph` object in `list(Dataset.graphs())` and I've found the `Dataset.default_graph` reference to be convenient.

OTOH, the Store API could be changed. AIUI RDFLib now has only context-aware Stores and I believe that uncontextualized statements are just quads with NULL as the fourth element (ISTR it's implemented with "" as a key).

ajnelson-nist:
FWIW, I have a static-rebuild motivation for requesting this change too.  I think.

I include built graph files as part of unit test output, and version-control those files to show result evolution.  One of those generated formats is JSON-LD.  If I try to write the least code possible and use `rdfpipe` to handle joining or format-converting graph files, I end up with an artifact that's a bit of a nuisance for independent results reproduction.  The path to the input file is embedded in the output as what I thought would be the "default" graph's IRI.  For instance, one workflow I just tried took a single Turtle file as input, converted it with `rdfpipe --output-format json-ld in.ttl > out.json`, and I got this form of output (excerpted):

```json
[
  {
[...]

PR #2867, opened 2024-08-01
Change some more internal usages of ConjunctiveGraph to Dataset to silence warnings

ashleysommer:
This was originally a very simple patch to change the use of `ConjuctiveGraph` to `Dataset` in the hextuples and nquads graph parsers. 

The motivation was to silence the deprecation warnings that are emitted when simply parsing a nquads file or hextuples file, even if you were not personally using a ConjunctiveGraph, the warnings were being emitted for the use of `ConjunctiveGraph` within the parsers.

One of the key differences between ConjuctiveGraph and Dataset is how the `default_graph` works. When creating a new  `ConjunctiveGraph` (`cg`), you can pass an `identifier` that becomes `cg.identifier` and also becomes the identifier of the default_graph. It uses a BNode as the identifier for the default graph if you don't pass one. A named graph with BNode as identifier is actually considered an "unnamed graph".

On the other hand, you cannot pass an `identifier` to a `Dataset`. The Dataset's internal identifier is _always_ a new BNode, (that isn't even used for anything_. The identifier of `ds.default_graph` is the constant `DATASET_DEFAULT_GRAPH_ID` (`"urn:x-rdflib:default"`). So the difference is ConjunctiveGraph by default has an Unnamed graph as its default_graph, and Dataset has a Named Graph (with a URI) as its default_graph.

Changing `ConjunctiveGraph` to `Dataset` in the Hextuples and NQuads parsers inadvertently exposed an underlying existing bug in both parsers. If the destination graph (aka, the "sink") already had a Named default_graph assigned (like in the case of every Dataset), the parser would ignore that and parse into a _new_ default_graph, causing one extra graph in the resulting dataset than you were expecting. That was fixed in this PR.

Another key difference between ConjuctiveGraph and Dataset is that `ConjunctiveGraph` always sets `default_union` to False unless you force it True, but `Dataset` is easily selectable, with the default `False`.
So for compatibility sake, we set `default_union` to True on the usage of Dataset in these two parsers.

Finally, after those changes were made I started getting failing tests in the seemingly unrelated JSON-LD Serializer. This was caused by the change of nquads to use `Dataset` in the parser. Specifically, it was an "Unnamed default-graph" vs "Named default-graph" issue. There was an existing bug in the JSON-LD serializer that treated all named-graphs as "non-default" graphs, where it wraps them in a `@graph{}` wrapper and adds an `@id` with the named-graph identifier. All "Unnamed" graphs were treated as "default-graph" and were added to the top-level context in the JSON-LD file. The problem is it was treating the Dataset named graph with name `DATASET_DEFAULT_GRAPH_ID` as a "non-default" graph because it has a URI as its identifier. This was triggering a lot of JSON-LD serializer tests to fail because the expected output was different.

So this PR fixes that issue in the JSON-LD serializer too, so it now treats Dataset default_graph as an un-named graph.
Note, this is directly related to https://github.com/RDFLib/rdflib/issues/2445 which still needs to be fixed.

### DIFF
--- a/rdflib/plugins/parsers/hext.py
+++ b/rdflib/plugins/parsers/hext.py
@@ -11,7 +11,7 @@ import warnings
 from io import TextIOWrapper
 from typing import TYPE_CHECKING, Any, BinaryIO, List, Optional, TextIO, Union
 
-from rdflib.graph import ConjunctiveGraph, Graph
+from rdflib.graph import ConjunctiveGraph, Dataset, Graph
 from rdflib.parser import InputSource, Parser
 from rdflib.term import BNode, Literal, URIRef
 
@@ -32,21 +32,22 @@ __all__ = ["HextuplesParser"]
 class HextuplesParser(Parser):
     """
     An RDFLib parser for Hextuples
-
     """
 
     def __init__(self):
+        super(HextuplesParser, self).__init__()
+        self.default_context: Optional[Graph] = None
         self.skolemize = False
 
     def _parse_hextuple(
-        self, cg: ConjunctiveGraph, tup: List[Union[str, None]]
+        self, ds: Union[Dataset, ConjunctiveGraph], tup: List[Union[str, None]]
     ) -> None:
         # all values check
         # subject, predicate, value, datatype cannot be None
         # language and graph may be None
         if tup[0] is None or tup[1] is None or tup[2] is None or tup[3] is None:
             raise ValueError(
-                "subject, predicate, value, datatype cannot be None. Given: " f"{tup}"
+                f"subject, predicate, value, datatype cannot be None. Given: {tup}"
             )
 
         # 1 - subject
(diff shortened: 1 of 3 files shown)


nicholascar (approved):
Great work @ashleysommer

ashleysommer:
I know the big wall of text in the original comment is hard to follow. The concepts are difficult, and the changes are hard to describe. `ConjuctiveGraph` vs `Dataset` usage is resulting in hidden bugs around the use of default_union, default_context, and `identifier` that are treated differently on Dataset than on ConjunctiveGraph.

Simply changing ConjunctiveGraph to Dataset is exposing these hidden bugs in unexpected areas, thankfully our test suit seems to be catching them. This issue is going to get worse as we approach full removal of ConjunctiveGraph, before it gets better.

The connection between the change in nquads parser, and the filing JSON-LD serializer tests, is because those tests read in the test data from the test suite on disk in nquads format. When the nquads parser changed from parsing using ConjunctiveGraph to parsing using Dataset, it begain using the Dataset's default_graph (DATASET_DEFAULT_GRAPH_ID), as the Named default graph, and that exposed the bug in JSON-LD where it could not handle that (seems like it was simply never tested to serialize a Dataset instance that uses the original default_graph.)

edmondchuc (approved):
Nice, looks good to me.

Issue #2812, opened 2024-06-29
JSON-LD serialization apparently loses XSD type information.


lu-pl:
JSON-LD serialization apparently loses XSD type information. 

```python
from rdflib import Graph, Literal, URIRef, XSD


graph: Graph = Graph()
graph.add(
    (
        URIRef("https://test.subject"),
        URIRef("https://test.predicate"),
        Literal("test type", datatype=XSD.string),
    )
)

## 1. check literal type of constructed graph object
# pass
assert all(o.datatype for o in graph.objects())

## 2. check literal type of xml parsed graph object
xml_serialized: str = graph.serialize(format="xml")
graph_xml_parsed: Graph = Graph().parse(data=xml_serialized, format="xml")

# pass
assert all(o.datatype for o in graph_xml_parsed.objects())

## 3. check literal type of json-ld parsed graph object (fails)
json_ld_serialized: str = graph.serialize(format="json-ld")
graph_json_parsed: Graph = Graph().parse(data=json_ld_serialized, format="json-ld")

# fail
assert all(o.datatype for o in graph_json_parsed.objects())
```

nicholascar:
The default JSON-LD & Turtle datatypes _are_ `xsd:string`. As per the JSON-LD [type coercion section](https://www.w3.org/TR/json-ld/#type-coercion).

So 

```json
[
  {
    "@id": "https://test.subject",
    "https://test.predicate": [
      {
        "@value": "test type",
        "@type": "http://www.w3.org/2001/XMLSchema#string"
      }
    ]
  }
]
```

is equivalent to 

```json
[
  {
    "@id": "https://test.subject",
    "https://test.predicate": [
      {
        "@value": "test type"
      }
    ]
  }
]
```

If you apply any sort of special datatype to the literal, JSON-LD & Turtle etc will preserve it.

So nothing is lost and I see no issue here.

lu-pl:
True, if a `Literal.datatype` attribute is `None` it means the default datatype, `XSD.string`.

Still I find it confusing that if `XSD.string` is explicitly passed as datatype to a `Literal` object, the JSON-LD *serializer* is not explicit about it while the Turtle and XML serializers are.

E.g. serializing the above `graph` object to turtle results in 

```ttl
@prefix ns1: <https://> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ns1:test.subject ns1:test.predicate "1"^^xsd:string .
```

serializing to JSON-LD results in 

[...]

PR #2889, opened 2024-08-13
Issue #2812: Reflect explicitly XSD-typed Literals in JSON-LD serialization

lu-pl:
# Summary of changes

This PR removes a single code path in the JSON-LD serializer responsible for dropping  XSD type information explicitly supplied in an `rdflib.Literal`.

Without the change, the following drops the `XSD.string` type for the JSON-LD serialization, but not for Turtle and XML serializations.

```python
from rdflib import Graph, Literal, URIRef, XSD

graph = Graph()

graph.add(
    (
        URIRef("https://test.subject"),
        URIRef("https://test.predicate"),
        Literal("test value explicit type", datatype=XSD.string),
    )
)
graph.add(
    (
        URIRef("https://test.subject"),
        URIRef("https://test.predicate"),
        Literal("test value implicit type"),
    )
)

print(graph.serialize())
print(graph.serialize(format="xml"))
print(graph.serialize(format="json-ld"))
```

Note: A test in the JSON-LD test-suite had to be modified for this change; this test will fail without the change.

Closes: #2812

# Checklist

  the same change.
  

ashleysommer:
@lu-pl 
I thought in an RDFLib Literal there is no way to distinguish between a provided datatype and an inferred datatype.

How does this know whether to emit a datatype or not?

lu-pl:
> @lu-pl I thought in an RDFLib Literal there is no way to distinguish between a provided datatype and an inferred datatype.
> 
> How does this know whether to emit a datatype for only user-provided datatypes?

Thanks for the reply! 

Is there such a thing as an inferred *explicit* datatype in `rdflib.Literal`? 

A datatype is either explicitly assigned, in which case `rdflib.Literal.datatype` holds the datatype URI, or not, in which case `rdflib.Literal.datatype` is `None`.

```python
from rdflib import Liteal, XSD

literal_implicit_type = Literal("value")
literal_explicit_type = Literal("value", datatype=XSD.string)

print(literal_implicit_type.datatype)
print(literal_explicit_type.datatype)
```

The problem I am trying to address is that the JSON-LD *serializer* does not retain datatypes defined in an `rdflib.Graph` instance (regardless of how that graph got populated).

lu-pl:
One thing that deems me worthy of discussion is the intended purpose of [line 400](https://github.com/RDFLib/rdflib/blob/6882f5fa64a36b43d5a541a58382738204bc7ec3/rdflib/plugins/serializers/jsonld.py#L400) in `plugins.serializers.jsonld.py`.

This code path is responsible for dropping type info for explicitly typed Literals (#2812), so the PR removes it. But I don't see any reason for that code path to exist in the first place.

PR #2892, opened 2024-08-19
jsonld - Improve handling of URNs in norm_url

avillar:
# Summary of changes

Improve handling of URNs in JSON-LD's `norm_url`. This mainly makes it possible to use URNs in `@base`. 

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugins/shared/jsonld/util.py
+++ b/rdflib/plugins/shared/jsonld/util.py
@@ -230,12 +230,12 @@ def norm_url(base: str, url: str) -> str:
     if parsed_url.scheme:
         # Assume full URL
         return url
-    if parsed_base.scheme in ('urn', 'urn-x'):
+    if parsed_base.scheme in ("urn", "urn-x"):
         # No scheme -> assume relative and join paths
-        base_path_parts = parsed_base.path.split('/', 1)
-        base_path = '/' + (base_path_parts[1] if len(base_path_parts) > 1 else '')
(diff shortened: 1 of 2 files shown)


ashleysommer:
Thanks @avillar  This looks good. This will make it into the upcoming v7.1.0 release.

Quick question, is it common to have URN identifiers in JSON-LD? I've never seen it before personally. It it even part of the JSON-LD spec?

I do however see the need for this fix in RDFLib, because we support URNs in all other serializations, and we need to be able to round-trip conversions to JSON-LD and back again, so we need to ensure URNs are preserved in the serialization.

avillar:
I'm not sure how common they are "in the wild", but the JSON-LD spec [specifically notes that IRIs are not limited to URLs](https://www.w3.org/TR/json-ld/#iris), using an example of a UUID URN. In any case, as you mention, as JSON-LD is a concrete RDF syntax, so support for IRIs (including URNs) is desirable.

In our case, we use them for plain JSON + JSON-LD context = RDF conversions that we the run through a SHACL validator, so they're useful for identifying the focus nodes while signaling to the user that we're not talking about resolvable resources.

ashleysommer:
> In our case, we use them for plain JSON + JSON-LD context = RDF conversions that we the run through a SHACL validator, so they're useful for identifying the focus nodes while signaling to the user that we're not talking about resolvable resources.

Nice example use case, thanks!

PR #2901, opened 2024-09-01
Replace deprecated method in csv2rdf

### LINKED ISSUE #2618: csv2rdf cannot parse config file
When trying to pass a settings file as an argument to -f to csv2rdf.py like this:
`python csv2rdf.py -f <path to settings> <path to csv file>`
it raises an exception and displays the following
```
  File "D:\Watos\env\Lib\site-packages\rdflib\tools\csv2rdf.py", line 553, in <module>
    main()
  File "D:\Watos\env\Lib\site-packages\rdflib\tools\csv2rdf.py", line 468, in main
    config.readfp(open(opts["-f"]))
    ^^^^^^^^^^^^^
AttributeError: 'ConfigParser' object has no attribute 'readfp'. Did you mean: 'read'?
```

I checked the code of csv2rdf.py and tried changing that call from config.readfp to config.read, but it still fails and won't parse the csv file

ageorgou:
# Summary of changes

Fixes #2618.

The code for tne `csv2rdf` command-line tool calls `ConfigParser.readfp`, which has been deprecated for a while and [has been removed in Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#configparser). The `read_file` method is a straightforward replacement.

I have checked manually that the old and new version give the same result for a sample config file.

I've also taken the opportunity to add another small test, since the coverage for that tool is quite low. The new test doesn't check that the contents of the config file are processed correctly, only that it is opened.

# Checklist

  the same change.
  - No change besides future-proofing.

### DIFF
--- a/rdflib/tools/csv2rdf.py
+++ b/rdflib/tools/csv2rdf.py
@@ -482,7 +482,7 @@ def main():
 
     if "-f" in opts:
         config = configparser.ConfigParser()
-        config.readfp(open(opts["-f"]))
+        config.read_file(open(opts["-f"]))
         for k, v in config.items("csv2rdf"):
             if k == "out":
                 csv2rdf.OUT = codecs.open(v, "w", "utf-8")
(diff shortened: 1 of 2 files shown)


ashleysommer (approved):

ashleysommer:
@ageorgou 
Thank you for this fix.

PR #2911, opened 2024-09-25
Replace html5lib with html5lib-modern

### LINKED ISSUE #2749: RFE: move away from deprecated `html5lib`
**Is your feature request related to a problem? Please describe.**
It would be nice tu cut tail of some legacy modules decencies.
One of those modules is `html5lib`.

**Describe the solution you'd like**
it wold be good to remove use od=f the html5lib deprecated  `html5lib` like it has been  done with pip ~2 years ago.
https://github.com/pypa/pip/pull/11259

**Describe alternatives you've considered**

**Additional context**
`html5lib` depends on `six` which is on list of deprecated modules even longer implanting this RFE would make easier kill two birds using one stone 😋

ashleysommer:
## Replace html5lib with html5lib-modern
This removes another source of `six` dependency.
Fixes #2749

Note however, this is not the last source of `six` sub-dependency.

The `isodate` module used in RDFLib also depends on `six`, and also hasn't been updated in over 3 years, (though it is not marked as deprecates like `html5lib`)

The path to fix, upgrade or replace `isodate` is not so clear.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,7 +21,7 @@ build:
     # the readthedocs environment.
     - pip install -r devtools/requirements-poetry.in
     post_install:
-    - poetry export --only=main --only=docs --extras=html -o requirements.txt
+    - poetry export --only=main --only=docs -o requirements.txt
     - pip install --no-cache-dir -r requirements.txt
     - pip install .
     - python -c "from rdflib import Graph; print(Graph)"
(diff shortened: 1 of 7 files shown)


ashleysommer:
Update on the `isodate` issue, looks like that package is maintained by our colleague Gerhard at TERN, so he might be able to push out a new version with no `six` dependency.

ashleysommer:
If there is no opposition, I will merge this on Monday.

nicholascar:
Waiting to see if the new html5lib release removes six

ashleysommer:
> Waiting to see if the new html5lib release removes six

Actually I misinterpreted the situation. The original author has not put out a new html5lib release, and more projects are now moving to html5-modern.

PR #2925, opened 2024-10-09
Fix for reassigned term aliases

avillar:
# Summary of changes

In JSON-LD, when a term is aliased to one of the `NODE_KEYS` (e.g, `id` to `@id`), but is later reassigned in a child context (e.g., `id` to `dct:identifier`), rdflib incorrectly keeps the original alias and uses it for the binding (e.g., `@id` instead of `dct:identifier`).

This commit removes any aliases for a term inherited from parent contexts when that term is bound in the current context.

# Checklist

  the same change.


ashleysommer:
Looks like a great fix, thanks @avillar

ashleysommer (approved):

PR #2929, opened 2024-10-15
Redo XSD Datetime, Date, Time, Duration parser and serializers

ashleysommer:
This PR introduces a XSD-specific datetime parser/serializer module that acknowledges there are subtle but important differences between iso8601 and XSD date strings.

RDFLib had been relying on some undocumented (and faulty) parsing behaviour in the old version of  `isodate` that coincidentally allowed parsing of strings that are XSD-compliant but not iso8601 compliant.

The new version of isodate fixes this issue, that means many tests were now failing in RDFLib due to XSD strings no longer parsing (eg, dates with a timezone "2024-01-31+10:00" that are XSD compliant but not iso compliant, the old version of isodate dropped the tz off the end by accident, so it worked fine, the new version does not allow it).

The preferred solution would be to switch to the new `datetime.fromisoformat()` function in Python stdlib in Python 3.11+, but there are three issues to solve with that plan:
1) We still need a solution for Python 3.8, 3.9, 3.10 users
2) The stdlib parser throws similar errors to the new `isodate` version, it also doesn't support non-standard iso formats.
3) There is no stdlib replacement for the `isodate`'s `Duration` class and corresponding `duration_isoformat` and `parse_duration` functions.
    - stdlib `timedelta` doesn't support deltas with years and months components, needed in XSD durations.
    - stdlib `timedelta` doesn't have `to_isoformat()` or `fromisoformat()` formats, needed to parse XSD duration strings (based on iso8601 duration strings).

So this PR addresses all of the above.

1) A new `xsd_datetime.py` module is added to RDFLib to house all of this XSD-datetime/duration-specific parsing/serializing logic.
2) The `Duration` class and `duration_isoformat()` and `parse_duration()` functions are absorbed from `isodate` into RDFLib.
3) For Python 3.11+, the dependency on external `isoformat` library is removed
    - All parsing of XSD_Date, XSD_Datetime, XSD_Time are done by the stdlib `fromisoformat()` function
    - XSD-specific quirks (eg, Timezones in dates) are handled in the `xsd_datetime.py` module _before_ calling into the stdlib function.
4) For Python < 3.11, we now depend on the new v0.7.2 release of `isodate`
    - All parsing of XSD_Date, XSD_Datetime, XSD_Time are done by the updated `isodate` utility
    - XSD-specific quirks (eg, Timezones in dates) are handled in the `xsd_datetime.py` module _before_ calling into `isodate`.
    - This update removes our last subdependency on `six` in our dependency tree!
5) Shortcut serializers are added to `xsd_datetime.py` based on `strftime` that can output XSD-specific XSD_Duration and XSD_Datetime strings.

### DIFF
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -46,7 +46,7 @@ from rdflib.term import (
     URIRef,
     Variable,
 )
-from rdflib.xsd_datetime import Duration, parse_xsd_datetime
+from rdflib.xsd_datetime import Duration, parse_datetime  # type: ignore[attr-defined]
 
 
 def Builtin_IRI(expr: Expr, ctx: FrozenBindings) -> URIRef:
@@ -692,7 +692,7 @@ def default_cast(e: Expr, ctx: FrozenBindings) -> Literal:  # type: ignore[retur
         if x.datatype and x.datatype not in (XSD.dateTime, XSD.string):
             raise SPARQLError("Cannot cast %r to XSD:dateTime" % x.datatype)
         try:
-            return Literal(parse_xsd_datetime(x), datatype=e.iri)
+            return Literal(parse_datetime(x), datatype=e.iri)
         except:  # noqa: E722
             raise SPARQLError("Cannot interpret '%r' as datetime" % x)
 
--- a/rdflib/plugins/sparql/operators.py
+++ b/rdflib/plugins/sparql/operators.py
@@ -21,7 +21,6 @@ from functools import reduce
 from typing import Any, Callable, Dict, NoReturn, Optional, Tuple, Union, overload
 from urllib.parse import quote
 
-import isodate
 from pyparsing import ParseResults
 
 from rdflib.namespace import RDF, XSD
@@ -47,6 +46,7 @@ from rdflib.term import (
     URIRef,
     Variable,
 )
+from rdflib.xsd_datetime import Duration, parse_xsd_datetime
 
 
 def Builtin_IRI(expr: Expr, ctx: FrozenBindings) -> URIRef:
@@ -521,8 +521,13 @@ def Builtin_TZ(e: Expr, ctx) -> Literal:
     if not d.tzinfo:
         return Literal("")
     n = d.tzinfo.tzname(d)
-    if n == "UTC":
+    if n is None:
+        n = ""
+    elif n == "UTC":
         n = "Z"
+    elif n.startswith("UTC"):
+        # Replace tzname like "UTC-05:00" with simply "-05:00" to match Jena tz fn
+        n = n[3:]
     return Literal(n)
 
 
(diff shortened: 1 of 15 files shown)


ashleysommer:
Finally after battling with the type checker, the linter, the ruff formatter, the doctring tests, the autodocs generator errors, then the linter again, then the type checker again, then the autodocs generator again, and finally the ruff formatter again, this is now passing all tests.

test/test_sparql/test_functions.py
```diff
@@ -167,7 +167,7 @@ def test_function(expression: str, expected_result: Identifier) -> None:
     if isinstance(expected_result, type):
         assert isinstance(actual_result, expected_result)
     else:
-        assert expected_result == actual_result
+        assert actual_result == expected_result
```
ashleysommer, line 170: This change was made to help the test suite output.
PyTest always assumes when asserting equality that LHS is actual, and RHS is expected, but this test harness compared them the other way around, so PyTest output was backwards that made debugging more difficult.

test/test_sparql/test_datetime_processing.py
```diff
@@ -94,8 +94,8 @@ def test_datetime_duration_subs():
     SELECT (?d - ?duration AS ?next_year)
     WHERE {
         VALUES (?duration ?d) {
-            ("P1Y"^^xsd:yearMonthDuration"2019-05-28T12:14:45Z"^^xsd:dateTime)
-            ("P1Y"^^xsd:yearMonthDuration"2019-05-28"^^xsd:date)
+            ("P1Y"^^xsd:yearMonthDuration "2019-05-28T12:14:45Z"^^xsd:dateTime)
+            ("P1Y"^^xsd:yearMonthDuration "2019-05-28"^^xsd:date)
```
ashleysommer, line 98: This was an error I found in the datetime test function, there was a missing space between the literal values, causing incorrect tests to be run on durations.

rdflib/plugins/sparql/operators.py
```diff
@@ -521,8 +521,13 @@ def Builtin_TZ(e: Expr, ctx) -> Literal:
...
         return Literal("")
     n = d.tzinfo.tzname(d)
-    if n == "UTC":
+    if n is None:
+        n = ""
+    elif n == "UTC":
         n = "Z"
+    elif n.startswith("UTC"):
+        # Replace tzname like "UTC-05:00" with simply "-05:00" to match Jena tz fn
+        n = n[3:]
```
ashleysommer, line 530: This was a required change because previous implementation used tzinfo instances (timezone instances) from `isodate`, but the new implementation always uses timezones from the stdlib. The SPARQL builtin `TZ()` will return the "timezone name" if known, in python this uses the `tzname()` function.
The difference is that stdlib timezones and isodate timezones have different `tzname()` generation pattern. In `isodate` an unnamed timezone with "-5H" offset will be have tzname of "-05:00", but in stdlib an unnamed timezone with "-5H" offset will have tzname of "UTC-05:00" (more correct IMHO).

So for consistence and backwards compatibility, this was change was added to normalize the tzname output, I tested Jena's `TZ()` SPARQL function too, and it also outputs "-05:00" like `isodate` did.

nicholascar (approved):
Just make sure to add yourself, @ashleysommer, to the contributor notes at the top of `xsd_datetime.py`

PR #2931, opened 2024-10-16
Bovlb patch 1

bovlb:
# Summary of changes

The following SPARLQ where clause:
```
?s :hasIngredient [:name "chicken"], [:name "butter"] .
```

should be parsed as:

```
_:1 :name "butter" .
_:2 :name "chicken" .
?s :hasIngredient _:1 .
?s :hasIngredient _:2
```

but was instead being parsed as:

```
_:1 :name "chicken" .
_:1 :name _:2 .
_:2 :name "butter" .
?s :hasIngredient _:1
```

I fixed the bug and added a new test.

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/plugins/sparql/parser.py
+++ b/rdflib/plugins/sparql/parser.py
@@ -59,6 +59,7 @@ def expandTriples(terms: ParseResults) -> List[Any]:
     Expand ; and , syntax for repeat predicates, subjects
     """
     # import pdb; pdb.set_trace()
+    last_subject, last_predicate = None, None  # Used for ; and ,
     try:
         res: List[Any] = []
         if DEBUG:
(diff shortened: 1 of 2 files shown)


bovlb:
pre-commit.ci autofix

bovlb:
pre-commit.ci autofix

ashleysommer:
Wow. Thanks. I can't believe a bug like that in the SPARQL parser has existed undetected for so long.
Was the bug only when the two bnodes are on the same line? Did it have the same problem when be bnodes were separated over two lines like this?
```
?s :hasIngredient [:name "chicken"],
                  [:name "butter"] .
```

bovlb:
> Wow. Thanks. I can't believe a bug like that in the SPARQL parser has existed undetected for so long. 

I didn't believe it at first myself.

> Was the bug only when the two bnodes are on the same line? Did it have the same problem when thee bnodes were separated over two lines like this?

That doesn't change the token sequence, so it's the same problem.

The bug was that we were assuming a comma implied we should repeat the third-last and second-last results as the subject and object of the new triple.  This assumption didn't hold when we had inserted a triple from a blank node.

PR #2943, opened 2024-10-21
Fix parser bug and add test

bovlb:
# Summary of changes

The where clause:

```
[ :p1 ?x ] :p2 ?y .
```

should be treated identically to:

```
[ :p1 ?x ; :p2 ?y ] .
```

Instead it fails with:
> TypeError: 'in <string>' requires string as left operand, not CompValue

I believe the first pattern is valid based on the fourth example given in [§4.1.4](https://www.w3.org/TR/sparql12-query/#QSynBlankNodes) of the SPARQL 1.2 spec.

# Checklist

  the same change.
  

ashleysommer:
Thanks for fixing the unrelated error. I've pushed the fix into `main` for that import ordering bug, so that should free up PRs to pass correctly.

rdflib/plugins/sparql/parser.py
```diff
@@ -81,7 +81,7 @@ def expandTriples(terms: ParseResults) -> List[Any]:
                 if len(t) > 1:
                     res += t  # Don't update last_subject/last_predicate
                 # is this bnode the subject of more triples?
-                if i + 1 < l_ and terms[i + 1] not in ".,;":
+                if i + 1 < l_ and not any(terms[i + 1] == token for token in ".,;"):
```
ashleysommer, line 84: I've tried to interpret this code change several times, and it seems to be the logic of the pre-fix state and post-fix state is functionally identical.
1) If this term is not any of the characters in this string, _vs_
2) None of the characters in this string are my term.

Clearly it has fixed the issue, the tests are passing, but I'm not clear on _how_ the change works.

bovlb: The issue is that the terms are not always strings. The __eq__ method on CompValue successfully returns False in this case, but the __contains__ method on str apparently raises a type error, although I can’t find any documentation of this behaviour. Compare with the similar test on line 72.

PR #2951, opened 2024-10-23
Replace html5lib with html5rdf, make it an optional dependency

ashleysommer:
This commit undoes some of what was merged in the v7.1 release that made HTML tokenisation of rdf:HTML literals non-optional.
This makes it an optional extra feature of RDFLib again, and moves away from using the deprecated html5lib (and the alias replacement html5-modern), to using the new html5rdf that lives in the RDFLib org.

This also fixes the tests that comparison code paths that were crashing, that was the reason for making HTML tokenisation non-optional in v7.1. Now that these are fixed, its safe to not have html5rdf installed when using RDFLib.

### DIFF
--- a/devtools/constraints.min
+++ b/devtools/constraints.min
@@ -6,6 +6,6 @@ pyparsing==2.1.0
 importlib-metadata==4.0.0
 berkeleydb==18.1.2
 networkx==2.0
-html5lib-modern==1.2.0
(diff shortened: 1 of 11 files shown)


ashleysommer:
@nicholascar  @edmondchuc 
This is ready for review now.

nicholascar (approved):

PR #2971, opened 2024-11-05
chore: type annotations - convert uses of Optional and some Union usage to union operator |

edmondchuc:
This PR is a follow-up to https://github.com/RDFLib/rdflib/pull/2963. Largely, it updates usages of `typing.Optional` and `typing.Union` type annotations to the more modern and preferred union operator `|`.

Some other notes:
- Some w3c sparql10 tests now passing, but not sure how. Results from the earl report has changed.

# Summary of changes

# Checklist

  the same change.
  

ashleysommer:
@edmondchuc 
Did you need to fix any typing errors that arose from these changes and needed to be fixed? I expect if they are all simply replacing Union and Optional with their alternate forms, then MyPy should not have generated any new errors.

Something else to note, on the topic of `Optional[X]` vs `X | None` personally I prefer the former. Even though now the Python developers discourage its use and advocate to use the `| None` form, I personally tend to still use `Optional[]` in my projects, as a matter of personal taste. I think its because it reminds me of the Rust `Option<>` type.

ashleysommer (approved):

edmondchuc:
I didn't run into any mypy errors. No runtime changes and no semantic changes to the codebase.

> Something else to note, on the topic of `Optional[X]` vs `X | None` personally I prefer the former. Even though now the Python developers discourage its use and advocate to use the `| None` form, I personally tend to still use `Optional[]` in my projects, as a matter of personal taste. I think its because it reminds me of the Rust `Option<>` type.

Yeah, that's fair. I definitely was tossing up between the two, but the python docs helped me over the line given that their official stance is to prefer the union operator now.

For others reading this. From the [python typing](https://docs.python.org/3/library/typing.html#typing.Union) docs:

> To define a union, use e.g. Union[int, str] or the shorthand int | str. Using that shorthand is recommended.

Keep in mind that `Optional` is shorthand for `Union`.

> Optional[X] is equivalent to X | None (or Union[X, None]).

PR #2989, opened 2024-11-21
Fixed incorrect ASK behaviour for dataset with one element

jcbiddle:
# Summary of changes
 - Fixed incorrect behaviour for ASK queries with a single FROM or FROM NAMED statement.
 - Previously, the dataset clause would be a single element instead of a singleton list, leading to the query failing.

# Checklist

  the same change.


nicholascar (approved):
Thanks @jcbiddle: every little fix counts!

Issue #2767, opened 2024-04-15
Bug in `longturtle` serialization


mschiedon:
The `longturtle` serializer fails to emit a whitespace separator between a predicate and a list of objects if one of these objects is a blank node (and the blank node cannot be 'inlined', i.e. is used more than once). The problem can be reproduced using this Python code:

```python
from rdflib import Graph

input = '''\
@prefix ex: <https://example.org/> .

ex:1 a ex:Thing ;
    ex:relatedTo ex:3, _:bnode0 .

ex:2 a ex:Thing ;
    ex:relatedTo _:bnode0 .

_:bnode0 a ex:Thing .
'''

graph = Graph().parse(data=input, format='turtle')
output = graph.serialize(format='longturtle')
print(output)
assert output.find('relatedTo_:') == -1, \
    'Missing whitespace separation between predicate' \
    ' and the first blank node of a list of objects.'
```

The resulting Turtle with the bug looks like below. Note the missing space between the predicate `ex:relatedTo` and blank node `_:n40fef3a41a034be9a7116df126afd613b1` for the `ex:1` case. The `ex:2` case does correctly use a space separator when serializing because it's a single object and not a list.

```turtle
PREFIX ex: <https://example.org/>

ex:1
    a ex:Thing ;
    ex:relatedTo_:n40fef3a41a034be9a7116df126afd613b1 ,
        ex:3 ;
.

ex:2
    a ex:Thing ;
    ex:relatedTo _:n40fef3a41a034be9a7116df126afd613b1 ;
.

_:n40fef3a41a034be9a7116df126afd613b1
    a ex:Thing ;
.
```

I believe the issue might be solved by adding an additional indent in the `longturtle.py` source code on [this line](https://github.com/RDFLib/rdflib/blob/7c99182845a156ba511a5240e266badc905ab73a/rdflib/plugins/serializers/longturtle.py#L295), as shown in the code below.

```python
    def objectList(self, objects):
        count = len(objects)
        if count == 0:
            return
        depthmod = (count == 1) and 0 or 1
        self.depth += depthmod
        first_nl = False
        if count > 1:
            if not isinstance(objects[0], BNode):
                self.write("\n" + self.indent(1))
                # BUG: Gave below line an extra indent.
                first_nl = True
        self.path(objects[0], OBJECT, newline=first_nl)
        for obj in objects[1:]:
            self.write(" ,")
            if not isinstance(obj, BNode):
                self.write("\n" + self.indent(1))
[...]

Issue #1890, opened 2022-05-06
Sorting Turtle output?


ajnelson-nist:
Hi all, probably esp. @nicholascar,

[PR 1425](https://github.com/RDFLib/rdflib/pull/1425) added the `ttl2` serializer, which since became `longturtle`.  One of the goals was to reduce Git difference noise.

I unfortunately find that most of the noise in Git diffs comes from Turtle output not being sorted.  This is an effect both on nodes with IRIs and with blank nodes that (if I'm guessing right) may be sorting based on their skolemized IDs.

Could an option to either of the Turtle serializers be added to sort the output, for those willing to pay the `.serialize` time cost?  I personally like Git-tracking demonstration results, so in many of my uses I'm willing to pay the time cost of sorting graphs.

aucampia:
Would be happy with such an option, and it would be useful with other formats also. My only concern is that our current approach for adding options  (i.e. using **kwargs) is a bit adhoc - it would be nice to consider an option that provides validation and type safety somehow.

gjhiggins:
> it would be nice to consider an option that provides validation

Assuming you mean validation of URIs, informed opinion on [stackoverflow](https://stackoverflow.com/a/18724720) is that URI validation is best left to a dedicated library so that would suggest the necessity of acquiring a dependency but I guess that can be handled via the pip `extras` mechanism.

aucampia:
> it would be nice to consider an option that provides validation

No I mean that provides a way to validate the options independently, maybe something like:

plugins.get("json-ld", Serializer).make_opts(...) -> JSONLDOptions

and then make it possible to pass serializer options into serialize

not entirely sure it is that critical though.

gjhiggins:
> No I mean that provides a way to validate the options independently..  and then make it possible to pass serializer options into serialize

Ah, so. Yes, that would address the needs of those wishing to exert more control over the JSONLD serializer.

ajnelson-nist:
Interesting.  I had thought that typing for keyword arguments would be compatible with a `**kwargs` pass-through, but it seems at least with `mypy` that's not the case.  Here are some notes in case others thought like I did.

The following program runs fine through Python, print `3`.  It passes `mypy --strict` review.  But it includes what look like to me to be three type errors (`f()` for 4--6).  Python 3.8.9, mypy 0.950.

```python
#!/usr/bin/env python3

# This software was developed at the National Institute of Standards
# and Technology by employees of the Federal Government in the course
[...]

PR #2731, opened 2024-03-12
fix: gh actions on PR

edmondchuc:
# Summary of changes

All actions are now passing.

Most of the file changes are due to black formatting.

The readthedocs build is now working again. It stopped working after the readthedocs poetry version was updated to 1.8.x. 

A small change was made to the longturtle serializer to fix ruff linting errors with trailing whitespace. Mypy also reported that the removed code was unreachable. This change doesn't appear to affect the longturtle serializer tests.

# Checklist

- [x] Checked that there aren't other open pull requests for
  the same change.
- [x] Checked that all tests and type checking passes.
- If the change adds new features or changes the RDFLib public API:
  
  - [ ] Created an issue to discuss the change and get in-principle agreement.
  - [ ] Considered adding an example in `./examples`.
- If the change has a potential impact on users of this project:
  
  - [ ] Added or updated tests that fail without the change.
  - [ ] Updated relevant documentation to avoid inaccuracies.
  - [ ] Considered adding additional documentation.
- [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

PR #2997, opened 2024-11-28
feat: sort longturtle blank nodes

edmondchuc:
# Summary of changes

### Fixes https://github.com/RDFLib/rdflib/issues/1890 - Sorting Turtle output?

This change improves git diffing Turtle data serialized with the `longturtle` serializer. Previously, blank nodes were not sorted, and round-trips using RDFLib's `turtle` parser and `longturtle` serializer would have blank node objects flip flop around, making it difficult to read real changes using git diff.

This PR fixes the above by implementing a sort on values where triples in the object position are blank nodes. The blank nodes are sorted by grabbing their concise-bounded description graph and sorting it as a string in their `longturtle` serialization.

This adds an additional cost to the `longturtle` serializer, but I think the cost is worth it if we are after a deterministic output with blank nodes.

Note that this depends on RDF data parsed using the `turtle` parser and its behaviour of how it assigns blank nodes. Exact serialization with data added via the graph object cannot be guaranteed. This would require implementing RDF Canonicalization to guarantee the same blank node identifiers in the graph.

Once RDF Canonicalization is implemented, sorting by the blank node identifier directly will be enough to guarantee deterministic serialization, and the expensive CBD sorting can be removed.

Update: looks like top-level blank nodes with no inbound relationships don't get sorted. I think this makes sense since we're only applying the sort to blank nodes in the object position. If we cared about sorting blank nodes in the subject position, we can apply the same kind of sorting based on the text serialization of the CBD onto the subject blank nodes.

https://github.com/RDFLib/rdflib/blob/08dd4b79f84638250529adb650c75a3305bad66d/rdflib/plugins/serializers/longturtle.py#L106-L112

### Fixes https://github.com/RDFLib/rdflib/issues/2767 - Bug in `longturtle` serialization

This PR also fixes the missing trailing whitespace in the special case described by @mschiedon.

> The longturtle serializer fails to emit a whitespace separator between a predicate and a list of objects if one of these objects is a blank node (and the blank node cannot be 'inlined', i.e. is used more than once)

This fix was previously fixed in PR https://github.com/RDFLib/rdflib/pull/2700 but was inadvertently reverted in PR https://github.com/RDFLib/rdflib/pull/2731 when I was trying to fix the ruff linting rule in the test case. To get around the ruff linting rule, I've now moved the target result of the test case into a text file.

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/plugins/serializers/longturtle.py
+++ b/rdflib/plugins/serializers/longturtle.py
@@ -293,7 +293,34 @@ class LongTurtleSerializer(RecursiveSerializer):
     def verb(self, node, newline=False):
         self.path(node, VERB, newline)
 
+    def sortObjects(
+        self, values: list[URIRef | BNode | Literal]
+    ) -> list[URIRef | BNode | Literal]:
+        """
+        Perform a sort on the values where each value is a blank node. Grab the CBD of the
+        blank node and sort it by its longturtle serialization value.
+
+        Identified nodes come first and the sorted blank nodes are tacked on after.
+        """
+        bnode_map: dict[BNode, list[str]] = {}
+        objects = []
+        for value in values:
+            if isinstance(value, BNode):
+                bnode_map[value] = []
+            else:
+                objects.append(value)
+
+        for bnode in bnode_map:
+            cbd = self.store.cbd(bnode).serialize(format="longturtle")
+            bnode_map[bnode].append(cbd)
+
+        sorted_bnodes = sorted(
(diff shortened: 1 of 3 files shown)


edmondchuc:
@ajnelson-nist would be good to see if you can test this with some of the data you work with and confirm that this PR fixes the blank node sorting issues in your workflows.

nicholascar (approved):
Awesome stuff @edmondchuc!

PR #1978, opened 2022-06-01
Sort Turtle output

ajnelson-nist:
# Summary of changes

This patch series starts an API for sorting graph serializations, beginning with Turtle.  The main objective is to produce consistent Turtle output, no matter the order of RDF triples being added to a `Graph`.

This PR will close [Issue 1890](https://github.com/RDFLib/rdflib/issues/1890).

The initial pair of patches only starts the PR.  Some discussion will be needed to design the remaining patches.

The PR's total effects are expected to be additive and preserve backward compatibility.

# Checklist

- [x] Checked that there aren't other open pull requests for the same change.
- [x] Added tests for any changes that have a runtime impact.
- [ ] Checked that all tests pass.  *(Test added, known to not yet pass, but should pass by end of PR development.)*
- [ ] Checked that all type checking passes.  *(Type signatures are being added to touched code.)*
- For changes that have a potential impact on users of this project:
  - [ ] Updated relevant documentation to avoid inaccuracies.  *(Not yet; unknown what to update.)*
  - [ ] Considered adding additional documentation.  *(Not yet; unknown what to update.)*
  - [ ] Considered adding an example in `./examples` for new features.  *(Not yet; unknown what to update.  Hard-coded sorted graph in new unit test can be copied to desired location.)*
  - [ ] Considered updating our changelog (`CHANGELOG.md`).
- [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so maintainers can fix minor issues and keep your PR up to date.

PR #2500, opened 2023-07-25
Longturtle improvements

nicholascar:
# Summary of changes

Fixes some ugliness in the longturtle format, in particular:

- double newline at the end of the file
- inconsistent Blank Node bracket positions - some line training, some newlining, now all newline
- Blank Node incorrect indentation - now they are all 1 indentation further right than their parents and their content is another further indent right

# Checklist

- [x] Checked that there aren't other open pull requests for
  the same change.
- [x] Checked that all tests and type checking passes.
- If the change adds new features or changes the RDFLib public API:
  - N/A
- If the change has a potential impact on users of this project:
  
  - [x] Added or updated tests that fail without the change.
  - [x] Updated relevant documentation to avoid inaccuracies.
  - [x] Considered adding additional documentation.
- [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

PR #3008, opened 2024-12-13
feat: deterministic longturtle serialisation using RDF canonicalization + n-triples sort

edmondchuc:
This PR improves upon https://github.com/RDFLib/rdflib/pull/2997 to remove the bespoke object blank node sorting technique to instead use sorted n-triples str lines after applying the RGDA1 graph canonicalisation algorithm. Fixes https://github.com/RDFLib/rdflib/issues/1890.

It's necessary to read in the sorted n-triples lines with `skolemize=True` to preserve the blank node identifiers from the canonicalisation algorithm.

Now that we can sort reliably by the blank node identifiers, this implementation works for all blank node positions in an RDF statement, no matter if it's in the subject or object position. It even works for blank nodes at the top-level.

@ajnelson-nist, I've added your blank node test from [Sort Turtle output (#1978)](https://github.com/RDFLib/rdflib/pull/1978) and it's now passing, yay!

Update: this also fixes the double up of semicolons bug when the subject is a top-level blank node. See https://github.com/RDFLib/rdflib/pull/3008/commits/412fb5d63d79c5d5816e05751e6be6bd791b1385.

### Checklist

  the same change.
  
### DIFF
--- a/test/test_serializers/test_serializer_longturtle_sort.py
+++ b/test/test_serializers/test_serializer_longturtle_sort.py
@@ -84,31 +84,31 @@ ns1:B
 []    ns1:has
         [
             rdfs:seeAlso ns1:A ;
-        ] ; ;
+        ] ;
 .
 
-[]    rdfs:seeAlso ns1:B ; ;
+[]    rdfs:seeAlso ns1:B ;
 .
 
 []    ns1:has
         [
             rdfs:seeAlso ns1:C ;
-        ] ; ;
+        ] ;
 .
 
-[]    rdfs:seeAlso ns1:A ; ;
+[]    rdfs:seeAlso ns1:A ;
 .
 
-[]    rdfs:seeAlso ns1:C ; ;
+[]    rdfs:seeAlso ns1:C ;
 .
 
-[]    rdfs:seeAlso ns1:B ; ;
+[]    rdfs:seeAlso ns1:B ;
 .
 
 []    ns1:has
         [
             rdfs:seeAlso ns1:B ;
-        ] ; ;
+        ] ;
 .
 
 """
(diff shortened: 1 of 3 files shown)


test/test_serializers/test_serializer_longturtle_sort.py
```diff
@@ -0,0 +1,117 @@
...
+    rdfs:comment "Thing B" ;
+.
+
+[]    ns1:has
+        [
+            rdfs:seeAlso ns1:A ;
+        ] ; ;
+.
+
+[]    rdfs:seeAlso ns1:B ; ;
```
ajnelson-nist, line 90: Two semicolons?

ajnelson-nist:
@edmondchuc Thank you for porting my test!

Though, I noticed there were some spots semicolons got doubled-up.

test/test_serializers/test_serializer_longturtle_sort.py
edmondchuc: That's strange. I've never seen that before. I just tried serialising the same data with rdflib v7.0.0 and 7.1.1, and it's producing the same result.

edmondchuc: The extra semicolons were introduced in this commit https://github.com/RDFLib/rdflib/pull/2500/commits/ba585bd820c2ac6610d8a93026065b602a7a1910. It appears that they were added to ensure that every statement in an object ended with a semicolon. But because the `s_squared ` method was already writing a semicolon, it would double up for subjects that are blank nodes.

https://github.com/RDFLib/rdflib/blob/e8f61d4b5019040f097e560c3c1c8ecb41dd3fac/rdflib/plugins/serializers/longturtle.py#L184-L190

Fixed in https://github.com/RDFLib/rdflib/pull/3008/commits/412fb5d63d79c5d5816e05751e6be6bd791b1385.

edmondchuc:
@ajnelson-nist thanks for catching the double-up of the semicolons! I think it's fixed now.

nicholascar (approved):

PR #3060, opened 2025-01-29
feat: proposed Dataset API changes

recalcitrantsupplant:
This draft PR is intended promote discussion/reach a consensus on the proposed changes to the Dataset API, by concretely describing what the interfaces would look like. As such, at this point in time, it is not intended that all of the required changes to dependencies/tests are implemented.

The following discussion gives context:
https://github.com/RDFLib/rdflib/discussions/2591
Additionally:
- There is a general writeup in /dataset_api.md
- Examples in examples/datasets.py

Summary of changes:
- Created a separate dataset.py module which contains only the Dataset class
- Removed the inheritance of Dataset from ConjunctiveGraph/Graph
- Stubbed methods / function signatures with expected input and output types
  - Added methods proposed/discussed here https://github.com/RDFLib/rdflib/discussions/2591#discussion-5619172
  - Added stubs for other expected methods based on existing and inherited methods.
  - Updated methods for triples, quads, subjects, subjects_objects etc. to accept "slicing" where iterables of terms can be passed in to filter on a set of terms.
  - Added graph= parameter and corresponding Enum for GraphType to allow filtering operations over the Default Graph or all Named Graphs
- Removed ConjunctiveGraph class
- _Started_ removing the identifier attribute on Graph. There will still be many references to this.


dataset_api.md
```diff
@@ -0,0 +1,155 @@
...
+# Add a single triple to the Default Graph, and a single triple to a Named Graph
+g1 = Graph()
+g1.add(
+    (
+        URIRef("http://example.com/subject-a"),
+        URIRef("http://example.com/predicate-a"),
+        Literal("Triple A")
+    )
+)
+d.add_graph(g1)
```
Tpt, line 50: This method name is confusing imho. I first thaught it was adding a new named graph into the dataset with an auto-generated blank node for graph name. I would find cleaner to have a `Dataset.defaut_graph` accessor and then do things like `dataset.default_graph += g1` to merge data into the default graph and maybe allow to write `dataset.default_graph = g1` to replace the default graph.

dataset_api.md
```diff
@@ -0,0 +1,155 @@
...
+# Add a Graph to a Named Graph in the Dataset.
+g2 = Graph()
+g2.add(
+    (
+        URIRef("http://example.com/subject-b"),
+        URIRef("http://example.com/predicate-b"),
+        Literal("Triple B")
+    )
+)
+d.add_named_graph(uri=URIRef("http://example.com/graph-B"), g2)
```
Tpt, line 61: I would maybe rename `uri` to `graph_name` or `name` because RDF 1.1 allows also blank nodes to be used.

Also is it replacing the named graph or merging into it? `add` might mean both.

dataset_api.md
```diff
@@ -0,0 +1,155 @@
+Incorporate the changes proposed from Martynas, with the exception of graphs(), which would now return a dictionary of graph names (URIRef or BNode) to Graph objects (as the graph's identifier would be removed).
+
+```
+add add_named_graph(uri: IdentifiedNode, graph: Graph) method
+add has_named_graph(uri: IdentifiedNode) method
+add remove_named_graph(uri: IdentifiedNode) method
+add replace_named_graph(uri: IdentifiedNode, graph: Graph)) method
+add graphs() method as an alias for contexts()
+add default_graph property as an alias for default_context
```
Tpt, line 9: Is it returning the default `Graph` or the name of the default graph? I would tend to prefer the former for easy lookup of the default graph. Also, do we want the default graph to have a graph name like currently or do we want to stop doing it?

Tpt (commented):
Thank you so much for pushing this!

I allowed myself a few API nits but I have not read the code carefully, please ignore them if they are not relevant

dataset_api.md
recalcitrantsupplant: add_graph was an existing method but I agree that this is confusing, especially if add_named_graph is introduced and identifier is removed from Graph.

I think just having `Dataset.defaut_graph` is cleaner too - there's a method there already, renamed from `default_context`.

**If others see a benefit in retaining add_graph please comment.**

dataset_api.md
recalcitrantsupplant: "name" instead of "uri" sounds sensible, I'll update this.

The method is a merge, the proposed methods were (https://github.com/RDFLib/rdflib/discussions/2591#discussion-5619172):
```
add_named_graph(uri: IdentifiedNode, graph: Graph) method
remove_named_graph(uri: IdentifiedNode) method
replace_named_graph(uri: IdentifiedNode, graph: Graph)) method
```
We could change the method name to something like"merge_named_graph" or "add_to_named_graph". I would prefer leaving the method as "add_named_graph", so a user would need to familiarise themselves with the different methods/read the docstrings.

dataset_api.md
recalcitrantsupplant: It would return the default graph not the name.

In the current discussion we concluded that the default graph should **not** have a name; for consistency with the spec. How achievable this is I don't know, we may be able to use `None` for the name where code/storage requires it, not sure.

dataset_api.md
recalcitrantsupplant: updated `uri` to `name` in in 2cc3590d5c27553c774a17841c636f781fba02a4

dataset_api.md
Tpt: Thanks! 100% agreed this is the good behavior and the default graph shouldn't have a name

ajnelson-nist:
Re:

> * Removed ConjunctiveGraph class

Does it make sense to apply removal of `ConjunctiveGraph` in response to #3064 , in advance of accepting this PR?  If not globally, then what about in just the JSON-LD serializer?

I'm asking without having reviewed all of the proposed typing here yet.

dataset_api.md
```diff
@@ -0,0 +1,158 @@
...
+# Get the Default graph
+# ============================================
+
+dg = d.default_graph  # same as current default_context
+
+# ============================================
+# Iterate on triples in the Default Graph only
+# ============================================
+
+for triple in d.triples(graph="default"):
```
namedgraph, line 96: I question the usefulness of this. Why not simply:
```python
d.default_graph.triples()
```
?

dataset_api.md
```diff
@@ -0,0 +1,158 @@
...
+    print(triple)
+# Output:
+(rdflib.term.URIRef('http://example.com/subject-a'), rdflib.term.URIRef('http://example.com/predicate-a'), rdflib.term.Literal('Triple A'))
+
+# ============================================
+# Access quads in Named Graphs only
+# ============================================
+
+for quad in d.quads(graph="named"):
```
namedgraph, line 105: Shouldn't this be equivalent to simply `d.quads()`? Since the default graph does not produce quads.

namedgraph:
My general comment is that I don't like the new arguments of `triples()` and `graphs()`. IMO the equivalent result should be achieved using `default_graph` and `get_named_graph()`. There shouldn't be multiple ways to achieve the same thing.

dataset_api.md
```diff
@@ -0,0 +1,158 @@
...
+    print(triple)
+# Output:
+(rdflib.term.URIRef('http://example.com/subject-a'), rdflib.term.URIRef('http://example.com/predicate-a'), rdflib.term.Literal('Triple A'))
+(rdflib.term.URIRef('http://example.com/subject-b'), rdflib.term.URIRef('http://example.com/predicate-b'), rdflib.term.Literal('Triple B'))
+
+# ============================================
+# Access quads in the Default Graph and specified Named Graphs.
+# ============================================
+
+for quad in d.quads(graph=["default", URIRef("http://example.com/graph-B")]):
```
namedgraph, line 134:
```python
for quad in (q for q in d.quads() if q[3] in (None, URIRef("http://example.com/graph-B"))): 
```
not much longer really.

dataset_api.md
```diff
@@ -0,0 +1,158 @@
...
+    for quad in d.quads(graph=ng_name):
+        print(quad)
+# Output:
+(rdflib.term.URIRef('http://example.com/subject-b'), rdflib.term.URIRef('http://example.com/predicate-b'), rdflib.term.Literal('Triple B'), rdflib.term.URIRef('http://example.com/graph-B'))
+
+# ============================================
+# Access triples in the Default Graph and specified Named Graphs.
+# ============================================
+
+for triple in d.triples(graph=["default", URIRef("http://example.com/graph-B")]):
```
namedgraph, line 124: `d.triples()` doesn't really make sense? There should be `Graph.triples()` and `Dataset.quads()` only?

namedgraph:
One more question arised while I commenting: what graph component for the default graph in `ds.graph()` result?
It used to be `DATASET_DEFAULT_GRAPH_ID = URIRef("urn:x-rdflib:default")`, but now that the `Graph.identifier` is gone, what is it?
It would make sense to use `None`, but then the signature of `Dataset.quad()` becomes ambiguous in the case of `quads(None, None, None, None)` -- is the last `None` a wildcard or does it signify the default graph?

recalcitrantsupplant:
> Re:
> 
> > * Removed ConjunctiveGraph class
> 
> Does it make sense to apply removal of `ConjunctiveGraph` in response to #3064 , in advance of accepting this PR? If not globally, then what about in just the JSON-LD serializer?
> 
> I'm asking without having reviewed all of the proposed typing here yet.

I wouldn't think so - as ConjunctiveGraph is used in a few different places and there's an inheritance hierarchy that removing it would break, so it's not a simple change. The JSON-LD serializer could potentially switch to Dataset while leaving ConjunctiveGraph in though, I haven't looked at it.

dataset_api.md
recalcitrantsupplant: Yes the proposal is to have the "graph" of triples in the default graph set to None.

dataset_api.md
recalcitrantsupplant: I'm comfortable with it - SPARQL queries in triplestores where named graphs are used frequently omit the graph, only having basic graph patterns, and we understand this to be across all graphs?

dataset_api.md
recalcitrantsupplant: Yes I think this is the point to get a broader consensus on. The way I see it, if including the graph parameter:
Pros:
- can restrict "named", "default" enums to only be used in the graph= attribute, and not in the quads methods.
- can separate concerns a bit better, similar to how dataset clauses are used in SPARQL. E.g. set up an instance with graph= to restrict the scope to certain named graphs, then at runtime graphs can be passed in using quads
- provides a convenience/clean interface for what is a common pattern (for me at least!)
Cons:
- two ways to do the same thing, as you've pointed out.

recalcitrantsupplant:
> One more question arised while I commenting: what is the graph component for the default graph in `ds.graph()` result? It used to be `DATASET_DEFAULT_GRAPH_ID = URIRef("urn:x-rdflib:default")`, but now that the `Graph.identifier` is gone, what is it? It would make sense to use `None`, but then the signature of `Dataset.quad()` becomes ambiguous in the case of `quads(None, None, None, None)` -- is the last `None` a wildcard or does it signify the default graph?

Good question. I would suggest None is an expected result/output from methods and can means that results are from the default graph, but that we are not in fact making them equivalent. When providing input to functions, None != default graph, and a user should not use it to mean this. The following methods can be used to unambiguously refer to the default graph:

ds.triples(graph="default")
ds.quads(graph="default")
ds.default_graph.triples()
etc.

dataset_api.md
namedgraph: Union graph is an extension feature though, not a feature of an RDF dataset.

dataset_api.md
namedgraph: Maybe I'm too used to Jena where `Dataset` is used via `getDefaultModel` and `getNamedModel`, but I don't really see myself needing the new parameters 🤷‍♂️

PR #3076, opened 2025-02-21
remove old hacks against 2to3

a-detiste:
I noticed this while reviewing the last traces of `2to3` in Debian Testing.

Greetings

This reverts e2fb491a3da80f9e01f3303b3df24881ab41eefa from 2011.


ashleysommer:
@a-detiste Note, the `main` branch is the v7.x maintenance branch.
Can you check if this same change needs to be done to v8 branch, where the latest work is?

PR #3086, opened 2025-03-23
Allow multi subjects & objects in graph funcs

nicholascar:
This PR allows `graph.subjects()` & `graph.objects()` to receive a list of objects & subjects respectively, rather than just single values


ashleysommer:
@nicholascar Is this a feature for v7.x or v8.0?

Is this too much feature-overlap with `graph.triples_choices()`?

nicholascar:
> @nicholascar Is this a feature for v7.x or v8.0?
> 
> Is this too much feature-overlap with `graph.triples_choices()`?

Good pick on that overlap. I do want to slim down all the various graph functions so we have fewer but more capable ones than now. This is because it's too hard to remember many functions. Easier to remember fewer and then see their permissive signatures for all the options. 

So I'd like to boil down `triples_choices()` into just `triples()` perhaps and this PR helps me think about how to get there by doing just the simple `subjects()` and `objects()` equivalents first

rdflib/graph.py
```diff
@@ -847,26 +847,32 @@ def set(
     def subjects(
         self,
         predicate: Union[None, Path, _PredicateType] = None,
-        object: Optional[_ObjectType] = None,
+        object: Optional[Union[_ObjectType, List[_ObjectType]]] = None,
```
ajnelson-nist, line 850: @nicholascar : Was there a reason this used `List[...]` instead of `Sequence[...]`?

ajnelson-nist: The use case I have in mind: I often load `set`s with `URIRef` instead of `list`s to de-duplicate.

rdflib/graph.py
```diff
@@ -847,26 +847,32 @@ def set(
     def subjects(
         self,
         predicate: Union[None, Path, _PredicateType] = None,
```
ajnelson-nist, line 849: If `object` gets a plurality expansion, why not `predicate` too (and likewise with other functions)?  Something to do with `Path`?

Issue #3078, opened 2025-02-25
Python conversion of xsd:gYear and xsd:gYearMonth typed Literals should not return datetime.date objects


lu-pl:
Currently, the `rdflib.Literal.toPython` conversion of `xsd:gYear` and `xsd:gYearMonth`-typed `rdflib.Literals` returns `datetime.date` objects with the day (`xsd:gYearMonth`) or day and month (`xsd:gYear`) arguments set to 1.

```python
from rdflib import Literal, XSD

year = Literal("2025", datatype=XSD.gYear)
year_month = Literal("2025-02", datatype=XSD.gYearMonth)

year_month.toPython()    # datetime.date(2025, 2, 1)
year.toPython()   # datetime.date(2025, 1, 1)
```

I think this is plain wrong, `xsd:gYear` is not equivalent to `datetime.date(2025, 1, 1)`.

Note that issue [#1379](https://github.com/RDFLib/rdflib/issues/1379) addressed this already, but focused more on a serialization problem and was then closed without really discussing  the datetime conversion of `xsd:gYear` and `xsd:gYearMonth`.

A practical solution for this would be to just return an `rdflib.Literal` with the `datatype` property set; this way, the XSD type information would still be available.

I would be happy to implement the necessary changes and open a PR if this the suggested/some solution is agreed upon.

WhiteGobo:
this should be related to #2419

WhiteGobo:
I dont expect any set behaviour from `toPython` so im not per-se against this change. These are just some arguments for consideration.
I would think it is reasonable to express `xsd:gYear` as `date` when you convert the data to python. When `datetime` isnt supporting any class, where this information, that its just a year with no set month, is expressed, then maybe we should not do something different, when handling data in context of `toPython`. And at least i wouldnt expect any guarantee not to lose such information during conversion. I dont think(im not sure) xsd-standard set any standards, when eg calculating the timedifference between a date expressed in `xsd:gYear` and `xsd:gYearMonth`.

lu-pl:
Thanks for input and sorry for my late reply. 

As mentioned in the initial post, imo there is no way to express `xsd:gYear` and `xsd:gYearMonth` in Python, maybe timedeltas could be an option, but this too is ontologically not correct. I think returning an `rdflib.Literal` with the `datatype` property set is best way to handle this problem, so if anyone could sanity-check this approach and give me a "go", I would open an PR with that change.

lu-pl:
[...]

PR #3115, opened 2025-04-16
fix: remove Literal.toPython date conversion for gYear/gYearMonth

lu-pl:
# Summary of changes

Issue #3078 reports, `that toPython`-casting of `xsd:gYear` and
`xsd:gYearMonth` to datetime objects is not possible, as there is no
appropriate Python equivalence for those types. The current
implementation casts `xsd:gYear` and `xsd:gYearMonth` to datetime objects
assuming Jannuary 1st for `xsd:gYear` and the 1st day of the given month
for `xsd:gYearMonth`. This is plain wrong.

The change re-implements `parse_xsd_gyear` and `parse_xsd_gyearmonth` so
that XSD-types are checked against a regex and `xsd:gYear` and `xsd:gYearMonth`
are not converted to Python datetime objects anymore but simply return
the value unchanged after checking.

Please note that this is a draft. I haven't fully figured out how to best adapt the tests and it might also be viable to simply remove `xsd:gYear` and `xsd:gYearMonth` from the `XSDToPython` mapping `rdflib.term`, if validation of those XSD types is not necessary.

# Checklist

  the same change.
  
- [ x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),


lu-pl:
I set this to be ready for review, I don't know why the pipeline fails though.

Issue #2528, opened 2023-08-15
Performance issues with rdflib.compare


william-vw:
We're having some performance issues with using rdflib.compare. While the test files contains a lot of blank nodes, they are relatively small (50, 75, 100 lines); even then one can see the performance drop with order(s) of magnitude. (Our actual files have close to 6000 lines).

Checkout the repo here with the test code: https://github.com/william-vw/rdflib_compare_test

On my machine, I get the following times:

```
file: test1.ttl
isomorphic: 0.06714916229248047
canonical: 0.03405618667602539

file: test2.ttl
isomorphic: 3.1934521198272705
canonical: 1.6439540386199951

file: test3.ttl
isomorphic: 17.75542116165161
canonical: 8.694658994674683
```

I'm basing myself on [this example](https://rdflib.readthedocs.io/en/stable/_modules/rdflib/compare.html) to perform the compare (for testing purposes I'm comparing a graph with itself).

aucampia:
Hi @william-vw, sadly the code for compare is quite complex, but any improvements will be welcomed and merged.

ddeschepper:
@william-vw, @aucampia I've done some research into this issue and made PR https://github.com/RDFLib/rdflib/pull/3119 that makes a big difference in the canonicalization (and isomorphization) performance. See the PR for results produced by @william-vw's test script.

ddeschepper:
If you want to use the performance improvements right now (at your own risk!), here's a monkey patch that applies them that works with v7.1.4:

```python
from rdflib.compare import Color, Stats, _TripleCanonicalizer

def patched_traces(self, coloring: list[Color], stats: Stats | None = None, depth: list[int] = [0]) -> list[Color]:
    if stats is not None and "prunings" not in stats:
        stats["prunings"] = 0
    depth[0] += 1
    candidates = self._get_candidates(coloring)
    best_coloring = None
    best_score = None

    for candidate, color in candidates:
        coloring_copy: list[Color] = []
        color_copy = None
        for c in coloring:
            c_copy = c.copy()
            coloring_copy.append(c_copy)
            if c == color:
                color_copy = c_copy
        new_color = self._individuate(color_copy, candidate)
        coloring_copy.append(new_color)
        refined_coloring = self._refine(coloring_copy, [new_color])
        color_score = tuple(c.key() for c in refined_coloring)

        if best_score is None or best_score < color_score:
            best_coloring = refined_coloring
            best_score = color_score

    if best_coloring is None:
[...]

PR #3119, opened 2025-04-25
Improve canonicalization performance

ddeschepper:
We're noticing big performance issues when using longturtle serialization on some graphs. I've been able to narrow this down to the performance of canonicalization, which is also tracked in issue https://github.com/RDFLib/rdflib/issues/2528.

Looking into it I found that the current implementation of the `_traces` method of `_TripleCanonicalizer` causes much of the performance impact.

This PR reduces the complexity of `_traces`, which leads to a performance gain of at least an order of magnitude in our worst cases (100s -> 4s). All rdflib tests still pass, and additionally, I've tested these changes with our set of a few hundred examples that are longturtle serialized, which causes no changes in the serialization output.

The author of the linked issue has created a performance test that, with the current code, gives the following results on my machine:
```
file: test1.ttl
isomorphic: 0.07787537574768066
canonical: 0.03909921646118164

file: test2.ttl
isomorphic: 3.528538942337036
canonical: 1.8337273597717285

file: test3.ttl
isomorphic: 20.140648365020752
canonical: 9.535402774810791
```
where my new version results in:
```
file: test1.ttl
isomorphic: 0.012566566467285156
canonical: 0.006159543991088867

file: test2.ttl
isomorphic: 0.15960264205932617
canonical: 0.09874987602233887

file: test3.ttl
isomorphic: 0.531768798828125
canonical: 0.2606654167175293
```


rdflib/compare.py
```diff
@@ -438,42 +429,19 @@ def _traces(
...
-            if last_coloring:
-                generator = self._create_generator(  # type: ignore[unreachable]
-                    [last_coloring, experimental], generator
-                )
-            last_coloring = experimental
-            if best_score is None or best_score < color_score:  # type: ignore[unreachable]
-                best = [refined_coloring]
+            color_score = tuple(c.key() for c in refined_coloring)
+
+            if best_score is None or best_score < color_score:
```
edmondchuc, line 434: The right-hand side of the `or` is never evaluated as `best_score` is always true here. Can you please review this.

nicholascar:
Agreed that there are large performance issues using `longturtle` instead of `turtle` and it would be great to fix them!

A few changes to `compare.py` have come through other PRs so there's a conflict to be resolved here. I'll try and resolve this now and then see how the PR looks.

edmondchuc:
Hi @ddeschepper, I will put the use of the canon algorithm behind a flag when calling the longturtle serializer and set it to false by default. This should address the performance issues for those who don't need deterministic outputs.

edmondchuc:
Hi @ddeschepper, please can you review this PR https://github.com/RDFLib/rdflib/pull/3197 and see if it addresses your concerns with performance? By default, canonicalization is no longer applied when using the `longturtle` serializer unless the `canon` flag is set to `True`.

Issue #3126, opened 2025-05-01
de_skolemize fails when handling literals


ddeschepper:
`Graph.de_skolemize()` incorrectly tries to deskolemize literals, which fails in some edge cases, like below:

```python
from rdflib import Graph

nt = '<http://example.com> <http://example.com> "http://example.com [some remark]" .'
graph = Graph().parse(data=nt, format="nt").de_skolemize()
```

Which raises a `ValueError` with message `"Invalid IPv6 URL"`.

PR #3127, opened 2025-05-01
Fix incorrect deskolemization of literals

ddeschepper:
Fixes issue https://github.com/RDFLib/rdflib/issues/3126.

Graph.de_skolemize() incorrectly tries to deskolemize literals, which fails in some edge cases. Limiting deskolemization of objects to `URIRef`s only fixes this behaviour.


edmondchuc:
pre-commit.ci autofix

PR #3201, opened 2025-09-04
Term improvements

multimeric:
# Summary of changes

Rebase of #3169 against `7-maintenance`.

**All these changes should have no impact at runtime**, they're purely to improve UX through static analysis.

* Convert `Term` to a typed `NamedTuple` from a `namedtuple`. This makes static analysis easier in that your IDE and type checker can understand which fields are present on the type, the type of each field, and give each field its own documentation
* Document the meaning of each field in `Term` for usability
* Move the type definition of `Context` fields into the class body. This is also for static analysis and allows the user to get tab completion etc when using these fields.

My long term goal is to make this context API usage as a public API, so that I can generate and interact with contexts programmatically instead of just by defining these in JSON.

# Checklist

  the same change.


edmondchuc:
pre-commit.ci run

edmondchuc:
pre-commit.ci autofix

edmondchuc:
Thanks for rebasing this @multimeric. I think the tests are currently failing due to some dependency bumps in the lockfile. Any chance we can roll those back?

multimeric:
> Thanks for rebasing this @multimeric. I think the tests are currently failing due to some dependency bumps in the lockfile. Any chance we can roll those back?

Done

edmondchuc:
Great, thanks for following up and making those changes.

The CI is now showig a few mypy/type errors. Would you mind having a look and seeing if you can address those as well?

edmondchuc:
Just an FYI, I'm targeting the `7.x` branch now as it's the branch intended for all v7 related features and fixes.

multimeric:
Thanks. The type errors are a bit annoying because it's making me fix everything in the JSON-LD parser, including things that I didn't change. I'll try to do so but it might not be soon.

nicholascar:
@multimeric could you please drop me a "hi" on my email address which you can find on my GitHub profile? Thanks

multimeric:
How come?

nicholascar:
> How come?

I'd like to chat to you about your use of RDFLib and what we might be able to do to assist with some of the ROCrates and similar interests you have.

Also, I'd like to introduce you to the [Australian Government Linked Data Working Group](https://www.linked.data.gov.au/)... which I guess I've just done, in a way!

multimeric:
I'm getting back to this PR, which is still quite important to me. The trouble is, it's nigh impossible to resolve the type checker errors without either heavily using `# ignore`, or otherwise using type assertions with `raise Exception()`, but this won't be backwards compatible.

An example of a weird edge case is, "what if we have an RDF `@list` that contains a subgraph"? The data type is `Node` so we can't assume it's strictly an `Identifier`.

PR #3241, opened 2025-09-27
Remove trivial code from turtle serializer

eyusupov:
# Summary of changes

firstTime is alway true when checked in the last if statement, so it can be removed.

# Checklist


edmondchuc:
Thanks. We should make the same change to the `longturtle` serializer too.

eyusupov:
> Thanks. We should make the same change to the `longturtle` serializer too.

Updated the PR with longturtle changes.

edmondchuc (approved):

Issue #3246, opened 2025-10-05
OPTIONAL causes WHERE to be re-evaluated between updates


maxcutlyp:
```
>>> print(rdflib.__version__)
7.2.1
```

Hi, I noticed that it seems like having `OPTIONAL` in the `WHERE` clause produces some unexpected behaviour. It seems like it's re-evaluating the WHERE clause after updating the graph for each solution to get a new value for `?oldValue`. See the below example:

```python
from rdflib import Graph, Namespace, Literal

EX = Namespace('http://example.com/')

g = Graph()
g.bind('ex', EX)

g.add((EX.foo, EX.value, Literal(1)))
g.add((EX.foo, EX.value, Literal(11)))

g.add((EX.bar, EX.value, Literal(3)))

g.update('''
DELETE {
    ex:bar ex:value ?oldValue .
}
INSERT {
    ex:bar ex:value ?newValue .
}
WHERE {
    ex:foo ex:value ?instValue .
    OPTIONAL { ex:bar ex:value ?oldValue . }
    BIND(COALESCE(?oldValue, 0) + ?instValue AS ?newValue)
}
''')

print(g.serialize(format='turtle'))
```

Running this prints:

```
@prefix ex: <http://example.com/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ex:bar ex:value 15 .

ex:foo ex:value 1,
        11 .
```

However, changing the line

```
    OPTIONAL { ex:bar ex:value ?oldValue . }
```

to

```
    ex:bar ex:value ?oldValue .
```

causes it to print:

```
@prefix ex: <http://example.com/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ex:bar ex:value 4,
        14 .

ex:foo ex:value 1,
        11 .
```

Additionally, if the `WHERE` clause is changed to:

```
WHERE {
    ex:foo ex:value ?instValue .
    {
        OPTIONAL { ex:bar ex:value ?oldValue . }
    } UNION {
        ex:bar ex:value ?oldValue .
    }
    BIND(COALESCE(?oldValue, 0) + ?instValue AS ?newValue)
}
```

then it prints the same as the version without the `OPTIONAL` clause.

The expected behaviour would be that all versions produce the result that the version without `OPTIONAL` currently produces. (That is how Apache Jena behaves, FWIW).

Let me know if you need any more info from me, or if this is working as intended.

Thanks!

[...]

PR #3261, opened 2025-10-16
patch for reevaluation in sparql modify between update loops. with test

WhiteGobo:
# Summary of changes

patch and tests for issue #3246
sparql modify loop now creates a list instead of an iterator from its initial query.

# feedback wanted

I want a little feedback on if this is performant enough. It could be possible to still use an iterator if i could find why the optional keyword breaks the query.

# also
Also its likely there is an mypy warning of unused `#type ignore` in the tests for version 8.x

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -1969,7 +1969,7 @@ class ConjunctiveGraph(Graph):
     All queries are carried out against the union of all graphs.
     """
 
-    default_context: _ContextType
+    _default_context: _ContextType
 
     def __init__(
         self,
@@ -1991,9 +1991,17 @@ class ConjunctiveGraph(Graph):
         )
         self.context_aware = True
         self.default_union = True  # Conjunctive!
-        self.default_context: _ContextType = Graph(
+        self._default_context: _ContextType = Graph(
             store=self.store, identifier=identifier or BNode(), base=default_graph_base
         )
+
+    @property
+    def default_context(self):
+        return self._default_context
+
+    @default_context.setter
+    def default_context(self, value):
+        self._default_context = value
 
     def __str__(self) -> str:
         pattern = (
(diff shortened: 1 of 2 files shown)


WhiteGobo:
patch works also on current main branch

rdflib/plugins/sparql/update.py
```diff
@@ -181,7 +181,7 @@ def evalModify(ctx: QueryContext, u: CompValue) -> None:
             g = ctx.dataset.get_context(u.withClause)
             ctx = ctx.pushGraph(g)
 
-    for c in res:
```
edmondchuc: Do you mind explaining why changing this to a list fixes the issue?

WhiteGobo: The Iterator `res` returned by evalPart (which should be a normal query) is still working on all matches, while the graph is updated in the (`for c in res`)-loop. In the test example the order of operations should look like this (predicate is ommited):
-> res finds {(ex:foo 1) (ex:bar 3)}
-> updates with (ex:bar 4)
-> res finds {(ex:foo 11) (ex:bar 4)}
-> updates with {ex:bar 15}

when replacing it with list the order of operations will look like:
-> res will result in {(ex:foo 1) (ex:bar 3)} and {(ex:foo 11) (ex:bar 3)}
-> updates with (ex:bar 4) 
-> updates with {ex:bar 14}

Why this is not the case, when optional is omitted, im not sure. This would need further testing.

edmondchuc: Thanks for the explanation! I understand the problem now. Basically, the SPARQL Update query is mutating the graph as it loops through, and the bounded var reflects this on each eval. Using `list()` here materialises the values so they don't mutate. Unless we can think of a better way to handle this, I am okay with this performance hit. For me, it's better that we have semantically correct SPARQL than not.

PR #3263, opened 2025-10-17
fix: Dataset.parse now returns Self

edmondchuc:
# Summary of changes

Both `ConjunctiveGraph` and `Dataset`'s `parse` method now correctly returns `Self` like their superclass `Graph`.

The test confirms that now the return value has the same reference as the original dataset that calls `parse`.

# Checklist

  the same change.
  
### DIFF
--- a/test/test_dataset/test_dataset_add.py
+++ b/test/test_dataset/test_dataset_add.py
@@ -99,16 +99,3 @@ def test_dataset_parse_return_value():
     )
     assert len(ds)
     assert return_value is ds
-
-
-def test_dataset_graphs_excludes_default_graph():
-    """
-    The Dataset.graphs() method should exclude the default graph.
-    """
-    ds = Dataset()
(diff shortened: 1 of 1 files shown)


rdflib/plugins/parsers/nquads.py
```diff
@@ -7,7 +7,7 @@
 >>> g = ConjunctiveGraph()
 >>> data = open("test/data/nquads.rdflib/example.nquads", "rb")
 >>> g.parse(data, format="nquads") # doctest:+ELLIPSIS
-<Graph identifier=... (<class 'rdflib.graph.Graph'>)>
+<Graph identifier=... (<class 'rdflib.graph.ConjunctiveGraph'>)>
```
edmondchuc: The doctest is updated to the correct class type now that ConjunctiveGraph's parse method returns Self.

PR #3276, opened 2025-10-23
Convert Define into an enum

multimeric:
# Summary of changes

Convert `Defined` into an enum. The justification of this [is PEP 484](https://peps.python.org/pep-0484/#support-for-singleton-types-in-unions), which suggests that singletons should be represented as enums with a single element. The reason why this matters is because it lets you say `if x is not UNDEF` and the type checker will correctly narrow the type to be no longer `Defined`, because all enum variants have been excluded.

# Checklist

  the same change.
  

edmondchuc:
pre-commit.ci autofix

multimeric:
Closed so I can recreate this against `main`

Issue #3102, opened 2025-03-29
Runtime error querying a dataset using multiple graph variables


jaw111:
I run into `RuntimeError: Set changed size during iteration` when iterating over results from a query and was able to reduce the problem down to the following example code that reproduces the error:

```python
import rdflib

# create a dataset
dataset = rdflib.Dataset()

# create a graph
g = dataset.graph(identifier="urn:example:graph")

sparql_query = """
SELECT *
WHERE {
  GRAPH ?g_1 { }
  GRAPH ?g_2 { }
}"""

results = dataset.query(sparql_query)

print(len(results))
```

The traceback is:

```
Traceback (most recent call last):
  File "C:\Users\XXXX\query_test.py", line 19, in <module>
    print(len(results))
          ^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\rdflib\query.py", line 347, in __len__
    return len(self.bindings)
               ^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\rdflib\query.py", line 242, in bindings
    self._bindings += list(self._genbindings)
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\rdflib\plugins\sparql\evaluate.py", line 562, in <genexpr>
    return (row.project(project.PV) for row in res)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\rdflib\plugins\sparql\evaluate.py", line 142, in evalLazyJoin
    for a in evalPart(ctx, join.p1):
  File "C:\Python312\Lib\site-packages\rdflib\plugins\sparql\evaluate.py", line 225, in evalGraph
    for graph in ctx.dataset.contexts():
  File "C:\Python312\Lib\site-packages\rdflib\graph.py", line 2529, in contexts
    for c in super(Dataset, self).contexts(triple):
  File "C:\Python312\Lib\site-packages\rdflib\graph.py", line 2154, in contexts
    for context in self.store.contexts(triple):
  File "C:\Python312\Lib\site-packages\rdflib\plugins\stores\memory.py", line 565, in <genexpr>
    return (context for context in self.__all_contexts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Set changed size during iteration
```

jaw111:
[...]

PR #3281, opened 2025-10-27
fix: set changed size when iterating the store's graphs

edmondchuc:
# Summary of changes

Fix: https://github.com/RDFLib/rdflib/issues/3102

The `RuntimeError: Set changed size during iteration` occurs because a Dataset's default graph is lazily evaluated and only added to the store's context dict when accessed. 

~Since the default graph is always present in a dataset, this new change adds the default graph to the context dict on dataset creation.~ - this change breaks a bunch of tests unrelated to Dataset, like namespace manager etc.

We materialise the iterator using a `list()` call. This way, when the default graph is added to the contexts dict, we don't get the runtime error.

We also avoid calling the super class (ConjunctiveGraph)'s methods as it's redundant and we'll be removing the class hierarchy soon. This now calls `self.store.contexts` directly.

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/graph.py
+++ b/rdflib/graph.py
@@ -2831,7 +2831,7 @@ class Dataset(ConjunctiveGraph):
             stacklevel=2,
         )
         default = False
-        for c in self.store.contexts(triple):
+        for c in super(Dataset, self).contexts(triple):
             default |= c.identifier == DATASET_DEFAULT_GRAPH_ID
             yield c
         if not default:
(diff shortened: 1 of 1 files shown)


edmondchuc:
I'm a bit puzzled as to why this only fails on the 3.8 tests in GitHub Action while it passes fine on python 3.8 locally.

edmondchuc:
To reproduce the failing test locally, run:

```sh
./with-fuseki.sh poetry run pytest -v test/test_dataset/test_dataset.py
```

edmondchuc:
The current Python 3.8 tests are extensive and include spinning up a Fuseki instance during the test run. These tests conditionally register the SPARQLUpdateStore plugin only if the Fuseki endpoint is reachable.

The issue arises from a change where the Dataset.contexts() method was modified to call the store’s contexts() method directly in https://github.com/RDFLib/rdflib/pull/3281/commits/63caed55d250832c58d0b95a5ce933ea8999ff51. However, not all stores guarantee that their contexts() implementation returns graph objects. For example, SPARQLUpdateStore returns graph names as URIRefs instead. To ensure contexts() always returns graph objects, we delegate to ConjunctiveGraph.contexts(), which converts graph name URIRefs into graph objects.

nicholascar (approved):

PR #3287, opened 2025-10-29
feat: v7 mkdocs

edmondchuc:
# Summary of changes

This PR backports PR https://github.com/RDFLib/rdflib/pull/3143 from `main` to v7.

# Checklist

  the same change.
  
### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1712,4 +1712,4 @@ orjson = ["orjson"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.8.1"
-content-hash = "bcdcfe65969feaa5c97b414fa88625282f18286faac13bc78eddb01442784c26"
+content-hash = "35ce70402138519fd62cf5ec9901125cab2c36fc6965ef13c66020637c8c8bd4"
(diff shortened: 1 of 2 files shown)


edmondchuc:
One thing to note: since we need to support python 3.9, we cannot install the latest major versions of the various mkdocs plugins. This results in some pages not generating all of the content from the python modules/scripts.

For example, see the examples page in this PR:

image

Compared to the examples page in `main` branch:

image


I think it has something to do with the `docs/gen_ref_pages.py` script or the python version or the mkdocs plugins. If I set the minimum python version to 3.9 and use the same plugin versions as `main`, it works. I will have to investigate this some more.

edmondchuc:
I've noticed readthedocs uses python 3.13 to perform the doc builds. For some reason, I thought we were running the doc builds on the lowest supported python version. I've now fixed the above issue by updating the mkdocs dependencies to the latest with a python version constraint of `>=3.11`, which is the python version we use to test the doc builds in the "validate" github actions.

edmondchuc:
To clarify the comment above, we now only run doc build test on python 3.11 in the validate workflow matrix. There's no point running the docs build on different python versions when readthedocs only uses the latest stable python version anyway. I've arbitrarily chosen 3.11 but readthedocs currently uses 3.13 for all builds.

Issue #3307, opened 2025-11-07
Better documentation for plugins, like parser and serializer


WhiteGobo:
The documentation for plugins, for example documentation about, which parameters are usable for `graph.parse` are a little difficult to find and understand. I would like to put a little effort that the documentation automatically gathers data about:
- a list of available plugins for at least parser and serializer
- an easy way to find out, which parameters are available, when using eg `graph.parse`

I havent done any automatic scripting, for documentation, so im not sure how possible this is.

Also i want to change the representation of `rdflib.plugin.Plugin` so that `rdflib.plugin.plugins()` returns a more informative list, like:
```
>>>from rdflib import plugin
>>>list(plugin.plugins())
[Plugin('turtle', ''other usable information'), ...]
```

WhiteGobo:
I guess the patch serializer implements the Jelly-Patch format?
As I try to link corresponding standards to each plugin, is there any good link for that format?

nicholascar:
Yes, work is needed to improve the documentation around parsers/serializers. We get a new one or other every year or so: LongTurtle, HexTuples, HTP, now Jelly.

There is work on enabling Jelly in pySHACL, a downstream package from RDFLib here: https://github.com/RDFLib/pySHACL/pull/303.

Ostrzyciel:
> I guess the patch serializer implements the Jelly-Patch format?

I don't think so, at least we haven't ported Jelly-Patch to Python yet.

Jelly-Patch itself is just a binary version of RDF Patch ([documented here](https://afs.github.io/rdf-delta/rdf-patch.html)).

So yeah, I think this proves the point that better docs are needed 😆

PR #3323, opened 2025-11-21
reworked documentation of JSONLDSerializer.

WhiteGobo:
Still work in correspondence to #3307. 
Added missing documentation arguments. 
Working towards that all arguments are accesible with `inspect`.

# Summary of changes

Added more documentation to module and `JsonLDSerializer.serialize`.
Set default of `use_native_types` to `False`.
Moved Arguments to serialize arguments, instead of extracting them from `**args`.
No functional change, but because i dont extract anything from `**args` anymore, there are some changes in the program code. 
Still missing explanation to argument `context`.
Added docstring to `PLAIN_LITERAL_TYPES`.
Added crosslink to `graph.serialize`

# Checklist

  the same change.
  
### DIFF
--- a/rdflib/plugins/serializers/jsonld.py
+++ b/rdflib/plugins/serializers/jsonld.py
@@ -1,5 +1,10 @@
-"""
-This serialiser will output an RDF Graph as a JSON-LD formatted document. See http://json-ld.org/
+"""This serialiser will output an RDF Graph as a JSON-LD formatted document.
+
+Implements serialization for [JSON-LD version 1.1](https://www.w3.org/TR/json-ld11/).
+For more information, see [json-ld.org](http://json-ld.org/).
+Additional `args` support by [`Graph.serialize`][rdflib.graph.Graph.serialize]
+are described by
+[`JsonLDSerializer.serialize`][rdflib.plugins.serializers.jsonld.JsonLDSerializer.serialize].
 
 Example:
     ```python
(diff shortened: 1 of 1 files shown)


rdflib/plugins/serializers/jsonld.py
```diff
@@ -64,29 +70,55 @@ def serialize(
...
+        if encoding is None:
+            encoding = "utf-8"
+        elif encoding not in ("utf-8", "utf-16"):
             warnings.warn(
-                "JSON should be encoded as unicode. " f"Given encoding was: {encoding}"
+                f"JSON should be encoded as unicode. Given encoding was: {encoding}"
             )
 
-        context_data = kwargs.get("context")
-        use_native_types = (kwargs.get("use_native_types", False),)
```
edmondchuc, line 78: I just want to point out that `use_native_types` was previously a tuple`(False,)` whereas the new code changes it to a `bool`. 

This alters the runtime behaviour, so it may not be strictly backwards compatible. I'm not sure whether the original tuple was intentional or an oversight.

@WhiteGobo what are your thoughts on this?

WhiteGobo: oh sorry, i dont know how i missed your answer.
I'm sure the change in `use_native_types` from `(bool,)` to `bool` is an oversight by me. I didn't want to change any possible behaviour or data in documentation changes, unless i explicitly stated it somewhere.

WhiteGobo: `use_native_types` is expected as `bool` in `from_rdf`. It seems like a bug, that it defaulted to `(bool,)` instead.
The only time it seems to be used is in line 418 on main, and there it result in a wrong result:
```python
native = self.use_native_types and o.datatype in PLAIN_LITERAL_TYPES
# native can be True, when use_native_types == (False,)
```
The behaviour wil change if we set the default value to `False`. Im certain with default value on `True` it will behave the same, but i think the behaviour wasnt intended.

I would suggest keeping the change to `bool` and `False`, but moving the merge to `8.0` instead.
Also i wonder if a test should check for this, as no test fails on changing `use_native_types`

PR #3324, opened 2025-11-21
sparqlconnector sends correct accept header in case of construct query

### LINKED ISSUE #3332: [SPARQLConnector] CONSTRUCT queries should use RDF media types in Accept header
Hi,
I would like to change the HTTP Accept header that the [SPARQLConnector](https://github.com/RDFLib/rdflib/blob/main/rdflib/plugins/stores/sparqlconnector.py) sends for SPARQL CONSTRUCT queries.

**current behavior:** SPARQLConnector sets the Accept header to the user-defined 'returnFormat' or default media types. The resulting Accept header uses media types for SPARQL results, e.g. `application/sparql-results+json` for returnFormat=json

**desired behavior:** SPARQLConnector selects only RDF graph media types, e.g. `application/ld+json` for CONSTRUCT or DESCRIBE queries which return a graph

I created a [small PR](https://github.com/RDFLib/rdflib/pull/3324) for checking the query type and selecting the media types accordingly.

#### [SPARQL 1.1 protocol](https://www.w3.org/TR/sparql11-protocol/)
```
The response body of a successful query operation with a 2XX response is either:
    a SPARQL Results Document in [XML](https://www.w3.org/TR/rdf-sparql-XMLres/), [JSON](https://www.w3.org/TR/sparql11-results-json/), or [CSV/TSV](https://www.w3.org/TR/sparql11-results-csv-tsv/) format (for SPARQL Query forms [SELECT](https://www.w3.org/2001/sw/DataAccess/rq23/#select) and [ASK](https://www.w3.org/2001/sw/DataAccess/rq23/#ask)); or,
    an RDF graph [[RDF-CONCEPTS](https://www.w3.org/TR/sparql11-protocol/#rdf-concepts)] serialized, for example, in the [RDF/XML syntax](https://www.w3.org/TR/rdf-syntax-grammar/) [[RDF-XML](https://www.w3.org/TR/sparql11-protocol/#rdf-xml)], or an equivalent RDF graph serialization, for SPARQL Query forms [DESCRIBE](https://www.w3.org/2001/sw/DataAccess/rq23/#describe) and [CONSTRUCT](https://www.w3.org/2001/sw/DataAccess/rq23/#construct)).
```
In section 3.1.3, there is an example of an HTTP request with a construct query:
* the request uses the HTTP header `Accept: text/turtle, application/rdf+xml`
* the response has the HTTP header `Content-Type: text/turtle`

Thank you for your help!

smeckler:
resolves #3332

# Summary of changes

**For SPARQL construct queries, send mime types for RDF instead of for SPARQL results in the header of the HTTP request.**
Some SPARQL servers (in my case Anzo) respond with a 400 'Bad Request' response if the Accept header does not describe RDF for construct or describe queries.

sparqlconnector.py
* get the type of SPARQL query from the passed query string
* if it's a construct or describe query, use an HTTP Accept header that represents an RDF serialization

plugin.py
* register the TurtleParser for the mime type "application/x-turtle" (for compatibility with legacy systems)

# Checklist

  the same change.

**update:**
`7600 passed, 592 skipped, 331 xfailed, 36 xpassed, 36012 warnings in 139.91s (0:02:19)`

### DIFF
--- a/rdflib/plugins/stores/sparqlconnector.py
+++ b/rdflib/plugins/stores/sparqlconnector.py
@@ -224,10 +224,13 @@ class SPARQLConnector:
         else:
             return ", ".join(rdf_mimetype_map)
 
-    def __get_query_type__(self, query: str) -> str:
-        q = prepareQuery(query)
-        algebra = getattr(q, "algebra", None)
-        name = getattr(algebra, "name", None) # e.g. 'SelectQuery', 'ConstructQuery', 'DescribeQuery', 'AskQuery'
-        return name
+    def __get_query_type__(self, query: str) -> str | None:
+        try:
+            q = prepareQuery(query)
+            algebra = getattr(q, "algebra", None)
+            name = getattr(algebra, "name", None) 
+            return name # e.g. 'SelectQuery', 'ConstructQuery', 'DescribeQuery', 'AskQuery'
+        except Exception:
+            log.debug(f"cannot parse query: {query}")
 
 __all__ = ["SPARQLConnector", "SPARQLConnectorException"]
--- a/rdflib/plugins/stores/sparqlconnector.py
+++ b/rdflib/plugins/stores/sparqlconnector.py
@@ -10,6 +10,7 @@ from urllib.parse import urlencode
 from urllib.request import Request, urlopen
 
 from rdflib.plugin import plugins
+from rdflib.plugins.sparql import prepareQuery
 from rdflib.query import Result, ResultParser
 from rdflib.term import BNode
 from rdflib.util import FORMAT_MIMETYPE_MAP, RESPONSE_TABLE_FORMAT_MIMETYPE_MAP
@@ -91,6 +92,11 @@ class SPARQLConnector:
             params["default-graph-uri"] = default_graph
 
         headers = {"Accept": self.response_mime_types()}
+        
+        # change Accept header to an RDF mime type in case of a construct query
+        qtype = self.__get_query_type__(query)
+        if qtype in ("ConstructQuery", "DescribeQuery"):
+            headers.update({"Accept": self.response_mime_types_rdf()})
 
         args = copy.deepcopy(self.kwargs)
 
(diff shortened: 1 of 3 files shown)


rdflib/plugins/stores/sparqlconnector.py
```diff
@@ -205,5 +211,26 @@ def response_mime_types(self) -> str:
...
 
+    def response_mime_types_rdf(self) -> str:
+        """Construct a HTTP-Header Accept field to reflect the supported mime types for SPARQL construct/describe queries that return a graph.
+
+        If the return_format parameter is set, the mime types are restricted to these accordingly.
+        """
+        rdf_mimetype_map = [mime for mlist in FORMAT_MIMETYPE_MAP.values() for mime in mlist]
+
+        # use the matched returnType if it matches one of the rdf mime types
+        if self.returnFormat in FORMAT_MIMETYPE_MAP:
```
WhiteGobo: As far as i remember we have different parsers available for Construct-queries and Select-queries. So i would rather see use of a variable `self.returnFormatConstruct` than reuse `self.returnFormat`. Of course, you would have to add it to `__init__` too.

WhiteGobo (commented):
I suggest also adding a test like `test/test_store/test_store_sparqlstore_query.test_query_construct_format`, where you check, that `turtle` is in `request.headers['Accept']`.

WhiteGobo:
Because of the failing testrun, you can take a look into [developers guide](https://rdflib.readthedocs.io/en/latest/developers/)
you can run black and mypy for the exact errors and the autoformatting
```bash
poetry run mypy
poetry run black .
```

Apart from that, this looks fine, to me at least. Sorry for my misunderstanding in your issue description.

rdflib/plugins/stores/sparqlconnector.py
edmondchuc: @smeckler Are there plans to address this?

edmondchuc:
@smeckler Is this PR backwards compatible with existing v7 releases? If so, we should rebase onto `7.x` and target the `7.x` branch.

PR #3344, opened 2025-12-06
Implement __hash__ and __eq__ methods in IdentifiedNode

### LINKED ISSUE #1332: A `URIRef` subclass breaks SPARQL querying

### LINKED ISSUE #3338: Memory store stores the resource object, not the semantic.
Hello,

I'm not sure how to describe the issue, so here is a sample code

```python
from rdflib import FOAF, Graph, Literal, URIRef


class MyId(URIRef): ...


g = Graph()

g.add((MyId("subject"), FOAF.name, Literal("bob")))

print(g.serialize())

assert (MyId("subject"), FOAF.name, Literal("bob")) in g  # ok
assert (URIRef("subject"), FOAF.name, Literal("bob")) in g  # fails
```

And the output

```turtle
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

<subject> foaf:name "bob" .
```

I was expecting the second assert to be True because in both case I'm looking for the subject `<subject>`.

PS: on the why I'm subclassing `URIRef`, it's to leverage static type checking with `mypy` when my functions are expecting ids that carry a special meaning.

Is it the expected behavior and should I take care of always passing `URIRef` object ?

This fix the issue but is a bit cumbersome.

```python
g.add((URIRef(MyId("subject")), FOAF.name, Literal("bob")))

print(g.serialize())

assert (URIRef(MyId("subject")), FOAF.name, Literal("bob")) in g  # ok
assert (URIRef("subject"), FOAF.name, Literal("bob")) in g  # ok
```

ticapix:
# Summary of changes

Attempt to fix #1332 and #3338

# Checklist

### DIFF
--- a/rdflib/term.py
+++ b/rdflib/term.py
@@ -253,6 +253,15 @@ class IdentifiedNode(Identifier):
 
     __slots__ = ()
 
+    def __hash__(self) -> int:
+        return super().__hash__()
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, IdentifiedNode):
+            # Compare the N3 notation and not the string representation.
+            return self.n3() == other.n3()
+        return super().__eq__(other)
+
     def __getnewargs__(self) -> Tuple[str]:
         return (str(self),)
 


edmondchuc:
@ticapix would you be able to add some tests to demonstrate the changes address the two issues mentioned?

PR #3367, opened 2026-01-16
Improve Graph.cbd

mgberg:
# Summary of changes

I noticed that the existing implementation of the CBD was not correct.  As stated [here](https://www.w3.org/Submission/CBD/) (and in the docstring of the method itself), the behavior with respect to reification should be as follows:

> 3. Recursively, for all statements included in the subgraph thus far, for all reifications of each statement in the source graph, include the concise bounded description beginning from the rdf:Statement node of each reification.

Prior behavior was to include reifications of any statement in the source graph for any subject of a triple in the subgraph. This meant:
- Reifications of ALL statements for a subject were included, even if the statement they reify was not in the subgraph. In fact, one of the tests had an example of this occuring.
- The CBD was not applied for reifications as stated in the linked source, only the triples with a subject of the rdf:Statement node of each reification were added to the subgraph.

This modifies the logic to conform to the description of the algorithm. It now includes reifications of statements only in the subgraph thus far and also performs a CBD on the rdf:Statement node of each reification.

Additionally, a new keyword argument is added which, if changed from the default, would disable the reification portion of the algorithm. This would result in performing blank node closure on the provided resource only. This is useful when only a blank node closure is desired and can also improve performance when reified statements are known to not be present. The default for this new argument is to include the reifications, so it should be backwards compatible in terms of whether reifications appear in the output or not.

# Checklist

  the same change.
  
### DIFF
--- a/test/test_graph/test_graph_cbd.py
+++ b/test/test_graph/test_graph_cbd.py
@@ -130,7 +130,7 @@ def testCbdReified(get_graph):  # noqa: N802
         format="turtle",
     )
 
-    assert len(g.cbd(EX.R6)) == (4 + 5 + 5 + 0), "cbd() for R6 should return 12 triples"
+    assert len(g.cbd(EX.R6)) == (4 + 5 + 5 + 0), "cbd() for R6 should return 14 triples"
     assert len(g.cbd(EX.R6, include_reifications=False)) == (
         4 + 0 + 0 + 0
     ), "cbd() for R6 with no reifications should return 4 triples"


nicholascar:
Hi @mgberg this is a great improvement! 

Running tests again to see if the pass now.

edmondchuc:
Nice one @mgberg! Since this fix is backwards compatible, could we rebase the changes onto the `7.x` branch (instead of `main`)? We're planning to make another `7.x` release in a couple of weeks and it'd be great to include this fix.

mgberg:
Sure! I'll do that.

PR #3444, opened 2026-06-11
Fix scheme-only base IRI parsing

### LINKED ISSUE #1216: Parsing N3 with a @base IRI which does not include a slash

vibhor-aggr:
## Summary

Fixes #1216.

Allows scheme-only base IRIs such as `local:` to resolve relative references like `<Category>` to `local:Category` instead of raising `ValueError`. The existing guard for opaque non-empty bases such as `mid:foo@example` is preserved.

## Changes

- Update the Notation3 URI join logic for scheme-only bases.
- Add an N3 regression test for `@base <local:>`.
- Update the existing `schema_only_base` variants and remove the obsolete xfails.

## Verification

```bash
uv run --with pytest python -m pytest -q test/test_n3.py rdflib/plugins/parsers/notation3.py --doctest-modules
uv run --with pytest python -m pytest -q test/test_graph/test_variants.py
uv run --with pytest python -m pytest -q test/test_parsers/test_parser_turtlelike.py
uv run --with ruff ruff check rdflib/plugins/parsers/notation3.py test/test_n3.py test/test_graph/test_variants.py
```


gjhiggins:
LGTM. On behalf of the community of RDFLib users, thank you for your contribution.

PR #3446, opened 2026-06-12
Update mypy and fix some type errors in main

multimeric:
# Summary of changes

`main` has many type errors that are causing unrelated PRs to look bad. The main cause of these type errors are that `mypy` is outdated and so doesn't infer correctly. I've updated mypy and fixed some remaining type errors. The only unfixed error here is fixed by #3276.

# Checklist

  the same change.
  

gjhiggins:
FTR, dependabot raised an as-yet-unaccepted PR [build(deps-dev): bump mypy from 1.18.2 to 1.19.1](https://github.com/RDFLib/rdflib/pull/3356) on Dec 22nd last year.

multimeric:
Happy for that to be merged first. Whichever way works. My PR adds some extra type fixes on top of that change though.

poetry.lock
```diff
@@ -1064,6 +1167,78 @@ install-types = ["pip"]
 mypyc = ["setuptools (>=50)"]
 reports = ["lxml"]
 
+[[package]]
+name = "mypy"
+version = "1.20.2"
+description = "Optional static typing for Python"
+optional = false
+python-versions = ">=3.10"
```
gjhiggins, line 1175: Is this consonant with python-versions specified elsewhere (e.g. Classifiers?) I'm not really familiar with the involved version semantics of pyproject.toml but this looks like it entails a python versions requirement bump.

multimeric: Hmm this is tricky. The lock format is not documented very well by poetry (I typically use `uv` these days). I would be surprised if updating the lockfile in this way broke anything considering it's not the definitive version declaration. 

Having said that, it may be the case that mypy 1.20.x which I tested against is Python 3.10 only. However it's worth considering that 3.9 [is end-of-life anyway](https://devguide.python.org/versions/). So if you would like me to separately change the minimum Python for this repo to 3.10 in another PR so that this isn't an issue then I can do that.

gjhiggins: If the as-yet-unaccepted dependabot mypy version bump [build(deps-dev): bump mypy from 1.18.2 to 1.19.1](https://github.com/RDFLib/rdflib/pull/3356) is accepted it would obviate the mypy version bump commit part of this PR. The poetry.lock component of the dependabot bump doesn't include any change to python versions so maybe it's an unrelated (or even irrelevant) issue - I only mentioned it because I'm half-expecting edmund to clear the dependabot backlog first just to get a contemporary level playing field, version-bump-wise.

*Edit* - ah, no your conjecture about mypy 1.20.x is correct, it *does* require Python>=3.10 according to its [PyPi page](https://pypi.org/project/mypy/).

PR #3447, opened 2026-06-14
Add initial RDF 1.2 support

mgberg:
# Summary of changes

Adds the first chunk of RDF 1.2 support, as I didn't see any other PRs looking to do so. 

This does **not** include parser, serializer, or SPARQL engine changes. It focuses exclusively on the data model and related utilities.

### What's included

- **Triple Terms**: A new `TripleTerm(Node)` class representing an RDF triple used as a term in the object position of another triple. Supports equality, hashing, ordering, n3 serialization (`<<( s p o )>>`), pickling, and nesting.
- **Directional Language-Tagged Strings**: A  `direction` parameter (`"ltr"` / `"rtl"`) is added to `Literal` for RDF 1.2 directional language-tagged strings. When set, the datatype is automatically `rdf:dirLangString` and `n3()` outputs the `@lang--dir` syntax.
- **RDF Namespace updates**: Adds `rdf:reifies` (property) and `rdf:dirLangString` (datatype) to the RDF DefinedNamespace.
- **`Graph.reify()` helper**: Convenience method that creates a `(reifier, rdf:reifies, <<( s p o )>>)` triple, with optional assertion of the underlying triple (conceptually similar to Turtle 1.2's annotation syntax).
- **Graph utility updates**: `skolemize()` / `de_skolemize()` recursively transform BNodes inside TripleTerms. `cbd()` follows BNodes embedded in TripleTerms and discovers RDF 1.2 reifiers linked via `rdf:reifies`.
- **`util.from_n3()` parsing**: `from_n3()` now round-trips both triple terms (`<<( s p o )>>` syntax, including nested triple terms) and directional language-tagged strings (`"text"@lang--dir`). A tokenizer handles quoted literals with escapes, IRIs, blank nodes, and recursive triple term delimiters within the inner content.
- **Isomorphism / canonicalization**: `compare.py` updated to enable graph diff and isomorphism checks to work when Triple Terms are used.

***NOTE: I am not super familiar with the `compare` module and am not confident the changes I made are fully correct. The tests (existing and new) pass but someone should look at it more closely.***

### Out of scope

Parsers, serializers, and SPARQL engine.

# Checklist

  the same change.
  
### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -107,6 +107,7 @@ Mark Watts
 Mark van der Pas
 Martin Wendt
 Martin van der Werff
+Matt Goldberg
 Matthias Urban
 Maurizio Nagni
 Maxim Kolchin
--- a/rdflib/__init__.py
+++ b/rdflib/__init__.py
@@ -62,6 +62,7 @@ __all__ = [
     "IdentifiedNode",
     "Literal",
     "Node",
+    "TripleTerm",
     "Variable",
     "Namespace",
     "Dataset",
@@ -200,7 +201,15 @@ from rdflib.namespace import (
     XSD,
     Namespace,
 )
-from rdflib.term import BNode, IdentifiedNode, Literal, Node, URIRef, Variable
+from rdflib.term import (
+    BNode,
+    IdentifiedNode,
+    Literal,
+    Node,
+    TripleTerm,
+    URIRef,
+    Variable,
+)
 
 from rdflib import plugin, query, util  # isort:skip
 from rdflib.container import *  # isort:skip # noqa: F403
--- a/rdflib/compare.py
+++ b/rdflib/compare.py
@@ -119,10 +119,91 @@ from typing import (
 )
 
 from rdflib.graph import ConjunctiveGraph, Graph, ReadOnlyGraphAggregate, _TripleType
-from rdflib.term import BNode, IdentifiedNode, Node, URIRef
+from rdflib.term import BNode, IdentifiedNode, Literal, Node, TripleTerm, URIRef
 
 if TYPE_CHECKING:
     from _hashlib import HASH
+
+
+def _get_bnodes_from_term(term: Node) -> Set[BNode]:
+    """Extract all blank nodes from a term, recursing into TripleTerms.
+
+    RDF 1.2 introduces triple terms (see
+    https://www.w3.org/TR/rdf12-concepts/#section-triple-terms) which may
+    contain blank nodes in their subject and object positions. For graph
+    canonicalization and isomorphism checking, we need to discover *all*
+    blank nodes that participate in a graph -- including those embedded
+    inside triple terms that appear as objects of asserted triples.
+
+    Per the RDF 1.2 abstract model:
+      - A triple term's **subject** is always an IRI or blank node (never
+        a nested TripleTerm, since triple terms can only appear in the
+        object position of a triple).
+      - A triple term's **predicate** is always an IRI.
+      - A triple term's **object** may be an IRI, blank node, literal, or
+        another (nested) triple term.
+
+    This function therefore:
+      1. Returns ``{term}`` immediately if the term itself is a BNode.
+      2. For a TripleTerm, collects BNodes from the subject and object
+         positions, recursing into nested TripleTerms in the object.
+      3. Returns the empty set for IRIs and Literals.
+
+    Args:
+        term: Any RDF node (BNode, URIRef, Literal, or TripleTerm).
+
+    Returns:
+        The set of all BNodes reachable within this term.
+    """
+    if isinstance(term, BNode):
+        return {term}
+    elif isinstance(term, TripleTerm):
+        result: Set[BNode] = set()
+        if isinstance(term.subject, BNode):
+            result.add(term.subject)
+        if isinstance(term.object, BNode):
+            result.add(term.object)
+        elif isinstance(term.object, TripleTerm):
+            result.update(_get_bnodes_from_term(term.object))
+        return result
+    return set()
+
+
+def _remap_triple_term(tt: TripleTerm, mapping: Dict[BNode, BNode]) -> TripleTerm:
+    """Apply a blank node mapping to a TripleTerm, producing a new TripleTerm.
+
+    During graph canonicalization and isomorphism checking, blank nodes are
+    relabeled (e.g., to canonical ``cb<hash>`` identifiers or to a mock BNode
+    for similarity checks). When a blank node appears *inside* an RDF 1.2
+    triple term -- either as the subject or the object -- those embedded
+    BNodes must be remapped consistently with how they are remapped at the
+    graph level.
+
+    This function walks the TripleTerm structure and replaces any BNode that
+    appears in the provided ``mapping`` dict. Because the RDF 1.2 spec allows
+    nested triple terms only in the **object** position, we recurse into the
+    object when it is itself a TripleTerm, but never into the subject.
+
+    Args:
+        tt: The TripleTerm whose embedded BNodes should be remapped.
+        mapping: A dictionary mapping original BNodes to their replacements.
+
+    Returns:
+        A new TripleTerm with BNodes replaced according to the mapping.
+        If no BNodes in the TripleTerm appear in the mapping, the returned
+        TripleTerm may still be a new instance.
+    """
+    s: Union[URIRef, BNode] = (
+        mapping.get(tt.subject, tt.subject)
+        if isinstance(tt.subject, BNode)
+        else tt.subject
+    )
+    o: Union[URIRef, BNode, Literal, TripleTerm] = tt.object
+    if isinstance(o, BNode):
+        o = mapping.get(o, o)
+    elif isinstance(o, TripleTerm):
+        o = _remap_triple_term(o, mapping)
+    return TripleTerm(s, tt.predicate, o)
 
 
 def _total_seconds(td):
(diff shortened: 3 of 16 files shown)


gjhiggins:
An excellent start! Just for info/surfacing, RDF1.2 adds two syntactical forms that parsers will need to handle:

### reifying triple (double chevron):
```
PREFIX : <http://example/>

:s :p :o .
<<:s :p :o>> :q 123 .
```
### triple annotation (brace and vbar)
```
PREFIX : <http://example/>

:s :p :o {| :r :z |} .
```
(examples taken from the [W3 RDF 1.2 test suite](https://github.com/w3c/rdf-tests/tree/main/rdf/rdf12))

> Out of scope
> Parsers, serializers, and SPARQL engine.

I believe Xuguang Song's work on RDFStar parsers and serializers could be adapted for use with RDF 1.2 - by the looks of the [RDFStar ebnf grammar included in the turtlestar parser](https://github.com/RDFLib/rdflib-rdfstar/blob/bceab0a444a482d0b2f58715bad4a2b7af93cb0f/rdflib/plugins/parsers/turtlestar.py#L405), both double chevron and brace+vbar syntax extensions are common to RDFStar and RDF 1.2

fwiw, Store API works just fine (*pace* some re-typing required):

```python
def test_reifying_triple_object_is_stored() -> None:
    """The object of the reifying triple is a TripleTerm and is correctly stored."""
    from rdflib.plugins.stores.sqleepycat import SQLeepycat
    store = SQLeepycat(identifier=EX.store0)
    path = "/tmp/sqleepycat-rdf1.2"
    store.open(configuration=path, create=True)
    g = Graph(store=store, identifier=EX.context0)
    g.reify(EX.alice, EX.knows, EX.bob, reifier=EX.r1)
    assert list(g.store.quads((None, None, None, None))) == [
        (
            EX.r1,
            RDF.reifies,
            TripleTerm(EX.alice, EX.knows, EX.bob),
            EX.context0
        )
    ]
    g.store.close()

    del store
    del g

    store = SQLeepycat(identifier=EX.store0)
    store.open(configuration=path, create=False)
    assert list(store.quads((None, None, None, None))) == [
        (
            EX.r1,
            RDF.reifies,
            TripleTerm(EX.alice, EX.knows, EX.bob),
            EX.context0
        )
    ]
    store.close()
    store.destroy(path)
```

mgberg:
I intentionally haven't touched the annotation syntax yet, as it only makes sense in the context of the serialization formats that support it (i.e. not the `n3()` methods or `from_n3()`). Although I agree, I think it would probably a good idea to start with that RDF* parser as it gets us a good chunk of the way there- probably would just need to add support for reifiers, dirLangStrings, version announcement, and perhaps a couple other details.

One of the reasons I wanted to stay away from the parsers/serializers is I'm not sure what the correct way would be to add them in the context of rdflib. Would we update the existing parsers/serializers to support 1.2? Would we add separate parsers/serializers for 1.2? I assume a discussion should be started about that.

As for typing- I agree, some typing adjustments should be made. I'm actually not sure what branch this PR should be against. Currently it is against `7.x`, which has, e.g., `_SubjectType = _PredicateType =_ObjectType = Node`. I'm not a huge fan of that, but these changes technically fall within the scope of those definitions.

That being said, `main` has the following:
```python
# RDFLib official stance is Subject can be a Literal
# If this ever changes, this part will be one of the first lines to modify.
_SubjectType: te.TypeAlias = Union[IdentifiedNode, Literal, Variable]
_PredicateType: te.TypeAlias = Union[IdentifiedNode, Variable]
_ObjectType: te.TypeAlias = Union[IdentifiedNode, Literal, Variable]
```
If I rebased against `main`, certainly we would want to add `TripleTerm` to `_ObjectType`. I chose to base this PR on `7.x` initially because it included some updates to `Graph.cbd` I made a while ago (#3367) I wanted to update which didn't exist in `main`.

gjhiggins:
> I intentionally haven't touched the annotation syntax yet, as it only makes sense in the context of the serialization formats that support it (i.e. not the `n3()` methods or `from_n3()`). 

Fair point.

> ... I think it would probably a good idea to start with that RDF* parser as it gets us a good chunk of the way there- probably would just need to add support for reifiers, dirLangStrings, version announcement, and perhaps a couple other details.

Having taken a closer look at the nturtle-star implementation, I no longer consider it to be a suitable starting point. For entirely pragmatic reasons of maintenance load, RDF 1.2 is very likely to mark a parting of the ways of RDFLib's Notation 3 and Turtle RDF 1.2 parsers.

> One of the reasons I wanted to stay away from the parsers/serializers is I'm not sure what the correct way would be to add them in the context of rdflib. Would we update the existing parsers/serializers to support 1.2? Would we add separate parsers/serializers for 1.2? I assume a discussion should be started about that.

Yes, a discussion is appropriate, or rather a continuation of [niklasl's](https://github.com/RDFLib/rdflib/pull/2115#issuecomment-1250030932) observation:

>> ... I did some simple benchmarking, and both Pymantic and TRLD parse Turtle about _ten times faster_ than RDFLib right now (that could be more due to the internal data model (or namespace handling) than the actual string wrangling though).  Properly wiring Turtle/TriG/-star EBNF into RDFLib graphs ought to reduce the code a lot too (following the lead of Pymantic, and attributing them for the approach if we're looking at their code).

I too, did some simple benchmarking but wasn't able to replicate the speed-up that @niklasl saw.

- Pymantic `test_parse_turtle_example_1` Parsed 500043 triples: **36.883801s**
- RDFLib lark `test_larkturtle_basic` Parsed 500043 triples: **29.521048s**
- RDFLib extant `test_turtle_basic` Parsed 500043 triples: **17.064235s**

I explored some RDFLib solutions in futher detail:

- Lark Turtle classic parser: Parsed 500043 triples: **48.53468s**
- Lark-Cython Turtle classic parser: Parsed 500043 triples: **46.90212s**
- Lark Turtle in-place parser: Parsed 500043 triples: **35.65342s**
- Lark-Cython Turtle in-place parser: Parsed 500043 triples: **31.73599s**
- Extant Turtle parser: Parsed 500043 triples: **19.03719s**

Where "classic" references the approach Pilon and Carrothers have taken in [Pymantic](https://github.com/norcalrdf/pymantic/) of parsing the source into triples and then adding the triples to a graph:

```python
triples = self.lark.parse(string_or_stream)

graph.addAll(triples)
```

and "in-place" references passing a graph as a property of the BaseParser class and adding triples directly to the graph's store as they are parsed:

```python
def statement(self, children):
    s, p, o = children
    self._call_state.graph.store.add((s, p, o), self._call_state.graph)
    return children
```

However, that's not the full story ...

RDFLib extant turtle parser tests on the W3c Turtle test suite:

```shell
# Extant
$ .venv/bin/pytest test/test_w3c_spec/test_turtle_w3c.py
======= 262 passed, 29 xfailed in 1.13s =======
```
RDFLib lark-based turtle parser tests on the W3c Turtle test suite:

```shell
# Lark
$ .venv/bin/pytest test/test_experimental/test_parsers/test_lark/test_turtle_w3c.py
============= 285 passed, 2 xfailed in 1.09s ===========
```

And those two XFAILs are not actually XFAIL, it's just that the Exception that's expected to be - and is properly - raised is improperly being consumed in the w3c test suite harness somehow; running those two tests outside the harness produces the expected Exceptions.

So, a likely gain in terms of maintenance requirement effort at the cost of an approximate halving of speed. The RDFLib plugin architecture allows extant and lark parsers to co-exist, selectable by the `format` parameter provided to `Graph.parse()`.

After some reflection, I think I was probably in error when I stated that the Store API works just fine with TripleTerms because when I come to reflect more deeply on the RDF 1.2 ntriples parser code, storing a *proposition tripleterm* as the object of a triple is straightforwardly supported by `Store.add()`. However, annotations reference *asserted triples* and are another matter entirely. In order to preserve the semantics, the subject needs somehow to reference an asserted triple without actually embedding it.

Just to get a better idea of the issues in play, I adapted Pilon and Carrothers' implementation of a [lark-based ntriples parser](https://github.com/norcalrdf/pymantic/blob/main/src/pymantic/parsers/lark/ntriples.py) (worth a quick look to get an idea of how direct and compact lark Transformer-based parsers can be) to handle RDF 1.2 propositional tripleterms. The parser passes all 29 of the [W3C RDF 1.2 rdf-n-triples test suite syntax tests](https://github.com/w3c/rdf-tests/tree/main/rdf/rdf12/rdf-n-triples/syntax) and affords me the opportunity to explore the implications of RDF 1.2 at least in the domain of propositional tripleterms

```python

rdflib.plugin.register(
    "ntriplesrdf12",
    rdflib.parser.Parser,
    "rdflib._experimental.plugins.parsers.larkntriples_rdf12",
    "LarkNTriplesParser",
)

def test_larkntriples_rdf12_parser():
    rdf12data = (
        "<http://example/a> "
        "<http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies> "
        "<<( <http://example/s> <http://example/p> <http://example/o> )>> . "
    )
    testg = rdflib.Graph().parse(data=rdf12data, format="ntriplesrdf12")
    assert(list(testg.triples((None, None, None)))) == [
        (
            rdflib.term.URIRef('http://example/a'),
            rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies'),
            rdflib._experimental.term.TripleTerm(
                rdflib.term.URIRef('http://example/s'),
                rdflib.term.URIRef('http://example/p'),
                rdflib.term.URIRef('http://example/o')
            )
        )
    ]

```

And by "explore", I mean ...

> Finally, RDF terms that appear in a triple term have the same
> denotation as when they appear in an asserted triple in the graph.
> For example, the term `:Alice` from a triple term and `:Alice` in an
> asserted triple both denote the same resource. For this reason, we
> say that triple terms are *transparent*.

```python
rdf12data = (
        "<http://example/s> <http://example/p> <http://example/o> .\n"
        "<http://example/a> <http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies> "
        "<<( <http://example/s> <http://example/p> <http://example/o> )>> . "
)
testg = rdflib.Graph().parse(data=rdf12data, format="ntriplesrdf12")
assert sorted(list(testg.triples((None, None, None)))) == [
    (
        rdflib.term.URIRef('http://example/a'),
        rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies'),
        rdflib._experimental.term.TripleTerm(
            rdflib.term.URIRef('http://example/s'),
            rdflib.term.URIRef('http://example/p'),
            rdflib.term.URIRef('http://example/o')
        )
    ),
    (
        rdflib.term.URIRef('http://example/s'),
        rdflib.term.URIRef('http://example/p'),
        rdflib.term.URIRef('http://example/o')
    )
]
asserted_s = next(
    testg.subjects(
        rdflib.term.URIRef('http://example/p'),
        rdflib.term.URIRef('http://example/o')
    )
)
assert asserted_s == rdflib.term.URIRef('http://example/s')
proposed_s = next(
    testg.subject_objects(
        rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies')
    )
)[1].subject
assert proposed_s == rdflib.term.URIRef('http://example/s')
assert asserted_s == proposed_s  # Probably obvious but has to be tested
```

niklasl:
> After some reflection, I think I was probably in error when I stated that the Store API works just fine with TripleTerms because when I come to reflect more deeply on the RDF 1.2 ntriples parser code, storing a proposition tripleterm as the object of a triple is straightforwardly supported by Store.add(). However, annotations reference asserted triples and are another matter entirely. In order to preserve the semantics, the subject needs somehow to reference an asserted triple without actually embedding it.

That doesn't sound correct; I think you're on the right track (triple terms are not allowed as subjects in RDF 1.2). Note that RDF 1.2 annotations is strictly syntactic sugar; so e.g.:
```turtle
prefix ex: <http://example/>

ex:s ex:p ex:o {| ex:x ex:y |} .
```
simply expands to:
```turtle
<http://example/s> <http://example/p> <http://example/o> .
_:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies>
        <<( <http://example/s> <http://example/p> <http://example/o> )>> .
_:b0 <http://example/x> <http://example/y> .
```
Meaning that the triple is both asserted and referenced by a reifier.

It's great to see this happen btw! I'll at least be around to chime in, and ideally help out more substantially. (I'll see if I can dig up my quick benchmarking tests; I might have missed something wrt. Pydantic.)

For the approach; I'd personally avoid duplicating parsers and serializers, and ideally get the RDFLib Turtle/TriG parser up do date, and fully conformant.

(FWIW, apart from TRLD I've made another pretty-printer supporting Turtle/TriG 1.2, for use with pyoxigraph stores: [gnarly](https://github.com/niklasl/gnarly). Like TRLD it's 0BSD, so feel free to plunder as you like.)

gjhiggins:
> > After some reflection, ... in order to preserve the semantics, the subject needs somehow to reference an asserted triple without actually embedding it.

> That doesn't sound correct; I think you're on the right track (triple terms are not allowed as subjects in RDF 1.2). Note that RDF 1.2 annotations is strictly syntactic sugar; so e.g.:
> 
> ```turtle
> prefix ex: <http://example/>
> 
> ex:s ex:p ex:o {| ex:x ex:y |} .
> ```
> 
> simply expands to:
> 
> ```turtle
> <http://example/s> <http://example/p> <http://example/o> .
> _:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies>
>         <<( <http://example/s> <http://example/p> <http://example/o> )>> .
> _:b0 <http://example/x> <http://example/y> .
> ```
> 
> Meaning that the triple is both asserted and referenced by a reifier.

Ohhh, I see. Thank you for the insight, I'd totally missed the "both asserted and referenced" perspective.

> It's great to see this happen btw! I'll at least be around to chime in, and ideally help out more substantially. (I'll see if I can dig up my quick benchmarking tests; I might have missed something wrt. Pydantic.)

I'll happily admit to being a bit puzzled at the time, all I did was to clone Pymantic and point its turtle parser at a file of 500k triples. Of course, it would be typical of me if I had been holding it wrong.

> For the approach; I'd personally avoid duplicating parsers and serializers, and ideally get the RDFLib Turtle/TriG parser up do date, and fully conformant.

Nit-pickingly, the erstwhile duplication comes in the form of creating lark-based parser implementations for parsers that RDFLib already has.

> (FWIW, apart from TRLD I've made another pretty-printer supporting Turtle/TriG 1.2, for use with pyoxigraph stores: [gnarly](https://github.com/niklasl/gnarly). Like TRLD it's 0BSD, so feel free to plunder as you like.)

Noted, thank you for your generosity. Not sure whether I'll be able to make much of a contribution in that area --- my discipline is cogpsych not compsci, so I do tend to blunder around somewhat, code-wise. I'll freely confess to struggling with the turtle_rdf12 implementation, merely attempting to transcribe the W3C RDF 1.2 turtle ebnf into a lark-compatible format is taxing my limited skillset, currently wrestling with conflicts.

PR #3434, opened 2026-05-20
validation workflow fix and deprecation bumps

gjhiggins:
# Summary of changes
The persistent tox run failure issue with 3.11+ubuntu-latest appears to be related to the use of the firejail harness specified in the validate workflow yaml --- https://github.com/RDFLib/rdflib/blob/0fdac395078b81313f30eeab74b1a18eb7bb4db6/.github/workflows/validate.yaml#L54

The [firejail README](https://github.com/netblue30/firejail) makes specific recommendations for Ubuntu:
`>>>`
### Ubuntu

Note: The versions from the distribution and PPA are likely to be outdated, so currently we recommend either downloading and installing the .deb package from the latest release:
`<<<`

In the course of addressing this, I tracked down the advised bumps to address the GH deprecations warnings:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@v4, actions/upload-artifact@v5, arduino/setup-task@v2.

I found bumps for actions/cache@v4 and actions/upload-artifact@v5 (grounded in the accompanying commit messages) but I wasn't able to resolve arduino/setup-task@v2 satisfactorily. Others have noted that an as-yet-unaddressed issue was raised back in early March and have started to migrate to [go-task/setup-task@v2](https://github.com/go-task/setup-task)
which purports to be a drop-in replacement, except that [isn't the case for the RDFLib validation workflow](https://github.com/RDFLib/rdflib/actions/runs/26132139918)

# Checklist

- [✅] Checked that there aren't other open pull requests for
  the same change.
- [✅] Checked that all tests and type checking passes.
- [✅] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

PR #3448, opened 2026-06-15
Stop supporting Python 3.8 and 3.9

multimeric:
# Summary of changes

* Remove all the Python 3.9 and 3.8 from the CI
* Updated the pyproject specifier to require later Python versions
* Regenerated the lockfile
* Bump the minimum Python version in ruff
* Apply the 3.10 ruff style
* Fix new and existing type checker issues as a result of this
* Bump mypy to assist with this
* Bump poetry to `2.4.1`

These versions are [end of life](https://devguide.python.org/versions/).

The benefits of removing support are that we can use newer versions of dependencies. For example, mypy 1.20 is needed to solve some type errors but it doesn't work on 3.9 so Poetry won't let us use it. See https://github.com/RDFLib/rdflib/pull/3446.

# Checklist

  the same change.
  

edmondchuc:
Thanks for this! I've enabled the checks and will review this soon.

multimeric:
Cheers. I'm a bit confused by this error though: https://github.com/RDFLib/rdflib/actions/runs/27526741875/job/81568290275?pr=3448#step:10:283. I'm no expert with tox.

multimeric:
@edmondchuc could I get a re-run please.

edmondchuc:
Sure thing, and apologies, I am not sure why it requires a manual approval to run the checks now.

gjhiggins:
> Sure thing, and apologies, I am not sure why it requires a manual approval to run the checks now.

Maybe it's a result of the GH actions Node.js version upgrade, it's possible that any workflow still configured to use the deprecated version of Node.js has been migrated to an "on-demand" status. The raft of now-failing checks are all reported thus:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: go-task/setup-task@v1. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

I did add a [PR to fix this](https://github.com/RDFLib/rdflib/pull/3434) but I've since come to recognise that the dev-dependabot backlist needs processing first because some of the dev-dependabot PRs, if accepted, would obviate some the changes made in my PR. AFAICT, the only change that *isn't* covered in a pending dev-dependabot PR is the deletion of the `firejail` installation for Python 3.11 on ubuntu-latest.

[Edit] Meh, actually *reading* the warning ... [go-task/setup-task repos version info](https://github.com/go-task/setup-task#version) implies `v2` as a minimum and `3.x` as current default whereas the RDFLib validation workflow has `v1`.

gjhiggins:
> Cheers. I'm a bit confused by this error though: https://github.com/RDFLib/rdflib/actions/runs/27526741875/job/81568290275?pr=3448#step:10:283. I'm no expert with tox.

I was too, until I dug deeper. I believe that the actual issue is upstream and there's nothing amiss with the tox call, it appears to be the JS script that sets up the environment and calls tox that is now out-of-scope because GH is migrating to a later version of Node.js.

I base this conclusion on the fact that all the checks passed for my PR that attempted to address the issue by [bumping the relevant versions](https://github.com/RDFLib/rdflib/pull/3434/changes#diff-9237aa56078dfbd8fcac3198563f306ada599aabaa43aca884905129779c5846)

I did set `go-task/setup` to v2, so I don't know why the failing checks with your PR are being reported by GH as using go-task-setup v1 unless they've done everyone a favour by rewriting the references to arduino with references to go-task. Go figure (ah-ha).

If you select "View details" from the `...` dropdown to the right of each failing check (below) and then open the "Annotations" drop-down, you'll see the GH warning I referenced in my reply to @edmondchuc (above).

multimeric:
@edmondchuc I've pulled from the new main. Can I get a workflow run please?

edmondchuc:
@multimeric done!

multimeric:
I've added https://github.com/RDFLib/rdflib/pull/3456 to fix the validation workflow, which you can merge separately or via this PR (which branches off it)

multimeric:
This has got a bit out of hand. The reason is that by dropping support for Python <3.10, I should bump the minimum ruff version. This causes a lot of existing code to fail, since it's not in Python 3.10 style, for example because it uses `typing.List`. Then if I try to apply the 3.10 ruff style, the type checker fails because of all the latent issues. Consequently I also have to update mypy, which can't be done separately since that requires dropping Python 3.9 support! 

If it is preferred, I could skip over the ruff change and put that in a separate PR. Then this one would be quite minimal but not complete.

edmondchuc:
Nice work @multimeric on getting all of the checks green!

Thanks for working through all the ruff and mypy failures. I can see this turned into a much larger and more tedious PR than expected, so I really appreciate the effort.

As long as the additional changes preserve the existing behaviour and are limited to linting, formatting, or type hint updates, I’m fine with keeping those commits in this PR.

I'll do a more detailed review of the changes tonight.

multimeric:
> As long as the additional changes preserve the existing behaviour and are limited to linting, formatting, or type hint updates, I’m fine with keeping those commits in this PR.

That is the intent of the changes, but in a handful of cases I had to change things outside of "type space" to make more appropriate changes. The main ones I can remember:

* [Adding a type assertion](https://github.com/RDFLib/rdflib/pull/3448/changes#diff-125fb4263650351b7c79e27fdee8dde93d58897bc7d2b453609b346a929d0693R330-R333)
* [Tightening an if statement condition](https://github.com/RDFLib/rdflib/pull/3448/changes#diff-125fb4263650351b7c79e27fdee8dde93d58897bc7d2b453609b346a929d0693R301-R306)
* [Renaming a variable (`type`) since this isn't compatible with the simultaneous use of `type` in type space](https://github.com/RDFLib/rdflib/pull/3448/changes#diff-cf861e2f6652917703ec1252e4f89436c6a0d7879ddb474005e5dda161f30117R38)

multimeric:
This intermittent failure is an external network thing: 
`coveralls.exception.CoverallsException: Could not submit coverage: 504 Server Error: Gateway Timeout for url: https://coveralls.io/api/v1/jobs`

edmondchuc:
> This intermittent failure is an external network thing: `coveralls.exception.CoverallsException: Could not submit coverage: 504 Server Error: Gateway Timeout for url: https://coveralls.io/api/v1/jobs`

I've created an issue for it here. https://github.com/RDFLib/rdflib/issues/3469

The errors appear to vary between 502 and 504.

multimeric:
I don't think that error needs to be a blocker though, it's not caused or impacted by this PR.

.github/workflows/validate.yaml
```diff
@@ -24,23 +24,12 @@ jobs:
...
-            os: ubuntu-latest
-            extensive-tests: true
-          - python-version: "3.8"
-            os: ubuntu-latest
-            extensive-tests: true
-            PREPARATION: "sudo apt-get install -y libxml2-dev libxslt-dev"
-            suffix: "-min"
-            TOXENV_SUFFIX: "-min"
-          - python-version: "3.9"
-            os: ubuntu-latest
```
edmondchuc, line 43: Instead of removing them, should we update them to a supported python version so the extensive tests and -min dependencies are still covered?

edmondchuc:
Would you be able to update the python references in `docs/developers.md`, specifically in the "Compatibility" and "Using tox" sections?

After those changes, we should be good to go!

nicholascar:
Hi all, I'm going to close this PR in favour of #3510 which aims to do the same thing but has been created after a move to uv and Zensical (for documentation). I should have fixed this first and then done those changes, sorry.

Issue #3445, opened 2026-06-12
Repository branch renaming - contributions against RDFLib branches


edmondchuc:
Hi RDFLib community. I'd like to improve the visibility of the two-branch split we currently have between `main` and `7.x`, to minimise the friction and effort required for contributors to submit pull requests.

The split was introduced to separate the maintenance for the next major version, v8 on `main`, from the current major version, v7 on the `7.x` branch. My understanding is that non-breaking changes, including new features and fixes, should generally target `7.x` so they can be released as minor or patch versions. Those changes can then be merged into `main`. Breaking changes should target `main` and are tracked for v8 in https://github.com/RDFLib/rdflib/issues/3302.

I'm very happy to see the recent activity on RDFLib, with many new pull request contributions. At a quick glance, many of them appear to be non-breaking and therefore suitable for the `7.x` branch. However, due to the current lack of visibility around the `main` vs `7.x` split, most of these pull requests seem to target `main`.

I'd be happy to spend some time helping contributors re-target their pull requests to `7.x` where appropriate, but I'd also like to get community feedback on how best to communicate the branch split.

A few possible improvements:
- Update [`developers.md`](https://github.com/RDFLib/rdflib/blob/main/docs/developers.md) to document the two-branch workflow
- Update [`PULL_REQUEST_TEMPLATE.md`](https://github.com/RDFLib/rdflib/blob/main/.github/PULL_REQUEST_TEMPLATE.md) to remind contributors which branch to target
- Consider adding a short note in [`CONTRIBUTING.md`](https://github.com/RDFLib/rdflib/blob/7.x/docs/CONTRIBUTING.md)

Feedback welcome!

lisat-dstg:
I did find this confusing when I came here to make a small non-breaking fix, even though it's obvious in retrospect. A note right up top in the Contributing Guide (https://rdflib.readthedocs.io/en/latest/CONTRIBUTING/) or Developer Guide (https://rdflib.readthedocs.io/en/latest/developers) addressing the "which branch should I target?" question would have had me understand it straight away.

It is necessary to already understand which branch we're targetting by the time we've gotten to creating a pull request, so updating the pull request template is somewhat less valuable.

gjhiggins:
[...]

PR #3452, opened 2026-06-23
docs: update dev docs - branch contribution targets

edmondchuc:
# Summary of changes

This PR updates the developer docs to provide contributors with the context necessary to determine which branch, `main` or `next`, to target their changes.

After this PR is merged, maintainers should quickly perform the branch renames as agreed by the community in https://github.com/RDFLib/rdflib/issues/3445.

# Checklist

  the same change.
  

lisat-dstg (approved):
Great!

docs/developers.md
```diff
@@ -382,3 +399,4 @@ Once this is all done, create another post-release pull request with the followi
 
 * Set the just released version in `docker/latest/requirements.in` and run `task docker:prepare` to update the `docker/latest/requirements.txt` file.
 * Set the version in the `pyproject.toml` file to the next minor release with a `a0` suffix to indicate alpha 0.
+* Ensure the released changes from `main` are merged forward into `next`.
```
ajnelson-nist, line 404: This reminds me - it would be helpful to note in the "Which branch do I target" paragraphs above that `next` receives periodic (/frequent) updates from `main`, by the maintainers running catch-up merges.  At least, I assume that is the planned strategy?

ajnelson-nist (approved):
Looks good to me, though I left one opportunity for clarification on how and when `next` gets updates that targeted `main`.

multimeric (approved):
LGTM. I strongly support making `main` the non-breaking branch as this should be targeted by default.

This isn't part of the PR since it's a one-off change, but can you spell out what the maintainers will need to do? Not only will they have to rename the branches in GitHub, but they will have to set the new `main` as the default branch right? Do we also need instruction for developers to update their remotes in local repos?

docs/developers.md
edmondchuc: Yes, that is the strategy! I've addressed this in https://github.com/RDFLib/rdflib/pull/3452/changes/1200f29e4ea186004e536b7893ad96aa0e6aeb00.

edmondchuc:
@multimeric good point. I'll add something to the github issue https://github.com/RDFLib/rdflib/issues/3445.

PR #3455, opened 2026-06-29
fix failing workflow - bump go-task/setup-task version

gjhiggins:
bump go-task/setup-task version from (failing) v1 to latest default of v3.x, as advised on [go-task/setup-task repos README](https://github.com/go-task/setup-task)

- [✅] Checked that there aren't other open pull requests for  the same change.
- [❌] Checked that all tests and type checking passes.
- [✅] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork),
  so maintainers can fix minor issues and keep your PR up to date.

PR #3456, opened 2026-06-30
Fix syntax and evrsion erorrs in the go-task/setup-task

multimeric:
The current validation workflow is (ironically) not valid.

See https://github.com/RDFLib/rdflib/actions/runs/28414953379 for an example.

I believe this was introduced in #3455.


gjhiggins:
> The current validation workflow is (ironically) not valid.
> I believe this was introduced in #3455.

I agree. The advice was vague. I expected my PR to cause the validation workflow to execute but i t didn't.

edmondchuc (approved):
Yeah, that's weird that it didn't flag the syntax error. Thanks for the fix @multimeric!

PR #3457, opened 2026-06-30
Apply a constant FILTER whose value is Python-falsy (e.g. FILTER(false))

### LINKED ISSUE #3381: rdflib.Graph.query: Constant False ASK query returns True
The following "Constant False" ASK query returns `True`: 

```python
from rdflib import Graph

graph = Graph()
result = graph.query("ask where {filter(false)}")

assert not result.askAnswer
```

`filter(false)` should effectively negate the possibility of a solution, so I would consider a `result.askAnswer` of `True` actually incorrect.

gaoflow:
Fixes #3381.

### The bug

A `FILTER` whose expression is a single constant with a false effective boolean value — but which is *also* Python-falsy — is silently discarded, so the filter never applies:

```python
from rdflib import Graph
Graph().query("ASK WHERE { FILTER(false) }").askAnswer   # -> True   (should be False)
```

Affected: `FILTER(false)`, `FILTER(0)`, `FILTER(0.0)`, `FILTER("")`. It also corrupts `SELECT` — `SELECT ?s WHERE { BIND(1 AS ?s) FILTER(false) }` returns one row instead of zero. By contrast `FILTER(1 = 2)` (a non-constant expression) and two stacked `FILTER(false) FILTER(false)` already behave correctly, which is what makes the single-constant case stand out.

Per SPARQL 1.1 §17.2, the effective boolean value of `"false"^^xsd:boolean` is `false`, so the FILTER eliminates every solution and the ASK must be `False`.

### Root cause

In `translateGroupGraphPattern` (`rdflib/plugins/sparql/algebra.py`):

```python
filters = collectAndRemoveFilters(graphPattern.part)
...
if filters:                       # truthiness of an Expr, not presence
    G = Filter(expr=filters, p=G)
```

`collectAndRemoveFilters` returns `None` only when there are no filters; otherwise it returns `and_(*filters)`, which for a *single* filter returns the bare expression object. When that lone expression is a constant `Literal` that is Python-falsy (`bool(Literal(False)) is False`, `bool(Literal(0)) is False`, `bool(Literal("")) is False`), `if filters:` is `False`, so the `Filter` node is never built and the constraint disappears from the algebra. (Multiple filters produce an `And`, which is always truthy — hence only the single-constant case was affected.)

### The fix

The value is `Optional[Expr]`, so test for presence rather than truthiness:

```python
if filters is not None:
    G = Filter(expr=filters, p=G)
```

### Tests

`test_constant_filter_ask` parametrizes ASK over the Python-falsy constants (`false`, `0`, `0.0`, `""` → must be `False`), the truthy constants (`true`, `1` → `True`), and the non-constant expressions that already worked (`1 = 2` → `False`, `1 = 1` → `True`). `test_constant_filter_select_rowcount` asserts the corresponding SELECT row counts. The Python-falsy cases fail on `main` (ASK `True`, SELECT leaks a row) and pass with the fix.

`test/test_sparql/` stays green (442 passed; the one pre-existing `xpassed` is unrelated, confirmed present without this change), and the W3C compliance suites (`test_sparql10_w3c.py`, `test_sparql11_w3c.py`) pass (863 passed). `ruff check` and `black` clean.

### DIFF
--- a/rdflib/plugins/sparql/algebra.py
+++ b/rdflib/plugins/sparql/algebra.py
@@ -383,10 +383,6 @@ def translateGroupGraphPattern(graphPattern: CompValue) -> CompValue:
                 "Unknown part in GroupGraphPattern: %s - %s" % (type(p), p.name)
             )
 
-    # ``filters`` is None when there were no FILTERs, otherwise it is the
-    # (possibly constant) filter expression. Test for presence explicitly: a
-    # constant expression whose value is Python-falsy (e.g. ``FILTER(false)``)
-    # must still produce a Filter node rather than being silently dropped.
     if filters is not None:
         G = Filter(expr=filters, p=G)
 
--- a/rdflib/plugins/sparql/algebra.py
+++ b/rdflib/plugins/sparql/algebra.py
@@ -383,7 +383,11 @@ def translateGroupGraphPattern(graphPattern: CompValue) -> CompValue:
                 "Unknown part in GroupGraphPattern: %s - %s" % (type(p), p.name)
             )
 
-    if filters:
+    # ``filters`` is None when there were no FILTERs, otherwise it is the
+    # (possibly constant) filter expression. Test for presence explicitly: a
+    # constant expression whose value is Python-falsy (e.g. ``FILTER(false)``)
+    # must still produce a Filter node rather than being silently dropped.
+    if filters is not None:
         G = Filter(expr=filters, p=G)
 
     # Mark this graph pattern as translated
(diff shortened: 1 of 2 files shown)


rdflib/plugins/sparql/algebra.py
```diff
@@ -383,7 +383,11 @@ def translateGroupGraphPattern(graphPattern: CompValue) -> CompValue:
                 "Unknown part in GroupGraphPattern: %s - %s" % (type(p), p.name)
             )
 
-    if filters:
+    # ``filters`` is None when there were no FILTERs, otherwise it is the
```
gjhiggins: This lavish comment appears superfluous, given the modern Python idiom of "in general, avoid truthiness else stuff like this can happen"

binggao1230: Removed — the is-not-None form speaks for itself.

PR #3478, opened 2026-07-12
Support BCE XSD date literals

### LINKED ISSUE #3092: Date parsing: BCE year triggers error
BCE dates are supported by ISO 8601 (`-0002-04-12` is listed as an example) and XSD 1.1 but rdflib throws an error:

`Literal("-0002-04-12", datatype=XSD.date)`

For python version > 3.11:
```
Failed to convert Literal lexical form to value. Datatype=http://www.w3.org/2001/XMLSchema#date, Converter=<function parse_xsd_date at 0x10516fec0>
  File "/rdflib-test/.venv/lib/python3.13/site-packages/rdflib/term.py", line 2163, in _castLexicalToPython
    return conv_func(lexical)  # type: ignore[arg-type]
  File "/rdflib-test/.venv/lib/python3.13/site-packages/rdflib/xsd_datetime.py", line 593, in parse_xsd_date
    return parse_date(date_string if not minus else ("-" + date_string))
ValueError: Invalid isoformat string: '-0002-04-12'
```

For python version == 3.10:
```
Failed to convert Literal lexical form to value. Datatype=http://www.w3.org/2001/XMLSchema#date, Converter=<function parse_xsd_date at 0x1049b3370>
  File "/rdflib-test/.venv/lib/python3.10/site-packages/rdflib/term.py", line 2163, in _castLexicalToPython
    return conv_func(lexical)  # type: ignore[arg-type]
  File "/rdflib-test/.venv/lib/python3.10/site-packages/rdflib/xsd_datetime.py", line 593, in parse_xsd_date
    return parse_date(date_string if not minus else ("-" + date_string))
  File "/bbaw/rdflib-test/.venv/lib/python3.10/site-packages/isodate/isodates.py", line 193, in parse_date
    raise ISO8601Error("Unrecognised ISO 8601 date format: %r" % datestring)
isodate.isoerror.ISO8601Error: Unrecognised ISO 8601 date format: '-0002-04-12'
```
The isodate error is misleading, see [this](https://github.com/gweis/isodate/issues/79) issue.

I know that python's datetime library does not support BCE dates, however [flexidate](https://github.com/datopian/flexidate) does. But perhaps there is a better way than introducing a new dependency. 

@ashleysommer I see that you recently reworked the whole date parsing logic, do you have any idea?

vibhor-aggr:
Fixes #3092.

## Summary

- Add a small `XSDDate` value for XSD date literals whose year cannot be represented by Python's `datetime.date`.
- Parse valid year-zero and negative-year XSD dates without logging a conversion failure or marking the literal ill typed.
- Keep ordinary positive-year XSD dates on the existing `datetime.date` conversion path.

## Root Cause

`parse_xsd_date()` stripped XSD date timezones correctly, but then delegated the resulting lexical form to Python's date parser. Python date values only support years in `1..9999`, so valid XSD 1.1 BCE values such as `-0002-04-12` raised during literal construction.

## Validation

- `python -m pytest -q test/test_literal/test_datetime.py::TestRelativeBase::test_bce_xsd_date`
- `python -m pytest -q test/test_literal/test_datetime.py test/test_literal/test_literal.py`
- `git diff --check`


nicholascar:
Approved. I just added a note into the documentation about this new class too

PR #3481, opened 2026-07-13
Internalise OWLRL & expand()

nicholascar:
This PR internalises the [OWL-RL RDFLib Package](https://github.com/RDFLib/OWL-RL/) so once this is merged, that package will be archived.

It preserves the original functionality of OWL-RL which can be called in the _old way_ like this:

```python
# as per test/test_inference/test_basic_inference.py::test_basic_inference

from rdflib import Graph
from rdflib.inference import DeductiveClosure, OWLRL_Semantics

# create an RDF graph, load a simple OWL ontology and data
g = Graph().parse(Path(__file__).parent / "relatives.ttl")

# expand the graph with OWL-RL semantics
DeductiveClosure(OWLRL_Semantics).expand(g)
```

It adds a new way of applying inference, via the new `expand()` function for `Graph` and equivalent use of that to the above is:

```python
# as per test/test_inference/test_new_interface.py::test_basic_inference_owlrl

from rdflib import Graph
# create an RDF graph, load a simple OWL ontology and data
g = Graph().parse(Path(__file__).parent / "relatives.ttl")

# expand the graph with OWL-RL semantics
g.expand("OWLRL")
```


nicholascar:
Closing in favour of #3484

PR #3495, opened 2026-07-28
Fix/jsonld internal conjunctivegraph warning

### LINKED ISSUE #3064: JSON-LD parser warns end-user of deprecation of `ConjunctiveGraph`
In my unit testing that loads JSON-LD into `Graph`s, I get a warning raised from this line (RDFLib 7.1.3):

https://github.com/RDFLib/rdflib/blob/7.1.3/rdflib/plugins/parsers/jsonld.py#L168

The warning reads:

`DeprecationWarning: ConjunctiveGraph is deprecated, use Dataset instead.`

I'm aware there's some significant discussion around `Dataset`, particularly in [Discussion 2591](https://github.com/RDFLib/rdflib/discussions/2591).  Apologies that I'm not caught up on the whole thread and am just asking instead.

Is it feasible and/or desirable to adapt that line of the JSON-LD parser to use a `Dataset` instead as part of another 7.x release?  I'm aware the desire is to not do many more 7.x's, but it might be desirable to remove that warning for users who need to continue using 7.x for a while yet.

Papishushi:
# Summary of changes

Fixes #3064.

Parsing JSON-LD into an ordinary `Graph` currently constructs an internal `ConjunctiveGraph`. Because `ConjunctiveGraph` is deprecated, this exposes RDFLib's own deprecation warning to downstream users even though their code does not use the deprecated class directly.

This pull request replaces that internal construction with a small private `Graph`-based context adapter that provides only the functionality required by the JSON-LD parser:

* context-aware default graph routing;
* preservation of the original sink identifier;
* named graph creation through `get_context`;
* support for context-aware stores that are not graph-aware.

Using `Dataset` directly was considered, but `Dataset` requires a graph-aware store. That would break previously supported configurations such as `Graph(store=AuditableStore(Memory()))`, whose store is context-aware but not graph-aware.

Direct parsing into a `Dataset` now explicitly uses `Dataset.default_graph`, avoiding access through deprecated default-context APIs.

Regression tests cover:

* JSON-LD parsing into an ordinary `Graph` without the internal warning;
* direct parsing into a `Dataset`;
* preservation of the target graph identifier;
* preservation of JSON-LD named graphs;
* compatibility with context-aware, non-graph-aware stores.

No warning is filtered or suppressed. The JSON-LD parser no longer imports or constructs `ConjunctiveGraph`.

This is independent of the broader `Dataset` API work and of #3477, which addresses internal uses of `Dataset.default_context`.

This pull request does not introduce any public API changes or backwards-incompatible behaviour.

# Checklist

* [x] Checked that there aren't other open pull requests for the same change.
* [x] Checked that all tests and type checking passes.
* [x] Added or updated tests that fail without the change.
* [x] Updated relevant documentation to avoid inaccuracies. No documentation changes were required because this restores the intended existing behaviour without changing the public API.
* [x] Considered adding additional documentation. No additional documentation appeared necessary for this internal compatibility fix.
* [x] Considered granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so maintainers can fix minor issues and keep the PR up to date.


Papishushi:
pre-commit.ci autofix

nicholascar:
Thanks @Papishushi. Great to squash warnings we don't have to have!

PR #3497, opened 2026-07-30
fix: reject Turtle/N3 local names beginning with a dash

### LINKED ISSUE #3403: ttl parser: Local name must not begin with dash
version 7.1.1

[this should not parse](https://w3c.github.io/rdf-tests/rdf/rdf11/rdf-turtle/turtle-syntax-bad-ln-dash-start.ttl)
```python
>>> list(Graph().parse(data="@prefix : <http://www.w3.org/2013/TurtleTests/> \n .:s :p :-o ."))
[(rdflib.term.URIRef('http://www.w3.org/2013/TurtleTests/s'), rdflib.term.URIRef('http://www.w3.org/2013/TurtleTests/p'), rdflib.term.URIRef('http://www.w3.org/2013/TurtleTests/-o'))]
```

Sanjays2402:
Closes #3403

# Summary of changes

The Turtle/N3 `PN_LOCAL` production allows `-` inside a local name but not as its first character. The `notation3` qname scanner began consuming the local name immediately after `:`, so `:-o` parsed as the IRI `<http://example.org/-o>` instead of raising a syntax error:

```python
>>> list(Graph().parse(data="@prefix : <http://www.w3.org/2013/TurtleTests/> \n .:s :p :-o ."))
[(URIRef('...TurtleTests/s'), URIRef('...TurtleTests/p'), URIRef('...TurtleTests/-o'))]
```

`qname()` now raises `BadSyntax` when the character directly after the prefix colon is `-`. Local names that merely contain a dash (`:o-1`) and blank node labels are unaffected. This makes previously-invalid input an error, so it is not backwards compatible for documents that relied on the lenient behaviour, but those documents were never valid Turtle.

The W3C `turtle-syntax-bad-ln-dash-start` and `trig-syntax-bad-ln-dash-start` negative-syntax entries were marked `xfail`; those markers are removed since they now pass.

# Checklist

  the same change.


nicholascar (approved):
Another great syntax error improvement, thanks @Sanjays2402

Issue #3315, opened 2025-11-13
Dataset.__len__() not returning expected count


edmondchuc:
I would have expected the following to return a size of `4` for the dataset, but it returns `3`.

```python
def test_mem_size():
    data = f"""
            <http://example.com/s> <{RDF.type}> <{SKOS.Concept}> <urn:graph:a> .
            <http://example.com/s> <{SKOS.prefLabel}> "Label" <urn:graph:a> .
            <http://example.com/s> <{SKOS.prefLabel}> "Label" <urn:graph:b> .
            <http://example.com/s> <{SKOS.definition}> "Definition" .
        """
    ds = Dataset()
    ds.parse(data=data, format="nquads")
    assert len(ds) == 4

```

Not confirmed yet but the dataset appears to be calculating the total number of triples instead of quads. When I remove one of the two `skos:prefLabel` statements, the count is still 3.

edmondchuc:
This issue is likely related to https://github.com/RDFLib/rdflib/issues/3314.

ghost:
I haven't been able to fix it, but I noticed a few interesting things. method nquads (meaning 4 parameters) but skos definition has three arguments not four. I tried to add a blank node (BNode) to make the skos definition have four arguments, so it wouldn't be rejected in adding to the DataSet due to having only three arguments.

https://github.com/RDFLib/rdflib/compare/main...interestingcoder1:rdflib:main

kiki286:
I took a look and was able to reproduce:

```
len(ds) == 3
len(list(ds)) == 4
len(list(ds.quads((None, None, None, None)))) == 4
```
From what I can tell, Dataset  inherits __len__() from ConjunctiveGraph, whose implementation calls the store length directly in graph.py:2384-2386. For the default in-memory store, Memory.__len__() counts the triples stored in a context in memory.py:568-572, not quads
### Existing documentation

The [Dataset documentation](https://rdflib.readthedocs.io/en/latest/apidocs/examples.datasets/#examples.datasets.graph_1_id) appears to intentionally demonstrate this behavior. The source example in datasets.py:196-205 is the same.

> """
> http://example.com/subject-x, http://example.com/predicate-x, Triple X, http://example.com/graph-1
> http://example.com/subject-y, http://example.com/predicate-y, Triple Y, http://example.com/graph-2
> http://example.com/subject-z, http://example.com/predicate-z, Triple Z, http://example.com/graph-1
[...]

PR #3499, opened 2026-07-31
fix: restrict ConjunctiveGraph.quads() to the requested context

### LINKED ISSUE #3314: Dataset.quads() does not restrict results to specified graph
I'm not sure if the following behaviour is expected or a bug in either `Dataset`/`ConjunctiveGraph` or the `MemoryStore`.

The test below shows that when iterating over quads from a specific graph, statements from other graphs leak into the result, except when using `DATASET_DEFAULT_GRAPH_ID`.

Only the third parameterized test (with `DATASET_DEFAULT_GRAPH_ID`) passes as expected. The first two tests (for named graphs `urn:graph:a` and `urn:graph:b`) incorrectly include statements from other graphs.

```python
import pytest

from rdflib import RDF, SKOS, Dataset, Graph, Literal, URIRef
from rdflib.graph import DATASET_DEFAULT_GRAPH_ID


@pytest.mark.parametrize(
    "graph_name, expected_quads",
    [
        [
            URIRef("urn:graph:a"),
            {
                (
                    URIRef("http://example.com/s"),
                    RDF.type,
                    SKOS.Concept,
                    URIRef("urn:graph:a"),
                ),
                (
                    URIRef("http://example.com/s"),
                    SKOS.prefLabel,
                    Literal("Label"),
                    URIRef("urn:graph:a"),
                ),
            },
        ],
        [
            URIRef("urn:graph:b"),
            {
                (
                    URIRef("http://example.com/s"),
                    SKOS.prefLabel,
                    Literal("Label"),
                    URIRef("urn:graph:b"),
                ),
            },
        ],
        [
            DATASET_DEFAULT_GRAPH_ID,
            {
                (
                    URIRef("http://example.com/s"),
                    SKOS.definition,
                    Literal("Definition"),
                    DATASET_DEFAULT_GRAPH_ID,
                ),
            },
        ],
    ],
)
def test_mem_quads(
    graph_name: URIRef, expected_quads: set[tuple[URIRef, URIRef, Literal, URIRef]]
):
    data = f"""
        <http://example.com/s> <{RDF.type}> <{SKOS.Concept}> <urn:graph:a> .
        <http://example.com/s> <{SKOS.prefLabel}> "Label" <urn:graph:a> .
        <http://example.com/s> <{SKOS.prefLabel}> "Label" <urn:graph:b> .
        <http://example.com/s> <{SKOS.definition}> "Definition" .
    """
    ds = Dataset().parse(data=data, format="nquads")
    quads = set(ds.quads((None, None, None, Graph(identifier=graph_name))))
    assert quads == expected_quads

```

Sanjays2402:
Closes #3314

# Summary of changes

`ConjunctiveGraph.quads()` passed the requested context to `Store.triples()`, but then yielded one quad per context in the store's returned context generator. `Memory.triples()` reports every context a triple appears in, so a triple present in more than one graph leaked quads for the other graphs — `Dataset.quads((None, None, None, <graph>))` returned statements belonging to unrelated named graphs.

When a context is given, only that context is now yielded; the unrestricted case is unchanged. This is a bug fix and should not be backwards incompatible for correct usage.

`test_aggregate2` asserted 6 quads for a graph holding 3 triples that are also present in another graph — it was encoding the leak — and is corrected to 3.

# Checklist

  the same change.


nicholascar:
@Sanjays2402 thanks for fixing this. I think this is one of the (many) small things involved with our long incorrect handling of Dataset, so every PR like this that finds and fixes these issues is important as we try and get comprehensive, correct, handling.

Sanjays2402:
The current head (dd70f7e) is ready to merge. It restricts `Dataset.quads()` to the requested context while preserving the existing all-context behavior, and the regression coverage includes both a shared triple and a context-specific triple. The focused dataset and aggregate-graph tests are green.

edmondchuc:
Hi @Sanjays2402, thank you for the fix. Do you know whether this also fixes https://github.com/RDFLib/rdflib/issues/3315?

Sanjays2402:
No — this PR does not fix #3315. #3499 changes `ConjunctiveGraph.quads()`/`Dataset.quads()` when a context is explicitly requested, so a shared triple is returned only for that requested context. `Dataset.__len__()` still delegates directly to `self.store.__len__()` (the code path involved in #3315), and this PR does not change it. The two issues are related in the broader Dataset handling, but #3315 needs a separate fix and regression test for counting quads across contexts.

rdflib/graph.py
```diff
@@ -2361,8 +2361,14 @@ def quads(
...
-            for ctx in cg:
-                yield s, p, o, ctx
+            if c is not None:
+                # A context was requested, so only that context is relevant;
+                # the store also reports the other contexts the triple
+                # appears in, which must not leak into the result.
+                yield s, p, o, c
+            else:
+                for ctx in cg:
+                    yield s, p, o, ctx
```
edmondchuc, line 2371: Could you please also add regression tests for the default graph case too?

Sanjays2402: Added the default-graph case to `test_quads_restricted_to_requested_graph`, including a triple also present in a named graph so the expected context is explicit. The focused dataset and aggregate-graph tests pass (11 total), and ruff is clean. The default-graph assertion also passes on the pre-fix tree because the Memory store already scopes that lookup; the named-graph cases remain the regression coverage for this change.

edmondchuc: Fantastic, thank you!

edmondchuc (approved):

PR #3500, opened 2026-07-31
fix: do not split XML namespace URIs into invalid qnames

### LINKED ISSUE #3490: XML Built-in namespace missing trailing slash
RDFLib version: 6.3.2
Python version: 3.12

I recently had an issue serializing an ontology using RDFlib and realized it is due to an inconsistency in the built-in namespace for XML.

```
rdflib.namespace._NAMESPACE_PREFIXES_CORE
{'owl': Namespace('http://www.w3.org/2002/07/owl#'),
 'rdf': Namespace('http://www.w3.org/1999/02/22-rdf-syntax-ns#'),
 'rdfs': Namespace('http://www.w3.org/2000/01/rdf-schema#'),
 'xsd': Namespace('http://www.w3.org/2001/XMLSchema#'),
 'xml': Namespace('http://www.w3.org/XML/1998/namespace')}
```

According to prefix.cc, the correct prefix for `xml` is `http://www.w3.org/XML/1998/namespace/` (note the additional trailing slash)

Based on this built-in namespace, if I have a file with URI `<http://www.w3.org/XML/1998/namespace/>` and attempt to parse, then serialize it into turtle format, that URI is serialized in turtle format as `xml:/`, which subsequently breaks when trying to re-parse using RDFLib.

Minimal example to reproduce:
```
from rdflib import Graph
data = "<https://example.org> <http://www.w3.org/2002/07/owl#imports> <http://www.w3.org/XML/1998/namespace/> ."
g = Graph()
g.parse(data=data, format="ttl")
data_ttl = g.serialize(format="ttl")
# this breaks with bad syntax error due to the xml URI appearing as xml:/
g.parse(data=data_ttl, format="ttl")
```

As a workaround, I have temporarily fixed the issue by adding this line:
```g.namespace_manager.bind("xml", Namespace("http://www.w3.org/XML/1998/namespace/"), override=True, replace=True)```

Sanjays2402:
Closes #3490

# Summary of changes

The built-in `xml` namespace (`http://www.w3.org/XML/1998/namespace`) has no trailing delimiter, but `split_uri()` unconditionally split any URI starting with it. A URI such as `http://www.w3.org/XML/1998/namespace/` was therefore split into the local name `/` and serialized as `xml:/`, which is not a valid qname and fails to parse back.

The XML namespace shortcut now applies only when the remainder is empty or a valid NCName (which still covers `xml:lang`, `xml:base`, etc.); otherwise the URI falls through to the generic splitting logic and is serialized as a full IRI, so it round-trips. Backwards compatible.

# Checklist

  the same change.

The new `test_xml_namespace_with_trailing_slash_roundtrips` in `test/test_namespace/test_namespace.py` fails without the fix and passes with it; `test/test_namespace/`, `test/test_serializers/` and `test/test_graph/` are green (719 passed).


nicholascar (commented):
Thanks for another parsing improvement @Sanjays2402

nicholascar (approved):

PR #3501, opened 2026-08-02
fix: terminate RDF patch prev header with a dot

### LINKED ISSUE #3007: RDF patch log serialiser omitting DOT on prev header
Each operation in an RDF patch must be terminated by a `.`, but this is omitted for `prev` headers.

```turtle
H prev <...> .
```

https://github.com/RDFLib/rdflib/blob/e8f61d4b5019040f097e560c3c1c8ecb41dd3fac/rdflib/plugins/serializers/patch.py#L67-L68

See the [RDF patch format](https://afs.github.io/rdf-delta/rdf-patch.html) for more details.

Sanjays2402:
Closes #3007

# Summary of changes

Every operation line in an RDF patch must be terminated by a `.`, but `PatchSerializer` wrote the optional `H prev` header without one, unlike the `H id` header directly above it, so patches carrying a prev header were not valid RDF patch documents.

The existing `test_prev_header` assertion matched the truncated output, so it is updated to require the terminating dot and now guards the fix. Backwards compatible: only the serialised prev header line changes.

# Checklist

  the same change.

### DIFF
--- a/rdflib/plugins/serializers/patch.py
+++ b/rdflib/plugins/serializers/patch.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import warnings
 from typing import IO, Any, Optional
-from uuid import uuid4
 
 from rdflib import Dataset
 from rdflib.plugins.serializers.nquads import _nq_row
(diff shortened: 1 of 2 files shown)


edmondchuc:
Could you please align this PR with https://github.com/RDFLib/rdflib/pull/3141 by making the headers optional?

Sanjays2402:
Done — aligned with #3141. The serializer no longer generates a UUID `header_id` by default; the `H id` header is only emitted when `header_id` is passed explicitly. Added a test covering the omitted case.

PR #3504, opened 2026-08-05
fix: Turtle/longturtle collection syntax can silently corrupt shared RDF lists

jdsika:
## Bug

`TurtleSerializer.isValidList` / `LongTurtleSerializer.isValidList` only check that every cell of an `rdf:List` chain carries exactly one `rdf:first`/`rdf:rest` pair. They do not check how many statements point *into* the chain.

When a list cell (not just the list head) is referenced from more than one place, the inline `( ... )` collection form consumes that cell at a single reference and never writes it out anywhere else. Any other reference to that cell is then left dangling in the output, and the serialization silently loses data (round-trip is broken).

### Repro

```python
from rdflib import Graph, BNode, Literal, Namespace, RDF
from rdflib.compare import isomorphic

g = Graph()
ex = Namespace("http://example.org/")
tail, head = BNode(), BNode()
g.add((tail, RDF.first, Literal("b")))
g.add((tail, RDF.rest, RDF.nil))
g.add((head, RDF.first, Literal("a")))
g.add((head, RDF.rest, tail))
g.add((ex.s1, ex.p, head))
g.add((ex.s2, ex.p, tail))  # second, independent reference into the chain

ttl = g.serialize(format="turtle")
g2 = Graph(); g2.parse(data=ttl, format="turtle")
print(isomorphic(g, g2))  # False on current main
```

Before this fix, `g` serializes as:

```turtle
ex:s1 ex:p ( "a" "b" ) .
ex:s2 ex:p _:tail .   # _:tail is never defined -> data loss
```

## Fix

`isValidList` now also requires that every non-nil cell in the chain (including the head) is the object of exactly one statement in the store. When that does not hold, the serializer falls back to explicit `rdf:first`/`rdf:rest` statements, which always round-trip correctly. Private (unshared) lists are unaffected and keep using the compact `( ... )` syntax.

Fixed identically in both `TurtleSerializer` (`rdflib/plugins/serializers/turtle.py`) and `LongTurtleSerializer` (`rdflib/plugins/serializers/longturtle.py`), which duplicate this logic rather than sharing it.

## Testing

- Added regression tests for both serializers covering the shared-cell case (asserting `isValidList` rejects the shared nodes and that serialize/parse round-trips are isomorphic), plus a test confirming private lists still use compact collection syntax.
- Full existing test suite passes locally (7561 passed / 73 skipped / 331 xfailed / 36 xpassed; two unrelated pre-existing errors in `test_misc/test_plugins.py` due to a missing `no_cover` fixture in this environment, unrelated to this change).
- `ruff` and `mypy` clean on the changed files.

Opening as draft while we do a bit more testing against real-world graphs; happy to remove draft status once reviewed.


nicholascar:
@jdsika this is a great find! I'm running the tests on it now and when they pass, this needs to be merged in. Thanks again.

nicholascar (approved):

PR #3506, opened 2026-08-10
Raise BadSyntax on a literal with '^^' but no datatype IRI

eeshsaxena:
### Problem

In the Notation3 parser (`notation3.py`, used for the `n3`, `turtle` and `trig` formats), a literal followed by a `^^` datatype marker with no valid IRI raises `IndexError` instead of a `BadSyntax` parse error:

```python
from rdflib import Graph
Graph().parse(data='<a> <b> "x"^^ .', format='turtle')
# IndexError: list index out of range   (notation3.py, dt = res2[0])
```

`uri_ref2()` reads the datatype URI into `res2`, but when the IRI is missing/invalid `res2` stays empty, so `res2[0]` blows up.

### Fix

Raise `BadSyntax` when `res2` is empty after reading the datatype, consistent with the parser's other error handling.

### Test

Added `test_missing_datatype_iri_raises_badsyntax` covering `n3`, `turtle` and `trig`. It raises `IndexError` before this change and now raises `BadSyntax`; valid typed literals are unaffected. `test/test_n3.py` passes (19 tests).


nicholascar (approved):

nicholascar:
@eeshsaxena thanks for this. The errors raised by parsers of RDF have been pretty bad over the years and really don't help people much with understanding what’s wrong with their RDF data when loading fails, so PRs like this are a really useful!

PR #3515, opened 2026-08-17
fix: count Dataset quads in len() and update documentation

### LINKED ISSUE #3315: Dataset.__len__() not returning expected count

kiki286:
# Summary of changes

Fixes #3315.

`Dataset.__len__()` now returns the number of quads in the dataset rather than the number of distinct subject-predicate-object triples. This means identical triples stored in different named graphs are counted separately, consistent with Dataset iteration and `quads()`.

Added regression coverage in test_dataset.py and updated the Dataset example in datasets.py.

This changes the existing `len(Dataset)` behavior, but is intentional and addresses the reported issue. `Graph.__len__()` continues to count triples.

Dataset.__len__() sums len(graph) over Dataset.graphs(), so it asks the store for the list of graphs and then for the size of each graph.  For the in-memory store each per-graph count is a constant-time lookup, so the cost grows with the number of graphs rather than the number of quads. For remote stores such as SPARQLStore it issues one query to list the graphs plus one COUNT query per graph, which can be slower than a single store-level count for datasets with many graphs.

# Checklist

- [x] Discussed the change and got in-principle agreement: #3315.
- [x] Considered adding an example in examples.
- [x] Considered granting push permissions to the PR branch so maintainers can fix minor issues and keep the PR up to date.


rdflib/graph.py
```diff
@@ -2729,6 +2729,10 @@ def __str__(self) -> str:
         )
         return pattern % self.store.__class__.__name__
 
+    def __len__(self) -> int:
+        """Number of quads in the dataset."""
+        return sum(1 for _ in self.quads((None, None, None, None)))
```
edmondchuc: This is potentially expensive and may be suboptimal depending on the backing store. Could we calculate the count by delegating to the `Graph.__len__` implementation instead?

```python
def __len__(self) -> int:
    """Number of quads in the dataset."""
    return sum(len(graph) for graph in self.graphs())
```

test/test_dataset/test_dataset.py
```diff
@@ -210,6 +218,9 @@ def test_not_union(get_dataset):
 def test_iter(get_dataset):
     store, d = get_dataset
     """PR 1382: adds __iter__ to Dataset"""
+    if store == "SPARQLUpdateStore":
+        pytest.skip("SPARQLUpdateStore does not expose contexts through triples()")
```
edmondchuc: Could you please explain why this skip is required?

kiki286: Sorry that was Codex overreacting. I have converted to draft and will republish when it is actually fixed. SPARQLUpdateStore is not exposing the contexts because of my changed behaviour.

kiki286: I investigated the SPARQLUpdateStore failure and tried to remove the skip by changing Dataset.quads() to iterate through each graph’s triples() method. However, that per-graph Dataset.quads() rewrite changed established context behavior and caused widespread regressions, so I reverted that attempt.

The underlying context limitation appears to predate the PR: SPARQLUpdateStore.triples() yields None for the context, while ConjunctiveGraph.quads() relies on store-provided context information. It only surfaced in this PR because I added assert len(d) == i_new compared the correct graph-count result with the backend’s empty iterator result. Before that assertion, the test compared two empty iterators and passed without verifying that the dataset contents were actually being iterated. The Ubuntu 3.10 jobs are the only matrix entries that start a local Fuseki server and therefore enable the SPARQLUpdateStore tests.

The current PR remains limited to changing Dataset.__len__() to count quads and updating the related tests and documentation so I left backend-specific skip remains in place. @edmondchuc would you prefer the SPARQLUpdateStore context behavior to be addressed separately in its own issue or PR? If the SPARQL store behaviour should be changed, I may need some guidance from someone more familiar with the implications of that behaviour and any compatibility considerations.

kiki286: Rebased onto main and squashed so no functional changes. The SPARQLUpdateStore skip is still needed because its triples() doesn't report contexts, now tracked in #3556. Could someone approve the workflow runs? Thank you!

kiki286:
Rebased onto main and squashed so no functional changes. The SPARQLUpdateStore skip is still needed because its triples() doesn't report contexts, now tracked in #3556. Could someone approve the workflow runs? Thank you!

Issue #1709, opened 2022-02-09
Intoduce WGS-84 DefinedNamespace


kvjrhall:
WGS84 ([html](https://www.w3.org/2003/01/geo/), [rdf](http://www.w3.org/2003/01/geo/wgs84_pos#)) is a small vocabulary that is commonly used in documents from dbpedia, geonames, and others. The vocabulary hasn't changed since 2006, suggesting that ongoing maintenance effort for the namespace should be minimal.

nicholascar:
Fee free to contribute a DefinedNamespace of the WGS-84 content if you like. You can add it to the files in https://github.com/RDFLib/rdflib/tree/master/rdflib/namespace.

There is no RDF prefix given within a DefinedNamespace class, so it doesn't matter what people use when you create that. Also, we are about to remove the larger set of prefixes bound to graphs here, https://github.com/RDFLib/rdflib/blob/master/rdflib/namespace/__init__.py#L350-L376, since this is problematic for some users.

Also, note that on [prefic.cc](http://prefix.cc) the prefix `pos` is registered for WGS-84 and that's what's used in the WGS-84-> GeoSPARQL 1.1 mapping available https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_e_2_wgs84_geo_positioning_an_rdf_vocabulary_pos.

So you don't really have to worry much about `geo` v. `wgs84` v. `pos` in RDFLib itself as you don't have to declare it for a new DefinedNamespace but I'd recommend going with `pos` when you want to use a prefix generally since, in the future, we may better link up RDFLib with prefix.cc.

kvjrhall:
Thank you for carrying this across the finish line for me. I got pulled
into other pressing issues and revisiting the PR slipped my mind.

PR #3550, opened 2026-09-11
fix: use the W3C Basic Geo http namespace IRI for WGS

### LINKED ISSUE #3529: WGS DefinedNamespace uses https://, but the vocabulary's namespace IRI is http://
**Version:** 7.6.0 (also present on `main` as of today)

## Summary

`rdflib.namespace.WGS` is declared as:

```python
# rdflib/namespace/_WGS.py
_NS = Namespace("https://www.w3.org/2003/01/geo/wgs84_pos#")
```

The namespace IRI of the W3C Basic Geo vocabulary is `http://www.w3.org/2003/01/geo/wgs84_pos#`.

Namespace IRIs are opaque identifiers, so `http://…` and `https://…` are two different
namespaces regardless of what the W3C server does with a request. As a result `WGS` matches
nothing in any published dataset, and the `wgs` prefix binding never applies to real data.

The mismatch is visible inside the file itself: the generated term comments carry the `http`
form while `_NS` carries `https`.

```python
_NS = Namespace("https://www.w3.org/2003/01/geo/wgs84_pos#")
...
lat: URIRef  # http://www.w3.org/2003/01/geo/wgs84_pos#lat
```

The issue that introduced the namespace (#1709) also links the vocabulary as
`http://www.w3.org/2003/01/geo/wgs84_pos#`.

## Reproduction

```python
import rdflib
from rdflib.namespace import WGS

g = rdflib.Graph()
g.parse(data='''
@prefix wgs: <http://www.w3.org/2003/01/geo/wgs84_pos#> .
<https://example.org/Chamonix> wgs:lat "45.92" .
''', format="turtle")

print(g.value(rdflib.URIRef("https://example.org/Chamonix"), WGS.lat))
# None   -- expected Literal('45.92')

print(WGS.lat)
# https://www.w3.org/2003/01/geo/wgs84_pos#lat
```

## Second symptom: the `wgs` prefix is never used

Because `_NAMESPACE_PREFIXES_RDFLIB` maps `"wgs": WGS`, the shipped prefix binds the https
IRI. Serialising data that uses the real namespace therefore invents a numbered prefix:

```python
g.serialize(format="turtle")
# @prefix wgs1: <http://www.w3.org/2003/01/geo/wgs84_pos#> .
```

`wgs1:` rather than `wgs:`, on the namespace the binding was added for.

The failure is silent. `WGS.lat` is a valid `URIRef` and lookups simply return nothing, so code
reaching for the shipped namespace gets empty results with no error — the reason I noticed was
a geo feature quietly finding no coordinates in a thesaurus that plainly had them.

For reference, GEMET (published by the European Environment Agency) declares exactly one form:

```
http://www.w3.org/2003/01/geo/wgs84_pos#
```

## Suggested fix

```diff
-    _NS = Namespace("https://www.w3.org/2003/01/geo/wgs84_pos#")
+    _NS = Namespace("http://www.w3.org/2003/01/geo/wgs84_pos#")
```

Happy to open a PR if that would help.

cpruijsen:
# Summary of changes

Set `WGS._NS` to `http://www.w3.org/2003/01/geo/wgs84_pos#`, the W3C Basic Geo vocabulary namespace IRI. `rdflib.namespace.WGS` used `https://www.w3.org/2003/01/geo/wgs84_pos#`. Those are different terms, so `WGS.lat` matched nothing in published datasets and Turtle serialisation invented `wgs1:` for the real namespace.

`WGS.lat` / `long` / `location` follow from `_NS`. Prefix-binding tests that encoded the old value are updated. A regression test covers lookup against Turtle that uses the vocabulary IRI, and that serialisation binds `wgs:` rather than `wgs1:`.

The change is a value correction of an existing public constant, not a new API. It is not backwards compatible for callers who stored or compared the previous `https://` strings.

Closes #3529

## Decision

Set `WGS._NS` to `http://www.w3.org/2003/01/geo/wgs84_pos#` and target `main`. Alternative: keep `https://` and document it, or treat the string change as breaking and land on `next`. Namespace IRIs are opaque. The W3C vocabulary RDF (`rdf:about="http://www.w3.org/2003/01/geo/wgs84_pos#"`), the HTML spec examples, the RDF link in #1709, and the term comments already in `_WGS.py` all use `http://`. Every other W3C DefinedNamespace in this repo already uses the published vocabulary IRI. The contributing guide says to open on `main` and ask if the branch target is unclear.

Provenance for the IRI: the issue reporter; #1709 linking `http://www.w3.org/2003/01/geo/wgs84_pos#`; W3C Basic Geo HTML + `wgs84_pos.rdf`.

Can retarget `next` if a major-version bump is preferred for this string change.

# Checklist

  - [x] Updated relevant documentation to avoid inaccuracies. (No docs stated the WGS namespace IRI; the HTML spec link in the docstring is a webpage URL and stays `https://`.)
- [x] Considered granting push permissions to the PR branch so maintainers can fix minor issues and keep your PR up to date.


edmondchuc (approved):
Thank you!

edmondchuc:
Targeting main is fine, we are planning for a v8 major version release in the next release.

